PR #72, opened 2015-09-04
Fix VHDL primitive for CLaSH.Sized.Internal.Signed.size#

leonschoorl:


christiaanb:
The verilog and systemverilog primitives seem similarly broken, could you include a fix in your pull request for those too?

leonschoorl:
Here you go.

I thought I had check them, but must have forgotten.

christiaanb:
Thanks for the patch

PR #83, opened 2015-10-06
Reference the version of ghc that clash was built with

ggreif:
And a bit of refactoring :-)


christiaanb:
Thanks for the patch!

PR #86, opened 2015-10-15
Disjoint expression consolidation

christiaanb:
A Transformation that lifts applications of the same function out of alternatives of case-statements. e.g.

```haskell
case x of
  A -> f 3 y
  B -> f x x
  C -> h x
```

is transformed into:

```haskell
let f_arg0 = case x of {A -> 3; B -> x}
    f_arg1 = case x of {A -> y; B -> x}
    f_out  = f f_arg0 f_arg1
in  case x of
      A -> f_out
      B -> f_out
      C -> h x
```


Ericson2314:
:D thanks!

PR #94, opened 2015-11-19
Add a stack.yaml to install easily

junjihashimoto:
Hi,

Add a stack.yaml for compiling this project with stack.
(Cabal  sometimes happens build-dependency-problems.)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,13 @@
+resolver: lts-3.11
+
+packages:
+- 'clash-lib'
+- 'clash-verilog'
+- 'clash-systemverilog'
+- 'clash-vhdl'
+- 'clash-ghc'
+
+extra-deps:
+- clash-prelude-0.10.3
+- ghc-typelits-extra-0.1
+


christiaanb:
A couple of questions, since I'm not really familiar with `stack` (yet): 

* Why is the `resolver` set to `lts` instead of the `nightly`? 
* Why doesn't the `extra-deps` section include `ghc-typelits-natnormalise` and `ghc-tcplugins-extra`?

junjihashimoto:
```nightly``` is just snapshot.
```lts``` is stable one. In genenal, ```lts``` is better.
https://github.com/fpco/lts-haskell#readme shows the details.

When a package of the ```lts``` does not match the version of cabal-file-dependency(for example a a package of ```lts```  is old) or  the ```lts``` does not have the package,
```extra-deps``` is needed

In our case, ```clash-prelude``` is old and ```ghc-tcpplugins-extra``` is not included.

christiaanb:
Thanks for the patch and the explanation.

junjihashimoto:
Thx, too.

PR #100, opened 2015-12-02
Differential clock

ggreif:
Should be pulled together with the clash-prelude PR.


christiaanb:
Thanks for the patch!

PR #108, opened 2016-01-07
Add support for Haskell types: Word(N), Int(N), and Char

christiaanb:
Add support for:
- `Int8`
- `Int16`
- `Int32`
- `Int64`
- `Word`
- `Word8`
- `Word16`
- `Word32`
- `Word64`
- `Char`

### DIFF
--- a/clash-ghc/src-ghc/CLaSH/GHC/Evaluator.hs
+++ b/clash-ghc/src-ghc/CLaSH/GHC/Evaluator.hs
@@ -28,38 +28,53 @@ reduceConstant tcm isSubj e@(collectArgs -> (Prim nm ty, args))
       [Literal (IntegerLiteral i), Literal (IntegerLiteral j)]
         | i == j    -> Literal (IntegerLiteral 1)
         | otherwise -> Literal (IntegerLiteral 0)
+      [Literal (IntLiteral i), Literal (IntLiteral j)]
+        | i == j    -> Literal (IntLiteral 1)
+        | otherwise -> Literal (IntLiteral 0)
       _ -> e
   | nm == "GHC.Prim.>#" || nm == "GHC.Integer.Type.gtInteger#"
   = case (map (reduceConstant tcm isSubj) . Either.lefts) args of
       [Literal (IntegerLiteral i), Literal (IntegerLiteral j)]
         | i > j     -> Literal (IntegerLiteral 1)
         | otherwise -> Literal (IntegerLiteral 0)
+      [Literal (IntLiteral i), Literal (IntLiteral j)]
+        | i > j     -> Literal (IntLiteral 1)
+        | otherwise -> Literal (IntLiteral 0)
       _ -> e
   | nm == "GHC.Prim.<#" || nm == "GHC.Integer.Type.ltInteger#"
   = case (map (reduceConstant tcm isSubj) . Either.lefts) args of
       [Literal (IntegerLiteral i), Literal (IntegerLiteral j)]
         | i < j     -> Literal (IntegerLiteral 1)
         | otherwise -> Literal (IntegerLiteral 0)
+      [Literal (IntLiteral i), Literal (IntLiteral j)]
+        | i < j     -> Literal (IntLiteral 1)
+        | otherwise -> Literal (IntLiteral 0)
       _ -> e
   | nm == "GHC.Prim.<=#" || nm == "GHC.Integer.Type.leInteger#"
   = case (map (reduceConstant tcm isSubj) . Either.lefts) args of
       [Literal (IntegerLiteral i), Literal (IntegerLiteral j)]
         | i <= j    -> Literal (IntegerLiteral 1)
         | otherwise -> Literal (IntegerLiteral 0)
+      [Literal (IntLiteral i), Literal (IntLiteral j)]
+        | i <= j    -> Literal (IntLiteral 1)
+        | otherwise -> Literal (IntLiteral 0)
       _ -> e
   | nm == "GHC.Prim.>=#" || nm == "GHC.Integer.Type.geInteger#"
   = case (map (reduceConstant tcm isSubj) . Either.lefts) args of
       [Literal (IntegerLiteral i), Literal (IntegerLiteral j)]
         | i >= j    -> Literal (IntegerLiteral 1)
         | otherwise -> Literal (IntegerLiteral 0)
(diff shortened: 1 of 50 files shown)


Ericson2314:
Do we support `Natural`? There are a bunch of changes in clash-prelude I wanted to make switching things from Integer/Int to Natural/Word.

christiaanb:
`Natural` is just as bad as `Integer`, it can only be improperly supported in CLaSH. I mean, I don't know what size `Integer` is, so I just make it 32 bits.

Ericson2314:
Oh, now that I think about it, the integer/natural use-cases I cared about we didn't synthesize anyways, (such as underlying type for `BitVector`). So this should be all that is needed.

Ericson2314:
On a different note, I am concerned about the final commit, that means two different developers on different arch could synthesize same code differently. Since clash is conceptually a cross-compiler, I am not sure inheriting host-arch stuff like that is a good idea.

At the very least, might be good to ensure the prelude doesn't use any word/int as an implementation detail that would trigger such behavior.

Ericson2314:
For example of both: https://github.com/clash-lang/clash-compiler/blob/master/clash-systemverilog/primitives/CLaSH.Prelude.RAM.json#L7-L8 ought to be unsigned, and *definitely* shouldn't be host-arch-dependent as a wrapper function hides the use of `Int` as an implementation detail.

Ericson2314:
https://github.com/clash-lang/clash-prelude/blob/master/src/CLaSH/Sized/Vector.hs#L981 same scenario here.

christiaanb:
I will take a look at the primitives/prelude. Until now, I've just been kinda hoping that people don't use `Vec`tors or `blockRAMs` of size `2^31-1`. That's the only time things can go wrong: VHDL simulation and synthesis tools don't support vectors/blockrams that are larger, I think. The reason is that those vectors could not be indexed, as `2^31-1` is the largest integer value: in VHDL, you can only index with values of type `integer`.

christiaanb:
Also, the reason I made the translation of `Int` and `Word` architecture-dependent, is overflow behaviour: I want overflow behaviour to be the same as the Haskell simulation on the system on which we are synthesising.

Ericson2314:
Re post 1: that all makes sense, sounds like we should be using `Word32

Ericson2314:
Re post 2: Maybe the thing to do is just not support `Int` or `Word` at all, or give a warning? The first especially is rather dogmatic, but either host-arch-dependence or a simulation-synthesis mismatch are *really* infuriating bugs to fall victim to.

Ericson2314:
To elaborate, there was some talk with Rust on what the default integer type should be (https://github.com/rust-lang/rfcs/pull/212 and maybe other places). One thing I learned is that 32-bit math is actually faster / just as fast on most x86-64 chips. IMO the only reason to use `Word/Int` is pointer arithmetic or speed. If `Word32` is just as fast, VHDL effective has a "pointer-size" of 32-bits, and any larger blockrams/bitvectors/etc would be grossly inefficient to simulate in Haskell regardless, I see no valid reason to ever want `Int` or `Word` with CLaSH.

christiaanb:
Well, yes, I don't want to use `Int` (or `Word`) either, but `Int` and `Integer` are Haskell's "staple" data types. They show up everywhere. I need to support `Int` and `Word`, or otherwise I can't support most of the standard Haskell prelude. The reason I `index_int` takes an `Int` argument, is because I want to be able to index into `Vec`tors with any data type that is an instance of `Enum`. And `Enum`'s `fromEnum` type is:

```haskell
fromEnum :: Enum a => a -> Int
```

Ericson2314:
> Well, yes, I don't want to use Int (or Word) either, but Int and Integer are Haskell's "staple" data types. They show up everywhere. I need to support Int and Word, or otherwise I can't support most of the standard Haskell prelude.

Then I would do the warning so users can audit when these show up. Somewhat relatedly, how does `fromIntegral = fromInteger . toInteger` work when the input and output types are >32 bits?

> The reason index_int takes an Int argument, is because I want to be able to index into Vectors with any data type that is an instance of Enum. And Enum's fromEnum type is...

Does using `Int` mean we rely on the backends to treat negative indices the same way? Negative indices, overflow, VHDL's 32 bit restriction---seems to me like we have the perfect storm of things to surmount.  This seems just more bullet-proof to me:

```haskell
index_word32 :: KnownNat n => Vec n a -> Word32 -> a
index_word32 xs i@(W32# n0) = sub xs n0
  where
    sub :: Vec m a -> Word32# -> a
    sub Nil              _ = undefined
    sub (y `Cons` (!ys)) n = if isTrue# (n ==# 0#)
                             then y
                             else sub ys (n -# 1#)
{-# NOINLINE index_word32 #-}

(!!) :: (KnownNat n, Enum i) => Vec n a -> i -> a
xs !! i = if
   | i' < 0 = error "CLaSH.Sized.Vector.(!!): negative index"
   | i' > fromIntegral (maxBound :: Word32) :: Int = error "CLaSH.Sized.Vector.(!!): indices over 2^32-1 unsupported due to VHDL restrictions"
   | i' > maxIndex xs = error (P.concat [ "CLaSH.Sized.Vector.(!!): index "
                                        , show i
                                        , " is larger than maximum index "
                                        , show (maxIndex xs)
                                        ])
   | otherwise = index_word32 xs 
 where i' = fromEnum i
{-# INLINE (!!) #-}
```

christiaanb:
So for the above code, I have no choice but to give a warning then... because It uses `Int`, due to:

```haskell
fromEnum :: Enum a => a -> Int
```

I'm not saying I'm happy with the current situation. But it's a compromise. And like I said, the only time the `Int`'s in the primitives become a problem is when people start creating vectors long lenght `2^31`!

christiaanb:
I will add a flag, with which you can set the number of bits that are used to encode 'Int' and 'Integer', where it defaults to the "native" width of the machine the compiler is build for. And I'll update the documentation to highlight this aspect of the translation of standard Haskell data types.

Ericson2314:
Good point about the would-be `int`-warning in my revised code. Heh, I had been proposing the warning and that change and completely forgot about the interaction. Furthermore, my experience just using Verilog and VHDL for synthesis led me to miss that for simulation purposes, Verilog and VHDL do something sane with out of bounds indexing, so there is no need for us to handle negative numbers instead of the output language.

Seeing all that, I think the flag you propose is the best we can do. Thanks!

PR #110, opened 2016-01-12
Fix GHC head warnings about about redundant constraints

ggreif:
Looks like these are not needed any more (due to Unbound?)

Let's see what the bots say...


christiaanb:
Thanks

PR #118, opened 2016-01-22
Clean up preprocessor directives

ggreif:
GHC 7.10.x is mandatory, so these #ifdefs
are not needed any more.


christiaanb:
Thanks

PR #133, opened 2016-03-15
Prepend 'work.' to use declaration of _types file

expipiplus1:
This seems to have been removed in 35689c4. Without qualifying using the
package name, xst is unable to find the imported module.


christiaanb:
Meh, I don't understand why ghdl didn't complain about this. I'll release a new version of clash which includes this fix tonight.

PR #152, opened 2016-06-20
Error handling for install script

j-marjanovic:


Ericson2314:
How about just `set -e`? This script should be simple enough for the variations of that not to cause problems.

christiaanb:
The install script has been removed

PR #175, opened 2016-09-09
Call generatePrimMap after loadModules

expipiplus1:
This is necessary in the case that the modules generate primitive
templates during compilation, for example with Template Haskell.

### DIFF
--- a/CLaSH.hs
+++ b/CLaSH.hs
@@ -9,7 +9,6 @@ import CLaSH.Rewrite.Types
 import CLaSH.GHC.Evaluator
 import CLaSH.GHC.GenerateBindings
 import CLaSH.GHC.NetlistTypes
-import CLaSH.Primitives.Util
 import CLaSH.Backend
 import CLaSH.Backend.SystemVerilog
 import CLaSH.Backend.VHDL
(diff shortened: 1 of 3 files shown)


christiaanb:
Could you also update https://github.com/clash-lang/clash-compiler/blob/0.6/CLaSH.hs#L41 in a similar way that you've updated InteractiveUI? Thanks

expipiplus1:
Done. I wasn't sure how to run CLaSH.hs, but if it compiles I'm sure it'll work.

```
$ nix-shell '<nixpkgs>' -p 'haskell.packages.ghc7103.ghcWithPackages (p: [p.clash-ghc])'
$ ghci CLaSH.hs -iclash-ghc/src-ghc
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help

clash-ghc/src-ghc/CLaSH/GHC/LoadModules.hs:19:2:
     error: #error TOOL_VERSION_ghc undefined
     #error TOOL_VERSION_ghc undefined
      ^
phase `C pre-processor' failed (exitcode = 1)
```

christiaanb:
Yeah, the way to call it is by executing `clash-dev`, which is located in the same top-level directory.

expipiplus1:
Thanks!

PR #190, opened 2016-11-29
Capture next stable GHC

ggreif:


ggreif:
This is obsolete now!

PR #209, opened 2017-03-25
Add 'clashi' program and 'clash-ghc' package

### LINKED ISSUE #208: Add a 'clashi' executable
Currently, `clash` works by default in batch mode, and can be invoked in interactive mode with `clash --interactive`. This is a wonderful feature that puts it leagues ahead of most traditional HDLs, of course.

However, this is extremely tedious to type and I can feel my fingers crying every time I do it. A `clashi` executable that mimics `ghci` would be fantastic.

**Pros**:

  - Incredible bloat decrease: only %20 more effort than `clash`, as opposed to 280% for `clash --interactive`. Amazing.

**Cons**:

  - Kind of stupid to implement. I suppose A) `Setup.hs` for `clash-ghc` could install a script to do this for us; B) the `clash-ghc` package could have a new exe `clashi` that uses `process` to invoke `clash --interactive <remaining args>`, or C) somehow thread --interactive through the current `Main.hs` for `clash-ghc`, but this is complicated because doing that is GHC's responsibility when parsing arguments to create `DynFlags`.

---

I think approach 2) (simply `execve(2)` the `clash` binary) is the right approach. ([This is actually what GHC for Windows does, since it cannot use a shell script.](https://github.com/ghc/ghc/blob/master/driver/ghc/ghc.c))

thoughtpolice:
Fixes #208.

This actually takes a better approach than the suggested paths in #208 -- it splits the `clash-ghc` executable into *another* new, `clash-ghc` library. The library is then used by two simple executables, with simply pass along their arguments -- `clashi` then simply prepends `--interactive` to the argument list.

This has advantages to the suggested paths:

  - It's easy to implement with minimal overhead thanks to `System.Environment.withArgs` (which I forgot about!)
  - It's generally Quite Difficult to know where cabal will install an executable at build-time, meaning that you simply have to execute `clash` out of `$PATH` (or get your binary path and execute the `./clash` executable next to it -- but this is all unreliable.) This means users can probably get themselves into various weird states and is more complicated.
  - There's now a new module `CLaSH.Main` with a single function, `defaultMain :: [String] -> IO ()`, which could be reused by *any* haskell code to run Clash programmatically, just like you would from the command line! (For example, inside Shake.) Just use the `clash-ghc` library.

I think this is superior to every single alternative.

I also took the time to move the cbits code around to keep things a bit cleaner.


ggreif:
(C) 2017 for new files?

thoughtpolice:
@ggreif Thanks for the catch! Fixed.

christiaanb:
Thanks @thoughtpolice

PR #218, opened 2017-04-21
Sum-of-product types: unused bits now "don't-care"

### LINKED ISSUE #212: Pad sum-of-products with U / X
Currently, when sum-of-product types are translated to HDL, we pad with `0` for all constructors but the constructor with the largest fields. We should pad with `U` or `X`, so that logic synthesis has more options for optimisation.

cbiffle:
Sum-of-product types include common types like Maybe and Either.  The
representations of values of such types often have bit positions that
are unused in some cases.

For example, the type 'Maybe (Unsigned 8)' is represented as a 9-bit
type, with a 1-bit "discriminator" that is set for Just and clear for
Nothing, and an 8-bit 'Unsigned 8' payload.  In the 'Nothing' case, the
8 bits of payload are unused.

Previously, these unused bits were explicitly initialized to zero.  This
isn't wrong, but it is overly conservative, and effectively instructs
the synthesis tool to put work into ensuring the value of bits that will
never be used.

This commit switches the representation of bits to use each backend
language's "don't-care" or "unknown" representation, giving the
synthesis tool more freedom and (in my tests) producing better results.

Fixes #212

### DIFF
--- a/clash-systemverilog/src/CLaSH/Backend/SystemVerilog.hs
+++ b/clash-systemverilog/src/CLaSH/Backend/SystemVerilog.hs
@@ -724,7 +724,7 @@ expr_ _ (DataCon ty@(SP _ args) (DC (_,i)) es) = assignExpr
     argExprs   = zipWith toSLV argTys es
     extraArg   = case typeSize ty - dcSize of
                    0 -> []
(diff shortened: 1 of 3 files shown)


christiaanb:
Thanks!

Issue #219, opened 2017-05-03
Verilog rules for bit widths in expressions are subtle and easy to get wrong


cbiffle:
On CLaSH 0.7.2, Verilog output (and probably SystemVerilog output) for certain constructs involving a sum type `t` can accidentally concatenate a platform-Int-sized bitstring in place of a `BitSize t`-size bitstring.

I haven't found the precise cause and I don't have a fix yet.  It was actually difficult to write a reduction (see below).

Here is an example of the offending output:

```verilog
  output [31:0] result;
  wire signed [63:0] result_app_arg;
  assign result_app_arg = $signed(64'sd8);
  
  assign result = {({({({({({7'b0000000,(5'd1)}),(5'd0)}),3'b010}),(5'd0)}),($unsigned(result_app_arg))}),2'b11};
```

`result` is a 32-bit value, and in Haskell it receives a `BitVector 32` result correctly.  In Verilog, however, it's being assigned from a **91-bit value**.  This is legal in Verilog, because Verilog hates you, but that's beside the point.  We wind up taking the bottom 32 bits of a 91 bit value, which are wrong.

The problem is `result_app_arg`.  It is the result of the following structure:

```haskell
data Opcode = .... | OpStore | ...
instance BitPack Opcode where
  type BitSize Opcode = 5
  pack = fromIntegral . fromEnum

something = pack ( ... OpStore ... )
```

`OpStore`'s enum value is 8, so `fromEnum OpStore` is giving the `64'sd8` value you see above... and then the 64-bit Int version is concatenated instead of the 5-bit version.

I believe the problem is [here in the backend](https://github.com/clash-lang/clash-compiler/blob/master/clash-verilog/src/CLaSH/Backend/Verilog.hs#L434-L435).  If you compare to [the equivalent VHDL backend code](https://github.com/clash-lang/clash-compiler/blob/master/clash-vhdl/src/CLaSH/Backend/VHDL.hs#L812-L816), you'll note that the VHDL version has a `resize` operation to explicitly truncate the value.  The Verilog is missing this.

Is the fix a simple matter of appending an index selector to the Verilog expression (the closest thing we've got to VHDL resize)?  Maybe.  I'll try in a bit.

[...]

PR #221, opened 2017-05-03
Verilog primitive fixes

cbiffle:
This mostly addresses #219 and fixes it in my use case, though I am uncomfortable declaring the fix "complete" because there are other appearances of `Int` and `Integer` remaining in primitive templates.

I also fixed a syntax error while I was in there (second commit).


christiaanb:
Thanks!

PR #223, opened 2017-05-05
Verilog: fix error in Index.fromInteger#

### LINKED ISSUE #222: Verilog: Index.fromInteger# truncation primitive is wrong
Here's a bug I introduced!

In 7c096f09f, I added code [here](https://github.com/clash-lang/clash-compiler/blob/master/clash-verilog/primitives/CLaSH.Sized.Internal.Index.json#L76) to truncate `Integer` when converting to `Index`.  However, this uses the type-level natural `n` as a bit width, when it is in fact a *vector size.*

So what we need here is not `~LIT[0]` but `~SIZE[~TYPO]`, i.e. the size in bits of the output type.

This is a result of me mechanically changing all the sized integer types to fix #219 and mindlessly applying the same technique to `Index` that I used on `BitVector`.

cbiffle:
Fixes #222

### DIFF
--- a/clash-systemverilog/primitives/CLaSH.Sized.Internal.Index.json
+++ b/clash-systemverilog/primitives/CLaSH.Sized.Internal.Index.json
@@ -73,7 +73,7 @@
 , { "BlackBox" :
     { "name"      : "CLaSH.Sized.Internal.Index.fromInteger#"
     , "type"      : "fromInteger# :: KnownNat n => Integer -> Index n"
-    , "templateE" : "$unsigned(~ARG[1][(~LIT[0]-1):0])"
+    , "templateE" : "$unsigned(~ARG[1][(~SIZE[~TYPO]-1):0])"
     }
   }
 , { "BlackBox" :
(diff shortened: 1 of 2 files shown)


christiaanb:
Would it be possible to add a regression test?

cbiffle:
I imagine so!  I don't see precedent in the test suite for checking that Verilog output matches e.g. a regular expression, but I can probably concoct a circuit that would test this behavior.

cbiffle:
Because the existing testsuites are large-scale integration tests with GHC, it's proving difficult to write a repro case that I'm confident will be robust in the face of future changes.

The class of bugs in #219 manifest only when an `Integer` is truncated to a fixed-width type in an expression context.  This behavior is unstable.  Small changes in the structure of the input source cause the output to diverge in one of two ways that hide the bug -- either

1. `fromInteger#` gets evaluated early and the primitive is never invoked, or
2. The result of `fromInteger#` is given a name at the Core level, causing it to get a separate wire/assign in Verilog, which hides the problem due to implicit coercions in Verilog.

A more targeted unit test of the backend, itself, should be more precise and robust, but may require additions to the test suite.  I'm looking into it.

christiaanb:
If it's too much work, you can just raise an issue about a regression test, and then I'll just merge this fix.

PR #227, opened 2017-05-16
clash-lib: Generate a slightly more helpful message on failure

expipiplus1:
This code used to generate __VOID__ with no explanation, this appends some information about the expression


christiaanb:
Closed in favour of the solution in #228

PR #247, opened 2017-09-02
clash-lib: make Verilog output tidier

thoughtpolice:
Mostly this makes the port declarations look nicer and much more readable, and uses Verilog-2001 compatible syntax for declaring input/output port types.


christiaanb:
Merged

PR #249, opened 2017-09-03
Add support for cabal new-build

thoughtpolice:
This adds support for `cabal new-build`, with freeze files as well, and blows away some old code. The main commit is 66becde, while the others are clean up. Here's the message from that commit.

# cabal: add new-build project files

This includes an accompanying freeze file, with a locked index state, so that
every build has fixed dependencies and may be upgraded atomically, independent
of others. The freeze file and index state should be updated in lockstep in
order to ensure the solver always picks exactly the same install plan (modulo
local modifications and local dependencies).

The configuration is designed to build all packages in a subdirectory under the
root directory, including any dependencies (preferred over whatever the solver
might pick otherwise). This means any patched dependencies or forks can simply
be checked out under the `clash-compiler` directory and `cabal new-build` will
do the right thing.

At the moment, `new-build` does not support cloning git repositories in order to
fetch custom dependencies. In particular, this is problematic for the ghc-*
plugin extensions needed by the compiler.

As a workaround, you can use the same workflow for patched dependencies: you can
simply clone those projects using `git` into the root of the `clash-compiler`
repository, and `new-build` will pick them up automatically.

Assuming you do this, you can build the clash executable and run the compiler on
an example in a single command:

    [a@ganon clash-compiler]$ cabal new-run clash-ghc:clash -- --help

Or, the equivalent command, in two steps:

    [a@ganon clash-compiler]$ cabal new-build clash-ghc
    [a@ganon clash-compiler]$ cabal new-run -- clash --help

As a bonus, cabal new-build supports GHC 8.2+ environment files. This means
after new-build has been run, it will write out a file under the name
`.ghc.environment-<platform>` which is automatically read by `ghci` to the root
of the repository.

At this point, simply executing `ghci` anywhere puts all packages and modules
for all dependencies available in the GHCi environment. e.g. you can run Clash
itself directly from GHCi:

    [a@ganon clash-compiler]$ ghci
    GHCi, version 8.2.1: http://www.haskell.org/ghc/  :? for help
    Prelude> import CLaSH.Main
    Prelude CLaSH.Main> defaultMain []
    <interactive>: no input files
    Usage: For basic information, try the `--help' option.
    *** Exception: ExitFailure 1


christiaanb:
Thanks!

PR #264, opened 2017-10-24
Fix pack#/unpack# of Signed in Verilog.

cbiffle:
This explicitly alters the signedness of values when moving between
BitVector and Signed, to ensure that BitVectors are interpreted as
unsigned and Signed as, well, signed.

This cropped up in the compilation of code like the following:

    signedLessThan :: BitVector 8 -> BitVector 8 -> Bool
    signedLessThan a b = signed a < signed b
      where signed :: BitVector 8 -> Signed 8
            signed = unpack

which wound up using an unsigned comparison in the Verilog output.

While in the prim definitions, I fixed two cases where an unsigned bit
literal was masquerading as a signed value, which would have produced
the same issue if a signed value were compared to maxBound/minBound.


christiaanb:
Thanks for the fix!

PR #265, opened 2017-11-07
Handle strict evaluation of Sized types

leonschoorl:


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -557,11 +559,20 @@ caseCast _ e = return e
 
 -- | Push a cast over a Letrec into it's body
 letCast :: NormRewrite
-letCast _ (Cast (Letrec b) ty1 ty2) = do
+letCast _ c@(Cast (Letrec b) ty1 ty2) | shouldPushDown c = do
   let (binds,body) = unsafeUnbind b
   changed $ Letrec $ bind binds (Cast body ty1 ty2)
 letCast _ e = return e
 
+-- | Is this a 'Cast' that you be pushed down?
```
christiaanb: Elaborate why some casts should or shouldn't be pushed down

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -557,11 +559,20 @@ caseCast _ e = return e
...
+shouldPushDown :: Term -> Bool
+shouldPushDown (Cast _ _ty1 ty2) = isIntegerTy ty2
+shouldPushDown _ = False
+
+argCastSpec :: NormRewrite
+argCastSpec ctx e@(App _ c@(Cast _ _ _)) | shouldPushDown c = specializeNorm ctx e
```
christiaanb: Shouldn't we ensure that the cast is work-free?? so let-bind any work before we specialise.

clash-lib/src/Clash/Core/Type.hs
```diff
@@ -513,3 +515,33 @@ isIntegerTy :: Type -> Bool
...
+-- For example: Vec (3+2) (Unsigned (3+1)) normalizes to Vec 5 (Unsigned 4)
+normalizeType :: HashMap TyConOccName TyCon -> Type -> Type
+normalizeType tcMap = go
+  where
+  go ty = case tyView ty of
+    TyConApp tcNm args
+      | name2String tcNm == "Clash.Signal.Internal.Signal"
+      , [_,elTy] <- args
+      -> go elTy
+      | name2String tcNm == "Clash.Sized.Internal.BitVector.BitVector" ||
```
christiaanb: Add a comment that you are looking through newtypes, but that clash' primitive number types are implemented as newtypes as well and you don't want to normalize them away.

clash-lib/src/Clash/Core/Evaluator.hs
```diff
@@ -297,6 +301,10 @@ unwind eval gbl tcm h k v = do
     Instantiate ty               -> return (instantiate h k' v ty)
     PrimApply nm ty tys vals tms -> primop eval gbl tcm h k' nm ty tys vals v tms
     Scrutinise _ alts            -> return (scrutinise h k' v alts)
+    CastTo ty1 ty2               -> return (castTo h k' v ty1 ty2)
```
christiaanb: This is wrong.... what it should be doing e.g. in a cast from `Integer` to `Signed`, is to add the `Signed.fromInteger#` primitive! The reason being, that primitives operating on `Signed` are expecting to see this primitive, they're not expecting to see a cast.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -538,6 +542,58 @@ bindConstantVar = inlineBinders test
...
+  where
+    test _ (_, Embed (Cast (Var _ _) _ _)) = return True
+    test _ _ = return False
+
+-- | Eleminate two back to back casts where the type going in and coming out are the same
+--
+-- @
+--   (cast :: b -> a) $ (cast :: a -> b) x   ==> x
+-- @
+eliminateCastCast :: NormRewrite
```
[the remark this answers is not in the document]
leonschoorl: It's good to check tyA and tyB.
But we need to do type normalization before comparing them, because they could be for example: `Signed 7` and `Signed (6+1)`.

I'm changing the case where tyA and tyC don't match, to produce an error.
Introducing a new cast here will produce a cast that can never be eliminated.
And then we will get an error during netlist generation that there was still some cast left over, but at that point we won't know where it came from.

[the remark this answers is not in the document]
leonschoorl: Also in the test case that I have this doesn't actually provide any source location.

leonschoorl:
I addressed all your comments in a series of patches in the [seq-on-sized-types-reviewed](https://github.com/leonschoorl/clash-compiler/commits/seq-on-sized-types-reviewed) branch.
Then I merged those patches into the original patches and force-pushed the result overwriting the  [seq-on-sized-types](https://github.com/leonschoorl/clash-compiler/commits/seq-on-sized-types).

But the Github UI seems very confused by this.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -538,6 +548,116 @@ bindConstantVar = inlineBinders test
...
+    test _ _ = return False
+
+-- | Eliminate two back to back casts where the type going in and coming out are the same
+--
+-- @
+--   (cast :: b -> a) $ (cast :: a -> b) x   ==> x
+-- @
+eliminateCastCast :: NormRewrite
+eliminateCastCast _ c@(Cast (Cast e tyA tyB) tyB' tyC)
+  | tyB == tyB' = do
```
christiaanb: I think we want to compare normalised types here as well, no?

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -538,6 +548,116 @@ bindConstantVar = inlineBinders test
...
+    tcm <- Lens.view tcCache
+    let tyA' = normalizeType tcm tyA
+        tyC' = normalizeType tcm tyC
+    if tyA' == tyC' then changed e
+                    else throwError
+  | otherwise = throwError
+  where throwError = do
+          (nm,sp) <- Lens.use curFun
+          throw (ClashException sp ($(curLoc) ++ showDoc nm ++ ": Found 2 nested casts whose types don't line up:\n" ++ showDoc c) Nothing)
```
christiaanb: Can we try to stay closer to a 80-column window in our formatting so that code is easier to read on smaller terminals?

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -538,6 +548,116 @@ bindConstantVar = inlineBinders test
...
+        tcm <- Lens.view tcCache
+        (nm',var) <- mkTmBinderFor tcm (mkDerivedName ctx (name2String $ varName nm)) e'
+        changed [(nm',Embed e')
+                ,(nm, Embed $ Cast var ty1 ty2)
+                ]
+      _ -> return [x]
+
+splitCastWork _ e = return e
+
+
```
christiaanb: These extra newlines aren't needed

clash-lib/src/Clash/Normalize/Transformations.hs
leonschoorl: Well `tyB` and `tyB'` should always be `Integer`.
But I'll add a check just to be sure.

PR #274, opened 2017-12-01
Add test of fromJust

bgamari:
I have found that fromJust fails to synthesize on GHC 8.2.1 due to missing
unfolding. Namely, the Data.Maybe interface file contains,
```
4cb6dd3ae4cd37241b07963360b4332c
  fromJust :: Maybe a -> a
  {- Arity: 1, Strictness: <S,1*U>,
     Unfolding: InlineRule (1, True, False)
                (\ @ a (ds :: Maybe a) ->
                 case ds of wild { Nothing -> fromJust1 @ a Just x -> x }) -}
53c35d567be116685e810e0ba1a76ff0
  fromJust1 :: a
  {- Strictness: x -}
```

### DIFF
--- a/testsuite/Main.hs
+++ b/testsuite/Main.hs
@@ -50,6 +50,7 @@ main =
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "ClassOps"            (["","ClassOps_testBench"],"ClassOps_testBench",True)
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "CountTrailingZeros"  (["","CountTrailingZeros_testBench"],"CountTrailingZeros_testBench",True)
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "DivMod"              ([""],"DivMod_topEntity",False)
+            , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "FromJust"            ([""],"FromJust_topEntity",False)
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "IrrefError"          ([""],"IrrefError_topEntity",False)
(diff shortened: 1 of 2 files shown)


thoughtpolice:
Oh, I see, the test /should/ fail, based on clash-lang/clash-prelude#137. Then it's just been marked incorrectly in the harness.

bgamari:
Ideally the test would pass but since we rely on `fromJust` from `base`, which doesn't have an unfolding, it fails. Consequently, I think the test is most accurately described as "expect broken".

I would argue that this situation of relying on libraries which weren't compiled with `-fexpose-all-unfoldings` is quite untenable. The only way to fix this IMHO is to suggest (or even insist) that the user uses a GHC built with `-fexpose-all-unfoldings`. Anything less is bound to be fragile.

testsuite/Main.hs
```diff
@@ -50,6 +50,7 @@ main =
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "ClassOps"            (["","ClassOps_testBench"],"ClassOps_testBench",True)
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "CountTrailingZeros"  (["","CountTrailingZeros_testBench"],"CountTrailingZeros_testBench",True)
             , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "DivMod"              ([""],"DivMod_topEntity",False)
+            , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "FromJust"            ([""],"FromJust_topEntity",False)
```
christiaanb: Change this line to:

```haskell
, runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "FromJust"            (["FromJust_topEntity"],"FromJust_topEntity",False)
```

or, remove the `TopEntity` annotation in your test, and rename `dut` to `topEntity`.

martijnbastiaan:
Closing this PR. I don't think this issue will be caught by our CI anyway.. If this is a mistake, please reopen.

PR #275, opened 2017-12-02
Fix heap shortcutting

bgamari:
This branch fixes a bug in the compile-time evalutor's heap management. The bulk of the change is described in e86d71b. I would like to stress that I'm not yet completely confident that I understand the precise trigger of the bug. That being said, I am much more confident in the implementation after this change than before.

I have also included a variety of patches improving error messages, which I found quite useful in the course of debugging.


christiaanb:
Fixed by: https://github.com/clash-lang/clash-compiler/commit/92dcfe581cc204283074e409a58dc6482d219c43

PR #277, opened 2017-12-02
CHANGELOG: Fix reference to -fclash-debug

bgamari:
This was the only reference to this flag that I was able to find, but it
wasn't quite correct.

### DIFF
--- a/clash-ghc/CHANGELOG.md
+++ b/clash-ghc/CHANGELOG.md
@@ -290,7 +290,7 @@
 * New features:
   * Inlining limit is configurable, run with `-clash-inline-limit=N` to set the inlining limit to `N`
   * Specialisation limit is configurable, run with `clash-spec-limit=N` to set the inline limit to `N`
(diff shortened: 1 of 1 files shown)


leonschoorl:
When the flag was introduced in v0.5.6 it was called `-clash-debug`.
Only much more recently has the `f` been added in #253.

But there hasn't been a changelog entry about this, nor are all the flags documented in one place.
Both should probably be done.

thoughtpolice:
(Also, I figured @christiaanb would be the one to write the changelog entries upon release.)

thoughtpolice:
Actually, sorry for triple-posting, but this is probably the better place to fix it, on the new docs page: http://clash-lang.readthedocs.io/en/latest/

But it should be mentioned in the changelog; I guess we can do that once we fork the new release branch...

bgamari:
https://github.com/thoughtpolice/clash-docs/pull/1 adds some documentation.

PR #279, opened 2017-12-05
CI: Haddock testing speedups

leonschoorl:
Lately a lot of our test runs on Travis have been timing out.

This patch combines the haddock generation pass with the build pass, thereby avoiding stack building everything twice.
Also disables haddock generation of all our dependencies.

This should speedup testing on Travis quite a lot.


leonschoorl:
:(
Looks like tests are broken until https://github.com/commercialhaskell/stack/issues/3624 gets sorted out

PR #283, opened 2017-12-12
Support Unsigned LIT in IF statements

basile-henry:
This is to prevent the following error:

```
<no location info>: error:                                                                     
    Clash error call:                                                                          
    Clash.Netlist.BlackBox.Util(261): IF: LIT must be a numeric lit:(DataCon (Unsigned 64) (DC 
(Void,-1)) [Literal (Just (Unsigned 64,64)) (NumLit 3)],Unsigned 64,True)
```


christiaanb:
Thanks

PR #286, opened 2017-12-26
Add primitive warnings

basile-henry:
This PR adds the possibility to specify warnings in the Clash primitive blackbox declarations. These will be shown to the user when generating a NetList using such primitives.

A Clash flag `-fclash-no-prim-warn` will remove the warning (which are enabled by default). Maybe these warnings shouldn't be shown by default?

The main use-case for these warnings is to warn the user that the primitive they are directly or indirectly using is not synthesizable. It is obviously fine to use such primitives in a `testbench` (and maybe the warnings should actually be disabled in testbenches).

The only blackboxes that currently have the warnings enabled are:

- `clockGen`
- `tbClockGen`

I will add warnings to more primitives as I think about them. I welcome any suggestions to improve this PR.

My own suggestions for improvements (I will make further commits to address them if they are of interest to other Clash contributors):

- Print warning to `stderr` instead of `stdout`
- Coloured warnings (orange?)
- Disable warnings for the `testbench` netlist generation
- Remove duplicate warnings (for the same primitive)

### DIFF
--- a/Clash.hs
+++ b/Clash.hs
@@ -41,7 +41,7 @@ doHDL b src = do
   let prepStartDiff = Clock.diffUTCTime prepTime startTime
   putStrLn $ "Loading dependencies took " ++ show prepStartDiff
   generateHDL bindingsMap (Just b) primMap tcm tupTcm (ghcTypeToHWType WORD_SIZE_IN_BITS True) reduceConstant topEntities
-    (ClashOpts 20 20 15 0 DebugFinal False True WORD_SIZE_IN_BITS Nothing HDLSYN True True False ["."]) (startTime,prepTime)
+    (ClashOpts 20 20 15 0 DebugFinal False True True WORD_SIZE_IN_BITS Nothing HDLSYN True True False ["."]) (startTime,prepTime)
 
 main :: IO ()
 main = genVHDL "./examples/FIR.hs"
--- a/clash-ghc/src-bin/Clash/Main.hs
+++ b/clash-ghc/src-bin/Clash/Main.hs
@@ -141,6 +141,7 @@ defaultMain = flip withArgs $ do
                              , opt_inlineConstantLimit = 0
                              , opt_cachehdl    = True
                              , opt_cleanhdl    = True
+                             , opt_primWarn    = True
                              , opt_intWidth    = WORD_SIZE_IN_BITS
                              , opt_hdlDir      = Nothing
                              , opt_hdlSyn      = Other
--- a/clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
@@ -43,6 +43,7 @@ flagsClash r = [
   , defFlag "fclash-hdlsyn"                  $ SepArg (setHdlSyn r)
   , defFlag "fclash-nocache"                 $ NoArg (liftEwM (setNoCache r))
   , defFlag "fclash-noclean"                 $ NoArg (liftEwM (setNoClean r))
+  , defFlag "fclash-no-prim-warn"            $ NoArg (liftEwM (setNoPrimWarn r))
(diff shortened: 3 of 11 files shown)


basile-henry:
This is a *Work in Progress* PR, I would love to get some feedback as to whether or not these warnings would be wanted, if the default should be different and so on.

This is the current warning I get from compiling `examples/Calculator.hs` (notice the `Warning:` line):

```
➜ stack exec -- clash -iexamples examples/Calculator.hs --vhdl                                            
Loading dependencies took 1.703727888s
Compiling: Calculator.topEntity
Applied 258 transformations
Normalisation took 0.672176668s
Netlist generation took 0.007352866s
Compiling: Calculator.testBench
Applied 304 transformations
Testbench normalisation took 0.290521178s
Warning: Clash.Signal.Internal.tbClockGen is not synthesizable!
Testbench netlist generation took 0.00730226s
Total compilation took 2.731448584s
```

Maybe this is an issue, or at least it's something I need to think about, but the warnings only appear on the first compilation because Clash then uses the cached results:

```
➜ stack exec -- clash -iexamples examples/Calculator.hs --vhdl
Loading dependencies took 1.737586829s
Compiling: Calculator.topEntity
Using cached result for: Calculator.topEntity
Compiling: Calculator.testBench
Using cached result for: Calculator.testBench
Total compilation took 1.740647645s
```

bgamari:
It seems to me that this could be quite useful.

martijnbastiaan:
Sorry for the (extremely) late reply to this PR. This is something I'd like to see in Clash too. I'll work on it shortly to get it in a state where we can merge it. As of your suggestions: makes sense, I'll  try and implement them.

martijnbastiaan:
I've rebased your work on top of master and started working on it here: https://github.com/clash-lang/clash-compiler/pull/379, so I'm closing this one.  Thanks for the PR :).

basile-henry:
@martijnbastiaan Great! Thanks for your help. Tell me if there's any of these suggestions you would want me to work on! :smile:

martijnbastiaan:
Maybe you could look into the colored output? It's a bit more complicated than just outputting a single color, as terminals might have dark or light backgrounds depending on their configuration. I've looked into it a few years ago and I think I remember terminals having color palettes with names for various types of messages. 

In the meantime I've refactored a bit and added a piece of code checking for duplicate warnings. I'm keeping a list on my own PR. Aaand... I'm actually not quite sure what the best way to work together is here. I think it might be best to pull my branch and force push it to your own (so you're rebased on master as well). I'll keep a watch on this pull request and pull the changes into my own. I'm open for other workflows :).

martijnbastiaan:
I've implemented the other suggestions. Colors is still an open issue, although it does print bold/underlined text now. I'm not really sure how to print colors correctly without doing it incorrectly for at least some people (groups running dark background and light backgrounds, for example).

basile-henry:
@martijnbastiaan Sorry I didn't get back to you earlier. Your PR is looking good! :+1: 
For the colours I wouldn't worry too much about what they exactly look like on other people's terminals. The way AINSI escape sequences specify coloured text is by using indices into the terminal's colour palette. So it's really up to the user to have a palette that is visible for them. If we use something bold and say yellow (or whatever their definition of yellow is) I think it should be fine for most people:

```bash
\\033[1;33mWarning\\033[0m
```

What might be useful on the other hand would be to give an option to remove all escape sequences (both bold and colour): something like `--no-colour`/`--no-color`. This can be useful when piping the output of `clash` to other tools.

martijnbastiaan:
No problem! 

>  So it's really up to the user to have a palette that is visible for them.

I agree, but defaults are important too. I've tested the default light themes for Konsole and gnome-terminal, and both perform poorly when using the "Yellow" color. I noticed GHC uses "Magenta", which seems to work well, so that's what I implemented too.

> What might be useful on the other hand would be to give an option to remove all escape sequences (both bold and colour): something like `--no-colour/--no-color`

good suggestion!

PR #288, opened 2018-01-11
Allow backends to have multiple primitive directories

bgamari:
This will allow us to share primitives between backends more easily (although this is left for future work).


bgamari:
Closing in favor of #290.

PR #290, opened 2018-01-12
Common up all identical black-boxes

bgamari:
This uses the support introduced in #288 to deduplicate primitives where possible.


christiaanb:
Thanks!

PR #322, opened 2018-06-21
Expose multiple include files in primitives

basile-henry:
This PR makes it possible to use more that 1 include file per primitive.


basile-henry:
Thanks for your help @christiaanb in getting this PR working! :smile:

Issue #320, opened 2018-06-03
Not able to create primitive


adamwalker:
I need an asymmetric width block ram in my design. The one in the Prelude is symmetric only so I decided to create a primitive. 

However, most of the time, my primitive is ignored. Where I would expect the primtive to be placed all I see is: 

```
  assign result = {32 {1'bx}};
```
The files to reproduce this are here: https://gist.github.com/adamwalker/cc30dc3f2b4d0ad6a4f37f66ab4b41d1.

If I replace the topEntity definition with:
```
topEntity = asymBlockRam# 
```
i.e. I remove the clk argument, then the ram is generated correctly. This is the only case where I have seen it generate the correct verilog. I am using the latest master (e1bc893).

leonschoorl:
Try changing your definition of asymBlockRam# to:
```haskell
asymBlockRam# clk = undefined
```

adamwalker:
Yeah, that works. As long as I list all the arguments to asymBlockRam it works fine. 

The problem is that this is unexpected. I think that when most people first create a primitive they will not bother with a Haskell implementation and simply leave it undefined without listing all the arguments, as I did. Clash will appear to ignore the primitive and they will be quite confused. 

On a related note, I now have an asymmetric block ram primitive that appears to work. If you would like me to incorporate that into the Prelude, I'd be happy to.

leonschoorl:
Yes, you are right, you should always list all the arguments to your custom primitives (if they're implemented with `error` or `undefined`).
And this is unexpected and annoying.

**TDLR**: When defining your own primitive and leaving the Haskell implementation undefined, list all the arguments of your primitive on the left side of the `=`
Like this: `myPrimitive _ _ = undefined`


The problem is we would like GHC(the frontend of Clash) to treat primitives as blackboxes, but when we do:
```haskell
myPrimitive :: Bool -> Bool -> Bool
myPrimitive = undefined
{-# NOINLINE myPrimitive #-}

topEntity a b = myPrimitive a b
```

We would expect GHC to translate `topEntity` to:
```haskell
topEntity = \a b -> myPrimitive a b
```
But GHC looks into `myPrimitive` and sees that we're applying `a` and `b` as the arguments to `undefined`.
 knows that that will never produce a (meaningful) result and produces:
```haskell
topEntity = \a b -> case myPrimitive a b of
  -- EmptyCase
```
Even if Clash replaces `myPrimitive` with a HDL template that is defined, `topEntity` is now undefined.

[...]

PR #325, opened 2018-07-03
WIP: Warn about problematic primitive definitions

leonschoorl:
Warns when a primitive:
  * isn't marked NOINLINE
  * produces an error when evaluating its result to WHNF
  * isn't using all its arguments

This should help with problems like #320.

Currently this produces a lot of warnings about not being marked NOINLINE, thinks like:
* GHC.Classes.eqInt
* GHC.Classes.geInt
* GHC.Classes.gtInt
* GHC.Classes.leInt
* GHC.Classes.ltInt
* GHC.Classes.neInt
* GHC.Classes.not
* GHC.Classes.&&
* GHC.Classes.||
* GHC.Integer.Type.eqInteger
* GHC.Integer.Type.neqInteger

Could that ever cause problems, or would it be safe to ignore/silence those warnings?

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
@@ -14,6 +14,7 @@ module Clash.GHC.GenerateBindings
 where
 
 import           Control.Lens            ((%~),(&),(^.),_1,_2)
+import           Control.Monad           (unless, when)
 import           Control.Monad.State     (State)
 import qualified Control.Monad.State     as State
 import           Data.Either             (lefts, rights)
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
After some internal discussion we've decided to pick this up again. Concerning the open question in this PR: we've decided to silence any warnings originating from `GHC.*` as we've never had problems with those functions and we don't control them anyway.

The proper solution is still to write a core2core plugin as mentioned in https://github.com/clash-lang/clash-compiler/issues/320#issuecomment-399997994, but this is an excellent band aid in the meantime.

martijnbastiaan:
Working on it in: https://github.com/clash-lang/clash-compiler/tree/leonschoorl-warn-problematic-prims

martijnbastiaan:
Merged, thanks!

PR #355, opened 2018-09-14
Rough code to derive BitPack for sum types.

lonetech:
Should be compatible for Bool and Maybe, and add support for Either and similar types.

Sample tests:
clashi -XDeriveGeneric -XDeriveAnyClass
import GHC.Generics
instance (BitPack a, BitPack b, KnownNat (BitSize a), KnownNat (BitSize b)) => BitPack (Either a b)
data Four = One | Two | Three | Four deriving (Generic, Show, Read, BitPack)
pack (Left 768 :: Either Int Bool)
unpack $ pack (Right True :: Either Int Bool) :: Either Int Bool
unpack $ pack Two :: Four

### DIFF
--- a/clash-prelude/src/Clash/Class/BitPack.hs
+++ b/clash-prelude/src/Clash/Class/BitPack.hs
@@ -39,6 +39,7 @@ import Data.Int
 import Data.Word
 import Foreign.C.Types                (CUShort)
 import GHC.TypeLits                   (KnownNat, Nat, type (+))
+import GHC.TypeLits.Extra             (Max)
 import Numeric.Half                   (Half (..))
 import GHC.Generics
 import Prelude                        hiding (map)
@@ -49,7 +50,7 @@ import Clash.Sized.BitVector
   (Bit, BitVector, (++#), high, low)
 import Clash.Sized.Internal.BitVector
   (BitVector (BV), pack#, split#, checkUnpackUndef, undefError, undefined#,
-   unpack#, unsafeToInteger)
+   unpack#, unsafeToInteger, resize#, msb#)
 
 {- $setup
 >>> :set -XDataKinds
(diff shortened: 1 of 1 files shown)


Jhana1:
Nice, I wanted to add this when I contributed the product type code. I wasn't sure how to avoid urnary encoding the type info. Do you think that's possible?

lonetech:
> Nice, I wanted to add this when I contributed the product type code. I wasn't sure how to avoid urnary encoding the type info. Do you think that's possible?

If you refer to my initial guess that Four would encode into something like 000, 100, 110, 111 (find first clear encoding), it turns out it didn't; GHC arranged the constructors into a tree producing 00, 01, 10, 11. This is noted in [GHC docs](http://hackage.haskell.org/package/base-4.11.1.0/docs/GHC-Generics.html#g:9): 

> The compiler is free to choose any nesting it prefers. (In practice, the current implementation tries to produce a more-or-less balanced nesting, so that the traversal of the structure of the datatype from the root to a particular component can be performed in logarithmic rather than linear time.)

It still has its quirks, for instance data Four = One (Signed 4) | Two (Unsigned 4) | Three Bool | Four deriving (Generic, Show, Read, BitPack) leads to Four encoding as 10_0010, as Three only needed one bit. Thus the discriminator bit between One and Two isn't in the same place as the one between Three and Four. This also won't merge values, so the first-clear encoding style would return for something like Maybe (Maybe (Maybe ())). 

Look to pull request #339 for a way to specify such things.

lonetech:
I managed to write a type class to count the constructors, such that e.g. `constructors (from True) == 2` and `constructor (from (Just ())) == 1`. I'm pretty sure the `constructors` function can be lifted into type computation, but `constructor` can't. 

    {-# LANGUAGE TypeOperators #-}
    {-# LANGUAGE ScopedTypeVariables  #-}

    import GHC.Generics

    class GCountConstructors t where
      constructors :: t -> Int
      constructor  :: t -> Int

    instance GCountConstructors (K1 i c p) where
      constructors _ = 1
      constructor  _ = 0

    instance GCountConstructors (V1 p) where
      constructors _ = 0
      constructor  _ = 0

    instance GCountConstructors (U1 p) where
      constructors _ = 1
      constructor  _ = 0

    instance (GCountConstructors (l p), GCountConstructors (r p)) =>
             GCountConstructors ((l :+: r) p) where
      constructors _ = constructors (undefined :: l p) + constructors (undefined :: r p)
      constructor (L1 v) = constructor v
      constructor (R1 v) = constructor v + constructors (undefined :: l p)

    instance (GCountConstructors (f p)) => GCountConstructors (M1 i c f p) where
      constructors m = constructors (unM1 m)
      constructor  m = constructor  (unM1 m)

This would allow making the encoding independent of the tree balance the compiler chooses. Is that an important property?

martijnbastiaan:
Thanks @lonetech ! I've written a PR based on the ideas presented in this thread over here: https://github.com/clash-lang/clash-compiler/pull/571. When merged, Clash should be able to derive `BitPack` instances for any Generic types. The result should match the format Clash already used internally. I think that that PR superseedes this one, so I'll be closing it. Thanks again!

PR #377, opened 2018-10-11
Tiny Macro fix in Core/Subst.hs

blaxill:
The space breaks at least MacOS GHC 8.4.3, as GHC interprets ASSERT as a
unknown datatype


martijnbastiaan:
Thank you!

PR #386, opened 2018-10-17
Generate sensitivity list correctly. Fix #365

### LINKED ISSUE #365: Sensitivity list incomplete in blocks generated for vector updates
Here is my minimized CLaSH source:

```
module CHIP8 where

import Clash.Prelude

data Phase
    = Init
    | Exec
    | Halt

{-# NOINLINE topEntity #-}
{-# ANN topEntity
  (Synthesize
    { t_name   = "CHIP8"
    , t_inputs =
          [ PortName "CLK"
          , PortName "RESET"
          ]
    , t_output = PortName "OUTPUT"
    }) #-}
topEntity
    :: Clock System Source
    -> Reset System Asynchronous
    -> Signal System (Unsigned 8)
topEntity = exposeClockReset board
  where
    board = (!!0) . snd <$> s

    s = register (Init, pure 0) $ f <$> s
      where
        f :: (Phase, Vec 16 (Unsigned 8)) -> (Phase, Vec 16 (Unsigned 8))
        f (Init, regs) = (Exec, regs)
        f (Halt, regs) = (Halt, regs)
        f (Exec, regs) = (Halt, replace 0 (regs!!0 + 1) regs)

```
So all it does is it updates the 0th value in a vector by adding 1 to it.

Running the resulting VHDL in a simulator, it seems the 0th element of the vector becomes `XXXXXXXX` after the update. In the CLaSH simulator, `sampleN 10 $ hideClockReset topEntity` shows `[0, 0, 1, 1, 1..]` as I would expect.

There are several small changes one can make to the program so that it does not corrupt the register's value:
* if I don't use the old value in computing the new one, e.g. `replace 0 1 regs` instead of `replace 0 (regs!!0 + 1) regs`. Perhaps notably, using the old value from a different index, `replace 0 (regs!!1 + 1) regs`, does NOT fix the corruption.
* Changing the program to us a single `Unsigned 8` not in a `Vec`tor. Something like `Vec 1 (Unsigned 8)` is not enough, it has to be an unwrapped `Unsigned 8`.

martijnbastiaan:
See issue #365. We only considered blackbox inputs, but this is incorrect - we need to look at variables IN the expression.

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
@@ -29,7 +29,7 @@ import           Data.List                       (nub)
 #if !MIN_VERSION_base(4,11,0)
 import           Data.Monoid
 #endif
-import           Data.Maybe                      (mapMaybe)
+import           Data.Maybe                      (mapMaybe, maybeToList)
 import           Data.Semigroup.Monad
 import qualified Data.Text
 import           Data.Text.Lazy                  (Text)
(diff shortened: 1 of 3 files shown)


clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -727,8 +719,63 @@ prettyElem (Template bbname source) = do
...
+usedVariables :: Expr -> [Identifier]
+usedVariables (Identifier i _)  = [i]
+usedVariables (DataCon _ _ es)  = concatMap usedVariables es
+usedVariables (DataTag _ e')    = [either id id e']
+usedVariables (Literal {})      = []
+usedVariables (ConvBV _ _ _ e') = usedVariables e'
+usedVariables (BlackBoxE _ _ _ _ t bb _) = nub (sList ++ sList')
+  where
+    -- Get input holes:
+    usedInputHoles = go
```
christiaanb: perhaps:

```haskell
usedInputHoles = go (\case {I _ i -> [i]; _ -> []}) 
usedVars = go (\case {Var [C v] _ -> [Text.toStrict v]; _ -> []})
go f = \case
  D (Decl _ args) -> ...
  ...
  DevNull es -> concatMap (go f) es
  e -> f e
```

Issue #383, opened 2018-10-16
fir filter from front page does not sythesize if len of fir is 3


ra1u:
```haskell
import Clash.Prelude

fir coeffs x = dotp coeffs (window x)
  where
    dotp as bs = sum (zipWith (*) as bs)

topEntity
  :: Clock System Source
  -> Reset System Asynchronous
  -> Signal System (Signed 8)
  -> Signal System (Signed 8)
topEntity = exposeClockReset ( fir v)

v = 1 :> 2 :> 3 :> Nil -- NOK
-- v = 1 :> 2 :> 3 :> 4 :> Nil -- OK
```

I am using clash-ghc from cabal `CLaSH, version 0.99.3 (using clash-lib, version: 0.99.3)`

command and output

    clash fir.hs --verilog
    Loading dependencies took 0.886579595s
    Compiling: Main.topEntity

    <no location info>: error:
        Clash error call:
        Clash.Rewrite.Util(409): Name {nameSort = Internal, nameOcc = x6658, nameLoc = UnhelpfulSpan "<no location info>"} not found

verbose start of clash
    -> % clash -v                           
    Glasgow Haskell Compiler, Version 8.4.2, stage 2 booted by GHC version 8.0.2
    Using binary package database: /usr/local/lib/ghc-8.4.2/package.conf.d/package.cache
    Using binary package database: /home/ralu/.ghc/x86_64-linux-8.4.2/package.conf.d/package.cache
    package flags []
    loading package database /usr/local/lib/ghc-8.4.2/package.conf.d
    loading package database /home/ralu/.ghc/x86_64-linux-8.4.2/package.conf.d
    wired-in package ghc-prim mapped to ghc-prim-0.5.2.0
    wired-in package integer-gmp mapped to integer-gmp-1.0.2.0
    wired-in package base mapped to base-4.11.1.0
    wired-in package rts mapped to rts
    wired-in package template-haskell mapped to template-haskell-2.13.0.0
    wired-in package ghc mapped to ghc-8.4.2
    wired-in package dph-seq not found.
    wired-in package dph-par not found.
    *** Deleting temp files:
    Deleting: 
    *** Deleting temp dirs:
    Deleting: 
    clash: no input files
    Usage: For basic information, try the `--help' option.

martijnbastiaan:
Interesting pattern..

| N  | Status                                   |
|----|------------------------------------------|
| **1**  | **OK**                                       |
| **2**  | **OK**                                       |
[...]

Issue #364, opened 2018-09-30
Simple design fails to compile


adamwalker:
Using Clash 0.99.3, compilation of the design below fails with the error:

```
<no location info>: error:
    Clash error call:
    Clash.Normalize(189): Expr belonging to bndr: x1536 not found
```
```
module Mod where

import Clash.Prelude

func
    :: BitVector 8
    -> BitVector 8
    -> BitVector 32
func d c
    | c == 0    = resize d
    | otherwise = 0

mod'
    :: forall dom gated sync. HiddenClockReset dom gated sync
    => Signal dom (BitVector 32)
mod' = o
    where

    x :: Signal dom (BitVector 8) = pure 0 
    y :: Signal dom (BitVector 8) = (+1) <$> z
    z :: Signal dom (BitVector 8) = pure 0

    f :: Signal dom (BitVector 8) -> Signal dom (BitVector 32)
    f = foldl1 (\x y o -> liftA2 (.|.) (x o) (y o)) 
        $  liftA2 func x
        :> liftA2 func y
        :> liftA2 func z
        :> Nil

    o = f (pure 0)

topEntity clk rst = withClockReset @System @Source @Synchronous clk rst mod'
```

Jhana1:
I think there's something broken with `resize`
See #352

adamwalker:
I'm not sure. If I remove the `resize` the bug is still there.

leonschoorl:
So far I've reduced the test case to this:
```haskell
module Mod where

import Clash.Prelude

func
    :: BitVector 8
    -> BitVector 8
    -> BitVector 32
func d c = 0

mod'
    :: forall dom gated sync. HiddenClockReset dom gated sync
    => Signal dom (BitVector 32)
mod' = o
    where

    x :: Signal dom (BitVector 8) = pure 0

    f :: Signal dom (BitVector 8) -> Signal dom (BitVector 32)
    f = foldl1 (\x y q -> liftA2 (.|.) (x q) (y q))
        $  liftA2 func x
        :> liftA2 func x
        :> liftA2 func x
        :> Nil

    o :: Signal dom (BitVector 32)
    o = f (pure 0)

topEntity clk rst = withClockReset @System @Source @Synchronous clk rst mod'
```

There seems to be something weird going on in Clash.Core.Evaluator during a caseCon transformation. (Adding `-fclash-debug DebugApplied` will make it complain about free variables being added)

[...]

PR #391, opened 2018-10-22
Bind heap variables

leonschoorl:
When the `Clash.Core.Evaluator.whnf'` returns with bindings left on the (pure)heap,
those should be bound so they are in scope for the (simplified) term returned by the evaluator.

Previously we just dropped this heap, leading to missing bindings, like in #364 and #383.

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations.hs
@@ -79,7 +79,7 @@ import           GHC.Integer.GMP.Internals   (Integer (..), BigNat (..))
 import           BasicTypes                  (InlineSpec (..))
 
 import           Clash.Core.DataCon          (DataCon (..))
-import           Clash.Core.Evaluator        (whnf')
+import           Clash.Core.Evaluator        (PureHeap, whnf')
 import           Clash.Core.Name
   (Name (..), NameSort (..), mkUnsafeSystemName)
 import           Clash.Core.FreeVars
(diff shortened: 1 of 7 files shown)


martijnbastiaan:
You can rebase this on master if you want to get rid of the Travis error. Sorry about that..

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -430,6 +433,33 @@ caseCon ctx e@(Case subj ty alts) = do
...
+    bndrs = map toLetBinding $ toListUniqMap heap
+    heapIds = map fst bndrs
+    is1 = extendInScopeSetList is0 heapIds
+    ctx' = TransformContext is1 (LetBody heapIds : ctxs)
+
+    toLetBinding :: (Unique,Term) -> LetBinding
+    toLetBinding (uniq,term) = (nm, term)
+      where
+        ty = termType tcm term
+        nm = mkId ty (mkUnsafeInternalName "x" uniq)
```
christiaanb: Add a comment saying why it's safe to use `mkUnsafeInternalName` here; that way, if ever there's a bug caused by this line, we have an idea which assumptions were wrong.

leonschoorl: Done
And I changed it to `mkUnsafeSystemName`, because the original name was generated with `mkUniqSystemId`

PR #395, opened 2018-10-24
Fix assertion behavior in outputVerifier*

basile-henry:
Assertions would fail even after the `outputVerifier` was done, which is not how modelsim and other HDL simulation software does assertions.


christiaanb:
Thanks

PR #404, opened 2018-11-06
 Enable DeriveGeneric and DeriveAnyClass by default

leonschoorl:
To make it less of a hassle to derive Undefined on your own types.


martijnbastiaan:
Maybe add them here as well: 

https://github.com/clash-lang/clash-compiler/blob/master/testsuite/Main.hs#L596

Issue #403, opened 2018-11-06
DSignal & explicit module heirarchy 


blaxill:
Currently some of the DSignal functionality that is not explicitly clocked is under the explicitly clocked subdirectories of clash-prelude. My understanding now is that the `Explicit` module hierarchy is for functions that take `=> Clock domain gated ->` value parameters, but DSignal functions don't seem to always follow this. 

For example

- The Signal data type and constructor are under non-explicit hierarchy [`Clash/Signal/Internal.hs`](https://github.com/clash-lang/clash-compiler/blob/27302687c62cdf8d9fcfe76748b8be6b0f69b6dd/clash-prelude/src/Clash/Signal/Internal.hs#L158)

- But DSignal data type and constructor, and functions such as

  1.  `feedback
  :: (DSignal domain n a -> (DSignal domain n a,DSignal domain (n + m + 1) a))
  -> DSignal domain n a`
  2. `unsafeFromSignal :: Signal domain a -> DSignal domain n a`

  etc, are under [`Clash/Explicit/Signal/Delayed.hs`](https://github.com/clash-lang/clash-compiler/blob/27302687c62cdf8d9fcfe76748b8be6b0f69b6dd/clash-prelude/src/Clash/Explicit/Signal/Delayed.hs#L85).

Is this unintentional and should these DSignal functions be under the non explicit hierarchy? I can start refactoring this around but wanted to confirm first before sending a pull request(s).

(Looking through `Clash/Explicit/Signal.hs` there are a couple of potentially misplaced functions `simulateB` and `simulateB_lazy` too)

blaxill:
I realize the current situation might be to simplify intermodule dependencies, and that placement of functions with no implicit or explicit clock might not be clear, but consistency could be better.

PR #406, opened 2018-11-08
Refactor non-explicit DSignal functions

blaxill:
This refactors the DSignal modules to:

```
clash-prelude/src/Clash/Explicit/Signal/Delayed.hs
clash-prelude/src/Clash/Signal/Delayed.hs
clash-prelude/src/Clash/Signal/Delayed/Internal.hs
clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
```

which is close to the Signal module layout. Specifically, it moves functions from out of Explicit hierarchy which arent relevant to it. No functions are removed or added. The choice of what is in the `Internal.hs` module was the most direct refactor of `Signal/Explicit/.../Delayed.hs`, as opposed to moving some functions into `Signal/Delayed.hs`  #403

### DIFF
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -112,7 +112,6 @@ Library
                       Clash.Explicit.Prelude.Safe
                       Clash.Explicit.Signal
                       Clash.Explicit.Signal.Delayed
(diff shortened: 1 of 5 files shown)


christiaanb:
Thanks!

PR #411, opened 2018-11-11
Additional DSignal functions

blaxill:
- delayN and delayI, delay a DSignal and leave the initial values after a
reset undefined. (Similar to Clash.Signal.delay)
- delayedFold, a tree fold by simple combinatorial application that
resolves its type level delay.

### DIFF
--- a/clash-prelude/src/Clash/Signal/Delayed.hs
+++ b/clash-prelude/src/Clash/Signal/Delayed.hs
@@ -5,20 +5,31 @@ License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 -}
 
-{-# LANGUAGE CPP                 #-}
-{-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE FlexibleContexts    #-}
-{-# LANGUAGE KindSignatures      #-}
-{-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE CPP                  #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE KindSignatures       #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE UndecidableInstances #-}
+#if __GLASGOW_HASKELL__ >= 806
+{-# LANGUAGE NoStarIsType #-}
+#endif
 
-{-# LANGUAGE Safe #-}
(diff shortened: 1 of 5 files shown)


blaxill:
This also fixes DSignal doctests

blaxill:
I've rebased onto your merged fixes (specifically keeping your Clash.Explicit.Signal.Delayed)

blaxill:
@christiaanb are there any other changes you'd like?

martijnbastiaan:
@blaxill Could you add a test case in https://github.com/clash-lang/clash-compiler/blob/master/testsuite/Main.hs? I'm willing to do it if you haven't got the time. You can basically copy the approach here: https://github.com/clash-lang/clash-compiler/blob/master/tests/shouldwork/Vector/VFold.hs. Run your testcase with `cabal new-run testsuite -- -p VFold`. (Where `VFold` is replaced by your test case name of course.) Don't worry if you get some failures due to missing simulators: some are annoyingly difficult to install, and Travis will run them for you anyway.

blaxill:
Sure, I think I'll have time in the next day or so.

martijnbastiaan:
Could you also rebase this branch on `master`? We've switched CI providers in the meantime.

clash-prelude/src/Clash/Signal/Delayed.hs
```diff
@@ -5,20 +5,31 @@ License    :  BSD2 (see the file LICENSE)
...
+{-# LANGUAGE CPP                  #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE KindSignatures       #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE TypeFamilies         #-}
+{-# LANGUAGE UndecidableInstances #-}
+#if __GLASGOW_HASKELL__ >= 806
```
martijnbastiaan: This is for GHC 8.0.6, right? We don't support <8.2, so I think you can remove this.

blaxill: Actually kind of confusingly, 806 refers to [8.6.0 (see here)](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/intro.html#ghc-version-numbering-policy). This case is needed for 8.6.0 and above as GHC starts interpreting all unqualified cases of * as `Type` (interfering with type level multiplication). We could also enable NoStarIsType in the cabal config or use ugly qualifications, but it looks like Clash is using this same construct elsewhere, and a general change like that would be better in its own PR? ([8.6.0 StarIsType migration notes](https://ghc.haskell.org/trac/ghc/wiki/Migration/8.6#StarIsType))

martijnbastiaan: I see, thanks for the pointer! And I agree, that change is better suited for another PR.

martijnbastiaan:
Thank you @blaxill !

PR #414, opened 2018-11-14
Move to GitLab CI

martijnbastiaan:
GitLab CI offers the ability to deploy our own test runners. This would speed up our test times significantly. Thanks to @leonschoorl for implementing this.


martijnbastiaan:
I've re-added Travis CI to handle pull requests and forks. The `.travis.yml` is quite clean so I don't expect much upkeep. With cached docker images, cabal store, and `dist-newstyle` it's reasonably fast. Probably not as fast as Circle CI, but allowed failures is still not an option there.. It might be a dumb idea, but we could combine all three:

* For forks and pull requests:
  * Circle for 8.2, 8.4, 8.6
  * Travis for HEAD with allowed fail (and maybe some short-circuit features; for example, we know singletons is failing to build for the time being, so we might just build it upfront)
* GitLab for everything else

We'd combine all positives of the three platforms:

- Cheap
- Tests everything, including PRs and forks
- Own builders
- Fast

Also, since we're using Docker we could set up a private repository including modelsim. Of course, forks would still use a public docker image.

leonschoorl:
For reference, the gitlab issue about building pull requests: https://gitlab.com/gitlab-org/gitlab-ee/issues/5667

martijnbastiaan:
I've added the CicleCI config file written by Leon and enabled it on the project. This should conclude this journey..

PR #419, opened 2018-11-19
Render delc attributes

cchalmers:


martijnbastiaan:
Could you provide a (small) example showing that this works? I'll add it to the testuite. Could you also fix the spelling mistake in the commit message? Thanks!

cchalmers:
@martijnbastiaan the only way I know how to test it is to make a black box template function. Is there infrastructure for testing this in clash or is there some other way I can test this?

@christiaanb yes, it looks like it will be rendered twice. I don't really know vhdl so I'm not sure what the correct thing to do here is. I tried understanding the comment at https://github.com/clash-lang/clash-compiler/blob/master/clash-lib/src/Clash/Backend/VHDL.hs#L618-L631 and looked at the links (the quartushelp link isn't valid anymore) but I don't even know what an `architecture` is.

martijnbastiaan:
> Is there infrastructure for testing this in clash or is there some other way I can test this?

It's very rudimentary, but it's there. See:

https://github.com/clash-lang/clash-compiler/blob/master/testsuite/Main.hs#L197

And

https://github.com/clash-lang/clash-compiler/blob/master/tests/shouldwork/TopEntity/PortProducts.hs#L42-L56

christiaanb:
Feature implemented by: https://github.com/clash-lang/clash-compiler/commit/67d36495cc108afad767e92d1ee21e12411cc9e2

PR #432, opened 2018-11-25
New inline strategy

christiaanb:
By default, Clash will now produce a single HDL file. To influence HDL file generation you can either:

* Add a `NOINLINE` pragma to a top-level function, so that it gets its own HDL file, except when they're constants.
* Add a `ANN Synthesize { .. }` pragma to a top-level function, so that it gets:
  * its own HDL file, 
  * in a separate directory, 
  * and a stable interface,
  * and is cached between runs.

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox/Types.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Types.hs
@@ -83,6 +83,9 @@ data Element = C   !Text         -- ^ Constant
              -- ^ Generated input hole, first argument is the scoping level
              | N   !Int          -- ^ Name hole
              | L   !Int          -- ^ Literal hole
+             | LC  !Int
+             -- ^ Assert that argument is a literal, but treat it like a normal
+             -- argument otherwise
              | Var [Element] !Int    --
              | Sym !Text !Int    -- ^ Symbol hole
              | Typ !(Maybe Int)  -- ^ Type declaration hole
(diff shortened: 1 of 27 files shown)


clash-lib/src/Clash/Netlist/BlackBox/Types.hs
```diff
@@ -83,6 +83,9 @@ data Element = C   !Text         -- ^ Constant
              -- ^ Generated input hole, first argument is the scoping level
              | N   !Int          -- ^ Name hole
              | L   !Int          -- ^ Literal hole
+             | LC  !Int
+             -- ^ Assert that argument is a literal, but treat it like a normal
```
martijnbastiaan: Where is this asserted?

martijnbastiaan: Ah, I guessed as much but I was thrown off by the docs:

```
-- | Determine if the number of normal/literal/function inputs of a blackbox
-- context at least matches the number of argument that is expected by the
-- template.
```

Maybe it's a good idea to add a small comment? Also: maybe add a `TODO` to get a more explicit error out instead of a general "can't match template".

martijnbastiaan:
The following code doesn't actually produce multiple files, even though I'd expect it would:

```haskell
module Test where

import Clash.Prelude

add' = add
{-# NOINLINE add' #-}

topEntity
  :: Clock  System Source
  -> Reset  System Asynchronous
  -> Signal System (Unsigned 17, Unsigned 17)
topEntity clk rst =
  pure
    ( add' (22 :: Unsigned 16) (0 :: Unsigned 0)
    , add' (0 :: Unsigned 0)   (22 :: Unsigned 16) )
{-# NOINLINE topEntity #-}
```

christiaanb:
@martijnbastiaan The example you posted now results in 3 components instead of 1: 2 versions of `add` and 1 `topEntity`

martijnbastiaan:
Closing in favor of #498.

PR #433, opened 2018-11-26
Further CI tweaks

leonschoorl:


leonschoorl:
Succes  on Gitlab, failure on Circle for the same test :frowning:
I'll investigate..

leonschoorl:
Gitlab initally ran the test on Friday when I pushed the branch and used `cabal-install version 2.4.0.0`.
CircleCI test ran on Monday after I submitted this pull request and used `cabal-install version 2.4.1.0`.

cabal-install-2.4.1.0 was released on Monday and changed its `.ghc.environment....` file generation behaviour.
By default it now only generates them on ghc >= 8.4.4
But the doctest fails to find the required modules without an environment file.

leonschoorl:
@martijnbastiaan It's currently using trusty because that was the only version for which I could find prebuild GHDL binaries. 
But we could build it ourself, just like we already do with iverilog.

martijnbastiaan:
I think we should build and publish it on our own launchpad, to be frank. It's quite tedious to have to jump through hoops on new machines. I think our users would agree.

christiaanb:
Note that we need GHDL 0.31, not newer, because those are broken/regressed wrt the VHDL clash generates.

martijnbastiaan:
Duly noted. In that case we can just copy the build instructions from the existing launchpad and tweak it somewhat.

martijnbastiaan:
@christiaanb any specific reason btw? Is it by any chance related to verilator issue?

christiaanb:
No, it fails parsing really dumb stuff, like when you add superfluous parenthesis around an expression it errors.

PR #435, opened 2018-11-27
Use `seqX` to define `sample`

JonFowler:
Current `sample` often throws an `X` error further than needed:
```
> showX <$> sampleN 5 ((5, errorX "") :- (1 :: Int,()) :- pure (errorX ""))
["X","(1,())","X","X","X"]
```
New version keeps the error in the field of the first element:
```
> showX <$> sampleN 5 ((5, errorX "") :- (1 :: Int,()) :- pure (errorX ""))
["(5,X)","(1,())","X","X","X"]
```


JonFowler:
:+1: Done.

PR #437, opened 2018-11-27
Improve support for zero-width constructs

martijnbastiaan:
This PR adds support for zero-width constructs for `BitVector`, `Signed`, and `Unsigned`.

Includes some (very minor) fixes.

### DIFF
--- a/testsuite/Test/Tasty/Clash.hs
+++ b/testsuite/Test/Tasty/Clash.hs
@@ -256,6 +256,9 @@ ghdlMake path modName subdirs libs entName =
   (testName, test)
   where
     args = concat [ ["-m"]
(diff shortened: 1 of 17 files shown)


testsuite/Test/Tasty/Clash.hs
```diff
@@ -256,6 +255,9 @@ ghdlMake path modName subdirs libs entName =
   (testName, test)
   where
     args = concat [ ["-m"]
+               -- TODO: Detect GHDL version
+               -- Enable flags when running newer versions of GHDL.
```
christiaanb: Doesn't have to do with GHDL, but with the version of the (GCC) linker.

martijnbastiaan:
@christiaanb does that comment conclude your review?

PR #447, opened 2018-12-12
Retry GitLab CI run on runner system failure

martijnbastiaan:
This restarts a GitLab CI job whenever there is a docker failure,
power loss, etc. This will allow us to restart our own CI runners,
without having to worry about our test cases failing.


leonschoorl:
Interestingly GHC is able to look through the virtualisation and see that the that the CPU has 64 threads:
```Default number of threads: 64```

I like hiding a lot of the test output, but now it also doesn't show any warnings anymore.

**Comment by martijnbastiaan** [comment]
The test times on CI now look pretty good. Some further improvements:

* Compile Clash itself (not its dependencies) with `-O0`. This reduces compile times on my machine with 50%. (1m44 vs 3m35, both without haddock). While the testsuite would run slower, my guess is that this would by offset by the fact that compiling is a quite sequential process while running the testsuite is "perfectly" parallel. I can't test it right now, as Clash doesn't work with `-O0` right now.

* Run Haddock in parallel to the tests. So it would look something like:

```
new-build
haddock &
testsuite
wait
test-haddock
```

* Make the testsuite lookup the `clash` binary, instead of using `cabal new-run` every time. Something like: `find dist-newstyle -type f -executable -name clash`.

**Comment by leonschoorl** [comment]
Part of why clash fails when build with `-O0` is that `-O0` implies `-fomit-interface-pragmas`.
And when the prelude is compiled with that GHC discards all the `NOINLINE`'s on our primitives.

**Comment by martijnbastiaan** [comment]
We should figure out what flags it needs to work then :).

It just hit me that the testsuite can simply call:

```cabal new-install clash --symlink-bindir=$testsuitetmp/bin```

and use the `clash` linked there.

**Comment by leonschoorl** [comment]
I could get the testsuite to pass with `-O1` on prelude and ghc-typelits-knownnat and `-O0` on the rest.

cabal.project(.local):
```
optimization: 0
package clash-prelude
  optimization: 1
package ghc-typelits-knownnat
  optimization: 1
```

martijnbastiaan:
Brilliant. I think it's spending most time in `clash-lib` and `clash-ghc` anyway!

leonschoorl:
We're probably tripping some memory limit on Circle.

Did some memory usage measurement on my machine (using `/usr/bin/time --format "%M"`) on some of the tasks that seem to be failing:

test      | memory usage
--------- | ------------------------
haddock  | 1.8 GB
doctests | 1.8 GB
TFold    | 2.7 GB
DFold    | 2.6 GB
DTFold   | 2.7 GB

So if we run haddock and doctest in parallel with 2 instances of clash compiling TFold (for different HDLs) we need about 9GB of RAM.

I can't find clear specification for the Circle machines, but [here](https://circleci.com/docs/2.0/executor-types/#using-machine) they're talking about 8GB.

martijnbastiaan:
Right. Well, reliability trumps speed, so I guess we'll have to disable it again. Or detect whether we're running Circle, but that seems a bit hackey..

martijnbastiaan:
It looks like disabling optimizations results in Clash using a large amount of memory :(. I'll drop the offending commits.

martijnbastiaan:
@christiaanb This is it for this PR. Could you check if you see any glaring errors, and press merge if not?

PR #457, opened 2018-12-21
Inst parameters

### LINKED ISSUE #413: generic map for netlist declarations
Currently `InstDecl` in `Clash.Netlist.Types.Declaration` doesn't have a field for the "generic map" which means it can't be used to generate code that instantiates a generic component.

I propose an extra field is added to `InstDecl` of type `[(Expr, Expr)]` (or whatever this should be for generic map). So that `inst` will generate:

```
my_component : component my_component
  generic map (
   param => val
  )
  port map (...);
```

for the VHDL backend and the equivalent `#(parameters ...)` syntax for the Verilog backend.

cchalmers:
Closes #413

### DIFF
--- a/clash-lib/src/Clash/Netlist/Types.hs
+++ b/clash-lib/src/Clash/Netlist/Types.hs
@@ -135,6 +135,8 @@ data HWType
   -- that vector.
   | String
   -- ^ String type
+  | Integer
+  -- ^ Integer type
   | Bool
   -- ^ Boolean type
   | Bit
--- a/clash-lib/src/Clash/Netlist/Types.hs
+++ b/clash-lib/src/Clash/Netlist/Types.hs
@@ -22,7 +22,7 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 
 module Clash.Netlist.Types
-  ( Declaration (..,NetDecl)
+  ( Declaration (..,NetDecl, InstDecl)
   , module Clash.Netlist.Types
   )
 where
@@ -208,7 +208,7 @@ data Declaration
   -- * Type of the scrutinee
   --
   -- * List of: (Maybe expression scrutinized expression is compared with,RHS of alternative)
-  | InstDecl EntityOrComponent (Maybe Identifier) !Identifier !Identifier [(Expr,PortDirection,HWType,Expr)]
+  | InstDecl' EntityOrComponent (Maybe Identifier) !Identifier !Identifier [(Expr,HWType,Expr)] [(Expr,PortDirection,HWType,Expr)]
   -- ^ Instantiation of another component
   | BlackBoxD
       -- Primitive name:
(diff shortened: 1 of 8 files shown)


clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -206,7 +208,7 @@ data Declaration
   -- * Type of the scrutinee
   --
   -- * List of: (Maybe expression scrutinized expression is compared with,RHS of alternative)
-  | InstDecl EntityOrComponent (Maybe Identifier) !Identifier !Identifier [(Expr,PortDirection,HWType,Expr)]
+  | InstDecl' EntityOrComponent (Maybe Identifier) !Identifier !Identifier [(Expr,HWType,Expr)] [(Expr,PortDirection,HWType,Expr)]
```
martijnbastiaan: Could you add documentation?

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -238,6 +240,15 @@ data WireOrReg = Wire | Reg
...
 instance NFData WireOrReg
 
+pattern InstDecl
+  :: EntityOrComponent
+  -> Maybe Identifier
+  -> Identifier
+  -> Identifier
+  -> [(Expr,PortDirection,HWType,Expr)]
+  -> Declaration
+pattern InstDecl eoc mi i2 i3 as = InstDecl' eoc mi i2 i3 [] as
```
martijnbastiaan: Is there a good reason to go with patterns here? As far as I can see, it adds an layer of abstraction with very little benefit. `InstDecl` is only used in ~10 locations, maybe you could just change those lines to include `[]`?

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -379,6 +379,7 @@ typeSize :: HWType
          -> Int
 typeSize (Void {}) = 0
 typeSize String = 1
+typeSize Integer = 1
```
martijnbastiaan: I'm confused. Shouldn't this be whatever integer width Clash is provided through `-fclash-intwidth`? (Also see other comment)

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -135,6 +135,8 @@ data HWType
   -- that vector.
   | String
   -- ^ String type
+  | Integer
+  -- ^ Integer type
```
martijnbastiaan: Why do we need `Integer` as a hwtype?

christiaanb: I guess the documentation should be extended that this `Integer` is for internal use only, and that it is not, nor should ever be used as the type representing the Haskell `Integer` type.

clash-lib/src/Clash/Netlist/Util.hs
christiaanb: Given that's internal use only, and doesn't actually a represent a Haskell `Int` or `Integer`, `0` is probably the better value. For historical reasons, the `typeSize` of `String` is `1`, but it should probably be made `0` as well.

This should be added as a comment.

clash-lib/src/Clash/Netlist/Types.hs
cchalmers: I did it this way because it's similar to how `InstDecl` is done. Also, even though this stuff is "internal", making this change would likely break code for anyone using the haskell blackbox api, which I wanted to avoid. But I'm not too fussed either way.

PR #458, opened 2018-12-25
 Account for zero-width constructs in `mkFunInput` 

martijnbastiaan:
`mkFunInput` doesn't account for zero-width constructs (correctly). This PR extends `coreTypeToHWType` by making it return what arguments it removes. This makes the following test case compile:

```
module Test where

import Clash.Prelude

topEntity
  :: Vec 4 (Int,Int)
  -> Vec 4 ((Int,Int),())
topEntity xs = zipWith (,) xs (repeat ())
```

`coreType*` functions now always return void removal information, thus forcing the caller to explicitly deal with it.


martijnbastiaan:
Per Christiaan:

* `CustomSP` / `CustomSum`
* Reconsider portnames in combination with vectors/rtrees
* Version of `coreTo*` that strips by default

PR #465, opened 2019-01-17
Add support for Vec-like GADTs

martijnbastiaan:
**TODO**: confirm that `InScopeSet`s are correctly extended.

This commit adds support for Vec-like GADT constructs, as well as
simpler ones. To understand what changed, let's consider a small
example:

    data T a where
      SBool :: Bool -> T Bool
      SInt  :: Int  -> T Int

I.e., a sum-of-products with two constructors: `SBool` and `SInt`. When
`T a` is constructed with the constructor `SBool`, we're sure that
`a ~ Bool`. Likewise, whenever `T a` is constructed with `SInt`, we know
that `a ~ Int`. The inverse is true too: whenever we match on each of
the constructors we gain type information. Consider:

    f :: T a -> a
    f t =
      case t of
        SBool b -> not b
        SInt i  -> i + 1

At first it might seem that GHC would reject this program: the two
alternatives yield wildly different types. But, as alluded to earlier,
the pattern match introduces some extra information:

    f :: forall a . T a -> a
    f t =
      case t of
        SBool {dt: a ~ Bool} b -> (not (b ▶dt)) ▶sym dt
        SInt  {dt: a ~ Int}  i -> ((i ▶dt) + 1) ▶sym dt

GHC inserts a proof (inside curly braces) that `a ~ Bool` when matching
on `SBool`, and a similar proof for `SInt`. On the right hand side of
each alternative it casts the polymorphic `a` to `Bool`/`Int` with
`b ▶dt`/`i ▶dt` respectively. Thanks to the given proof (which was
obtained during construction), this typechecks. Because the type of the
case has to be `a`, a symmetric cast is applied with `▶sym dt`.

So how do we use this information to compile seemingly polymorphic
functions? Well, consider what happens when `f` is specialized such that
it is monomorphic (like every sound Clash circuit). That is:

    f @int

The function definition now looks like:

    f :: forall Int . T Int -> Int
    f t =
      case t of
        SBool {dt: Int ~ Bool} b -> (not (b ▶dt)) ▶sym dt
        SInt  {dt: Int ~ Int}  i -> ((i ▶dt) + 1) ▶sym dt

Clearly, `Int ~ Bool` is absurd! We can therefore remove it (see
transformation `caseElemNonReachable`) leaving us with:

    f :: forall Int . T Int -> Int
    f t =
      case t of
        SInt  {dt: Int ~ Int}  i -> ((i ▶dt) + 1) ▶sym dt

..a case with a single alternative.

This is not enough to support `Vec`-like GADTs though. Why not? Well,
let's look at the definition of `Vec`:

    data Vec :: Nat -> Type -> Type where
      Nil  :: Vec 0 a
      (:>) :: a -> Vec n a -> Vec (n + 1) a

Similar to our previous function `f`, we can build a function pattern
matching on `Vec`:

    g :: Vec n Bool -> Bool
    g v =
      case v of
        Nil       -> False
        Cons x xs -> True

What would the annotated version look like? Well..

    g :: forall n . Vec n Bool -> Bool
    g v =
      case v of
        Nil  {dt: n ~ 0}                                     -> False
        Cons {n1:Nat} {dt: n~n1+1} (x :: a) (xs :: Vec n1 a) -> True

Alright, so let's do the same trick again: specialize `g`.

    g @0

By filling in the `n`s we get:

    g :: Vec 0 Bool -> Bool
    g v =
      case v of
        Nil  {dt: 0 ~ 0}                                     -> False
        Cons {n1:Nat} {dt: 0~n1+1} (x :: a) (xs :: Vec n1 a) -> True

So what gives? The alternative `Cons` is not _clearly_ absurd now. Of
course, intuition dictates that if `0 ~ n1+1` then `n1 ~ 0-1 ~ -1`. As
we're operating in the domain of naturals, this _is_ absurd! Our
intuition relied the fact that `-` is the inverse of `+`. This
property (+injectivity) is not being tracked by GHC/Haskell/Clash
though, so we can't really do anything in general. This commit adds
hardcoded logic for `+`, see `elemExistentials`.

Future work: make "polymorphic" GADTs synthesizable by treating them
(in as sense) as sum-of-product types.

### DIFF
--- a/clash-lib/src/Clash/Core/Util.hs
+++ b/clash-lib/src/Clash/Core/Util.hs
@@ -141,7 +141,7 @@ piResultTyMaybe
   -> Type
   -> Maybe Type
 piResultTyMaybe m ty arg
-  | Just ty' <- coreView m ty
+  | Just ty' <- coreView1 m ty
   = piResultTyMaybe m ty' arg
   | FunTy _ res <- tyView ty
   = Just res
@@ -181,7 +181,7 @@ piResultTys
   -> Type
 piResultTys _ ty [] = ty
 piResultTys m ty origArgs@(arg:args)
-  | Just ty' <- coreView m ty
+  | Just ty' <- coreView1 m ty
   = piResultTys m ty' origArgs
   | FunTy _ res <- tyView ty
   = piResultTys m res args
@@ -194,7 +194,7 @@ piResultTys m ty origArgs@(arg:args)
 
   go env ty' [] = substTy (mkTvSubst inScope env) ty'
   go env ty' allArgs@(arg':args')
-    | Just ty'' <- coreView m ty'
+    | Just ty'' <- coreView1 m ty'
     = go env ty'' allArgs
     | FunTy _ res <- tyView ty'
     = go env res args'
@@ -581,7 +581,7 @@ isClockOrReset
   :: TyConMap
   -> Type
   -> Bool
-isClockOrReset m (coreView m -> Just ty) = isClockOrReset m ty
+isClockOrReset m (coreView1 m -> Just ty) = isClockOrReset m ty
 isClockOrReset _ (tyView -> TyConApp tcNm _) = case nameOcc tcNm of
   "Clash.Signal.Internal.Clock" -> True
   "Clash.Signal.Internal.Reset" -> True
@@ -591,7 +591,7 @@ isClockOrReset _ _ = False
 tyNatSize :: TyConMap
           -> Type
           -> Except String Integer
-tyNatSize m (coreView m -> Just ty) = tyNatSize m ty
+tyNatSize m (coreView1 m -> Just ty) = tyNatSize m ty
 tyNatSize _ (LitTy (NumTy i))       = return i
 tyNatSize _ ty = throwE $ $(curLoc) ++ "Cannot reduce to an integer:\n" ++ showDoc (ppr ty)
 
--- a/clash-lib/src/Clash/Core/Util.hs
+++ b/clash-lib/src/Clash/Core/Util.hs
@@ -7,6 +7,7 @@
 -}
 
 {-# LANGUAGE CPP               #-}
+{-# LANGUAGE NamedFieldPuns    #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell   #-}
 {-# LANGUAGE ViewPatterns      #-}
@@ -19,8 +20,9 @@ import Control.Monad.Trans.Except              (Except, throwE)
 import           Data.Coerce                   (coerce)
 import qualified Data.IntSet                   as IntSet
 import qualified Data.HashSet                  as HashSet
-import Data.List                               (foldl', mapAccumR)
-import Data.Maybe                              (fromJust, mapMaybe)
+import Data.List
+  (foldl', mapAccumR, elemIndices)
+import Data.Maybe                              (fromJust, mapMaybe, catMaybes)
 import qualified Data.Text                     as T
 import           Data.Text.Prettyprint.Doc     (line)
 #if !MIN_VERSION_base(4,11,0)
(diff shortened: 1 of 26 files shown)


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -240,6 +245,77 @@ caseLet _ (Case (Letrec xes e) ty alts) =
...
+--    g :: STy Int -> Int
+--    g = f
+--
+-- @f@ is always specialized on @STy Int@. The SBool alternatives are therefore
+-- unreachable.
+caseElemNonReachable :: NormRewrite
+caseElemNonReachable _ case0@(Case scrut altsTy alts0) = do
+  tcm <- Lens.view tcCache
+
+  if any (isAbsurdAlt tcm) alts0 then do
```
christiaanb:
```
(altsAbsurd,altsNormal) = parttiion (isAbsurdAlt tcm) alts0
case altsAbsurd of
  [] -> return case0
  _ -> do case1 <- caseOneAlt (Case scrut altsTy altsNormal)
          changed case1
```

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1747,3 +1747,11 @@ splitGatedClock baseNm (Clock nm rt Gated) = do
     hwtys = [Clock nm rt Source,Bool]
     partNms = ["_clk","_clken"]
 splitGatedClock _ ty = error $ $(curLoc) ++ "splitGatedClock can't split: " ++ show ty
+
+bindsExistentials
```
christiaanb: Add comments

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -628,3 +783,61 @@ mkUniqInternalId (supply,inScope) (nm, ty) =
...
+-- | Given a DataCon and a list of types, the type variables of the DataCon
+-- type are substituted for the list of types. The argument types are returned.
+--
+-- The list of types should be equal to the number of type variables, otherwise
+-- @Nothing@ is returned.
+dataConInstArgTys :: DataCon -> [Type] -> Maybe [Type]
+dataConInstArgTys (MkData { dcArgTys, dcUnivTyVars, dcExtTyVars }) inst_tys =
+  let tyvars = dcUnivTyVars ++ dcExtTyVars in
+  if length tyvars == length inst_tys then
+    -- Notice it is safe to use substTyWith here, as ALL variables are replaced
```
christiaanb: It's safe to use subsTyWith here because:

1. The [Note] subst invaration says: FVs should be inScopeSet or in the range of the subst
2. The FVs of dcArgTys can only be tyvars
3. The range of the substitution is tyvars, thus satisfying the invariant of 1.

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -57,6 +60,158 @@ type Gamma = VarEnv Type
...
+  | AbsurdSolution
+  -- ^ A solution was found, but it involved negative naturals.
+  | NoSolution
+  -- ^ Given type wasn't an equation, or it was unsolvable.
+    deriving (Show)
+
+-- | Solve given equations and return the first non-absurd solution (if any)
+solveFirstNonAbsurd
+  :: [(Type, Type)]
+  -> Maybe (TyVar, Integer)
```
christiaanb: Maybe (TyVar, Type)

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -57,6 +60,158 @@ type Gamma = VarEnv Type
...
+    _ ->
+      Nothing
+ where
+  lhsLit x                 (LitTy (NumTy n)) = Just (n, x)
+  lhsLit (LitTy (NumTy n)) y                 = Just (n, y)
+  lhsLit _                 _                 = Nothing
+
+-- | Data type that indicates what kind of solution (if any) was found
+data AddSolution
+  = Solution (TyVar, Integer)
```
christiaanb: Solution (TyVar, Type)

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -628,3 +783,61 @@ mkUniqInternalId (supply,inScope) (nm, ty) =
...
+    (map (substTy subst) dcArgTys)
+
+ where
+  go
+    :: [TyVar]
+    -- ^ Existentials
+    -> [Type]
+    -- ^ Type arguments
+    -> Maybe [Type]
+    -- ^ Maybe (solved existentials, types of non-existentials)
```
martijnbastiaan: ^ Comment lies ^

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -57,6 +60,158 @@ type Gamma = VarEnv Type
...
+  -- ^ Variables in scope
+  -> [TyVar]
+  -- ^ List of existentials to apply the substitution for
+  -> (TyVar, Type)
+  -- ^ Substitution
+  -> [TyVar]
+substInExistentials is exts subst@(typeVar, _type) =
+  -- TODO: Is is actually possible that existentials shadow each other? If they
+  -- TODO: can't, we can remove this function
+  let eqOnName = (==) `on` varName in
```
christiaanb: you don't need this, variables have the same unique as their names.

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -628,3 +783,61 @@ mkUniqInternalId (supply,inScope) (nm, ty) =
...
+-- hence the extra argument @TyConMap@. WARNING: It will return the types
+-- of non-existentials only
+dataConInstArgTysE
+  :: HasCallStack
+  => InScopeSet
+  -> TyConMap
+  -> DataCon
+  -> [Type]
+  -> Maybe [Type]
+dataConInstArgTysE is0 tcm (MkData { dcArgTys, dcExtTyVars, dcUnivTyVars }) inst_tys =
```
christiaanb: Just use the Maybe monad,  so you can use the `let` without the `in`

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -57,6 +60,158 @@ type Gamma = VarEnv Type
...
+  -> [TyVar]
+substGlobalsInExistentials is exts substs0 = result
+  -- TODO: Is is actually possible that existentials shadow each other? If they
+  -- TODO: can't, we can remove this function
+  where
+    iss     = scanl extendInScopeSet is exts
+    substs1 = map (\is_ -> extendTvSubstList (mkSubst is_) substs0) iss
+    result  = zipWith substTyInVar substs1 exts
+
+substInExistentials
```
christiaanb: Add a top-level comment

martijnbastiaan:
Also: I should add tests to see if unsupported constructions are still failing properly (instead of generating broken HDL).

PR #493, opened 2019-02-01
Make clockResetError not error unnecessarily

cchalmers:


cchalmers:
For the record, I’d prefer if this check was just removed. It often creates annoying false warnings.

christiaanb:
@cchalmers I agree, the check should just be disabled for now, until we have a better implementation.

martijnbastiaan:
For prosperity: 

```
in principle you shouldn't define a function with two identical clocks
but a polymorphic function with two different clocks, but instantiated to the same, is okay
so the way `clockResetErrors` should be implemented is as a traversal of the entire hierarchy of functions
before normalisation starts
not as a post-normalisation check of a function
but that's just for clocks
the reset one is different
because only two asynchronous reset lines are bad
so two synchronous lines is jsut fine (edited) 
but that information is only known after normalisation, because types need to be propagated (edited) 
```

martijnbastiaan:
I've removed the check altogether. We should revisit this in the future.

Issue #499, opened 2019-02-15
Add `force` and `tryForce`


martijnbastiaan:
I cases such as:

```
topEntity = map (+1) (1 :> 2 :> 3 :> Nil)
```

Users might expect Clash to unroll the `topEntity` and insert the equivalent of:

```
topEntity = (2 :> 3 :> 4 :> Nil)
```

in the HDL. However, Clash considers this "too expensive" and will simply insert the definition of map. 

We should give users the ability to force Clash to reduce an expression to constant, if possible. This could help with readability, or help working around not-smart-enough HDL tooling in certain cases. A first idea is to add the following functions:

```haskell
-- | Reduce given expression to a constant in HDL. Errors if it could not reduce the
-- given expression to a constant.
force :: a -> a
force = id

-- | Try to reduce given expression to a constant in HDL. Will not error if
-- this does not succeed.
tryForce :: a -> a
tryForce = id
```

PR #501, opened 2019-02-19
Add test for (Num) constant folding

leonschoorl:
And fix various issues encountered along the way.

----
Evaluation doesn't work for:
* ~~`countLeadingZeros` and `countTrailingZeros` on any Sized Clash type
  These are implemented in terms of maps and folds which don't evaluate (well).
  An option could be to make them primitives, just like they are for general Haskell numbers.~~
  edit @martijnbastiaan : This has been implemented in https://github.com/clash-lang/clash-compiler/pull/501/commits/44f4dc7b6b88ed9814ab18bb179836aeb06d2298

* `popCount @Integer`
  And there is no blackbox either.
  There isn't really a sensible implementation possible.
  In Haskell `popCount (-3::Integer)` is  `-2`, while popCount is documentated as: "Return the number of set bits in the argument"

* Bits Natural

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 aliases:
   - &env
     TERM: xterm-color
-    THREADS: 2
+    THREADS: 1
     CABAL_JOBS: 1
 
   - &submodules
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
@@ -3282,8 +3282,9 @@ sizedLiteral szCon val = case val of
 bitLiterals
   :: [Value]
   -> [(Integer,Integer)]
-bitLiterals = typedLiterals' go
+bitLiterals = map normalizeBit . typedLiterals' go
  where
+  normalizeBit (msk,v) = (msk .&. 1, v .&. 1)
   go val = case val of
     PrimVal nm _ _ [Lit (IntegerLiteral m), Lit (IntegerLiteral i)]
       | nm == "Clash.Sized.Internal.BitVector.fromInteger##"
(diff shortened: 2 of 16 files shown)


martijnbastiaan:
@leonschoorl Could you take a look at https://github.com/clash-lang/clash-compiler/pull/501/commits/44f4dc7b6b88ed9814ab18bb179836aeb06d2298 ? If it makes sense to you, I think we can merge this PR.

martijnbastiaan:
@christiaanb Could you also take a look at https://github.com/clash-lang/clash-compiler/commit/44f4dc7b6b88ed9814ab18bb179836aeb06d2298? We're not sure if this is the correct approach. 

We want:

* The evaluator to pick up on `count{Leading,Trailing}Zeros#` so we can constant fold it

We don't want:

* To implement a blackbox for it

We did this

* by marking the functions as primitive with `primType"  : "Function"`

We're not sure what (unforeseen) consequences our solution has, even though the testsuite passes.

martijnbastiaan:
Well that was the question I suppose. But I see that Clash "silently" adds `NO_TRANSLATION_FOR` to the HDL when `primtype: Function` is there. So to rephrase the question: is there anyway we can add evaluator rules for non-primitives builtin to Clash? We know how to constant fold `countLeadingZeros`, but we're also fine with Clash translating it if it's not used on constants.

christiaanb:
Implement:

```
force# :: a -> a
force# x = x
```

with primitive template `~ARG[0]`

and an evaluator rule that is simply the identity as well.

then add it to the existing implement of CLZ and CTZ. Now, because the `foldr` and `foldl` are an argument to a primitive, so the evaluator will force them to WHNF.

martijnbastiaan:
Clever! We should probably take https://github.com/clash-lang/clash-compiler/issues/499 into account.

christiaanb:
Actually... I'm surprised this doesn't happen already, I mean... basically everything in https://github.com/clash-lang/clash-compiler/blob/580b058618ae4b51f32deb0b69b99ee0698fc430/clash-prelude/src/Clash/Sized/Internal/BitVector.hs#L612-L616 is a primitive. So adding a `force#` shouldn't help.

martijnbastiaan:
`inlineSmall` replaces the call with it's definition if I remember correctly, way before the evaluator gets a chance to look at it.

christiaanb:
Can we take out https://github.com/clash-lang/clash-compiler/pull/501/commits/44f4dc7b6b88ed9814ab18bb179836aeb06d2298 of this PR and then merge it. Leaving https://github.com/clash-lang/clash-compiler/pull/501/commits/44f4dc7b6b88ed9814ab18bb179836aeb06d2298 for a seperate PR.

martijnbastiaan:
yes we can!

PR #504, opened 2019-02-21
CircleCI: run single-threaded to try to avoid memory limits

leonschoorl:


martijnbastiaan:
Do we need to include a `-j$THREADS` here:

https://github.com/clash-lang/clash-compiler/blob/master/.ci/build.sh#L14

as well to have cabal build with X threads package based?

leonschoorl:
Good point

But we probably don't what to set both inter and intra package threading to `THREADS` or we always get `THREADS^2` threads.

leonschoorl:
clash-testsuite detects 36 cores:
```
+ cabal new-run -- clash-testsuite -j1
Up to date
Running in /root/project/.clash-test-tmp/clash-test--35e9d29e9ebb0dce
Making sure Clash is compiled.. 
Default number of threads: 36
```

If cabal uses the same logic core detection logic, it might be trying to build 36 packages at the same time.

martijnbastiaan:
Maybe always build dependencies with 1 thread? They can be cached and shouldn't change too often.

martijnbastiaan:
Yeah, the testsuite uses a standard Haskell function which looks in `/proc` instead of looking at what cgroups sets.

leonschoorl:
That's what was going on
Because we weren't configuring cabal parallelism, cabal was auto-detecting that (incorrectly).
And starting way to many builds in parallel.

You can see here that cabal is `Starting` to build 8 packages before the first one is `Completed`
https://circleci.com/gh/clash-lang/clash-compiler/1227

martijnbastiaan:
LGTM. Squash and merge?

PR #515, opened 2019-03-01
Class selectors for newtype dicts are id

christiaanb:
`mkSelectorCase`, which is used for most dictionaries, fails on newtype dictionaries because it looks through newtypes.

You get newtype dictionaries for single method classes without superclasses.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
@@ -46,7 +46,7 @@ import           Clash.Annotations.Primitive (HDL)
 
 import           Clash.Core.Term         (Term (..))
 import           Clash.Core.Type         (Type (..), TypeView (..), mkFunTy, splitFunForallTy, tyView)
-import           Clash.Core.TyCon        (TyConMap, TyConName)
+import           Clash.Core.TyCon        (TyConMap, TyConName, isNewTypeTc)
 import           Clash.Core.TysPrim      (tysPrimMap)
 import           Clash.Core.Util         (mkLams, mkTyLams)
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
I've tried to reproduce this issue by creating a typeclass with a single method, but I can't get Clash to stumble. How would you need to use the typeclass in order to trigger a failure?

christiaanb:
Using `Data.Default` triggered it for me. Make sure you NOINLINE the polymorphic function using class method so that GHC doesn’t specialize on the class constraint.

martijnbastiaan:
Yes, that's what I thought. I figured this would trigger it:

```haskell
module Test2 where

import Clash.Prelude
import Data.Default

data ABC a
 = A Int a
 | B Int

ffffffffffff :: Default a => Int -> ABC a
ffffffffffff 0 = B 7
ffffffffffff n = A (n + 5) def
{-# NOINLINE ffffffffffff #-}
```

```haskell
module Test where

import Test2
import Data.Default
import Clash.Prelude


topEntity :: Int -> (Int, Int)
topEntity n =
  case ffffffffffff n of
    A 5 (0 :: Int, 1 :: Int) -> (1, 2)
    B 7                      -> (2, 2)
    _                        -> error "What!"
```

GHC doesn't specialize it on the class constraint (I think?), as I see:

```
topEntity
  = \ (n_aqWR :: Int) ->
      case ffffffffffff
             @ (Int, Int)
             (Test.topEntity12
              `cast` (Sym (Data.Default.Class.N:Default[0] <(Int, Int)>_N)
                      :: ((Int, Int) :: *) ~R# (Default (Int, Int) :: Constraint)))
             n_aqWR
```

But to no avail..

Issue #314, opened 2018-05-02
Hard-coded -dynamic flag breaks build on openSUSE Linux


peti:
I just had to [patch the build of `clash-ghc` for openSUSE Linux/x86-64 with GHC 8.4.2](https://github.com/opensuse-haskell/configuration/commit/8843294a252be6e912d51492c709f6399205bd95), because the hard-coded use of `-dynamic` in the Cabal file broke our post-processing checks:

~~~~
[  412s] calling /usr/lib/rpm/brp-suse.d/brp-35-rpath
[  412s] ERROR: RPATH "/home/abuild/rpmbuild/BUILD/clash-ghc-0.99/dist/build" on /home/abuild/rpmbuild/BUILDROOT/clash-ghc-0.99-2.1.x86_64/usr/bin/clash is not allowed
[  413s] ERROR: RPATH "/home/abuild/rpmbuild/BUILD/clash-ghc-0.99/dist/build" on /home/abuild/rpmbuild/BUILDROOT/clash-ghc-0.99-2.1.x86_64/usr/bin/clashi is not allowed
~~~~

What happens is that `ghc` makes bad decisions about setting certain RPATH entries in the generated executables. It ought to be *Cabal* making those decisions: setting the compiler flag `-dynamic` is not the proper way to enable dynamic linking.

Is there a particular reason why this flag is hard-coded in the Cabal file?

christiaanb:
The issue is that if the executables are not dynamically linked we incur a startup time of around 3 seconds.

christiaanb:
I'll remove the hard-coded choice in the Cabal flag, and inform users who build Clash from source/cabal should use the `--enable-executable-dynamic` flag. But how would I inform distributors to build Clash using dynamic linking? Is it even possible to have dynamically linked Haskell executables in OpenSUSE taking the RPATH-mentioned issue into account?

peti:
Yes, dynamically linked binaries work fine on openSUSE if it's enabled through Cabal.

If you point out the performance gain in the readme or in the build instructions, then I'm sure people who distribute this package will notice it. And even if they don't, then their users will eventually. :-)

Incidentally, many distributions (like openSUSE and NixOS) disable dynamically linked binaries by default because these binaries tend to start up much slower that statically linked ones. I am surprised that it appears to be the other way round with `clash`.

PR #518, opened 2019-03-04
Small patches to allow building Debian packages

martijnbastiaan:
* Ubuntu 18.10 packages `exceptions` 0.10, so we need it to build binary package
* ~~Enable `-dynamic` by default, as we need it for speedy builds (can be disabled by `-f-dynamic`)~~ Add flag to enable `-dynamic`. Default is false.

### DIFF
--- a/clash-ghc/clash-ghc.cabal
+++ b/clash-ghc/clash-ghc.cabal
@@ -55,10 +55,18 @@ source-repository head
   type: git
   location: https://github.com/clash-lang/clash-compiler.git
 
+flag dynamic
+  description:
+    Build Clash binaries with GHC flag `-dynamic`. This flag should only be used for packaging purposes. Installations using cabal should use `--enable-executable-dynamic`!
+  default: False
+  manual: True
+
 executable clash
   Main-Is:            src-ghc/Batch.hs
   Build-Depends:      base, clash-ghc
   GHC-Options:        -Wall
+  if flag(dynamic)
+    GHC-Options: -dynamic
   extra-libraries:    pthread
   default-language:   Haskell2010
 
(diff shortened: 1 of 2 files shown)


clash-ghc/clash-ghc.cabal
```diff
@@ -55,17 +55,27 @@ source-repository head
...
+    Build the clash binaries with -dynamic
+  default: True
+  manual: True
+
 executable clash
   Main-Is:            src-ghc/Batch.hs
   Build-Depends:      base, clash-ghc
   GHC-Options:        -Wall
+  if flag(dynamic)
+    GHC-Options: -dynamic
```
christiaanb: No, cabal should handle dynamic binary handling using `--enable-executable-dynamic`, it has better RPATH handling than GHC. See also: https://github.com/clash-lang/clash-compiler/issues/314

martijnbastiaan: A few things:

1. The Debian build scripts use fakeroot and (probably?) won't generate bad RPATH directories the check in #314 (https://github.com/openSUSE/brp-check-suse/blob/master/brp-35-rpath) checks for.
2. The Debian build scripts don't support passing a flag such as `--enable-executable-dynamic` to cabal, it only allows boolean flags like the one above. Trying to "fix" this tool seems a waste of time if we can just use ghc-options IMO.
3. Either way, the problem in #314 is that it's a hardcoded flag. It doesn't have to be, as this patch proves.
4. We could set the default to false.
5. We could document the flag to indicate "normal" users probably don't want to touch it.
6. I'm not sure why we even care about better RPATH handling. Does it have any tangible benefits?

leonschoorl: What if we rename the flag to `force-dynamic` with a default state to `False`?

christiaanb: I can live with a `False`  default, and mention in the flag description that his flag is only used for packaging, and that uses should use `--enable-executable-dynamic`.

martijnbastiaan: Perfect!

Issue #516, opened 2019-03-01
Clash can't coveniently disable powerup values


martijnbastiaan:
With the addition of https://github.com/clash-lang/clash-compiler/pull/498, Clash now supports initial/powerup values. By default, register _reset_ values are also used as _initial_ values. For most designs, this is what we want. If a user doesn't want initial values, they can use `register#` to set a different (or no) initial value. This is, of course, not really a good answer: some designs can't support initial values. This report discusses some of the solutions to these problems.

-----------------------------------------------------

**Solution 1: implicit/default arguments**

We could use [implicit arguments in combinations with defaults](http://hackage.haskell.org/package/implicit-params) to specify whether a register should have a initial value. 

Pros:
 * Backwards compatible
 * Can change strategies within a design

Cons:
 * Easy to forget to propagate
 * Ugly, IMO
 * Doesn't feel clashy

---------------------------------------------------

**Solution 2: Add field to ~~`Reset`~~`Clock`**

Add an additional field `Initial` to ~~`Reset`~~`Clock`, such that it looks something like:

```haskell
data Clock (domain :: Domain) (gated :: ClockKind) (regInitials :: Bool) where
```

Pros:
 * Can change init strategies within a design
 * Polymorphic; users don't have to decide on a strategy up to `topEntity`
 * Easy to implement
 * No real maintenance costs

Cons:
 * Not backwards compatible
 * `Undefined` is back baby

---------------------------------------------------

**Solution 2a: Add field to ~~`ResetKind`~~`ClockKind`**

Same as _2_, but add field to ~~`ResetKind`~~`ClockKind`. With `PolyKinds` we can (probably?) make this completely backwards compatible.

Pros:
 * Can change init strategies within a design
 * Polymorphic; users don't have to decide on a strategy up to `topEntity`
 * Easy to implement
 * No real maintenance costs
 * Fully backwards compatible

Cons:
 * `Undefined` is back baby

-------------------------------------------------

**Solution 3: Add `Initial`, besides `Clock` and `Reset`**

Similar to having `Clock` and `Reset` arguments, add an `Initial` argument deciding whether to use initial values in registers.

Pros:
 * Can change init strategies within a design
 * Polymorphic; users don't have to decide on a strategy up to `topEntity`

Cons:
 * We'd need twice as many `Hidden` variants. `HiddenClock`, `HiddenIntial`, `HiddenClockInitial`, `HiddenClockResetInitial` etc..
 * Not backwards compatible

[...]

PR #527, opened 2019-03-07
Configurable initial register values

martijnbastiaan:
**Updated: 2019 June 14**

This is an implementation of configurable initial register values, discussed in https://github.com/clash-lang/clash-compiler/issues/516 and in the comments below. We've settled on the following design:

* All data types previously taking a `Domain`, now take a tag of kind `Symbol` denoting the domain name. For example, `Clock` is now: `Clock (dom :: Symbol)`.
* `Domain` is no more, but `DomainConfiguration` now exists specified by:

```haskell
-- | A domain with a name (@Symbol@). Configures the behavior of various aspects
-- of a circuits. See the documentation of this record's field types for more
-- information on the options.
--
-- See module documentation of "Clash.Explicit.Signal" for more information on
-- how to create custom synthesis domains.
data DomainConfiguration
  = DomainConfiguration
  { _tag :: Symbol
  -- ^ Domain name
  , _period :: Nat
  -- ^ Period of clock in /ps/
  , _edge :: ActiveEdge
  -- ^ Determines which edge of the clock registers are sensitive to
  , _reset :: ResetKind
  -- ^ Determines how components with reset lines respond to changes
  , _init :: InitBehavior
  -- ^ Determines the initial (or "power up") value of various components
  , _polarity :: ResetPolarity
  -- ^ Determines whether resets are active high or active low
  }
  deriving (Typeable)
```

* Domains are unique on their name. This is achieved by making `KnownDomain` a typeclass with a functional dependency on its name/configuration. Users should create their own domain by using a template function `createDomain`:

https://github.com/clash-lang/clash-compiler/blob/05d3fa82b24eb0f7512964fe514cf36567c4b0e7/clash-prelude/src/Clash/Signal/Internal.hs#L492-L512

* A function in need of one of these fields should indicate so by using a constraint `KnownDomain`. A `HiddenReset` or `HiddenClock` automatically introduces this constraint. Simple one-clock designs can continue to use `SystemClockReset` without making any changes.

* Reset polarity is carried in the domain, as well as its synchronicity. A `Reset` itself is now a very simple datatype.

* Clock gatedness as a concept does not exist anymore in Clash. Instead, users should use a separate `Enable` line to achieve the same. Almost all components in Clash have an Enable line.

-----------------------------------------------------------

To summarize the design's features:

* It should be easily extendible if we ever think of adding new domain configuration options.
* It doesn't require any new concepts over current Clash designs. Users are already familiar `KnownNat` constraints, the new `KnownDomain` constraint should feel familiar.
* Thanks to the use of `AppendSymbol`, these constraints are usable in a multi-clock designs. Coincidentally, we can use this strategy to support hidden clocks / resets in multi-clock designs too!

-----------------------------------------------------------

For reviewers:

* Core API changes can be found in https://github.com/clash-lang/clash-compiler/pull/527/commits/000db69f110a39eb7387c41349beab123c76d8bc.
* Core compiler changes can be found in https://github.com/clash-lang/clash-compiler/pull/527/commits/c4da0ac4ac1c256faeb7b64334b2aa5e88eac452.
* Primitive changes can be found in https://github.com/clash-lang/clash-compiler/pull/527/commits/5a4e108db3850979266f840887346ea477ebb8a0
* All other commits are just "fix up" commits. Nothing exciting should be in there.

-----------------------------------------------------------

Roadmap

- [x] Restructure 'clash-prelude'
- [x] Fix doctests in 'clash-prelude'
- [x] Fix documentation
- [x] Fix `clash-lib`
- [x] Fix `clash-ghc`
- [x] Fix `clash-cosim`
- [x] Repair testsuite
- [x] Fix tutorial
- [x] Rebase on master
- [x] Split commits in style / non-style commits (I had to touch the whole of `clash-prelude`, so figured I'd get it consistent with our style guide), or revert them
- [x] Write tests for multiple hidden clocks
- [x] Implement reset polarity in blackboxes
- [x] Fix DDR primitives (https://github.com/clash-lang/clash-compiler/pull/546 I believe)
- [x] Fix bug causing `HOClock` test to fail. Will be fixed in another PR.
- [x] Extend `createDomain` to create type aliases
- [x] Write a blogpost outlining the difference between all the different domain options and how to use them


oliverbunting:
Could the domain also encompass the reset behaviour? I've always felt uneasy about the ability to connect up logic on the same clock, with different resets, which are nominally in the same domain.

I think would would make reasoning about reset behaviour over any domain easier.

martijnbastiaan:
I think so. If I understand you correctly, this would induce the following change: 

```diff
 data Domain name period initKind edgeKind
   = Domain { domainName :: SSymbol name
            , domainPeriod :: Period name period
            , domainInit :: Init name initKind
            , domainEdge :: Edge name edgeKind
+           , domainReset :: Reset name resetKind
            }
```

As a bonus, we could freely add reset polarity too:

```haskell
data ResetPolarity 
  = ActiveHigh 
  | ActiveLow
```

```diff
-data Reset (domain :: Symbol) (synchronous :: ResetKind) where
-  Sync  :: Signal domain Bool -> Reset domain 'Synchronous
-  Async :: Signal domain Bool -> Reset domain 'Asynchronous
+data Reset (domain :: Symbol) (polarity :: ResetPolarity) where
+  Reset    :: Signal domain Bool -> Reset domain 'ActiveHigh
+  InvReset :: Signal domain Bool -> Reset domain 'ActiveLow

```

martijnbastiaan:
Where ResetKind (in `ResetNess` for the lack of a better word) would still be:

```haskell
data ResetKind
  = Synchronous
  | Asynchronous
```

oliverbunting:
data ResetKind
  = Synchronous
  | Asynchronous
  | None

martijnbastiaan:
3 options is hard, because you can't (easily?) write a function which "can deal with Synchronous and Asynchronous, but not with None". I'm having the same implementation problem with:

```haskell
data EdgeKind
  = Rising
  -- ^ Registers are rising edge (low-to-high) sensitive
  | Falling
  -- ^ Registers are falling edge (high-to-low) sensitive
  | Double
  -- ^ Registers are sensitive to both a rising edge (low-to-high) and a
  -- falling edge (high-to-low). [Not supported yet.]
```

I have to think about it some more.

oliverbunting:
Double is a bit weird. 
FPGA's certainly don't support that internally, although they do have DDR primitives at the pins.

martijnbastiaan:
We know at least some interest exists to use Clash for ASICs so I don't think we should only focus on FPGAs :). Anyway, I mainly added it to see how much sense it makes to introduce 3 options in *some* domain property. And so far I still don't know, so I'll keep experimenting.

leonschoorl:
We have some [DDR primitives](http://hackage.haskell.org/package/clash-prelude-0.99.3/docs/Clash-Explicit-DDR.html), their DDR signals might be annotated with edgekind Double.

martijnbastiaan:
After an offline discussion with @oliverbunting, we concluded that we just need a `delayEn`. "None" can simply be achieved by not routing `Reset`. We still might want to move reset kind to domain though, in order to  implement type-safe reset polarity.

leonschoorl:
If you want to describe a function that could be active on either edge, but not both, maybe something like:
```haskell
data ActiveEdge
  = OneEdge Edge
  | BothEdges

data Edge = Rising | Falling

f :: KnownEdge dom (OneEdge edge)
  => Signal dom a -> Signal dom a
```

martijnbastiaan:
I've confirmed something like that works @leonschoorl. The jury is out on its prettiness :P.

christiaanb:
My initial thoughts: I'm really not liking even more `KnownXYZ` constraints and GADTs. I feel like much of this information could be stored as normal data types inside the constructors for the `Clock` type; perhaps using some phantom types and "smart constructors".

christiaanb:
But I'd rather discuss this when I'm back in the office to understand the rational behind the current design.

martijnbastiaan:
I'm open to suggestions! Although I don't think the `Known*` are too bothersome, it wasn't what I set out to do. I have tried to do it with phantom types and "normal" constructors, but I couldn't manage without making them completely cumbersome to work with. I'll see if I can reproduce the process I went though, and detail why it didn't work out tomorrow.

martijnbastiaan:
Alright, didn't see your second comment just now. I'm fairly convinced that phantom types can't really work (conveniently), so I'll still try to write up the process tomorrow. If it fails we can try it face to face when you're back (and I'm back from the UK too, so probably in 2.5 weeks).

christiaanb:
@martijnbastiaan my apologies, I should've waited with reacting before reading the entire PR properly. When I saw all the new `KnownPeriod` stuff, I was afraid the same would happen for the `KnownInit`, `KnownEdge`, etc. But I think the solution is to give up slightly on type safety in the clock type:

```
-- | A clock signal belonging to a @domain@
data Clock (domain :: Symbol) (gated :: ClockKind) where
  Clock
    :: SSymbol nm
    -> Integer
    -- ^ Period
    -> Clock nm 'Source

  GatedClock
    :: SSymbol nm
    -> Integer
    -- ^ Period
    -> Signal nm Bool
    -> Clock  nm 'Gated
```

The same can be done for the `Edge` and the `InitKind`, they can just be stored as normal value inside clock, so the values can be used where needed, ie:

```
-- | Get the clock period of a clock (in /ps/) as a 'Num' from a 'Domain'
clockPeriod
  :: forall domain period gated a
   . Num a
  => Clock domain gated
  -> a
clockPeriod (Clock _ p) = fromInteger p
clockPeriod (GatedClock _ p _) = fromInteger p
```

christiaanb:
Another issue really needs solving properly is:

old
```
ddrIn 
  :: ( fast ~ 'Dom n pFast   
     , slow ~ 'Dom n (2*pFast))
```

vs

new:
```
ddrIn
  :: KnownPeriod fast pFast
  => KnownPeriod slow pSlow
  => 
```

where in the old/current API, when you use this function in a monomorphic context, and the periods do not exactly differ by `2*`, you get a type error right away.

with the new proposed API you simply propagate the constraint:

```
:: (KnownPeriod "DomA" pFast, KnownPeriod "DomB" (2*pFast)
```

and what's now stopping me from assigning period 4 to "DomA"  in one part of my design, and period 5 in another part? Using `withDomain (Domain "DomA" 4)`  in one part, and `withDomain (Domain "DomA" 5)` in another part. This is bound to happen in actual code, especially as you get larger teams.

This could never happen in the old/current design, because the period was an integral part of the domain annotation on Signals. I don't think I can accept an API where the above descirbed situation can legally happen.

martijnbastiaan:
Right, I had considered putting them in as "normal" values, but -ironically- thought that wouldn't have a chance of passing the Christiaan test :P. I'm slightly hesitant of it too, considering that you would store period/init/edge-iness in various places without type guarantees. 

> and what's now stopping me from assigning period 4 to "DomA" in one part of my design, and period 5 in another part?

The direct answer to this question is: we can reliably track what domains we've seen in the Clash compiler, and generate an error if we see _any_ differences between domains with the same name. As opposed to `Clock`/`Reset`/etc we don't (shouldn't?) export the `Domain` constructor so we could even do some simple tracking in `mkDomain`. I realize this postpones errors to the simulation or synthesis stage, but the fact that we can be 100% sure this works is valuable IMO. This is opposed to `Integer` etc, which we can neither guarantee to correspond to their type level siblings in simulation nor while synthesizing.

The indirect answer is that using `Integer` for period works in practice, but I'm not so sure the same tactic would work for Edginess (for lack of a better word) without some magic. Say you have:

```haskell
f :: Clock (domain :: NameAndPeriod) 'RisingEdge (gated :: ClockKind)
  -> Signal domain Int

g :: Clock (domain :: NameAndPeriod) 'FallingEdge (gated :: ClockKind)
  -> Signal domain Int
```

Now, you'd like to write:

```haskell
g :: Clock (domain :: NameAndPeriod) edge (gated :: ClockKind)
  -> Signal domain Int
```

I don't think you can do it without resorting to `unsafeCoerce`s. The same goes for `Init`. Because we never have to make a decision on `period`, we wouldn't have previously encountered this problem. We could probably put period back in domain (so instead of `Symbol`, you'd get `OldDomainType`), but it felt a bit arbitrary to leave it in after I had cooked up the current PR's method.

Last but not least, and I don't want to drag oude koeien uit de sloot, I think the current design avoids the conceptual problem of "where to put the `Edge` / `Init`", which helps from a pedagogic point of view.

To complete my sales pitch: I think the current PR proposes a method that's safe, easy to understand, easy to extend and implement. It only needs a single modification to Clash itself (that is, domain tracking). *If* a user manages to do something wrong, they'll get a clear and easy to understand error.

christiaanb:
I thought of a solution yesterday, but was too tired to go back on the laptop again.
Instead of using the `Hidden` class for things like `KnownPeriod`, make it an actual class like so:

```
class KnownPeriod (domain :: Symbol) (period :: Nat) | dom -> period where
  knownPeriod :: SNat period
```

Now there can be only 1 period per domain; and we could do the same for the active clock, etc.

Also, the only safe way to construct a `Clock` for the user would be to use `clockGen` and `tbClockGen`; which will use the `KnownPeriod`, `KnownEdge`, etc to fill the "untagged" (`Integer` vs `Period nat`) field of the `Clock` constructor. I a so-called "expert" user then goes and uses the `.Signal.Internal` to construct an illegal clock that's their problem.

martijnbastiaan:
That's an option @leonschoorl and I discussed, but for some reason we decided not to go with that one. I can't remember if we had a good reason or not. I'll see if it come back to me in the following days. Have a nice weekend in SF :)!

martijnbastiaan:
I did manage to think of the reasons we shot it down, but after rereading your comment I realize the designs are slightly different, so they don't really apply. I do think the `Hidden` approach would still have some upsides:

1. Case-ing on `Edge`/`Init` is guaranteed to work out, as we know type information will always propagate. (And we'd finally leverage GADT support! ;))
2. Functions can explicitly indicate what things they are sensitive to, with type guarantees that they won't touch things they promise not to touch. 

A minor complaint is that I don't think using instances (with domains leaking into a global namespace, even if you don't end up using it in your design) is the most elegant. It's not something I really care about though, as I doubt it would be a problem in practice.

Hopefully we can resolve this before Monday so I can freely work on it :)

christiaanb:
clock period and active flank really need to be globally unique for a domain, so a normal type-class (constraint) is really the proper solution is that part of the design space.

christiaanb:
This is too big of a change to be resolved before monday, as I really want to have a proper in-person discussion; and I won't be in until tuesday morning.

christiaanb:
But really I want is a solution where:

* The current `Clash.Prelude` remains the same, i.e. no extra class constraints should show up from using the new version. The existing API is already borderline too complicated/verbose for new (non-Haskell) users.
* `Clash.Explicit.Prelude`, which is more for intermediate/expert-level users, is allowed to become more complicated, i.e. introduce more class constraints.
* Things like the clock period and active flank really need to be globally unique for every domain

christiaanb:
Here's some of my initial test code, but requires more experimentation: https://gist.github.com/christiaanb/8eac9bc72fc2bf44e9cce2a8499a4db8

I has:
* Globally unique flank, period, etc per domain
* Multiple implicit clocks
* Supports monomorphic implicit clocks (needs more experimentation)
* Supports un-annotated functions that have implicit clock (i.e working. type inference)
* Unaltered `Clash.Prelude` API/Types seems likely

martijnbastiaan:
> Initial value handling,

Put it in `KnownConfiguration`, or handle it separately with a `KnownInit`.

>  Reset polarity, etc.

Move `ResetKind` to `KnownConfiguration`, replace current holder of `ResetKind` with a `Polarity` construct.

As far as I can tell, your proposal is what I've proposed here, with the Known-constraints consolidated in a single construct. Could you confirm this? Either way, it's fine by me, but like I said earlier, you'll lose:

> Functions can explicitly indicate what things they are sensitive to, with type guarantees that they won't touch things they promise not to touch.

This could be solved by reintroducing `KnownInit`, `KnownEdge`, with a single superconstraint `KnownConfiguration`.

martijnbastiaan:
One more thing

> Supports un-annotated functions that have implicit clock (i.e working. type inference)

I've experimented with this and it doesn't seem an issue in practice if GHC can't. You almost always pass a "real" argument anyway (mostly `Signal`s) that refer to the symbol allowing type inference to figure out what constraint belongs to what function call.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -1,68 +1,93 @@
...
 
 @
-data Domain = Dom { domainName :: 'GHC.TypeLits.Symbol', clkPeriod :: 'GHC.TypeLits.Nat' }
+'Domain'
+  { _tag :: 'GHC.TypeLits.Symbol'
+  -- ^ Domain name
+  , _period :: 'GHC.TypeLits.Nat'
+  -- ^ Clock period in /ps/
+  , _edge :: 'ActiveEdge'
+  -- ^ Determines which edge of the clock registers are sensitive to
```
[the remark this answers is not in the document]
martijnbastiaan: Edge! :P

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -120,13 +120,13 @@ We initially create a memory out of simple registers:
 
 @
 dataMem
-  :: Clock domain gated
-  -> Reset domain synchronous
-  -> Signal domain MemAddr
+  :: Clock tag
```
christiaanb: What about the enable? The rest of the docs for this module seem to missing the enable as well

clash-prelude/src/Clash/Explicit/Moore.hs
```diff
@@ -44,14 +45,14 @@ import           Clash.XException                 (Undefined)
 -- macT s (x,y) = x * y + s
 --
 -- mac
---   :: 'Clock' mac Source
---   -> 'Reset' mac Asynchronous
---   -> 'Signal' mac (Int, Int)
---   -> 'Signal' mac Int
--- mac clk rst = 'moore' clk rst macT id 0
+--   :: 'Clock' tag
+--   -> 'Reset' tag
```
christiaanb: missing enable

clash-prelude/src/Clash/Explicit/Moore.hs
```diff
@@ -44,14 +45,14 @@ import           Clash.XException                 (Undefined)
 -- macT s (x,y) = x * y + s
 --
 -- mac
---   :: 'Clock' mac Source
---   -> 'Reset' mac Asynchronous
---   -> 'Signal' mac (Int, Int)
---   -> 'Signal' mac Int
--- mac clk rst = 'moore' clk rst macT id 0
+--   :: 'Clock' tag
```
christiaanb: missing `KnownDomain tag dom` constraint, probably missing in other places as well.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -298,39 +355,45 @@ systemResetGen = asyncResetGen
...
+  => Clock tag
+  -> Reset tag
+  -> Enable tag
+  -> Reset tag
+resetSynchronizer clk rst en =
+  case knownDomain @tag of
+    SDomain _tag _period _edge SAsynchronous _init _polarity ->
+      let r1 = register clk rst en True (pure False)
+          r2 = register clk rst en True r1
+          r3 = bool (not <$> r2) r2 (isActiveHigh @tag)
```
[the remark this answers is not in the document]
martijnbastiaan: Yes!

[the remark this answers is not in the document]
martijnbastiaan: Although I think it should be exactly the other way around.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -73,89 +98,112 @@ meta-stability:
...
-        * The Clash compiler will give a warning whenever a function has a
-          type-signature similar to the one above.
-        * This is the reason why `unsafeFromAsyncReset` is prefixed with the
-          word /unsafe/.
-
     *   /Reset situation 3/:
 
         @
-        h :: Reset domain Asynchronous -> Reset domain Synchronous -> ..
+        h :: 'Reset' \"AsyncExample\" -> 'Reset' \"SyncExample\" -> ..
```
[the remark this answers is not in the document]
martijnbastiaan: Right, good catch

martijnbastiaan:
@christiaanb I've addressed your concerns and fixed them in a separate commit. I'll leave them up for a while and squash them whenever we're serious about actually merging the PR. Do you have any other comments?

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -93,104 +132,517 @@ where
...
+-- period, and creates an instance for it. As most users often want to update
+-- the system domain, a shortcut is available in the form:
+--
+-- > createDomain vSystem{vTag="System10", vPeriod=10}
+--
+data VDomainConfiguration
+  = VDomainConfiguration
+  { vTag    :: String
+  -- ^ Corresponds to '_name' on 'DomainConfiguration'
+  , vPeriod :: Integer
```
basile-henry: Any reason for using `Integer` rather than `Natural`?

martijnbastiaan: No! Thanks.

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -1271,10 +1288,18 @@ a general listing of the available template holes:
...
+* @~PERIOD[N]@: Clock period of given domain. Errors when called on an argument
+  which is not a 'KnownDomain'.
+* @~ISENABLED[N]@: Is the @(N+1)@'th argument a an Enable line set to a constant
+  True. Errors when called on an argument which is not an 'Enable'.
+* @~ISSYNC[N]@: Does synthesis domain of the @(N+1)@'th have synchronous resets. Errors
+  when called on an argument which is not a 'KnownDomain'.
+* @~ISINITDEFINED[N]@: Does synthesis domain of the @(N+1)@'th have defined initial
+  values. Errors when called on an argument which is not a 'KnownDomain'.
+* @~ISRISINGEDGE[N]@: Does synthesis domain of the @(N+1)@'th respond to rising
+  edges. Errors when called on an argument which is not a 'KnownDomain'.
```
leonschoorl: Similar to `~ISENABLED`,  `~ISRISINGEDGE` has a weird name particularly because you have to use in on a KnownDomain.

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -627,8 +627,10 @@ wantedLanguageExtensions df =
              , LangExt.ExplicitForAll
              , LangExt.ExplicitNamespaces
              , LangExt.FlexibleContexts
+             , LangExt.FlexibleInstances
              , LangExt.KindSignatures
              , LangExt.MagicHash
+             , LangExt.MultiParamTypeClasses
```
leonschoorl: If we really need these, they should also go into `Test.Tasty.Clash.outputTest'`

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -2276,31 +2384,36 @@ dotp :: SaturatingNum a
...
+{-# NOINLINE topEntity #-}
 
 testBench :: Signal System Bool
 testBench = done
   where
     testInput      = stimuliGenerator clk rst (2:>3:>(-2):>8:>Nil)
     expectedOutput = outputVerifier clk rst (4:>12:>1:>20:>Nil)
-    done           = expectedOutput (topEntity clk rst testInput)
-    clk            = tbSystemClockGen (not \<$\> done)
+    done           = expectedOutput (topEntity clk rst (enableGen) testInput)
```
leonschoorl: Drop the ()?

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -2276,31 +2384,36 @@ dotp :: SaturatingNum a
...
+    clk            = tbSystemClockGen (not <$> done)
```
leonschoorl: We need these escapes

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -2276,31 +2384,36 @@ dotp :: SaturatingNum a
...
+  :: ( HiddenClockResetEnable dom conf
+     , Default a
+     , KnownNat n
+     , SaturatingNum a
+     , Undefined a )
+  => Vec (n + 1) a -> Signal dom a -> Signal dom a
 fir coeffs x_t = y_t
   where
-    y_t = dotp coeffs \<$\> bundle xs
+    y_t = dotp coeffs <$> bundle xs
```
leonschoorl: We need these escapes

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -93,104 +133,520 @@ where
...
+-- | A clock (and reset) dom with clocks running at 100 MHz. Memory elements
+-- respond to the rising edge of the clock, and synchronously to changes in
+-- reset signals. It has defined initial values, and active-high resets.
+--
+-- See module documentation of "Clash.Explicit.Signal" for more information on
+-- how to create custom synthesis domains.
+type XilinxSystem = "XilinxSystem"
+
+-- | Same as SDomainConfiguration but allows for easy updates through record update syntax.
+-- Should be used in combination with 'vDomainConfiguration' and 'createDomain'. Example:
```
leonschoorl: `vDomainConfiguration` doesn't seem to exist (anymore?)

leonschoorl:
@martijnbastiaan I added 2 more commits
07b5fd67a65ed4de28dd591a73c250d2d7a5aa70 should probably be squashed into  cbc0c3a7a480a8cf8f82382d5d6f7f3103faffc0, but I'll leave that up to you

clash-prelude/src/Clash/Signal/Internal.hs
martijnbastiaan: Good catch

martijnbastiaan:
I agree with "Remove domain singleton from Reset". I'll remove it from `Clock` too.

clash-prelude/src/Clash/Tutorial.hs
martijnbastiaan: Fixed in https://github.com/clash-lang/clash-compiler/pull/527/commits/4b01f6f8b23bf815c531abe32734b4c99fb299f8

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -1271,10 +1288,18 @@ a general listing of the available template holes:
 * @~ISLIT[N]@: Is the @(N+1)@'th argument to the function a literal.
 * @~ISVAR[N]@: Is the @(N+1)@'th argument to the function explicitly not a
   literal
-* @~ISGATED[N]@: Is the @(N+1)@'th argument a gated clock, errors when called on
-  an argument which is not a 'Clock'.
-* @~ISSYNC[N]@: Is the @(N+1)@'th argument a synchronous reset, errors when
-  called on an argument which is not a 'Reset'.
+* @~TAG[N]@: Name of given domain. Errors when called on an argument which is not
+  a 'KnownDomain'.
```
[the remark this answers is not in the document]
martijnbastiaan: Fixed in https://github.com/clash-lang/clash-compiler/pull/527/commits/216a0a3725f16f3bf7db98389b14a4ffb6821f9f

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -1271,10 +1288,18 @@ a general listing of the available template holes:
...
+* @~PERIOD[N]@: Clock period of given domain. Errors when called on an argument
+  which is not a 'KnownDomain'.
+* @~ISENABLED[N]@: Is the @(N+1)@'th argument a an Enable line set to a constant
+  True. Errors when called on an argument which is not an 'Enable'.
```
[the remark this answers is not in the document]
martijnbastiaan: Fixed in https://github.com/clash-lang/clash-compiler/pull/527/commits/21314c80be1dac4adf3c30947cdd187a39b2afe6

PR #530, opened 2019-03-08
Fix count{Trailing,Leading}Zeros constant folding

martijnbastiaan:
See the discussion at https://github.com/clash-lang/clash-compiler/pull/501


christiaanb:
https://github.com/clash-lang/clash-compiler/pull/551 solves the same issue without needing a primitive

martijnbastiaan:
> #551 solves the same issue without needing a primitive

I've ran a quick test and it doesn't seem so unfortunately.

christiaanb:
Adding a primitive fixes the symptom, not the cause. If stuff still doesn't work after #551, then we need to investigate why.

rowanG077:
I think the issue is that `foldl` is not implemented in the evaluator and blackboxes.

martijnbastiaan:
Fixed in master.

martijnbastiaan:
Ah nvm, you're right @rowanG077, CTZ still doesn't constant fold.

PR #532, opened 2019-03-08
Add eval rule for `Data.Singletons.TypeLits.Internal.$s^_f`

martijnbastiaan:
edit: Circle CI interprets the `$s` in the branch name as a variable so it fails to test this :1st_place_medal:


leonschoorl:
Since this is dependent on the particular way singletons writes(or generates) its internal function and how GHC decides to specialize them, can we add a test for this?
So we known it's still correct.

martijnbastiaan:
Alright, so, from our internal discussion: there's no real reliable way to actually test this. We need a way to detect how GHC specializes functions and use that information in the evaluator. I'll add a separate issue for it, outlining what we should do.

PR #542, opened 2019-03-13
Various fixes

leonschoorl:
A bunch of random small changes I've been collecting, mostly cleanups and documentation tweaks.

There are 2 that could be considered API change:
 * b600a2d: on GHC >=8.6 I've disabled StarIsType
   So you can no longer use `*` as a Kind, but you can now write types like `Vec (2*n) Bool` without having to enable any extensions yourself.
 * 74ab734: the type variables of `delayedI` and `delayI` have been reordered to make the delay the first type argument, so we can easily use type application to specify the delay: `delayI @2`.
   But this would break any code that was already doing that.


martijnbastiaan:
LGTM. Small request: could you rename the commit "Wobble haddock" to something more descriptive?

PR #551, opened 2019-03-18
Eta-expand under-applied primitives in the evaluator

christiaanb:
We were already doing this for primtives that were partially applied to at least one argument, but not for ones that weren't applied at all.

This was an issue in e.g.:

```
map BitVector.unpack# (1 :> Nil)
```

where `BitVector.unpack#` wouldn't reduce to a value (a lambda in this case), and so the entire `map BitVector.unpack# (1 :> Nil)`  wouldn't reduce to a value. Now that we eta-expand, the first argument of `map` does reduce to a value: `\x -> BitVector.unpack# x`


martijnbastiaan:
Shouldn't we have looked into that segmentation fault before merging this..?

christiaanb:
It worked in ModelSim

PR #552, opened 2019-03-18
pack _|_ = undefined#

christiaanb:
Ensures that packing something that throws an `XException` doesn't throw an `XException` itself, but creates a BitVector with it's "undfinedness" bitmask set to all 1's.

So:

```
pack XException = undefined#
pack _|_ = _|_
```


martijnbastiaan:
Makes sense! And LGTM. A test case should make sure we can pack undefined things and don't regress in the future.

jonfowler:
Nice!

PR #554, opened 2019-03-18
Pretty-printing: Additional formatting options

omelkonian:
* Added options to the pretty-printer that toggles display of:
      1. Unique identifiers of names
      2. Types (+ type lambdas, type-level arguments, etc...)
      3. Qualifiers (things before the last '.') of names

  * Previously exported functions such as `ppr` and `showPpr` use the
    default options (= display everything), so as not to propagate
    changes to the rest of the compiler.
    Nonetheless, I added new primed versions `ppr'` and `showPpr'` that
    take the options as argument.

Feel free to suggest other options that could be useful (@leonschoorl mentioned that he is using
some pretty-printing tweaks locally).

The real change is that `PrettyPrec.pprPrec` now takes one more argument that holds the options,
but I could reduce the boilerplate by using `-XImplicitParams`. Let me know if you'd like that.

### DIFF
--- a/clash-lib/src/Clash/Core/Pretty.hs
+++ b/clash-lib/src/Clash/Core/Pretty.hs
@@ -16,9 +16,10 @@
 
 module Clash.Core.Pretty
   ( PrettyPrec (..)
-  , ppr
+  , PrettyOptions (..), defPrettyOptions
+  , ppr, ppr'
   , showDoc
-  , showPpr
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
Assigning @leonschoorl so his local changes might be added too :)

martijnbastiaan:
This can be merged once documentation has been added to `PrettyOptions`, IMO.

PR #571, opened 2019-04-04
Add Generic instance for BitVector

martijnbastiaan:
The instance in this PR will generate `pack`/`unpack` functions that match Clash's behavior when converting to `std_logic_vector`-like in HDL.

- [X] Create Generic instance
- [X] Write tests for a number of interesting datatypes
- [x] Test against manually written bitvectors (that is, don't use template haskell, but manually write `$$(bLit "0100....")`)
- [X] ~~Create Seperate module with "orphan" instances~~

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -54,3 +54,10 @@ package *
 -- The fail package is empty for GHC 8+, and haddock errors out on it
 package fail
   documentation: False
+
+-- | We need: a <=? Max a b ~ True
+source-repository-package
+  type: git
+  location: https://github.com/clash-lang/ghc-typelits-extra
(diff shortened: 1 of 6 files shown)


martijnbastiaan:
We've decided that all common types (that is, types popular in the Haskell ecosystem) will have a `BitPack` instance that reflects what Clash uses when it converts to `std_logic_vector` (and friends) in HDL. Other approaches are counter-intuitive when used in combination with custom bit representations. We _might_ add the ability to switch between different representation behaviors (through CPP?), but not before Clash 1.1.

Tagging @leonschoorl: see ^

martijnbastiaan:
Related issue: https://github.com/clash-lang/clash-compiler/issues/572

Issue #328, opened 2018-07-11
Synthesize annotations are ignored


adamwalker:
If I generate Verilog using Clash 0.99.2 for the Clash code below, the Synthesize annotation seems to be ignored. 

```
import Clash.Prelude

{-# ANN func
  (Synthesize
    { t_name   = "test"
    , t_inputs = [ PortName "clk"
                 , PortName "reset"
                 , PortName "x" ]
    , t_output = PortName "y"
    }) #-}
func 
    :: Clock System Source 
    -> Reset System Synchronous 
    -> Signal System Bool 
    -> Signal System Bool
func clk reset p = withClockReset clk reset $ register False p
```
I get the error:
```
<no location info>: error:
    Other error:
    No 'topEntity', nor function with a 'Synthesize' annotation found in root module: Main
```
Compilation succeeds if I rename the function to topEntity, but then the port and module names are wrong.

christiaanb:
Odd.. I can't replicate with `master`

adamwalker:
I rebuilt Clash from master and the problem went away. I had previously built Clash with ghc 8.4.2. I then upgraded to the latest Stackage LTS which uses 8.4.3 but I don't think I rebuilt Clash. Could that explain it?

christiaanb:
Possibly, I was building with GHC 8.4.1

adamwalker:
I rebuilt Clash with GHC 8.4.1, 8.4.2, and 8.4.3 and I can no longer reproduce this. This is super weird because I still have the terminal open and can see the error so I didn't imagine it. Anyway, since I can no longer reproduce this, and everything works great, it's probably safe to close the issue.

pacak:
Same problem with ghc 8.4.4 and 0.99.3. If name is not `topEntry` - compilation fails, if name is `topEntry` - anything inside annotation is ignored.

pacak:
It was achieved in clean  installation of ghc 8.4.4 (via ghcup), clash compiled with ` cabal new-install clash-ghc --enable-documentation --enable-executable-dynamic` (at least that's what I found in my command line history) then created cabal project with `build-depends:       base ^>=4.11.1.0, clash-prelude`.

Problem was fixed by installing clash-ghc with stack and using it as stack project. So something fishy is going on.

adamwalker:
I've seen this again a couple of times since I originally created the issue. Deleting the .stack-work directory and rebuilding seems to fix.

leonschoorl:
[...]

PR #574, opened 2019-04-08
Error when we're using a different clash-prelude then we were built with

leonschoorl:
Because if they differ clash won't be able to recognize any ANNotations.

Should help detect issues like #328

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -27,12 +27,15 @@ where
 -- External Modules
 import           Clash.Annotations.Primitive     (HDL, PrimitiveGuard)
 import           Clash.Annotations.TopEntity     (TopEntity (..))
+import           Clash.Util                      (ClashException(..), pkgIdFromTypeable)
 import           Control.Arrow                   (first, second)
 import           Control.DeepSeq                 (deepseq)
+import           Control.Exception               (throw)
 #if MIN_VERSION_ghc(8,6,0)
 import           Control.Exception               (throwIO)
 #endif
 import           Control.Monad.IO.Class          (liftIO)
+import           Data.Char                       (isDigit)
 import           Data.Generics.Uniplate.DataOnly (transform)
 import           Data.Data                       (Data)
 import           Data.Typeable                   (Typeable)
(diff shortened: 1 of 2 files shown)


clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -702,3 +706,27 @@ removeStrictnessAnnotations pm =
...
+preludePkgId :: String
+preludePkgId = $(lift $ pkgIdFromTypeable $ defSyn "")
+
+-- | Check that we're using the same clash-prelude as we were built with
+--
+-- Because if they differ clash won't be able to recognize any ANNotations.
+checkForInvalidPrelude :: Monad m => HscTypes.ModGuts -> m ()
+checkForInvalidPrelude guts = case filter isWrongPrelude pkgIds of
+                           [] -> return ()
+                           (x:_) -> throw (ClashException noSrcSpan (msgWrongPrelude x) Nothing)
```
martijnbastiaan: Did you mean this indention to start there? It seems more or less random. I usually prefer:

```haskell
checkForInvalidPrelude guts = 
  case filter isWrongPrelude pkgIds of
    [] -> return ()
    (x:_) -> throw (ClashException noSrcSpan (msgWrongPrelude x) Nothing)
```

or even

```haskell
checkForInvalidPrelude guts = 
  case filter isWrongPrelude pkgIds of
    [] -> 
      return ()
    (x:_) -> 
      throw (ClashException noSrcSpan (msgWrongPrelude x) Nothing)
```

martijnbastiaan:
LGTM! I'll add the error message to this PR to make it google-able:

```
<no location info>: error:
    Clash only works with the exact clash-prelude it was built with.
Clash was built with: clash-prelude-0.99-inplacedummy
So can't run with:    clash-prelude-0.99-inplace
```

Causes / solutions / workarounds are described in https://github.com/clash-lang/clash-compiler/issues/328.

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
leonschoorl: Thnx
At some point is was nicely indented with respect to the case, then refactoring happened and the name of the function got longer.
I'll fix it tomorrow

martijnbastiaan: Ah right! That's why I usually prefer to write it as my second suggestion; it really helps to reduce how much (annoying) realignment you need to do when refactoring. Maybe one day it'll be part of The Official Clash Style Guide :grin:

Issue #575, opened 2019-04-09
`stack build` fails as of 3abca477


gergoerdi:
Clean build using `stack build` on 3abca477 fails with the following error message. Is there some GHC `Nat` plugin that wasn't properly version-bumped in the dependencies?

```
    [29 of 78] Compiling Clash.Class.BitPack ( src/Clash/Class/BitPack.hs, .stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build/Clash/Class/BitPack.o )
    
    /home/cactus/prog/clash/import/clash-compiler/clash-prelude/src/Clash/Class/BitPack.hs:320:19: error:
        Couldn't match type ‘GFieldSize f
                             GHC.TypeNats.<=? Max (GFieldSize f) (GFieldSize g)’
                       with ‘'True’
          arising from a use of ‘undefined#’
        |
    320 |     let padding = undefined# :: BitVector (Max (GFieldSize f) (GFieldSize g) - GFieldSize f) in
        |                   ^^^^^^^^^^
    
    /home/cactus/prog/clash/import/clash-compiler/clash-prelude/src/Clash/Class/BitPack.hs:325:19: error:
        Couldn't match type ‘GFieldSize g
                             GHC.TypeNats.<=? Max (GFieldSize f) (GFieldSize g)’
                       with ‘'True’
          arising from a use of ‘undefined#’
        |
    325 |     let padding = undefined# :: BitVector (Max (GFieldSize f) (GFieldSize g) - GFieldSize g) in
        |                   ^^^^^^^^^^
    
    /home/cactus/prog/clash/import/clash-compiler/clash-prelude/src/Clash/Class/BitPack.hs:337:78: error:
        Couldn't match type ‘GFieldSize f
                             GHC.TypeNats.<=? Max (GFieldSize f) (GFieldSize g)’
                       with ‘'True’
          arising from a use of ‘split#’
        |
    337 |     (f, _ :: BitVector (Max (GFieldSize f) (GFieldSize g) - GFieldSize f)) = split# b
        |                                                                              ^^^^^^^^
    
    /home/cactus/prog/clash/import/clash-compiler/clash-prelude/src/Clash/Class/BitPack.hs:338:78: error:
        Couldn't match type ‘GFieldSize g
                             GHC.TypeNats.<=? Max (GFieldSize f) (GFieldSize g)’
                       with ‘'True’
          arising from a use of ‘split#’
        |
[...]

PR #577, opened 2019-04-10
Test Stack build on CI

martijnbastiaan:
Seeing we provide a `stack.yaml`, we should test if it can actually build Clash. To spare resources on `diepenheim` I've only added it to Circle CI. This should prevent issues such as https://github.com/clash-lang/clash-compiler/issues/575.

### DIFF
--- a/.ci/stack_build.sh
+++ b/.ci/stack_build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -xeo pipefail
+
+apt update
+apt install wget -y
+wget -qO- https://get.haskellstack.org/ | sh
+stack build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,7 @@ aliases:
       key: cabal-store-{{ .Environment.CIRCLE_JOB }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "cabal.project" }}
       paths:
         - cabal-store
+        - ~/.stack
 
   - &build
     run:
@@ -37,7 +38,12 @@ aliases:
         if [ -n "$CIRCLE_PR_NUMBER" ]; then
           .ci/build.sh
         fi
-            
+
+  - &stack_build
+    run:
+      name: Build dependencies and Clash itself with Stack
+      command: .ci/stack_build.sh
+
   - &run_tests
     run: 
       name: Run tests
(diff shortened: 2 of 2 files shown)


christiaanb:
Why `cabal new-install stack`, instead of the [official](https://docs.haskellstack.org/en/stable/README/#how-to-install): `curl -sSL https://get.haskellstack.org/ | sh`

martijnbastiaan:
No particular reason. It's just the way I've been installing `stack` on my own computers. Good to know there's a quicker option!

.ci/stack_build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -xeo pipefail
+
+apt install wget -y
+wget -qO- https://get.haskellstack.org/ | sh
```
leonschoorl: I know this is the "official way", but `wget ... | sh` just feel so wrong to me.

.circleci/config.yml
```diff
@@ -37,7 +34,12 @@ aliases:
         if [ -n "$CIRCLE_PR_NUMBER" ]; then
           .ci/build.sh
         fi
-            
+
+  - &stack_build
+    run:
+      name: Build dependencies and Clash itself with Stack
+      command: .ci/stack_build.sh
```
leonschoorl: Should this also have the `if [ -n "$CIRCLE_PR_NUMBER" ]; then`?

.ci/setup.sh
```diff
@@ -19,3 +19,14 @@ case "$GHC" in
     cp .ci/cabal.project.local-8.6 cabal.project.local
     ;;
 esac
+
+# run new-update first to generate the cabal config file that we can then modify
```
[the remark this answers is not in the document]
martijnbastiaan: Right!

.circleci/config.yml
martijnbastiaan: Well, I wasn't planning on adding it to GitLab CI (because `diepenheim` is busy enough as it is), so I think we should always test it. But maybe we should add it to GitLab CI anyway, as having it only on Circle means our own branches don't get tested until they're a PR..

martijnbastiaan:
> Should we (also) be running this through gitlab?

See comment ^

> Or we could skip install the system ghc, probably skip the whole setup.sh.

I think this is the better option. This is probably what most people will do.

> If we do this we might not want to cache ~/.stack/programs which is where stack's ghc lives.

OK!

PR #579, opened 2019-04-10
Add `rnfX` to `Undefined`

martijnbastiaan:
`NFDataX` enables users to use fully evaluate a structure containing `XException`s, without the `XException` bubbling  up. Piggybacked commits:

* Strictify Bitvector: WHNF now evaluates to NF.
* Add `Clash.Sized.Vector.seqV(X)` and `Clash.Sized.Vector.forceV(X)`: useful when one wants to evaluate all elements of a vector to WHNF. (Maybe we need `Clash.Sized.Vector.Strict`?)

### DIFF
--- a/clash-prelude/src/Clash/Explicit/BlockRam.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam.hs
@@ -404,7 +404,7 @@ import Clash.Signal.Internal
 import Clash.Signal.Bundle    (unbundle)
 import Clash.Sized.Unsigned   (Unsigned)
 import Clash.Sized.Vector     (Vec, toList)
-import Clash.XException       (maybeX, seqX, Undefined, deepErrorX)
+import Clash.XException       (maybeIsX, seqX, Undefined, deepErrorX)
 
 {- $setup
 >>> import Clash.Explicit.Prelude as C
@@ -779,11 +779,11 @@ blockRam# clk content rd wen = case clockEnable clk of
           o'   = if re then ram V.! r else o
       in  o `seqX` o :- (ret `seq` rt `seq` et `seq` wt `seq` dt `seq` go' ram' o' res rs en wr din)
 
-    upd ram we waddr d = case maybeX we of
-      Nothing -> case maybeX waddr of
+    upd ram we waddr d = case maybeIsX we of
+      Nothing -> case maybeIsX waddr of
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
-      Just True -> case maybeX waddr of
+      Just True -> case maybeIsX waddr of
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
       _ -> ram
--- a/clash-prelude/src/Clash/Explicit/BlockRam/File.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam/File.hs
@@ -110,7 +110,7 @@ import Clash.Sized.BitVector (BitVector)
 import Clash.Signal.Internal (Clock, Signal (..), (.&&.), clockEnable)
 import Clash.Signal.Bundle   (unbundle)
 import Clash.Sized.Unsigned  (Unsigned)
-import Clash.XException      (errorX, maybeX, seqX)
+import Clash.XException      (errorX, maybeIsX, seqX)
 
 
 -- | Create a blockRAM with space for 2^@n@ elements
@@ -243,11 +243,11 @@ blockRamFile# clk _sz file rd wen = case clockEnable clk of
           o'   = if re then ram V.! r else o
       in  o `seqX` o :- go' ram' o' res rs en wr din
 
-    upd ram we waddr d = case maybeX we of
-      Nothing -> case maybeX waddr of
+    upd ram we waddr d = case maybeIsX we of
+      Nothing -> case maybeIsX waddr of
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
-      Just True -> case maybeX waddr of
+      Just True -> case maybeIsX waddr of
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
       _ -> ram
--- a/clash-prelude/src/Clash/Explicit/RAM.hs
+++ b/clash-prelude/src/Clash/Explicit/RAM.hs
@@ -43,7 +43,7 @@ import Clash.Explicit.Signal ((.&&.), unbundle, unsafeSynchronizer)
 import Clash.Promoted.Nat    (SNat (..), snatToNum, pow2SNat)
 import Clash.Signal.Internal (Clock (..), Signal (..), clockEnable)
 import Clash.Sized.Unsigned  (Unsigned)
-import Clash.XException      (errorX, maybeX)
+import Clash.XException      (errorX, maybeIsX)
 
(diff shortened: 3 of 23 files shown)


clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -779,11 +779,11 @@ blockRam# clk content rd wen = case clockEnable clk of
           o'   = if re then ram V.! r else o
       in  o `seqX` o :- (ret `seq` rt `seq` et `seq` wt `seq` dt `seq` go' ram' o' res rs en wr din)
 
-    upd ram we waddr d = case maybeX we of
-      Nothing -> case maybeX waddr of
+    upd ram we waddr d = case maybeHasX we of
```
christiaanb: I think `maybeIsX` is sufficient here

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -779,11 +779,11 @@ blockRam# clk content rd wen = case clockEnable clk of
...
+      Nothing -> case maybeHasX waddr of
```
christiaanb: Also, I think `maybeIsX` is sufficient.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -779,11 +779,11 @@ blockRam# clk content rd wen = case clockEnable clk of
...
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
-      Just True -> case maybeX waddr of
+      Just True -> case maybeHasX waddr of
```
christiaanb: Same

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -243,11 +243,11 @@ blockRamFile# clk _sz file rd wen = case clockEnable clk of
           o'   = if re then ram V.! r else o
       in  o `seqX` o :- go' ram' o' res rs en wr din
 
-    upd ram we waddr d = case maybeX we of
-      Nothing -> case maybeX waddr of
+    upd ram we waddr d = case maybeHasX we of
```
christiaanb: `maybeIsX` ?

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -243,11 +243,11 @@ blockRamFile# clk _sz file rd wen = case clockEnable clk of
...
+      Nothing -> case maybeHasX waddr of
```
christiaanb: same

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -243,11 +243,11 @@ blockRamFile# clk _sz file rd wen = case clockEnable clk of
...
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
-      Just True -> case maybeX waddr of
+      Just True -> case maybeHasX waddr of
```
christiaanb: same

clash-prelude/src/Clash/Explicit/RAM.hs
```diff
@@ -132,11 +132,11 @@ asyncRam# wclk rclk sz rd en wr din =
           o    = ram V.! r
       in  o :- go ram' rs es ws ds
 
-    upd ram we waddr d = case maybeX we of
-      Nothing -> case maybeX waddr of
+    upd ram we waddr d = case maybeHasX we of
```
christiaanb: `maybeIsX`?

clash-prelude/src/Clash/Explicit/RAM.hs
```diff
@@ -132,11 +132,11 @@ asyncRam# wclk rclk sz rd en wr din =
...
+      Nothing -> case maybeHasX waddr of
```
christiaanb: Same

clash-prelude/src/Clash/Explicit/RAM.hs
```diff
@@ -132,11 +132,11 @@ asyncRam# wclk rclk sz rd en wr din =
...
         Nothing -> V.map (const (seq waddr d)) ram
         Just wa -> ram V.// [(wa,d)]
-      Just True -> case maybeX waddr of
+      Just True -> case maybeHasX waddr of
```
christiaanb: Same

clash-prelude/src/Clash/Explicit/BlockRam.hs
martijnbastiaan: You're right. `maybeHasX ~ maybeIsX` in this case, so I'm doing more work than needed!

martijnbastiaan:
@christiaanb I've processed your suggestions. Any other comments?

martijnbastiaan:
Actually, a lot of instances of `NFDataX` are wrong. For example:

```haskell
rnfX (undefined# :: (Int, Int)) = ⊥
```

while it should be


```haskell
rnfX (undefined# :: (Int, Int)) = ()
```

I'll look into this.

martijnbastiaan:
The current PR fixes these problems. I've added tests to make sure this is the case.

clash-prelude/src/Clash/XException.hs
```diff
@@ -374,9 +434,341 @@ instance GShowX UInt where
...
+--
+-- > forceX x = x `deepseqX` x
+forceX :: NFDataX a => a -> a
+forceX x = x `deepseqX` x
+{-# INLINE forceX #-}
+
+-- | 'deepseqX': fully evaluates the first argument, before returning the
+-- second. Does not propagate 'XException's.
+deepseqX :: NFDataX a => a -> b -> b
+deepseqX a b = rnfX a `seqX` b
```
martijnbastiaan: Should probably be `seq`, as `rnfX` is only allowed to return `()`.

clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
```diff
@@ -156,6 +157,7 @@ instance Show (Unsigned n) where
 instance ShowX (Unsigned n) where
   showsPrecX = showsPrecXWith showsPrec
 
+instance NFDataX (Unsigned n) where rnfX = rwhnfX
```
leonschoorl: Doesn't only evaluate the U constructor?

martijnbastiaan: Newtypes don't have a constructor at runtime, so I don't think so!

martijnbastiaan: Hmm.. the error GHC throws is confusing:

```
src/Clash/Sized/Internal/Unsigned.hs:139:5-37: error:
    • A newtype constructor cannot have a strictness annotation,
        but ‘U’ does
    • In the definition of data constructor ‘U’
      In the newtype declaration for ‘Unsigned’
    |
139 |     U { unsafeToInteger :: !Integer }
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So it seems to suggest that you can add it to `U`..?

martijnbastiaan: Ah nvm, I can't read! So it does seems like the `rnfX = rwhnfX` is correct in this case.

leonschoorl: Right, with the newtype it makes total sense.

PR #580, opened 2019-04-11
 Add `dumpReplayable` and `replay` to `Clash.Signal.Trace` 

martijnbastiaan:
These functions have been primarily added to aid debugging space leaks
in larger designs, whenever its hard or cumbersome to come up with
realistic test data. A typical debugging session would look like:

 * Record the inputs to a subcircuit of an existing design that exhibits
 space leaks.

 * Replay these values in a new session to only the subcircuit. Evaluate
 only parts of the output (if applicable) and determine whether the
 design still contains space leaks. If so, fix them, is not, goto the
 first step.

### DIFF
--- a/clash-prelude/src/Clash/Signal/Trace.hs
+++ b/clash-prelude/src/Clash/Signal/Trace.hs
@@ -1,5 +1,6 @@
 {-|
 Copyright  :  (C) 2018, Google Inc.
+                  2019, Myrtle Software Tld
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 
@@ -44,6 +45,7 @@ main = do
 @
 -}
 {-# LANGUAGE CPP                 #-}
+{-# LANGUAGE BangPatterns        #-}
 {-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE MagicHash           #-}
@@ -95,6 +97,8 @@ import           Clash.Class.BitPack   (BitPack, BitSize, pack)
 import           Clash.Promoted.Nat    (snatToNum, SNat(..))
 import           Clash.Signal.Internal (sample)
 import           Clash.XException      (deepseqX, Undefined)
+import           Clash.Sized.Internal.BitVector
+  (BitVector(BV))
 
 -- Haskell / GHC:
 import           Control.Monad         (foldM)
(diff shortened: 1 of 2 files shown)


clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -449,12 +494,70 @@ dumpVCD
...
+-- evaluate more values than were originally dumped.
+replay
+  :: forall a domain n
+   . Typeable a
+  => Undefined a
+  => BitPack a
+  => n ~ BitSize a
+  => KnownNat n
+  => ByteString
+  -> Either String (Signal domain a)
```
leonschoorl:
```suggestion
  :: forall a domain n
   . (Typeable a, Undefined a, BitPack a, n ~ BitSize a, KnownNat a)
  => ByteString
  -> Either String (Signal domain a)
```

clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -449,12 +494,70 @@ dumpVCD
...
+
+-- | Helper function of 'replay'. Decodes ByteString to some type with
+-- BitVector as an intermediate type.
+decodeSamples
+  :: forall a n
+   . BitPack a
+  => n ~ BitSize a
+  => KnownNat n
+  => ByteString
+  -> [Either String a]
```
leonschoorl:
```suggestion
  :: forall a n
   . (BitPack a, n ~ BitSize a, KnownNat n)
  => ByteString
  -> [Either String a]
```

clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -49,6 +49,7 @@ main = do
 {-# LANGUAGE DataKinds           #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE MagicHash           #-}
+{-# LANGUAGE MultiWayIf          #-}
```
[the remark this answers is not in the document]
martijnbastiaan: Correct!

clash-prelude/src/Clash/Signal/Trace.hs
martijnbastiaan: I much prefer to do it on multiple lines, because there are so many constraints (I think my mental/visual limit is around ~3). I'll keep the "tuple" notation of course, as we discussed for [the unofficial official style guide](https://github.com/clash-lang/clash-compiler/pull/584).

PR #583, opened 2019-04-16
Add Cabal flag `-fsuper-strict`

martijnbastiaan:
Enabling this makes Clash use `deepseqX` in register-like primitives,
instead of `seqX`. This can help to fight space leaks.


martijnbastiaan:
The first two commits are separate because I need the first one for another PR :slightly_smiling_face:

PR #587, opened 2019-04-23
Pretty-printer: Annotations

omelkonian:
* When pretty printing, we annotate the document
    with information regarding the current expression context
    and classify expressions in different types of syntax.

  * Hide options can now be realised as a post-processing step
    that simply hides sub-expressions marked with a specific
    type of syntax (e.g. qualifiers).

  * Since we require a fixed type of annotations `ClashAnnotation`,
    the method signature of `PprPrec` is changed, leading to
    cascading changes on usages of pretty-printing across the codebase.

  * Provide a manual `Eq` instance for `CoreContext`, since `AppArg`
    now contains information about the argument, which shouldn't
    influence equality of the surrounding contexts.

### DIFF
--- a/clash-lib/src/Clash/Core/Pretty.hs
+++ b/clash-lib/src/Clash/Core/Pretty.hs
@@ -11,17 +11,20 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell   #-}
 {-# LANGUAGE ViewPatterns      #-}
+{-# LANGUAGE LambdaCase        #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Clash.Core.Pretty
   ( PrettyPrec (..)
-  , PrettyOptions (..), defPrettyOptions
+  , PrettyOptions (..)
+  , ClashAnnotation (..)
+  , SyntaxElement (..)
   , ppr, ppr'
-  , showDoc
   , showPpr, showPpr'
   , tracePprId
   , tracePpr
+  , fromPpr
   )
 where
 
@@ -30,7 +33,7 @@ import Data.Text                        (Text)
 import Control.Monad.Identity
 import qualified Data.Text              as T
 import Data.Text.Prettyprint.Doc
-import Data.Text.Prettyprint.Doc.Render.String
+import Data.Text.Prettyprint.Doc.Internal
 import Debug.Trace                      (trace)
 import GHC.Show                         (showMultiLineString)
 import Numeric                          (fromRat)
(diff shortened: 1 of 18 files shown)


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -199,12 +200,11 @@ runRewriteSession :: RewriteEnv
...
                   -> a
-runRewriteSession r s m = traceIf True ("Clash: Applied " ++
-                                        show (s' ^. transformCounter) ++
-                                        " transformations")
-                                  a
-  where
-    (a,s',_) = runR m r s
+runRewriteSession r s m = unsafePerformIO $ do
+  let (a,s',_) = runR m r s
+  encodeFile "history.dat" (s' ^. rewriteSteps)
```
leonschoorl: Is this supposed to be in this commit?

omelkonian: Oops, you're right. Forgot to save the file when editing.

clash-lib/src/Clash/Unique.hs
```diff
@@ -97,15 +99,15 @@ instance Uniquable Int where
 newtype UniqMap a = UniqMap (IntMap a)
   deriving (Functor, Foldable, Traversable, Semigroup, Monoid, NFData, Binary)
 
-instance Pretty a => Pretty (UniqMap a) where
-  pretty (UniqMap env) =
+instance Pretty' a => Pretty' (UniqMap a) where
+  pr (UniqMap env) =
```
martijnbastiaan: Isn't it better to call this function `pretty'`? Seems to match the class rename better. Maybe even call it `ClashPretty`/`clashPretty` or something similar to clarify that this is _our_ implementation.

clash-lib/src/Clash/Core/Pretty.hs
```diff
@@ -38,42 +43,83 @@ import Numeric                          (fromRat)
...
--- | PrettyPrec printing Show-like typeclass
-class PrettyPrec p where
-  pprPrec :: Monad m => PrettyOptions -> Rational -> p -> m (Doc ann)
-
+-- | Options for the pretty-printer, controlling which elements to hide.
 data PrettyOptions = PrettyOptions
-  { displayUniques    :: Bool
-  , displayTypes      :: Bool
-  , displayQualifiers :: Bool
+  { displayUniques    :: Bool -- ^ whether to display unique identifiers
```
martijnbastiaan: Could you put the comments below the field name as outlined in our [non-official official Clash style guide](https://github.com/clash-lang/clash-compiler/blob/feature-styleguide/STYLE.md#top-level-definitions)? :)

clash-lib/src/Clash/Core/Pretty.hs
```diff
@@ -38,42 +43,83 @@ import Numeric                          (fromRat)
...
+  , displayTypes      :: Bool -- ^ whether to display type information
+  , displayQualifiers :: Bool -- ^ whether to display module qualifiers
   }
 
-defPrettyOptions :: PrettyOptions
-defPrettyOptions = PrettyOptions True True True
+-- | Annotations carried on pretty-printed code.
+data ClashAnnotation
+  = AnnContext CoreContext   -- ^ marking navigation to a different context
```
martijnbastiaan: Same here.

omelkonian:
Fixed some spacing and hiding issues.

martijnbastiaan:
`text-show` is acting up, I'll push a fix to master!

martijnbastiaan:
Done, you can rebase!

PR #592, opened 2019-04-30
 Add `Clash.Class.Exp` for exponentiation calculations 

martijnbastiaan:
Some users reported issues when using `^` in Clash: the compiler would slow down to a crawl trying to translate it. As a solution, this PR adds an extra typeclass `Exp`, implementing a version of `^` that can't overflow. More info can be found in the comments below.

This PR could be better in a couple of ways, but I can't realistically spend more time on it:

* I'm not sure what to do with `Integer` and `Int`. Should they have an implementation? 

* As we always know the exponent (`SNat n`) a synthesizable implementation always exists in the form of `foldl (*) 1 (replicate n (resize b))`. If we want to keep constant folding though, we need an actual blackbox for it - which is relatively difficult to write. Even if we could let Clash generate the definition mentioned earlier if it can't constant fold, we'd still be introducing a very slow version of exponentiation in simulation.. For now, whenever a exponentiation blackbox is instantiated it emits a warning pointing users to https://github.com/clash-lang/clash-compiler/issues/593.

* Tests could be more extensive.

Despite the issues I'm still going to request this PR to be merged, due to earlier mentioned time constraints. 

----------------

During the implementation of this PR I found a subtle oversight in one of the tests testing constant folding of operations on various number types. The fixes are piggybacked to this PR.

------------------

PR also adds minor cleanups / missing functions to `Evaluator.hs`.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
@@ -18,6 +18,7 @@
 
 module Clash.GHC.Evaluator where
 
+import           Control.Applicative (liftA2)
 import           Control.Concurrent.Supply  (Supply,freshId)
 import           Control.Monad              (ap)
 import           Control.Monad.Trans.Except (runExcept)
@@ -26,7 +27,7 @@ import           Data.Char           (chr,ord)
 import qualified Data.Either         as Either
 import qualified Data.IntMap         as IntMap
 import           Data.Maybe
-  (fromMaybe, mapMaybe)
+  (fromMaybe, mapMaybe, catMaybes)
 import qualified Data.List           as List
 import qualified Data.Primitive.ByteArray as ByteArray
 import           Data.Proxy          (Proxy)
@@ -35,10 +36,12 @@ import           Data.Text           (Text)
 import qualified Data.Text           as Text
 import qualified Data.Vector.Primitive as Vector
 import           Debug.Trace         (trace)
+import           GHC.Stack           (HasCallStack)
 import           GHC.Float
 import           GHC.Int
 import           GHC.Integer
-  (decodeDoubleInteger,encodeDoubleInteger,compareInteger)
+  (decodeDoubleInteger,encodeDoubleInteger,compareInteger,orInteger,andInteger,
+   xorInteger,complementInteger,absInteger,signumInteger)
 import           GHC.Integer.GMP.Internals (Integer (..), BigNat (..))
 import           GHC.Natural
 import           GHC.Prim
@@ -823,6 +826,16 @@ reduceConstant isSubj gbl tcm h k nm ty tys args = case nm of
   "GHC.Classes.divInt#" | Just (i,j) <- intLiterals args
     -> reduce (integerToIntLiteral (i `div` j))
 
+  -- modInt# :: Int# -> Int# -> Int#
+  "GHC.Classes.modInt#"
+    | [dividend, divisor] <- intLiterals' args
+    ->
+      if divisor == 0 then
+        let iTy = snd (splitFunForallTy ty) in
+        reduce (evalUndefined iTy)
+      else
+        reduce (Literal (IntLiteral (dividend `mod` divisor)))
+
   "GHC.Classes.not"
     | [DC bCon _] <- args
     -> reduce (boolToBoolLiteral tcm ty (nameOcc (dcName bCon) == "GHC.Types.False"))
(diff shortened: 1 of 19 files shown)


clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3211,6 +3244,9 @@ integerLiterals' = typedLiterals' integerLiteral
  where
   integerLiteral x = case x of
     Lit (IntegerLiteral i)      -> Just i
+    DC dc [Left (Literal (IntegerLiteral i))]
```
christiaanb: This shouldn’t exist! What is the name of this constructor??? The constructor with tag 1, `S#` (http://hackage.haskell.org/package/integer-gmp-1.0.2.0/docs/GHC-Integer-GMP-Internals.html#t:Integer), takes an Int literal, not an Integer literal. Is GHC giving us malformed Integers??? Or did we create them?

martijnbastiaan: Actually, I think this is just a mistake on my part after misreading debugging output and accidentally leaving it there. I'll double-check in the morning.

martijnbastiaan:
Hmm, you're right. Whenever the base is a power of 2 or the exponent is a constant, it should pretty much be synthesizable. (Any others?)

```haskell 
(^) :: a -> SNat n -> ExpResult a n
```

I guess this misses the case where the base is a power of two (thus being perfectly synthesizable without knowing the exponent). I'm tempted to say: leave the default `^` as `a -> Integer -> a` and at a second `exp` similar to `mul` and `add` that resizes its result. 

I'll rework it in the morning.

christiaanb:
Could even be the same type class, just give it 2 methods instead of 1. Also, the associated type could probably do with some `TypeFamilyDependencies` to improve type inference:

```haskell
class Exp a where
  type Exp a n = r | r -> a
  (^) :: a -> Integer -> a
  exp :: a -> SNat n -> Exp a n
```

martijnbastiaan:
`Unsigned n`, right? And thanks, excellent suggestions :).

christiaanb:
you might want to describe fixed point numbers using negative exponents, e.g `2^(-2)` instead of `0.25`. And so `(*2^(-n))` can have a cheap right shift as implementation.

martijnbastiaan:
Ah right. Well this isn't covered by `Prelude.^`, so it didn't come to mind. `Prelude` handles this by introducing a second function `^^`. Maybe we should do the same (with possibly a second typeclass `ExpFrac`)?

martijnbastiaan:
So the whole thing would become:

```
class Exp a where
  type Exp a n = r | r -> a
  (^) :: a -> Signed n -> a
  exp :: a -> Signed n -> Exp a n

class ExpFrac a where
  type ExpFrac a n = r | r -> a
  (^^) :: a -> Unsigned n -> a
  expFrac :: a -> Unsigned n -> ExpFrac a n
```

martijnbastiaan:
I've checked the VHDL reference manual and it says `**` is only available on floating point numbers when using a negative exponent. I think I'll leave `ExpFrac` out for now, and not deal with negative exponents at all - there's no (simple) implementation for now AFAIK.

martijnbastiaan:
VHDL-93 doesn't permit exponentiation at all. It only allows "based literals":

> A based literal is an abstract literal expressed in a form that specifies the base explicitly. The base must be at least two and at most sixteen.

Syntax:

```
    based_literal ::=
        base # based_integer [ . based_integer ] # [ exponent ]

    base ::= integer

    based_integer ::=
        extended_digit { [ underline ] extended_digit }

    extended_digit :: = digit | letter
```

But that ^ is mostly useless.. An actual power operator (`**`) was introduced in VHDL-2000.

christiaanb:
So if a synthesizable version of `^` is almost non-existant. And you always want it to be constant-folded, i.e. end up as a constant/literal, then:

```
class Exp a where
  type ExpResult a n = r | r -> a
  (^) :: a -> SNat n -> ExpResult a n
```

Gives the least surprising result: nobody intends/expects their constants to overflow the representation! What’s worse if this shows up during a long compilation process, and you just wasted valuable minutes of your time.

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
martijnbastiaan: Turns out I indeed misread the logs. I've scrapped ^this^.

martijnbastiaan:
The same turns out to be true for Verilog: exponentiation is supported up to 31 bits. Again, pretty useless.

> Gives the least surprising result: nobody intends/expects their  constants to overflow the representation! What’s worse if this shows up  during a long compilation process, and you just wasted valuable minutes  of your time.

Agreed.

martijnbastiaan:
@christiaanb I think my last commit concludes this PR. I'm aware of some issues, which I've outlined in the OP.

clash-prelude/src/Clash/Class/Exp.hs
```diff
@@ -0,0 +1,93 @@
...
+import           GHC.TypeLits
+  (KnownNat, Nat, type (^), type (*))
+
+-- | Type class implementing exponentiation with explicitly resizing results.
+class Exp a where
+  type ExpResult a (n :: Nat)
+
+  -- | Exponentiation with known exponent.
+  (^)
+    :: KnownNat n
```
christiaanb: Additional `KnownNat n` should not be needed because you already have an `SNat n`. So if you have a function:

```
f :: SNat -> a -> r
f sn x = ...
```

and GHC complains about not having an `KnownNat n` available somewhere in the context of `f`, then simply change the definition of f to:

```
f :: SNat -> a -> r
f sn@SNat x = ...
```

i.e. pattern-match on the `SNat` constructor to bring the `KnownNat` constraint into scope.

christiaanb: Something you already seem to be doing in all of your instances... so I'm not sure why the class method has this `KnownNat n` constraint?

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -26,7 +26,7 @@ normalization = rmDeadcode >-> constantPropgation >-> etaTL >-> rmUnusedExpr >-!
     rmUnusedExpr = bottomupR (apply "removeUnusedExpr" removeUnusedExpr)
     rmDeadcode = bottomupR (apply "deadcode" deadCode)
     bindConst  = topdownR (apply "bindConstantVar" bindConstantVar)
-    evalConst  = topdownR (apply "evalConst" reduceConst)
+    evalConst  = bottomupR (apply "evalConst" reduceConst)
```
christiaanb: Add the reason for this bottomup traversal as a comment/note `-- See [Note] bottomup traversal evalConst`.

Also, I don't completely agree with the nuance of the commit message. Basically, there are two issues:

1. 2-May-2019: There is a bug in the evaluator where all data constructors are considered lazy, even though their declaration says they have strict fields. This causes some reductions to fail because the term under the constructor is not in WHNF, which is what some of the evaluation rules for certain primitive operations expect. Using a bottom-up traversal works around this bug by ensuring that the values under the constructor are in WHNF
2. Using a bottomup traversal ensures that constants are reduced to NF, even if constructors are lazy, thus ensuring more sensible/smaller generated HDL.

clash-prelude/src/Clash/Class/Exp.hs
martijnbastiaan: Ah darn, I guess that's a left-over from when I was using `Unsigned n`! Thanks!

clash-lib/src/Clash/Normalize/Strategy.hs
martijnbastiaan: I see. I should have explicitly stated that the top-down traversal in and of itself is not a problem and that it only becomes a problem if you don't account for strictness annotations. Thanks!

martijnbastiaan: Just so we're clear, these are the only changes you'd like to see? You're fine with merging even with the issues I outlined in the OP?

christiaanb: Yes, we can fix them later.

PR #597, opened 2019-05-03
Catch DivideByZero exceptions in the Evaluator

### LINKED ISSUE #573: divide by zero error when compiling to verilog/vhdl
When compiling 'TopEntity.hs' (see attachment) to verilog or vhdl, clash exits with the following error:
```
<no location info>: error:
    Other error:
    divide by zero
```
We can run and sample topEntity with clashi, so it passes the typecheck.
We tried this with commit fff460, built with stack, and clash 0.99.3, built with cabal.
[Divide_by_zero_error.zip](https://github.com/clash-lang/clash-compiler/files/3053379/Divide_by_zero_error.zip)

leonschoorl:
Fixes #573

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
@@ -22,6 +22,8 @@ module Clash.GHC.Evaluator where
 
 import           Control.Applicative (liftA2)
 import           Control.Concurrent.Supply  (Supply,freshId)
+import           Control.DeepSeq            (force)
+import           Control.Exception          (ArithException(..), Exception, tryJust, evaluate)
 import           Control.Monad              (ap)
 import           Control.Monad.Trans.Except (runExcept)
 import           Data.Bits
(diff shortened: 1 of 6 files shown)


martijnbastiaan:
Is there a reason for catching the error (that is, doing unsafe IO) over just checking for zero?

leonschoorl:
We have to implement each primitive five times: in the three different in HDLs, in Haskell and in the evaluator.
I think we should reuse the Haskell implementation in the evaluator wherever possible.

And my idea was to (eventually) move the catching of the DivideByZero exception out the specific primitives and next to where the XExceptions are caught: https://github.com/clash-lang/clash-compiler/blob/2c8a3c96a9387d5683e111a5dc853c5d4a5a4440/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs#L3259-L3260

martijnbastiaan:
> I think we should reuse the Haskell implementation in the evaluator wherever possible.

@christiaanb mentioned this morning he wanted to work on this

PR #600, opened 2019-05-10
Fix clock period values

leonschoorl:
Our documentation states that clock periods are in picoseconds, but this wasn't actually true.
Generated VHDL had them in tens of picoseconds and (System)Verilog had them in nanoseconds.

Now they are actually in picoseconds.

Only problems is that some VHDL simulators (ModelSim and Vivado) always seem to calculate time in an integer number of picoseconds, even though in Verilog mode they'll handle it just fine.
This means that can't handle a clock period of 1ps.
And when your period isn't divisible by two that generated clock won't have an exact 50% duty cycle.

But since 1ps == 1THz, this probably won't be a problem any time soon.


martijnbastiaan:
I've fixed some linebreak issues in the commit messages causing them to be hard to read when using `git log`. Other than that LGTM, will merge soon :).

PR #612, opened 2019-06-03
Increase a lower bound to temporary-1.2.1 instead of 1.2.

judah:
Version 1.2.0.* of the temporary package doesn't  contain the function
`getCanonicalTemporaryDirectory`.


christiaanb:
Thanks!

PR #614, opened 2019-06-04
AppPropFast: inline, instead of let-bind, work-free arguments

christiaanb:
Normally `AppPropFast` let-binds arguments that are non-constant, because they might perform work, which we do not want to duplicate.

This patch allows us to describe exactly entails "work", including all the primitives. AppPropFast uses this information to determine whether to inline (substitute) or let-bind arguments,instead of using "constantness" as a heuristic.


christiaanb:
@martijnbastiaan Could you rebase on top of this PR to check whether it fixes the "not in normal form" issue you were having with the new prelude API?

PR #617, opened 2019-06-06
Make `eqArg` recognize more complex cases

martijnbastiaan:
In order for `recToLetRec` to do its work, it needs to determine if constructs are equal to some variable. For example, it needs to know that:

```a ~ (fst a, snd a)```

This PR adds the ability for `eqArg` to determine these properties for nested projections as well as type classes (which term-level constructs are equal if their types are equal).

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations.hs
@@ -13,6 +13,7 @@
 {-# LANGUAGE FlexibleContexts  #-}
 {-# LANGUAGE LambdaCase        #-}
 {-# LANGUAGE MagicHash         #-}
+{-# LANGUAGE MultiWayIf        #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell   #-}
 {-# LANGUAGE TupleSections     #-}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: clash-lib/src/Clash/Normalize/Transformations.hs [hunk 1]
```diff
@@ -1618,20 +1642,60 @@ recToLetRec (TransformContext is0 []) e = do
...
 
-    -- `isNthProjection s n c` checks that `c` is the `n`th projection
-    -- of `s`.
-    isNthProjection :: Term -> Int -> Term -> Bool
-    isNthProjection v n (Case v' _ [(DataPat _ _ xs, Var s)])
-      | v == v'
-      , Just n' <- List.elemIndex s xs
-      = n == n'
-    isNthProjection _ _ _ = False
+    -- Recursively check if data construction is equal to some variable.
```
christiaanb:
```whether
-- Recursively check whether a term /e/ is semantically equal to some variable /v/.
-- Currently it can only assert equality when /e/ is  syntactically equal
-- to /v/, or is constructed out of projections of /v/, importantly:
--
-- [Note: Breaks on constants and predermined equality]
-- ...
```

PR #620, opened 2019-06-06
Up version numbers to disambiguate dev from release

christiaanb:


martijnbastiaan:
What do you think of following the same scheme GHC does? `1.0` as release, `1.1` dev, `1.2` release, etc.

christiaanb:
I agree, but we don't have a 1.0 yet

PR #625, opened 2019-06-07
Improve (?) error message of `assert`

martijnbastiaan:
I've always thought the error message `assert` produces is a bit non-obvious. It, IMO, takes some time to understand what the clock(period) is. It additionally has some problems with visual alignment whenever values grow/shrink. Both issues still occasionally trip me up. This PR addresses my problems. I've labeled it as discussion, suggestions more than welcome.

Before:

```
>>> sampleN 8 testBench
[False,False,False,False,False
cycle(system10000): 5, outputVerifier
expected value: 14, not equal to actual value: 30
,False
cycle(system10000): 6, outputVerifier
expected value: 14, not equal to actual value: 46
,False
cycle(system10000): 7, outputVerifier
expected value: 14, not equal to actual value: 62
,False]
```

After:

```
>>> sampleN 8 testBench
[False,False,False,False,False
On cycle 5 of clock 'system10000', outputVerifier encountered an unexpected value:
  Expected: 14
  Actual:   30
,False
On cycle 6 of clock 'system10000', outputVerifier encountered an unexpected value:
  Expected: 14
  Actual:   46
,False
On cycle 7 of clock 'system10000', outputVerifier encountered an unexpected value:
  Expected: 14
  Actual:   62
,False]
```

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Testbench.hs
+++ b/clash-prelude/src/Clash/Explicit/Testbench.hs
@@ -62,14 +62,10 @@ assert clk _rst msg checked expected returned =
   (\c e cnt r ->
       if eqX c e
          then r
-         else trace (concat [ "\ncycle(" ++ show clk ++ "): "
-                            , show cnt
-                            , ", "
-                            , msg
-                            , "\nexpected value: "
-                            , showX e
-                            , ", not equal to actual value: "
-                            , showX c
+         else trace (concat [ "\nOn cycle ", show cnt, " of clock '", show clk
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
I'd like to note that what we really should do is change `assert` such that it saves up wrong values, and spits them out at the very end. This would give us much nicer formatting options. That's more work than just changing the message though :smile:  .

leonschoorl:
We should do something similar for the HDL assertions

martijnbastiaan:
Closing due to inactivity. I'll open an issue report.

PR #639, opened 2019-06-16
Fix `gUnpack` constructor index bug in `instance GBitPack (f :+: g)`

liyang:
The implementation of `gUnpack` in `instance GBitPack (f :+: g)` is returning the wrong `Rep` for larger sum types, and was only correct by coincidence for smaller (<= 3) numbers of constructors. Here's a minimal example:
```
module BitPackGeneric where

import Clash.Prelude

data T = A | B | C | D
    deriving (Generic, Enum, Bounded, Show, BitPack)

main :: IO ()
main = flip mapM_ [minBound .. maxBound :: T] $ \ t -> do
    let p = pack t
    print (t, p, unpack p :: T)
```
which outputs:
```
*BitPackGeneric> main
(A,00,A)
(B,01,C)
(C,10,C)
(D,11,D)
```

I think this change fixes it. (Ping @martijnbastiaan.)

### DIFF
--- a/clash-prelude/src/Clash/Class/BitPack.hs
+++ b/clash-prelude/src/Clash/Class/BitPack.hs
@@ -326,10 +326,10 @@ instance ( KnownNat (GFieldSize g)
     (sc, packed ++# padding)
 
   gUnpack c cc b =
-    if c <= cc then
+    let cLeft = snatToNum (SNat @(GConstructorCount f)) in
+    if c < cc + cLeft then
       L1 (gUnpack c cc f)
     else
-      let cLeft = snatToNum (SNat @(GConstructorCount f)) in
       R1 (gUnpack c (cc + cLeft) g)
 
    where


christiaanb:
Good catch, thanks! This was only recently added, so most of the code out there had hand-rolled instances for sum-types and is probably why we didn’t see it earlier.

martijnbastiaan:
Thanks @liyang! I'll see if I can either add a quicktest or something like https://github.com/clash-lang/clash-compiler/blob/41612fc6e4e401c084a3fd777d5d0a6634a1a1b7/clash-prelude/tests/Clash/Tests/Undefined.hs before merging to squat other dormant bugs in `pack`/`unpack`.

martijnbastiaan:
I've written a few tests, and your fix seems to work perfectly fine! I've opened another PR because I couldn't push to your master branch (usually, github allows maintainers of repositories you make PRs to, to push to that particular branch - apparently excluding master): https://github.com/clash-lang/clash-compiler/pull/642. I expect to merge that one soon, so I'll close this one.

Thanks again @liyang !

Issue #588, opened 2019-04-23
Clash should throw an error if it can't find function mentioned in primitive annotations


adamwalker:
Compiling this file: https://gist.github.com/adamwalker/f7d0afc30c416a43807b771509a91ba0
Results in this output: https://gist.github.com/adamwalker/5ab19410c8be6e67c9514f3a33025bbc

Notice that the arithmetic operation is a subtraction instead of an addition, as specified in the primitive annotation. 

This example is almost exactly the InlinePrimitive example from the documentation. The only difference is that I've swapped the addition for a subtraction in the clash code only and it lives in the top level file, not a module. 

I'm using the latest master (6891e6)

leonschoorl:
When you don't explicitly name your module, GHC will give it the name `Main`.
(This is why clash will create a it's output in a directory called Main)

So the fully qualified name of example is `Main.example`:
```diff
@@ -7,5 +7,5 @@
 {-# ANN example (InlinePrimitive VHDL $ unindent [i|
   [ { "BlackBox" :
-      { "name" : "example"
+      { "name" : "Main.example"
       , "kind": "Declaration"
       , "template" :
```

leonschoorl:
Ideally you should need write the name field for inline prims.
Clash should be able to infer the name from the fact that it's an annotation attached to the function.

martijnbastiaan:
Alternatively, it should throw and error if it can't find `example`.

martijnbastiaan:
For inspiration, look at: https://github.com/clash-lang/clash-compiler/pull/514

adamwalker:
Aha. Thanks. Dumb mistake on my part. 

Yeah, it would be nice if it threw an error. The file I provided is broken, but compilation succeeds anyway.

christiaanb:
Fixed by #641

martijnbastiaan:
#641 isn't ready to be merged yet! I'll work on it shortly.

martijnbastiaan:
Fixed in https://github.com/clash-lang/clash-compiler/pull/641.

PR #641, opened 2019-06-17
Throw an error if inline primitives refer to wrong primitive

martijnbastiaan:
Fixes issue #588. It should be noted that this merely "fixes" the issue
in the sense that users will be told to correct their mistake, while
they shouldn't have been able to make the mistake in the first place. We
should seriously think about ditching json-primitives entirely, and
offer a much more ergonomic API.

As a happy side-effect, this removes the inline_primitives hack that was
around to support inline primitives. ~~This commit also explicitly removes
the ability to use file references in inline primitives, as we can't
reliably determine where they are.~~

### DIFF
--- a/Clash.hs
+++ b/Clash.hs
@@ -19,9 +19,7 @@ import Clash.Annotations.BitRepresentation.Internal (buildCustomReprs)
 import Clash.Util
 
 import Control.DeepSeq
-import Control.Exception (finally)
 import qualified Data.Time.Clock as Clock
-import System.Directory (removeDirectoryRecursive)
 
 import GHC.Stack (HasCallStack)
 import Util (OverridingBool(..))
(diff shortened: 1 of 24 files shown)


christiaanb:
@martijnbastiaan What do you mean that we cannot reliably determine where files are? If I add them to the `data-files` stanza in the `.cabal` file, and reference them as we do in https://github.com/clash-lang/clash-compiler/blob/3787ce7ff71531210a6c112284cd486e03ff2ed3/clash-lib/src/Clash/Backend/SystemVerilog.hs#L100-L105 + some TemplateHaskell IO, then their location is completely unambiguous, no?

martijnbastiaan:
>  + some TemplateHaskell IO, then their location is completely unambiguous, no?

Yes, you're right!

martijnbastiaan:
I've removed the error. PR should be pretty much done now.

PR #655, opened 2019-06-28
 Reliably handle hidden resets in 'withReset' and siblings 

martijnbastiaan:
I was bitten by the surprising behavior of `withReset` today. This led @leonschoorl and I chasing lots of red herrings, even though -in retrospect- I had seen the issue before in an only slightly different context. The commit below describes what we did in order to mitigate the issues we found. Note that we don't have a satisfactory answer for bundled types yet.. 

-----------------

Before this commit, 'withReset' would behave very surprisingly in
combination with 'simulate' or 'sample':

    >>> rst0 = fromList [True, True, False, False, True, True]
    >>> rst1 = unsafeFromHighPolarity @System rst0
    >>> sig = withReset rst1 (register 'a' (pure 'b'))
    >>> sampleN @System 6 sig
    "aabbbb"

Even though rst1 is asserted on the 5th and 6th clock cycle, we never
see the reset value, a, after the initial two. But even the first two
as are suspicious. We'd expect to see three as (one powerup, two
reset) instead! It seems that 'sampleN' is generating a default reset
value and using that one. Indeed, when we inspect the type of sig we
see it is as we never applied the reset at all:

    >>> :t sig
    (HiddenReset dom, ..) => Signal dom Char

To figure out what's going on, let's inspect the type signature of
'withReset':

    withReset
      :: KnownDomain dom
      => Reset dom
      -> (HiddenReset dom => r)
      -> r

It turns out that GHC is perfectly happy to /only/ resolve r when
given the following as its second argument:

    reg :: (HiddenReset dom, ..) => Signal dom Char
    reg = register 'a' (pure 'b')

i.e., withReset would conceptually resolve to something like:

    withReset*
      :: forall dom r
       . KnownDomain dom
      => Reset dom
      -> (HiddenReset dom  => (HiddenReset dom0 => Signal dom0 Char))
      -> (HiddenReset dom0 => Signal dom0 Char)

thus completely discarding its given reset, and leaving it up to
sample to actually generate the reset. The reason GHC is free to do
this, is that nothing in r ties dom to the "dom inside r" simply
because it's not mentioned in the type signature.

This commit adds a few variants of 'withReset' that explicitly take
an argument, thus "tying" the `dom` in that argument to the result of
the function given to it.

'withReset0' behaves exactly like we expect it would:

    >>> rst0 = fromList [True, True, False, False, True, True]
    >>> rst1 = unsafeFromHighPolarity rst0
    >>> sig = withReset0 rst1 (register 'a' (pure 'b'))
    >>> sampleN @System 6 sig
    "aaabaa"

### DIFF
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -291,6 +291,7 @@ test-suite unittests
                  Clash.Tests.BitVector
                  Clash.Tests.DerivingDataRepr
                  Clash.Tests.DerivingDataReprTypes
+                 Clash.Tests.Signal
                  Clash.Tests.Undefined
 
 
--- a/clash-prelude/src/Clash/Signal.hs
+++ b/clash-prelude/src/Clash/Signal.hs
@@ -895,12 +895,7 @@ sample
   -- (and reset)
   -> [a]
 sample s =
-  case knownDomain @dom of
-    SDomainConfiguration dom _ _ _ _ _ ->
-      let clk = Clock dom in
-      let rst = Reset (True :- pure False) in
-      let en = S.enableGen in
-      S.sample (exposeClockResetEnable s clk rst en)
+  S.sample (exposeClockResetEnable s clockGen (resetGen @dom) enableGen)
 {-# NOINLINE sample #-}
 
 -- | Get a list of /n/ samples from a 'Signal'
@@ -921,13 +916,9 @@ sampleN
   -- ^ 'Signal' we want to sample, whose source potentially has a hidden clock
   -- (and reset)
   -> [a]
-sampleN n s =
-  case knownDomain @dom of
-    SDomainConfiguration dom _ _ _ _ _ ->
-      let clk = Clock dom in
-      let rst = Reset (True :- pure False) in
-      let en = S.enableGen in
-      S.sampleN n (exposeClockResetEnable s clk rst en)
+sampleN n s0 =
+  let s1 = exposeClockResetEnable s0 clockGen (resetGen @dom) enableGen in
+  S.sampleN n s1
 {-# NOINLINE sampleN #-}
 
 -- | /Lazily/ get an infinite list of samples from a 'Clash.Signal.Signal'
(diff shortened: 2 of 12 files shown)


christiaanb:
The issue seems to be that `AppendSymbol` is not injective, so when we try to type-check:

```haskell
{-# LANGUAGE DataKinds, TypeFamilies, TypeOperators #-}
module Test where

import GHC.TypeLits
import Data.Proxy

f :: (AppendSymbol dom "_rst" ~ "system_rst")
  => Proxy "system"
  -> Proxy dom
f = id
```

we get:

```
Test.hs:10:5: error:
    • Could not deduce: dom ~ "system"
      from the context: AppendSymbol dom "_rst" ~ "system_rst"
                   f :: forall (dom :: Symbol).
                        (AppendSymbol dom "_rst" ~ "system_rst") =>
                        Proxy "system" -> Proxy dom
        at Test.hs:(7,1)-(9,14)
      ‘dom’ is a rigid type variable bound by
        the type signature for:
          f :: forall (dom :: Symbol).
               (AppendSymbol dom "_rst" ~ "system_rst") =>
               Proxy "system" -> Proxy dom
        at Test.hs:(7,1)-(9,14)
      Expected type: Proxy "system" -> Proxy dom
        Actual type: Proxy dom -> Proxy dom
    • In the expression: id
      In an equation for ‘f’: f = id
    • Relevant bindings include
        f :: Proxy "system" -> Proxy dom (bound at Test.hs:10:1)
```

I think I can write a constraint-solver plugin to correct this behavior, hopefully obviating the need for this PR

martijnbastiaan:
Ah alright. Seems like I misidentified the real issue then! I'm wondering if I got the semantics of constraints right though? I.e., would GHC be free to conceptually do this:

```
withReset*
  :: forall dom r
   . KnownDomain dom
  => Reset dom
  -> (HiddenReset dom  => (HiddenReset dom0 => Signal dom0 Char))
  -> (HiddenReset dom0 => Signal dom0 Char)
```

even if it could deduce `dom ~ dom0`? That is, even though it might not do it in practice, would it be free to drop the first argument and simply propagate the `HiddenReset` constraint?

christiaanb:
If you're asking whether you can write `withReset*`, then probably yes. Class instances are unique, so if you solve one class constraint now, and another exactly the same at a later moment, the program will behave exactly the same. Now, implicit parameter constraints are different, since they can be instantiated to different values. However, the tag of an implicit parameter should be monomorphic, otherwise the behavior is "undefined". Since we're abusing implicit parameters using TypeFamilies and type variables inside of the tag and whatnot, we are in this "undefined" behavior territory.

martijnbastiaan:
Right, that's what I thought. Thanks!

christiaanb:
Okay, I implemented the constraint-solver plugin: https://github.com/clash-lang/ghc-typelits-appendsymbol Sadly, it didn’t fix the issue. Looking at the type-checker trace, I don’t thinks it’s solvable using a constraint-solver plugin (modulo questionable implementations), but perhaps a source plugin can fix it, I’ll investigate.

martijnbastiaan:
Ah! I was just trying to implement it myself :). Cool to see a simple real-world typechecker plugin. If it doesn't work out tonight, maybe we can take a look at it together tomorrow.

PR #669, opened 2019-07-14
Add type classes to reliably use hidden constructs

martijnbastiaan:
Rewritten: 2019-07-17

----------------------------------

This is a follow-up of PR #655. That PR identified unexpected behavior
when using 'withReset' in combination with 'sample'. It originally
proposed a solution with a type class "HasDomain" that solved the
unexpected behavior, but potentially introduced confusing error reports
if misused. We therefore settled on a simpler, though more cumbersome,
way of dealing with the issue: we introduced 'withReset0', 'withReset' -
that only worked on Signal dom a -> r -, and 'withResetProxy'. After
testing this change on larger codebases we concluded that this wasn't
satisfactory either. Upgrading to the new API still triggered confusing
errors and the new API seemed worse for no obvious reason.

This commit revisits the originally proposed solution. Once more, we
introduce type classes: HasSingleDomain and HasSpecificDomain. Both
type classes aim to tie the domain provided by the explicitely passed
clock, reset, or enable lines to the domain in the given construct.
This change differs from the original solution in that it tries to
generate clear error messages if the functions using it can't resolve
the constraint. For example, if a user tries to use 'withReset' on a
function whose first argument does not contain a domain, they will
get an error message explaining what happened:

	>>> withReset resetGen (3 :: Int)
	error:
	    • Could not find a domain in the following type:

		Int

	      This might happen if an instance for TryDomain is missing. Try to determine
	      which of the types miss an instance, and add them. Example implementations:

	       * type instance HasDomain t (MyVector n a)    = TryDomain t a
	       * type instance HasDomain t (MyCircuit dom a) = Found dom
	       * type instance HasDomain t Terminal          = NotFound

	      Alternatively, use one of the withSpecific* functions.

	      ------

	      You tried to apply an explicitely routed clock, reset, or enable line
	      to a construct with, possibly, an implicitely routed one. Clash failed to
	      unambigously determine a single domain and could therefore not route it.
	      You possibly used one of these sets of functions:

	       * with{ClockResetEnable,Clock,Reset,Enable}
	       * expose{ClockResetEnable,Clock,Reset,Enable}

	      These functions are suitable for components defined over a single domain
	      only. If you want to use multiple domains, use the following instead:

	       * withSpecific{ClockResetEnable,Clock,Reset,Enable}
	       * exposeSpecifc{ClockResetEnable,Clock,Reset,Enable}

	    • In the expression: withReset resetGen (3 :: Int)
	      In an equation for ‘it’: it = withReset resetGen (3 :: Int)

Error messages will try to be specific, although there is still a lot of
work to do.

	>>> withReset resetGen (Prelude.undefined :: Int -> Index n -> Signal XilinxSystem b)

	error:
	    • Could not determine whether the following type contained a domain:

		Index n

	      In the full type:

		Int -> Index n -> Signal XilinxSystem b

	      This might happen if an instance for TryDomain is missing. Try to determine
	      which of the types miss an instance, and add them. Example implementations:

	       * type instance HasDomain t (MyVector n a)    = TryDomain t a
	       * type instance HasDomain t (MyCircuit dom a) = Found dom
	       * type instance HasDomain t Terminal          = NotFound

	      Alternatively, use one of the withSpecific* functions.

	      ------

	      You tried to apply an explicitely routed clock, reset, or enable line
	      to a construct with, possibly, an implicitely routed one. Clash failed to
	      unambigously determine a single domain and could therefore not route it.
	      You possibly used one of these sets of functions:

	       * with{ClockResetEnable,Clock,Reset,Enable}
	       * expose{ClockResetEnable,Clock,Reset,Enable}

	      These functions are suitable for components defined over a single domain
	      only. If you want to use multiple domains, use the following instead:

	       * withSpecific{ClockResetEnable,Clock,Reset,Enable}
	       * exposeSpecifc{ClockResetEnable,Clock,Reset,Enable}

	    • When checking the inferred type [..]

Two `withReset` (and Clock/Enable) strategies now exist:

 1. `withReset`: will only work if it can determine a _single_ domain
    in a given type. Allows given reset domain to be inferred. That is
    a call such as `withReset resetGen f` will work as long as `f`
    contains a single domain.

 2. `withSpecific`: will look for a _specific_ domain in a given type.
    That is, the domain of the given clock/reset/enable must be known
    and present in the given type.

These two strategies should strike a happy balance between simple for
beginners, and fully usable for advanced users. Most importantly, it's
always safe to use.

### DIFF
--- a/clash-prelude/src/Clash/Class/HasDomain/CodeGen.hs
+++ b/clash-prelude/src/Clash/Class/HasDomain/CodeGen.hs
@@ -1,0 +1,66 @@
+{-|
+Copyright  :  (C) 2019, Myrtle Software Ltd
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+-}
+
+{-# LANGUAGE ViewPatterns #-}
+
+module Clash.Class.HasDomain.CodeGen
+  ( mkTryDomainTuples
+  , mkHasDomainTuples
+  ) where
+
+import           Language.Haskell.TH.Syntax
+
+mkTup :: [Type] -> Type
+mkTup names@(length -> n) =
+  foldl AppT (TupleT n) names
+
+-- | Creates an instance of the form:
+--
+--  type instance TryDomain t (a, b, c, d, e) = Merge t a (b, c, d, e)
+--
+-- With /n/ number of variables on the LHS.
+mkTryDomainTupleInstance :: Name -> Name -> Int -> Dec
+mkTryDomainTupleInstance tryDomainName mergeName n =
+  TySynInstD tryDomainName (TySynEqn [t, tupPat] tupBody)
+ where
+  bcde = map (VarT . mkName . ("a"++) . show) [1..n-1]
+  a    = VarT (mkName "a0")
+  t    = VarT (mkName "t")
+
+  -- Merge t a (b, c, d, e)
+  tupBody = ConT mergeName `AppT` t `AppT` a `AppT` (mkTup bcde)
+
+  -- (a, b, c, d, e)
+  tupPat = mkTup (a : bcde)
+
+mkTryDomainTuples :: Name -> Name -> Q [Dec]
+mkTryDomainTuples tryDomainName mergeName =
+  pure (map (mkTryDomainTupleInstance tryDomainName mergeName) [3..62])
+
+
+-- | Creates an instance of the form:
+--
+--  type instance HasDomain' dom (a, b, c, d, e) =
+--    Merge' (HasDomain' dom a) (HasDomain' dom (b, c, d, e))
+--
+-- With /n/ number of variables on the LHS.
+mkHasDomainTupleInstance :: Name -> Name -> Int -> Dec
+mkHasDomainTupleInstance hasDomainName mergeName n =
+  TySynInstD hasDomainName (TySynEqn [dom, tupPat] merge)
+ where
+  bcde = map (VarT . mkName . ("a"++) . show) [1..n-1]
+  a    = VarT (mkName "a0")
+  dom  = VarT (mkName "dom")
+
+  -- Merge dom a (b, c, d, e)
+  merge = ConT mergeName `AppT` dom `AppT` a `AppT` mkTup bcde
+
+  -- (a, b, c, d, e)
+  tupPat = mkTup (a : bcde)
+
+mkHasDomainTuples :: Name -> Name -> Q [Dec]
+mkHasDomainTuples hasDomainName mergeName =
+  pure (map (mkHasDomainTupleInstance hasDomainName mergeName) [3..62])
--- a/clash-prelude/src/Clash/Class/HasDomain/Common.hs
+++ b/clash-prelude/src/Clash/Class/HasDomain/Common.hs
@@ -1,0 +1,46 @@
+{-|
+Copyright  :  (C) 2019, Myrtle Software Ltd
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+-}
+
+{-# LANGUAGE DataKinds          #-}
+{-# LANGUAGE PolyKinds          #-}
+{-# LANGUAGE TypeFamilies       #-}
+{-# LANGUAGE TypeOperators      #-}
+
+module Clash.Class.HasDomain.Common
+  ( Unlines
+  , (:<<>>:)
+  , (:$$$:)
+  , (:++:)
+  ) where
+
+import           GHC.TypeLits               (Symbol)
+import           Type.Errors
+  (ErrorMessage(Text, ShowType, (:<>:), (:$$:)))
+
+type family ToEM (k :: t) :: ErrorMessage where
+  ToEM (k :: Symbol)       = 'Text k
+  ToEM (k :: ErrorMessage) = k
+  ToEM (k :: t)            = 'ShowType k
+
+infixl 5 :<<>>:
+type (:<<>>:) (k1 :: t1) (k2 :: t2) = ToEM k1 ':<>: ToEM k2
+
+infixl 4 :$$$:
+type (:$$$:) (k1 :: t1) (k2 :: t2) = ToEM k1 ':$$: ToEM k2
+
+
+{- | Combine multiple lines with line break. Type-level version of the @unlines@
+function but for ErrorMessage. -}
+type family Unlines (ln :: [k]) :: ErrorMessage where
+  Unlines '[] = 'Text ""
+  Unlines ((x :: Symbol) ': xs) = 'Text x ':$$: Unlines xs
+  Unlines ((x :: ErrorMessage) ': xs) = x ':$$: Unlines xs
+
+infixl 4 :++:
+type family (:++:) (as :: [k]) (bs :: [k]) :: [k] where
+  (:++:) a '[] = a
+  (:++:) '[] b = b
+  (:++:) (a ': as) bs = a ': (as :++: bs)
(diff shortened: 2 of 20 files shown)


.circleci/config.yml
```diff
@@ -84,10 +100,21 @@ aliases:
...
-      - *setup
       - *cache_restore
       - *stack_build
       - *cache_save
+  - &build_with_nix
+    docker:
+      - image: nixos/nix
+    steps:
+      # I can't get caching to work on Nix due to missing certificates. Doesn't
+      # seem to bad though, thanks to nix's binary cache of 99% of Clash's
```
cchalmers:
```suggestion
      # seem too bad though, thanks to nix's binary cache of 99% of Clash's
```

cchalmers: I use [cachix for diagrams](https://github.com/cchalmers/diagrams-dev/blob/master/.travis.yml) if you're interested. I kinda nice because users don't have to build anything.

clash-prelude/src/Clash/Class/HasDomain/CodeGen.hs
```diff
@@ -0,0 +1,36 @@
...
+domainEq :: String -> String -> String
+domainEq a b = a ++ " `DomainEq` " ++ b
+
+inst :: WhichDomain -> Integer -> IO ()
+inst d n = do
+  case d of
+    HasDomain -> do
+      putStrLn $ "instance (" ++ constraints ++ ") => HasDomain (" ++ tup ++ ") where"
+      putStrLn $ "  type GetDomain (" ++ tup ++ ") = GetDomain " ++ head vars
+      putStrLn ""
```
cchalmers: I feel like this is something you can do fairly easily with template Haskell without having to add ~700 lines of generated code to git.

clash-prelude/src/Clash/Class/HasDomain/Common.hs
```diff
@@ -0,0 +1,100 @@
...
+   , ""
+   , " * withClockResetEnableProxy, withClockProxy, withResetProxy, withEnableProxy"
+   , " * exposeClockResetEnableProxy, exposeClockProxy, exposeResetProxy, exposeEnableProxy"
+   ]
+
+type NoHasDomainInstance = Unlines
+   [ "You tried to apply an explicitely routed clock, reset, or enable line"
+   , "to a construct with an implicitely routed one. Clash failed to"
+   , "(unambigously) determine its domain and could therefore not route it."
+   , "You possibly used one of these sets of functions:"
```
cchalmers: A duplicated paragraph you could pull out to another type alias?

cchalmers:
Look like in interesting solution, I like the use of the custom error message.

clash-prelude/src/Clash/Class/HasDomain/CodeGen.hs
martijnbastiaan: Hmmyeah, I tried it with template Haskell, but it's _so slow_. I do mention this in `HasDomain{N,L,R}`:

```
-- Tuple instances. Generated with 'HasDomain/CodeGen.hs". Purposely generated
-- with a script to prevent long Template Haskell compile times.
```

but that comment should probably be added to `CodeGen.hs` too.

.circleci/config.yml
martijnbastiaan: Huh, interesting. I'll look into it (probably not for this PR though).

clash-prelude/src/Clash/Class/HasDomain/CodeGen.hs
martijnbastiaan: I ran some tests this morning: reducing the number of tuple instances that are generated by Template Haskell from 63 to 7 (in Bitpack/Bundle) halves the compile time. So I'll probably add a CPP flag developers can override in their `cabal.project.local` indicating how many instances should be generated..

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
```diff
@@ -0,0 +1,179 @@
...
+  :$$$: " * type instance HasDomain t (MyCircuit dom a) = Found dom"
+  :$$$: " * type instance HasDomain t Terminal          = NotFound"
+  :$$$: ""
+  :$$$: "Alternatively, use one of the withSpecific* functions."
+
+type Outro =
+         ""
+   :$$$: "------"
+   :$$$: ""
+   :$$$: "You tried to apply an explicitely routed clock, reset, or enable line"
```
acairncross:
```suggestion
   :$$$: "You tried to apply an explicitly routed clock, reset, or enable line"
```

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
```diff
@@ -0,0 +1,179 @@
...
+   :$$$: "to a construct with, possibly, an implicitely routed one. Clash failed to"
```
acairncross:
```suggestion
   :$$$: "to a construct with, possibly, an implicitly routed one. Clash failed to"
```

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
```diff
@@ -0,0 +1,179 @@
...
+   :$$$: "unambigously determine a single domain and could therefore not route it."
```
acairncross:
```suggestion
   :$$$: "unambiguously determine a single domain and could therefore not route it."
```

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
acairncross: Same applies to a couple of other modules

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
martijnbastiaan: And my editor even highlights spelling mistakes :sweat: . Thanks!

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
```diff
@@ -0,0 +1,179 @@
...
+   :$$$: "You possibly used one of these sets of functions:"
+   :$$$: ""
+   :$$$: " * with{ClockResetEnable,Clock,Reset,Enable}"
+   :$$$: " * expose{ClockResetEnable,Clock,Reset,Enable}"
+   :$$$: ""
+   :$$$: "These functions are suitable for components defined over a single domain"
+   :$$$: "only. If you want to use multiple domains, use the following instead:"
+   :$$$: ""
+   :$$$: " * withSpecific{ClockResetEnable,Clock,Reset,Enable}"
+   :$$$: " * exposeSpecifc{ClockResetEnable,Clock,Reset,Enable}"
```
JvWesterveld:
```suggestion
   :$$$: " * exposeSpecific{ClockResetEnable,Clock,Reset,Enable}"
```

martijnbastiaan: Thank you!

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
```diff
@@ -0,0 +1,179 @@
...
+  :$$$: ""
+  :$$$: "Alternatively, use one of the withSpecific* functions."
+
+type Outro =
+         ""
+   :$$$: "------"
+   :$$$: ""
+   :$$$: "You tried to apply an explicitly routed clock, reset, or enable line"
+   :$$$: "to a construct with, possibly, an implicitly routed one. Clash failed to"
+   :$$$: "unambigousley determine a single domain and could therefore not route it."
```
acairncross:
```suggestion
   :$$$: "unambigously determine a single domain and could therefore not route it."
```
Here's another! (appears twice)

PR #673, opened 2019-07-17
Add Empty InstDecl option

basile-henry:
This is useful to create an instanciation that doesn't specify "entity"
or "component". This is sometimes used in VHDL instantiations like the
ones in the Xilinx XPM libraries

### DIFF
--- a/clash-lib/src/Clash/Backend/VHDL.hs
+++ b/clash-lib/src/Clash/Backend/VHDL.hs
@@ -21,7 +21,7 @@
 module Clash.Backend.VHDL (VHDLState) where
 
 import           Control.Applicative                  (liftA2)
-import           Control.Lens                         hiding (Indexed)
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
Would this only be useful for VHDL, or for (System)Verilog too?

basile-henry:
@martijnbastiaan I forgot about the other two backends but after a quick look at both:

https://github.com/clash-lang/clash-compiler/blob/0303454baddd616d769fc9baef57027ca4780699/clash-lib/src/Clash/Backend/Verilog.hs#L529

https://github.com/clash-lang/clash-compiler/blob/0303454baddd616d769fc9baef57027ca4780699/clash-lib/src/Clash/Backend/SystemVerilog.hs#L899

It looks like they don't differentiate between components and entities anyway so I don't think there's anything to do for this new `Empty` constructor.

martijnbastiaan:
Alright, thanks!

PR #676, opened 2019-07-18
Add CPP flag to decrease number of TH generated tuple instances

martijnbastiaan:
By default, all tuple instances will still be build. This CPP flag gives developers the option to reduce the number of tuple instance will be generated. To do this, add the following to `cabal.project.local`:

```
package clash-prelude
  ghc-options: -DMAX_TUPLE_SIZE=7
```

This halves compilation time on my machine.


leonschoorl:
This is really nice
And I found that haddock also get about 2x faster when you add:
```
   haddock-options: --optghc=-DMAX_TUPLE_SIZE=7
```

Issue #678, opened 2019-07-22
Implement Bits class for Index type


rowanG077:
I was refactoring some code and changing some Unsigneds I'm using to Index since they better represent the use case. I'm using `shiftR x 1` as non-costly way to halve one of these unsigneds. But `Bits` is class is not implemented for Index.

Is there any specific reason I'm overlooking why the `Bits` class is not implemented for index?

leonschoorl:
Compared to other number types, `Index n` is somewhat special.

When `n` is not a power of 2 there exists bitpatterns which are not valid Index values.
In Haskell we check for these for out of range values and notify the user by throwing an exception:
```
Clash.Prelude> 9 + 1 :: Index 10
*** Exception: X: Clash.Sized.Index: result 10 is out of bounds: [0..9]
```
But in HDL there no such checks are done. You'll just get 10.

While this is a problem for arithmetic operations.
At least it's relatively easy to understand that `x + 9 :: Index n` is only  safe when `x <= (maxBound - 9)`. Also we have thing like [saturating arithmetic] to help.

But with bitwise operations it's very easy to get out of range.
When would it be safe to do `x .|. 9 :: Index n`?

A reason for not having an instance `Bits (Index n)` could be to discourage users from doing bitwise operation on Index values. 
Thereby hinting that maybe they'd better so another type like Unsigned if they're doing bitwise operations.

Of course it is perfectly safe to do `shiftR x 1 :: Index n`.
But I'm pretty sure you can rely on (almost all?) synthesis software to synthesize `shiftR x 1` and 
``x `div` 2`` to the same hardware implementation of a shift.

[Saturating arithmetic]: http://hackage.haskell.org/package/clash-prelude-0.99.3/docs/Clash-Class-Num.html#g:2

rowanG077:
Yeah you can get out of bounds very easily with `Index`. But I don't feel this is different then getting out of bounds with any other mathematical operation. Once you start doing bit operations you have to understand anyway what the underlying bit representation is. `Signed` has a `Bits` instance as well and there you also can get very weird result if you don't understand what is going on under the hood.

I'm not particularly married to having a bits instance if division will also work for my case. However I do think having the same API for `Unsigned` and `Index` would make sense because fundamentally they both are just unsigned number.

martijnbastiaan:
[...]

PR #679, opened 2019-07-22
Implemented size#, Bits class and FiniteBits class for Index type (#678)

rowanG077:
I copied most of the implementation from Unsigned.

### DIFF
--- a/clash-prelude/src/Clash/Sized/Internal/Index.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/Index.hs
@@ -36,6 +36,9 @@ module Clash.Sized.Internal.Index
     Index (..)
     -- * Construction
   , fromSNat
+  -- * Accessors
+  -- ** Length information
+  , size#
     -- * Type classes
     -- ** BitPack
   , pack#
@@ -74,6 +77,7 @@ module Clash.Sized.Internal.Index
 where
 
 import Control.DeepSeq            (NFData (..))
+import Data.Bits                  (Bits (..), FiniteBits (..))
 import Data.Data                  (Data)
 import Data.Default.Class         (Default (..))
 import Data.Proxy                 (Proxy (..))
@@ -94,7 +98,9 @@ import Clash.Class.BitPack        (BitPack (..), packXWith)
 import Clash.Class.Num            (ExtendingNum (..), SaturatingNum (..),
                                    SaturationMode (..))
 import Clash.Class.Resize         (Resize (..))
-import {-# SOURCE #-} Clash.Sized.Internal.BitVector (BitVector (BV),undefError)
+import Clash.Prelude.BitIndex     (replaceBit)
+import {-# SOURCE #-} Clash.Sized.Internal.BitVector (BitVector (BV), high, low, undefError)
+import qualified Clash.Sized.Internal.BitVector as BV
 import Clash.Promoted.Nat         (SNat, snatToNum, leToPlusKN)
 import Clash.XException
   (ShowX (..), Undefined (..), errorX, showsPrecXWith, rwhnfX)
@@ -127,6 +133,10 @@ newtype Index (n :: Nat) =
     -- synthesizable.
     I { unsafeToInteger :: Integer }
   deriving (Data, Generic)
+
+{-# NOINLINE size# #-}
+size# :: (KnownNat n, 1 <= n) => Index n -> Int
+size# = BV.size# . pack#
 
 instance NFData (Index n) where
   rnf (I i) = rnf i `seq` ()
@@ -336,6 +346,31 @@ quot#,rem# :: Index n -> Index n -> Index n
 {-# NOINLINE toInteger# #-}
 toInteger# :: Index n -> Integer
 toInteger# (I n) = n
+
+instance (KnownNat n, 1 <= n) => Bits (Index n) where
+  a .&. b           = unpack# $ BV.and# (pack# a) (pack# b)
+  a .|. b           = unpack# $ BV.or# (pack# a) (pack# b)
+  xor a b           = unpack# $ BV.xor# (pack# a) (pack# b)
+  complement        = unpack# . BV.complement# . pack#
+  zeroBits          = unpack# zeroBits
+  bit i             = unpack# $ bit i
+  setBit v i        = unpack# $ replaceBit i high (pack# v)
+  clearBit v i      = unpack# $ replaceBit i low  (pack# v)
+  complementBit v i = unpack# $ complementBit (pack# v) i
+  testBit v i       = testBit (pack# v) i
+  bitSizeMaybe v    = Just (size# v)
+  bitSize           = size#
+  isSigned _        = False
+  shiftL v i        = unpack# $ shiftL (pack# v) i
+  shiftR v i        = unpack# $ shiftR (pack# v) i
+  rotateL v i       = unpack# $ rotateL (pack# v) i
+  rotateR v i       = unpack# $ rotateR (pack# v) i
+  popCount i        = popCount (pack# i)
+
+instance (KnownNat n, 1 <= n) => FiniteBits (Index n) where
+  finiteBitSize        = size#
+  countLeadingZeros  i = countLeadingZeros  (pack# i)
+  countTrailingZeros i = countTrailingZeros (pack# i)
 
 instance Resize Index where
   resize     = resize#
--- a/tests/shouldwork/Numbers/NumConstantFolding.hs
+++ b/tests/shouldwork/Numbers/NumConstantFolding.hs
@@ -207,11 +207,12 @@ tBitVector16
     )
 
 tIndex
-  = ( cNum        @(Index 50000)
-    , cEq         @(Index 50000)
-    , cOrd        @(Index 50000)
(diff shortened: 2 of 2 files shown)


martijnbastiaan:
@rowanG077 Thanks for making the PR! 

I wonder if we could simply reuse the implementations for `Unsigned` directly. Something like (not tested):

```haskell
-- | Internal index to unsigned
i2s# :: Index n -> Unsigned (CLog 2 n)
i2s# = fromInteger . toInteger
{-# NOINLINE i2s# #-}  
{-# ANN hasBlackBox i2s# #-}  

-- | Internal unsigned to index
s2i# :: HasCallStack => Unsigned n -> Index m
s2i# = fromInteger_INLINE . toInteger
{-# NOINLINE s2i# #-}
{-# ANN hasBlackBox s2i# #-}  

instance (KnownNat n, 1 <= n) => Bits (Index n) where
  someUnaryOp a = s2i (U.someUnaryOp (i2s a))
  someBinaryOp a b = s2i (U.someBinaryOp (i2s a) (i2s b))
```

Where `i2s#` and `s2i#` would have "identity" HDL implementations. To make sure this constants folds, we should add it to the evaluator too. This way we wouldn't have to reimplement all the blackboxes for `Unsigned` for `Index` too, and we'd only have to write two evaluator rules (for `i2s#` and `s2i#`) to get constant folding for free.

What do you think?

leonschoorl:
Right now all Index operations that would generate out of range values throw exceptions in Haskell, so you know you've done something wrong.
```
Clash.Prelude> 9 + 1 :: Index 10
*** Exception: X: Clash.Sized.Index: result 10 is out of bounds: [0..9]
```

If we add extra operations on Index, those should also be protected.
This protection is done by [fromInteger_INLINE](https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-prelude/src/Clash/Sized/Internal/Index.hs#L240)
And Martijn's `s2i#` does that.

But we can make it even easier: if we reuse the implementations of BitVector we don't even have to write new conversion functions, because they already exist:
```Haskell
pack :: Index n -> BitVector (CLog 2 n)
unpack :: BitVector (CLog 2 n) -> Index n
```
And these are synthesizable, constant-folded and have the out-of-range exceptions built-in.

rowanG077:
Indeed using `pack#` and `unpack#` we would not have to create additional blackboxes. However if I look into the blackbox for `pack#` and `unpack#` they are not exactly identity functions. At least for VHDL it changes the signal from `std_logic_vector` to `unsigned` and vice versa. I don't think this will have impact after synthesis but I'm not sure.

martijnbastiaan:
@leonschoorl @rowanG077 I did consider using `pack`/`unpack`, but I settled for direct conversions to prevent simulation complexity (`BitVector`s contain a mask indicating unknown bits, that we're sure we'll never need in this case).

martijnbastiaan:
> If we can reuse the BitVector implementation for Index. We can also do the same for Unsigned which would cut down on the existing blackboxes and evaluator entries.

Makes sense to me. What do you think @leonschoorl, can you think of any issues this might cause?

leonschoorl:
@rowanG077 Just to explain a little bit what involved with primitive operations in general.

A primitive operation like `rotateR` on Unsigned has 1-5 different implementations.

These are:
- a haskell implementation:[Clash/Sized/Internal/Unsigned.hs]
  This is used when you're just simulating you design in Haskell.
  If this implementation can not be compiled to HDL by clash then we need to mark it `NOINLINE` and write blackboxes and evaluator rules.
- blackboxes for each of the HDLs we support: [prims/systemverilog/Clash_Sized_Internal_Unsigned.json], [prims/verilog/Clash_Sized_Internal_Unsigned.json], [prims/vhdl/Clash_Sized_Internal_Unsigned.json]
  These are manually written translations to HDL.
  And there can be 1-3 of these depending whether we can reuse the same HDL code for all HDLs(like [+#]), or the can share the Verilog and SystemVerilog implemantations(like [eq#] and [eq# vhdl]).
- an evaluator rule used to evaluate this operation on constants at compile-time: [Clash/GHC/Evaluator.hs]
  


[Clash/Sized/Internal/Unsigned.hs]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-prelude/src/Clash/Sized/Internal/Unsigned.hs#L399-L408
[prims/systemverilog/Clash_Sized_Internal_Unsigned.json]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-lib/prims/systemverilog/Clash_Sized_Internal_Unsigned.json#L13-L24
[prims/verilog/Clash_Sized_Internal_Unsigned.json]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-lib/prims/verilog/Clash_Sized_Internal_Unsigned.json#L13-L24
[prims/vhdl/Clash_Sized_Internal_Unsigned.json]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-lib/prims/vhdl/Clash_Sized_Internal_Unsigned.json#L190-L196
[Clash/GHC/Evaluator.hs]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs#L2173-L2179
[+#]: https://github.com/clash-lang/clash-compiler/blame/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-lib/prims/common/Clash_Sized_Internal_Unsigned.json#L1-L7
[eq#]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-lib/prims/commonverilog/Clash_Sized_Internal_Unsigned.json#L1-L7
[eq# vhdl]: https://github.com/clash-lang/clash-compiler/blob/6422262599ebf53b4bf97d44a5b063c92d3ff81b/clash-lib/prims/vhdl/Clash_Sized_Internal_Unsigned.json#L25-L31

martijnbastiaan:
Anyway, to bypass my distracting comments up to now: I'd be fine with merging it as is provided that it's added to the constant folding testbench. That is, the following:

```haskell
    , cBits           @(Index 50000)
    -- , cFiniteBits     @(Index 50000) -- broken
```

should be added over here:

https://github.com/clash-lang/clash-compiler/blob/b371224bee00caf69657e4d8350c6fe75659e751/tests/shouldwork/Numbers/NumConstantFolding.hs#L209-L217

leonschoorl:
I'm not sure it's a good idea to have `Bits (Index n)`, see my comments here: https://github.com/clash-lang/clash-compiler/issues/678#issuecomment-514190095

rowanG077:
@martijnbastiaan Thanks I will add those.

I will create a new pull request where I will refactor the `Unsigned` implementation of `Bits` to use the `BitVector` representation as well. Could `Signed` use the `BitVector`  implementation as well?

martijnbastiaan:
> I will create a new pull request where I will refactor the Unsigned implementation of Bits to use the BitVector representation as well. Could Signed use the BitVector implementation as well?

While I like the idea of having less blackboxes, I can't really oversee the consequences. @leonschoorl spent a lot of time a while ago fixing all the weird casting errors we had, so maybe he can comment?

rowanG077:
The unit tests are failing because Index is going out of bounds. The easy way I could fix this is make the size a power of 2. So

```
 tIndex 
   = ( cNum        @(Index 50000) 
     , cEq         @(Index 50000) 
     , cOrd        @(Index 50000) 
     , cIntegral   @(Index 50000) 
     -- no Bits, FiniteBits 
     , csClashSpecific @(Index 50000) 
     , cResize         @(Index 50000) 
     ) 
```

would become

```
tIndex
  = ( cNum        @(Index 50000)
    , cEq         @(Index 50000)
    , cOrd        @(Index 50000)
    , cIntegral   @(Index 50000)
    , cBits       @(Index 65536) -- use an index of 2^n to ensure index will not throw an exception.
    -- , cFiniteBits     @(Index 50000) -- broken
    , csClashSpecific @(Index 50000)
    , cResize         @(Index 50000)
    )
```

martijnbastiaan:
We did have some internal discussion about this PR. Ideally, we would like to move to a situation where all operations on `Index` are safe. Unfortunately, Haskell legacy stands in our way (most notably, we _have_ to implement `Num` and its unsafe operations to get index literals), so we can't really offer a nice API. We decided that the next best thing was to unify the API of `Unsigned` and `Index`, until we can implement it the "proper" way.

Btw, @rowanG077, in case you missed my comment earlier: 

>> I will create a new pull request where I will refactor the Unsigned implementation of Bits to use the BitVector representation as well. Could Signed use the BitVector implementation as well?
>
> We've discussed this internally, and we don't think it's necessary. Like I commented earlier, converting to bitvector is nice from a blackbox-perspective, but it induces simulation costs we don't currently have. Thanks for the offer anyway :).

PR #683, opened 2019-07-25
Nix dependency fixes

basile-henry:
This PR makes nix dependencies more explicit. This was an issue in particular for the `gitignore` package as that was defaulting to the user environment's version of nixpkgs.


basile-henry:
Thanks @Taneb for helping to track this issue!

PR #685, opened 2019-07-25
Scripts to build a binary distribution

christiaanb:
* Based on Nix. Requires Nix on the host, but _not_ on the target.

* bindist will contain
  * `/nix`: Nix-derivation of:
      1. bash
      2. ghc with clash-prelude and clash-lib
      3. clash
  * `/bin`: symlinks for the above executables
  * `nix-user-chroot`: fakechroot to map `./nix` to `/nix`
  * `clash`: shell-wrapper using fakechroot to `clash`
  * `clashi`: shell-wrapper using fakechroot to `clashi`

Simply run `./bindist/linux/mkBinDist.sh` to create a `./clash-bindist.tar.xz` archive.

Unpack said archive anywhere using:

`tar xf clash-bindist.tar.xz --one-top-level`

Then you can simply run:

`./clash-bindist/clash` or
`./clash-bindist/clashi`

or add `./clash-bindist` to you path.

### DIFF
--- a/bindist/linux/mkBinDist.sh
+++ b/bindist/linux/mkBinDist.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+set -e
+
+# Path to thhe nix file containing the derivation we want to package
+NIXFILE=$(dirname $0)/bindist.nix
+
+# "drv" will be set to the path inside /nix/store which contains the resulting the derivation
+drv=$(nix-build $NIXFILE)
+
+# output filename
+tarball=$(pwd)/clash-bindist.tar.xz
+
+# temporary working directory
+tmpdir=$(mktemp -d)
+trap "rm -rf $tmpdir" EXIT
+
+g++ -o $tmpdir/nix-user-chroot -DENV_PATH='"/bin"' $(dirname $0)/nix-user-chroot/main.cpp
+cp $(dirname $0)/wrappers/clash $tmpdir/
+cp $(dirname $0)/wrappers/clashi $tmpdir/
+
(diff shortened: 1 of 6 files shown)


bindist/linux/mkBinDist.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# create symlinks inside target's usr/local/bin to every executable in the derivation (non-transitively)
+mkdir -p bin
+for binary in $(find ${drv}/bin); do
+  ln -s ${binary} bin/$(basename ${binary})
+done
+
+# finally, pack the full closure of the derivation with the symlinks
+# we just created
+tar cJf $tarball --owner 0 --group 0 bin clash clashi nix-user-chroot $(nix-store -qR $drv)
```
martijnbastiaan: Maybe

```bash
tar -I pxz -cf $tarball [..]
```

For >1 thread.

bindist/linux/mkBinDist.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+set -e
+
+# Path to thhe nix file containing the derivation we want to package
```
TMVector:
```suggestion
# Path to the nix file containing the derivation we want to package
```

christiaanb: Fixed

bindist/linux/mkBinDist.sh
christiaanb: Fixed by other means

Issue #682, opened 2019-07-24
satAdd and satMul produce errors for Index 1


rowanG077:
`satAdd` and `satMul` with any of the `SaturationModes` results in out of bounds exceptions when used with `Index 1` even though the operations should be safe. `satSub` works fine.

```
satModes = [SatWrap, SatBound, SatZero, SatSymmetric]
operations = [satAdd, satSub, satMul]
n = 0 :: Index 1
results = showX [f n n | f <- operations <*> satModes] -- [X,X,X,X,0,0,0,0,X,X,X,X]
```

It's due to usage of the following code in each of those operations:
```
m :: AResult (Index 1) (Index 1)  or m :: MResult (Index 1) (Index 1) 
m = fromInteger# (natVal (Proxy @ n))
```
`m` thus has value `1` and since `AResult (Index 1) (Index 1) ~ Index 1` and `MResult (Index 1) (Index 1) ~ Index 1` this is never a valid value and the error is triggered.

See [here](https://github.com/clash-lang/clash-compiler/blob/7d0ae243734f7e4b3a61d7860067dc652e3221ef/clash-prelude/src/Clash/Sized/Internal/Index.hs#L284).

martijnbastiaan:
Good catch. So we should change or bound checking to 'greater than', instead of 'greater than or equal to'. For example, in `satAdd SatWrap`:

```diff
--- a/clash-prelude/src/Clash/Sized/Internal/Index.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/Index.hs
@@ -281,8 +281,8 @@ instance (KnownNat n, 1 <= n) => SaturatingNum (Index n) where
   satAdd SatWrap a b =
     leToPlusKN @1 @n $
       case plus# a b of
-        z | let m = fromInteger# (natVal (Proxy @ n))
-          , z >= m -> resize# (z - m)
+        z | let m = fromInteger# (natVal (Proxy @ (n - 1)))
+          , z > m -> resize# (z - m)
         z -> resize# z
   satAdd SatZero a b =
     leToPlusKN @1 @n $
```

..would presumably fix the issue.

rowanG077:
> Good catch. So we should change bound checking to 'greater than', instead of 'greater than or equal to'. For example, in `satAdd SatWrap`:
> 
> ```diff
> --- a/clash-prelude/src/Clash/Sized/Internal/Index.hs
> +++ b/clash-prelude/src/Clash/Sized/Internal/Index.hs
[...]

PR #686, opened 2019-07-26
Fixed bug that Saturating operations on `Index 1` were unsafe (#682).

rowanG077:
I have added constant folding test for the specific case `Index 1` case.

The parser doesn't detect yet whether it is actually constant folded since it only will trigger an error when it finds a specific number range. It's not possible to easily extend this with the correct number since for `Index 1` all operations take and result in 0.

Creating a parser to check that no real operations are taking place would be quite an undertaking I think. What do you think @martijnbastiaan?

### DIFF
--- a/clash-prelude/src/Clash/Sized/Internal/Index.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/Index.hs
@@ -5,6 +5,7 @@ License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 -}
 
+{-# LANGUAGE BangPatterns          #-}
 {-# LANGUAGE CPP                   #-}
 {-# LANGUAGE DataKinds             #-}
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
Hey @rowanG077, thanks for your second PR :)!

I think the following would suffice:

```haskell
  case (satAdd @(Index 1) SatWrap (lit 0) (lit 0)) of {0 -> 0; 1 -> 22270 :: Int}
  case (satSub @(Index 1) SatWrap (lit 0) (lit 0)) of {0 -> 0; 1 -> 22271 :: Int}
  ...
```

If Clash successfully constant folds this, you shouldn't see `22270` in the output. It might be that it _doesn't_ actually constant fold this, but that it strips out the case entirely due to `Index 1` being zero bits. I'm a bit fuzzy on the details, as it has been a while since I looked at this, but either way should be fine.

martijnbastiaan:
LGTM, thanks!

PR #692, opened 2019-08-01
Run `apt update` before `apt install` on CI

martijnbastiaan:
This prevents apt from fetching non-existing archives


leonschoorl:
Wouldn't it be better just to bake ksh into the docker image?

martijnbastiaan:
I like the idea of the image being _just_ a nix installation installed on a different path, but sure, we could bake `ksh` in. We would need the other `apt update` anyway (which was actually the one failing last night), so this commit is a bit less work.

PR #698, opened 2019-08-06
Allow Haskell blackbox to be used in HO primitives

martijnbastiaan:


clash-lib/src/Clash/Netlist/BlackBox.hs
```diff
@@ -455,28 +455,40 @@ mkFunInput
...
+                  -- Determine result type of this blackbox. If it's not a
+                  -- function, simply use its term type.
+                  let
+                    resTy0 = termType tcm e
+                    resTy1 =
+                      case splitFunTy tcm resTy0 of
+                        Just (_, t) -> t
+                        Nothing -> resTy0
+
+                  bbhRes <- func True (Right resId) pName args resTy1
```
martijnbastiaan: Point of attention: I'm not sure whether passing in `(Right resId)` is the correct thing to do here, but I'm not sure what else to do..?

christiaanb: No... this isn't right, that `resId` is the lhs in:

```haskell
lhs = map fun xs
```

While the BlackBoxHaskell is expecting an identifier to bind the result of `fun`.

martijnbastiaan: Right, that's what I suspected. Should we just remove the argument and let users use `~RESULT`?

martijnbastiaan: Actually, I'm pretty sure that's the only valid option here, as there is no singular result binder id. It depends on the instantiation of the blackbox later on.

christiaanb: Yeah... knowing the type of the result should be enough.

PR #705, opened 2019-08-13
Make MultipleHidden pass for VHDL

christiaanb:
Before, it would get stuck, probably because the clockGen keeps running forever in the VHDL simulators.


testsuite/Main.hs
```diff
@@ -77,7 +77,7 @@ runClashTest =
         , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "LambdaDrop"          ([""],"topEntity",False)
         , runTest ("tests" </> "shouldwork" </> "Basic") defBuild [] "LotOfStates"         (["","testBench"],"testBench",True)
         -- TODO: Fix for VHDL. ModelSim / GHDL simulate _extremely_ slowly
```
martijnbastiaan: This comment can go :P

PR #720, opened 2019-08-16
InlineCleanup should not inline recursive bindings

christiaanb:


martijnbastiaan:
@christiaanb We had a small example triggering this error right? Could we add that to the CI?

PR #721, opened 2019-08-16
Perform reduceConst before ANF

christiaanb:
Also reduce try and reduce anything that looks like a constant, i.e. has no local FVs, not just applications of primitives.

This allows e.g. the constant reduction of CTZ; and obviates #530


christiaanb:
Okay, this obviously brakes things, putting the expression into a form the ANF-transformation was not expecting.

Issue #595, opened 2019-05-02
Warn when compiling with MonoLocalBinds disabled


christiaanb:
Due to complex interactions with GHCs optimizations and Clash’ optimizations, turning off MonoLocalBinds can lead to highly surprising logic duplication. We should warn the user about this, until it is fixed.

martijnbastiaan:
@mmzx Did some work on this: Clash now warns whenever a user compiles with `-XNoMonoLocalBinds`. We should check for `{-# LANGUAGE NoMonoLocalBinds #-}` in modules too, so I'm leaving this issue open for now.

mmzx:
>  We should check for `{-# LANGUAGE NoMonoLocalBinds #-}` in modules too, so I'm leaving this issue open for now.

Oh yes! It seems I have the solution at hand already, I am going to make a PR soon for it.

martijnbastiaan:
Awesome! Looking forward to merging that one too!

martijnbastiaan:
@mmzx Would you think it's possible to make a PR in ~the coming week? We'd like to include a fix for this in Clash 1.0, which is to be release the 3rd of September. If that's too much, don't worry, we'll pick it up ourselves.

mmzx:
Absolutely, please stay tuned until Monday.

martijnbastiaan:
Fixed in https://github.com/clash-lang/clash-compiler/commit/1c93e09b8fc720f7dd0ae30d7ebd79125950b4b5, thanks to @mmzx

PR #722, opened 2019-08-18
Check whether MonoLocalBinds is enabled in modules (#595).

mmzx:


martijnbastiaan:
GHC environment files strike again :S. The last issue is caused by clash-dev detecting an instance in the package-db (in `clash-lib-0.9999:GHC.BasicTypes.Extra`) and locally (`GHC.BasicTypes.Extra`):

```
      Matching instances:
        two instances involving out-of-scope types
          instance NFData BasicTypes.InlineSpec
            -- Defined in ‘clash-lib-0.9999:GHC.BasicTypes.Extra’
          instance NFData BasicTypes.InlineSpec
            -- Defined in ‘GHC.BasicTypes.Extra’
```

I think this should solve it:

```diff
diff --git a/clash-dev b/clash-dev
index d161d050..5443aa76 100755
--- a/clash-dev
+++ b/clash-dev
@@ -1,2 +1,13 @@
 #!/bin/sh
-ghci -fobject-code -iclash-lib/src -iclash-ghc/src-ghc -Wall -DTOOL_VERSION_ghc=\"$(ghc --numeric-version)\" Clash.hs
+GHC_VERSION=$(ghc --numeric-version)
+GHC_MAJOR_VERSION=$(echo ${GHC_VERSION} | tr -d '.' | head -c2)
+GHC_BIN_SRC=$(ls -d clash-ghc/src-bin-${GHC_MAJOR_VERSION}*)
+
+ghci \
+  -fobject-code \
+  -iclash-ghc/src-bin-common/ \
+  -i${GHC_BIN_SRC} \
+  -iclash-lib/src \
+  -iclash-ghc/src-ghc \
+  -Wall -DTOOL_VERSION_ghc=\"${GHC_VERSION}\" \
+  Clash.hs
diff --git a/clash-ghc/src-bin-common/Clash/GHCi/Common.hs b/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
index 2c4ad077..ac5678af 100644
--- a/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
+++ b/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP        #-}
 {-# LANGUAGE LambdaCase #-}
 
 module Clash.GHCi.Common
```

martijnbastiaan:
I've created a new PR for the patch I just posted: https://github.com/clash-lang/clash-compiler/pull/723.

mmzx:
Sorry for the noise! I think should not use github from my phone...

martijnbastiaan:
No problem! I'll see if I can write some tests later today, and then I'll merge it. Thanks :).

mmzx:
Thanks @martijnbastiaan , please let me know if there is more to do on this issue. :)

mmzx:
Regarding ghc-8.8.1 build failing, at least now I see a return code from `PortNamesWithRTree`, the previous builds were only time-outs.

martijnbastiaan:
Hmm, `-9` usually means it got killed by the OOM killer..

martijnbastiaan:
Just ignore CircleCI. I've pushed it to our local CI too.

martijnbastiaan:
I've just ran a few tests, and it seems that the detection doesn't work for external packages. `modGraph2` only contains locally compiled modules. The way I tested it was:

1. Add `{-# LANGUAGE NoMonoLocalBinds #-}` to `Clash/Sized/Vector.hs`
2. Compile:

```haskell
module Test where

import Clash.Prelude

topEntity :: Vec 3 Int
topEntity = repeat 3
```

I think the check should also go into `LoadInterfaceFiles.loadExternalExprs` (or one of it's dependencies).

mmzx:
Right @martijnbastiaan , I didn't think about testing it with the interface files. Thanks!
Reproduced your sample, so now I am using it to solve this.

mmzx:
> I think the check should also go into `LoadInterfaceFiles.loadExternalExprs` (or one of it's dependencies).

@martijnbastiaan , I am reaching out to get ideas or suggestions, because there is a deadline on this and I seem to be stuck at it.

So far what I have concluded and found out (hopefully correctly):
- interface files (.hi) do not store the utilized language extension(s). So during de-serialization this is unavailable.
- `DynFlags` do store the Extensions utilized no other data type.
- however `DynFlags` does not carry `NoMonoLocalBinds` unless it is enabled in a local module or explicitly changed in the given ghc session.

Since a .hi file does not contain the extension list it would not trigger `DynFlags` to store `NoMonoLocalBinds`.

So I tried to go via:
data `ModIface` {mi_module} -> `moduleName` -> `getModSummary` -> `ms_hspp_opts` -> `DynFlags`.
But calling getModSummary tends to end up in exceptions, because of "Module not part of module graph exception."

Other this above chain I do not seem to find out how else to get to the utilized extensions of interface files.

martijnbastiaan:
@mmzx, nice detective work. Well I guess that's it. If `hi` files don't store what language extensions it was built with (which makes sense I guess), the only thing we can do is actually change GHC instead of Clash. I'll create a new issue pointing out that we should extend GHC to support this feature. 

Thank you very much for your work. This get's us reeeealy close to 1.0, which is certainly exciting!

mmzx:
Thank you @martijnbastiaan !

PR #728, opened 2019-08-20
Add ghc plugins to shell.nix

blaxill:
This tiny PR adds the ghc type plugins to the nix shell context. i.e. This addresses an issues that after entering nix-shell, running clash may result in the error `<command line>: Could not find module ‘GHC.TypeLits.Normalise’` if the type checker plugins aren't available from some larger nix context.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -9,6 +9,13 @@ mkShell {
     niv
 
     # For quick clash experimentation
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
Thanks!

PR #731, opened 2019-08-21
Refuse to compare types with type families

martijnbastiaan:
'caseElemNonReachable' considers a branch unreachable when a branch
coerces a type to another. For example, a branch coercing `Bool ~ Int`
would be considered unreachable. It would also consider `F Char ~ Int` to
be unreachable, even if `F` was a type family eventually resolving to
`Int`.

Regrettably, I haven't been able to find a small test case.

### DIFF
--- a/clash-lib/src/Clash/Core/Type.hs
+++ b/clash-lib/src/Clash/Core/Type.hs
@@ -44,6 +44,7 @@ module Clash.Core.Type
   , isPolyFunTy
   , isPolyFunCoreTy
   , isPolyTy
+  , isTypeFamilyApplication
   , isFunTy
   , isClassTy
   , applyFunTy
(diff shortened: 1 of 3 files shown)


clash-lib/src/Clash/Core/Type.hs
```diff
@@ -536,6 +537,11 @@ reduceTypeFamily tcm (tyView -> TyConApp tc tys)
 
 reduceTypeFamily _ _ = Nothing
 
+isUnresolvedTypeFamily ::  TyConMap -> Type -> Bool
+isUnresolvedTypeFamily tcm (tyView -> TyConApp tcNm _args)
```
christiaanb: Shouldn't this be `coreView`? Since `coreView` actually tries to solve type families. So If after `coreView` we still see a type family, that means it's stuck.

martijnbastiaan: Aha, that makes sense. Right now I was banking on the transformations reducing the type family at some point (after which `caseElemNonReachable` could do its job).

martijnbastiaan: Fixed, thank you.

martijnbastiaan:
@christiaanb Are you happy with the changes?

clash-lib/src/Clash/Core/Type.hs
christiaanb: Looking better at how `stuckTypeFamily` is actually used, I see that you're passing in types that have already gone through `coreView`, so it's kinda wasteful to call `coreView` again. I guess you should just call it `isTypeFamilyApplication` (using just `tyView`), and then inside `solveEq` add a comment that any remaining type family applications are stuck ones, as they should've been solved by `coreView` if they werent stuck.

martijnbastiaan: Yeah.. I thought I'd do it this way to make my intent very clear, but I comment will suffice too. Thanks!

PR #746, opened 2019-08-27
Auto release Clash on Hackage

martijnbastiaan:
This PR will make the CI release a candidate version of Clash every night. When a tag is added, a full release is made. (For the moment, it will release a candidate version when tagged too. I'm taking a conservative approach here as we can't delete anything once published..)

--------------------

I propose we add a check making sure `clash-prelude`, `clash-lib`, and `clash-ghc` always have the same version number for a single commit. While it's _technically_ true they're separate packages, changes often span multiple packages, and it might very well lead to confusion among users. This also makes more sense from a monorepo perspective, as we only have a single tag to handout for a given commit. Not coincidentally, this also fits the current release logic better. 

Edit: I've added the check.

### DIFF
--- a/.ci/publish_sdist.sh
+++ b/.ci/publish_sdist.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -eu -o pipefail
+
+PASSWORD=$(echo $HACKAGE_PASSWORD | base64 --decode --ignore-garbage)
+
+SDIST=$(find $1-*.tar.gz | grep -v docs)
+DDIST=$(find $1-*.tar.gz | grep docs)
+
+set +u
+
+if [[ "$HACKAGE_RELEASE" == "yes" ]]; then
+    # Release tag set, upload as release.
+    # TODO: Right now, this branch will do the same as the other: make a
+    # release candidate. We can change it to actually have it release after
+    # we've gained enough trust in the release logic.
+    echo "XXX: Release logic activated!"
+    cabal upload --username=${HACKAGE_USERNAME} --password=${PASSWORD} ${SDIST} 
+    cabal upload --documentation --username=${HACKAGE_USERNAME} --password=${PASSWORD} ${DDIST} 
+elif [[ "$HACKAGE_RELEASE" == "no" ]]; then
+    # Upload as release candidate
+    cabal upload --username=${HACKAGE_USERNAME} --password=${PASSWORD} ${SDIST} 
+    cabal upload --documentation --username=${HACKAGE_USERNAME} --password=${PASSWORD} ${DDIST} 
+else
+    echo "Unrecognized \$HACAKGE_RELEASE: $HACAKGE_RELEASE"
+    exit 1;
+fi
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -9,5 +9,38 @@ cabal --write-ghc-environment-files=always new-build all
 sed "s/^ghci/ghc -fno-code/" clash-dev > clash-dev-test
 sh clash-dev-test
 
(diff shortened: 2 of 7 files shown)


.ci/publish_sdist.sh
```diff
@@ -0,0 +1,23 @@
...
+SDIST=$(find $1-*.tar.gz | grep -v docs)
+DDIST=$(find $1-*.tar.gz | grep docs)
+
+set +u
+
+if [[ "$HACKAGE_RELEASE" == "yes" ]]; then
+    # Release tag set, upload as release
+    echo "Release!"
+    cabal upload --username=${HACKAGE_USERNAME} --password=${PASSWORD} ${SDIST} 
+    cabal upload --documentation --username=${HACKAGE_USERNAME} --password=${PASSWORD} ${DDIST}
```
leonschoorl: You're not uploading a release, but release candidate.
Is that intentional?

For now I think prefer it this way.
After checking it we can manually publish a candidate as a release through the hackage webinterface.

But if that is what we want, the comments should reflect that.

.ci/test.sh
```diff
@@ -9,5 +9,19 @@ cabal --write-ghc-environment-files=always new-build all
...
 
+# Check whether version numbers in snap / clash-{prelude,lib,ghc} are the same
+cabal_files="clash-prelude/clash-prelude.cabal clash-lib/clash-lib.cabal clash-ghc/clash-ghc.cabal"
+snapcraft_file="bindist/linux/snap/snap/snapcraft.yaml"
+versions=$(grep "^[vV]ersion" $cabal_files $snapcraft_file | grep -Eo '[0-9]+\.[0-9]+' | tr ' ' '\n')
+
+if [[ $(echo $versions | tr ' ' '\n' | wc -l) == 4 ]]; then
+    if [[ $(echo $versions | tr ' ' '\n' | uniq | wc -l) != 1 ]]; then
+        echo "Expected all distributions to have the same version number. Found: $versions"
+    fi
```
leonschoorl: If we're building a tag maybe we should check that the version is equal to tag (stored in `CI_COMMIT_TAG`)

.gitlab-ci.yml
```diff
@@ -50,6 +51,65 @@ ghc-head:
...
 
+hackage-sdist:
+  extends: .tests
+  variables:
+    GHC: ghc-8.8.1
+  script:
+    - export GHC="${GHC:-$CI_JOB_NAME}"
+    - export THREADS=$(nproc)
+    - export CABAL_JOBS=$(nproc)
+    - export
```
leonschoorl: Should we have this export print our secrets and rely on gitlab to filter them out?

leonschoorl: You could unset them in a subshell around export:
`    - ( unset SNAPCRAFT_LOGIN_FILE HACKAGE_PASSWORD ; export )`
Or just drop the export.

.ci/publish_sdist.sh
martijnbastiaan: Yes, I was trying to be conservative as we can't unpublish. You're right that it should have been a comment.

.ci/test.sh
martijnbastiaan: Good idea!

.gitlab-ci.yml
martijnbastiaan: We _should_ be able to, but yeah, dropping the export seems a lot safer.

.ci/test.sh
martijnbastiaan: Ooooof, I also forgot to include `exit 1`s :sweat:

PR #751, opened 2019-08-30
Simplify unsafe synchronizer

jonfowler:
**Edit:** It is impossible to avoid the following behaviour, the potential replacement is considerably simpler though.

The current "unsafeSynchronizer" sometimes can end up ahead of time. The function
```
unsafeSynchronizer @slowDomain @fastDomain . unsafeSynchronizer @fastDomain @slowDomain
```
currently results in a signal which is sometimes ahead of the original signal.

The proposed replacement is never ahead of time (but is still as "close" to the original signal) and is still idempotent when the clock period is the same.

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Signal.hs
+++ b/clash-prelude/src/Clash/Explicit/Signal.hs
@@ -500,6 +500,20 @@ unsafeSynchronizer _clk1 _clk2 =
 {-# INLINE unsafeSynchronizer #-}
 
 -- | Same as 'unsafeSynchronizer', but with manually supplied clock periods.
+--
+-- Note: this unsafeSynchronizer is defined to be consistent with the vhdl and verilog
+-- implementations however as only synchronous signals are represented in Clash this
+-- cannot be done precisely and can lead to odd behaviour. For example,
+-- @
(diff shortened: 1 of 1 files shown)


clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -480,7 +480,7 @@ freqCalc = toInteger . hzToPeriod
 -- @
 --
 -- >>> sampleN 37 (oversampling clk7 clk2 en7 en2 0 (fromList [(1::Int)..10]))
--- [0,0,1,1,1,2,2,2,2,3,3,3,4,4,4,4,5,5,5,6,6,6,6,7,7,7,8,8,8,8,9,9,9,10,10,10,10]
+-- [0,0,0,0,0,1,1,1,2,2,2,2,3,3,3,4,4,4,4,5,5,5,6,6,6,6,7,7,7,8,8,8,8,9,9,9,10]
```
christiaanb: But this is not how the actual circuit behaves

jonfowler:
@christiaanb I have changed the code to be what I think is consistent with the original/hdl simulation. I also documented a bit that this has some behaviour.

Your choice whether it goes in (I think it is a bit more readable than the original).

PR #790, opened 2019-09-11
 Don't overwrite existing binders when specializing 

martijnbastiaan:


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -597,11 +608,13 @@ mkFunction
...
   -- ^ Name with a proper unique and the type of the function
 mkFunction bndrNm sp inl body = do
-  tcm    <- Lens.view tcCache
+  tcm <- Lens.view tcCache
   let bodyTy = termType tcm body
-  bodyNm <- cloneName bndrNm
-  addGlobalBind bodyNm bodyTy sp inl body
-  return (mkGlobalId bodyTy bodyNm)
+  binders <- Lens.use bindings
+  bodyNm0 <- cloneName bndrNm
```
christiaanb: I think `cloneName` should call `uniqAway` wrt the inScopeSet you want to uniqAway from. There are other areas where we use `cloneName` that might have the same shadowing conflicts as we had `mkFunction`

martijnbastiaan: I've generalized it a bit now. `Uniqable` has an additional method `setUnique`, used by `uniqAway`. `cloneName` now always returns a unique unique ( :-) ).

PR #791, opened 2019-09-11
Bind more work free constructs in 'bindConstantVar'

martijnbastiaan:
~~TODO:~~

- [x] Add documentation (both to PR and code)
- [x] Use `WorkInfo` to determine work-free-y-ness


martijnbastiaan:
@christiaanb Do you think we should use `isWorkFreeIsh` in ANF too?

Issue #748, opened 2019-08-29
Generalize `Undefined` to work with non-errors too


martijnbastiaan:
`deepErrorX` builds the structure/spine of a datatype and puts an error at its "core". This can be generalized such that it "lazyfies" a given type instead. Consider:

```haskell
Clash.Prelude> x = errorX "foo" :: Vec 3 Int
Clash.Prelude> map (const 3) x
<*** Exception: X: foo
```

We could build `lazify` (for the lack of a better name) to achieve:

```haskell
Clash.Prelude> x = lazify (errorX "foo") :: Vec 3 Int
Clash.Prelude> map (const 3) x
<3,3,3>
```

`deepErrorX` can now be defined as:

```haskell
deepErrorX = lazify . errorX
```

christiaanb:
It's somewhat specific to products of a single type though... how do you generalize this to e.g. tuples?

martijnbastiaan:
I'm not sure I understand what you're saying. The implementation for a tuple would be something like:

```haskell
instance (Undefined a, Undefined b) => Undefined (a, b) where
  -- lazify ~(a, b) = (lazify a, lazify b) ?
  lazify ab = 
    ( let (a, _) = ab in lazify a
    , let (_, b) = ab in lazify b
    )
```

christiaanb:
So what's the `class` supposed to look like?

martijnbastiaan:
```haskell
class Undefined a where
  lazify :: a -> a
```

? Could you just explain what issues you're seeing?

christiaanb:
I don't see what the benefit is of having:

```
lazify (1,2,errorX "foo") :: (Int,Int,(Int, Int))
```

over

```
(1,2,deepErrorX "Foo" :: (Int,Int))
```

martijnbastiaan:
Ah right. It allows your to implement `deepLazyV`, if you will.

martijnbastiaan:
To be more specific, `deepLazyV` would be a type specialized version of `lazify`.

martijnbastiaan:
Does that make sense @christiaanb?

christiaanb:
Yeah, that makes sense

jonfowler:
@christiaanb a potential use case is when you have `mux`:
```
mux (errorX "undefined") (pure (repeat 0)) someVecSig
```
will produce `errorX "undefined"`. With `lazify` you could write:
```
lazify $ mux ...
```
which will give you back the structure.

PR #803, opened 2019-09-17
Add 'NFDataX.ensureSpine`

martijnbastiaan:
~~**Wanted**: a new name for `spinify`! Feel free to comment below~~ `ensureSpine` it is!

-----------------

Consider the following example:

```haskell
>>> x = undefined :: Vec 3 Char
>>> y = map (\_ -> 3) x
>>> y
<error>
```

Currently, this would fail, as `map` tries to pattern match on the constructor of vector `Cons`. This constructor doesn't exist however, and resolves to `undefined`. In this specific instance we could use `lazyV` to create the "spine" of `Vec` for us:

```haskell
>>> z = map (\_ -> 3) (lazyV x)
>>> z
<3,3,3>
```

`spinify` generalizes this behavior to all data structures having an instance for `NFDataX`.

-----------------

Fixes https://github.com/clash-lang/clash-compiler/issues/748

### DIFF
--- a/clash-prelude/src/Clash/Sized/Fixed.hs
+++ b/clash-prelude/src/Clash/Sized/Fixed.hs
@@ -280,6 +280,7 @@ instance NFDataX (rep (int + frac)) => NFDataX (Fixed rep int frac) where
   deepErrorX = Fixed . errorX
   rnfX f@(~(Fixed x)) = if isLeft (isX f) then () else rnfX x
   hasUndefined f@(~(Fixed x)) = if isLeft (isX f) then True else hasUndefined x
+  ensureSpine ~(Fixed x) = Fixed x
 
 -- | None of the 'Read' class' methods are synthesizable.
 instance (size ~ (int + frac), KnownNat frac, Bounded (rep size), Integral (rep size))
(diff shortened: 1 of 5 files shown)


christiaanb:
`spineDefined`, it ensures that the spine of a data-structure is always defined.

gergoerdi:
`ensureSpine`? `ensureShape`?

basile-henry:
`deepDefine`

rowanG077:
I like ensureSpine the best.

masterdezign:
proto?

martijnbastiaan:
Thanks everyone :). Seems like `ensureSpine` is the most popular. Leave a comment if you strongly disagree! If no one does, I'll change the name and merge this PR next week.

martijnbastiaan:
@christiaanb This is ready for review. I don't believe it's possible to deprecate "custom" implementations of type class functions (in this case `deepErrorX`), is it?

martijnbastiaan:
My original idea was to deprecate "custom" implementations of `deepErrorX`, as `deepErrorX` ~ `ensureSpine . errorX`. As we discussed online, that's not really possible, so I'll revert the deletion of custom implementations.

martijnbastiaan:
@christiaanb I've readded `withFrozenStack` too. I'll merge it if it passed the tests!

PR #807, opened 2019-09-19
TopEntityGeneration: fix for Vec's

blaxill:
- Fixed seen' bug
- Remove all Vec's from type tree during naming as:
  - Elements are type homogeneous so will generate the same name and
  - We don't try to provide number prefixes for elements
- Added Vec test

### DIFF
--- a/clash-prelude/src/Clash/Annotations/TH.hs
+++ b/clash-prelude/src/Clash/Annotations/TH.hs
@@ -1,12 +1,19 @@
 {-|
 
 This module can automatically generate TopEntity definitions from 'Clash.NamedTypes'
-annotations. Type/data families will resolve a single step (non-recusive).
+annotations. Annotations try to be as accurate to the shape of the Clash compiler output.
+Not all cases can be handled with automatic generation due to the difficulty of type manipulation
+in template Haskell. In particular annotations _inside_ the following is unlikely to work:
+
+- Data/type family constructors referencing other data/type families.
+- Recursive data types
+- Clock constraints other than a single HiddenClockResetEnable. (You can still
+  use arbitrary explicit clock/reset/enables!)
 
 See "Clash.Tests.TopEntityGeneration" for more examples.
 
 @
-import Clash.Annotations.TopEntityGen
+import Clash.Annotations.TH
 
 data Named
   = Named
@@ -72,6 +79,7 @@ import           Data.Foldable                  ( fold
                                                 , find
                                                 )
 import qualified Data.Set                      as Set
+import qualified Data.Map                      as Map
 #if !(MIN_VERSION_base(4,11,0))
 import           Data.Semigroup                as Semigroup
 #endif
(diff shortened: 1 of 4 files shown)


blaxill:
`seen` was used instead of `seen'` in some recursive calls, which bypassed checking if the current name had been seen (it was a typo). This caused the erroneous cases to not terminate. With that fixed, recursive datatypes will still unroll once, so I removed them during the first pass. Another solution might be to backtrack during `gatherNames` when discovering a directly recursive type, but this would not be perfect either as circular referencing types wouldnt be caught. Unhandled cases would terminate under both solutions but may result in unexpected portnames when annotations exist inside the recursive types (which doesn't have clear semantics anyway.)

blaxill:
That is to say

~~~Haskell
data Foo = Foo ("someName" :: Int)

Bar :: Vec _ Foo
==> Should now be e.g. PortName ""
~~~

but 

~~~Haskell
data Baz = Baz Foo Baz

Bar :: Baz
==> will be something like: PortProduct "" [PortName "someName"]
~~~

This is incorrect but a recursive type with annotation inside is either meaningless, or requires enumerating over to generate ordinal prefixes. I'm hesitant to do this because its going to get increasingly hairy to do in template haskell.

christiaanb:
I guess a refactor to `ReaderT (Set.Set Name) Q` might be worth it, so you can't make this mistake in the future.

blaxill:
Adding much broader / general parametrized type support. Types resolving to a single constructor after applying type parameters are the only candidates.

This brings the support of `makeTopEntity` closer to the support of clash compiler top entities. I'm going to make one more pass through this to clean stuff up a bit.

clash-prelude/src/Clash/Annotations/TH.hs
```diff
@@ -284,15 +375,15 @@ expandFamilies t = embed <$> sequence t
 
 -- | Runs 'expandFamilies' and then 'gatherNames'
 expandFamiliesAndGatherNames
-  :: Set.Set Name
+  :: -- Set.Set Name
   -- ^ Locally track seen names to handle recursive datatypes
```
martijnbastiaan: This might very well be what's breaking Haddock on CI

blaxill:
CI still looks fine so I'll remove remaining todos and squash

christiaanb:
@blaxill Could you add a line to a changelog for this new feature? (not persé  for this fix for Vec, but for this extension to the port naming feature in general)

blaxill:
> @blaxill Could you add a line to a changelog for this new feature? (not persé for this fix for Vec, but for this extension to the port naming feature in general)

Done and I've added a small summary of limitations for the feature too.

blaxill:
@martijnbastiaan gitlab CI failure looks unrelated? I'm not sure why it's only failing for the last commit

christiaanb:
@blaxill one of our dependencies is lacking an upper bound on `vty`, which has changed API

christiaanb:
@blaxill just created #826, that will hopefully solve the issue once merged

christiaanb:
Thanks!

PR #835, opened 2019-10-06
Verilog vec1bit output fix.

### LINKED ISSUE #810: Vec 1 Bit turned into single wire in Verilog
I have a Vivado constraints file that has a 16 wires wide `LED` bus. I can use a bottom slice easily from Clash without having to change the constraints file: I can connect, e.g., a `Vec 8 Bit` to it to drive only the lowest 8 bits.

However, this same method breaks down for `Vec 1 Bit`: At least when targeting Verilog, Clash 1.0 synthesizes that into a single `wire`, instead of a `wire [0:0]`, resulting in an unconnected output, since `LED` and `LED[0]` are not the same.

jhmcstanton:
Fixes #810. Change here is based on the discussion in the issue.


christiaanb:
Thanks!

PR #845, opened 2019-10-14
Examine work content of global binders in 'isWorkFree' 

martijnbastiaan:
Edit: **I'll test this on a larger codebase later this week. Please postpone merging until then.**

..and prevent new let-bindings in `flattenCallTree`, as this breaks Clash's NF.

### DIFF
--- a/clash-lib/src/Clash/Normalize/Util.hs
+++ b/clash-lib/src/Clash/Normalize/Util.hs
@@ -11,6 +11,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
 {-# LANGUAGE ViewPatterns      #-}
+{-# LANGUAGE MagicHash         #-}
 
 module Clash.Normalize.Util
  ( ConstantSpecInfo(..)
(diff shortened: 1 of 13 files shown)


martijnbastiaan:
Do you think we should backport this to 1.0 @christiaanb ?

clash-lib/src/Clash/Normalize/Util.hs
```diff
@@ -67,10 +69,47 @@ import           Clash.Rewrite.Types
...
+isWorkFree
+  :: Term
+  -> NormalizeSession Bool
+isWorkFree (collectArgs -> (fun,args)) = case fun of
+  Var i -> do
+    let isPoly = isPolyFunTy (varType i)
+    if isLocalId i && not isPoly then
+      pure True
+    else
+      not <$> isRecursiveBndr i
```
christiaanb: We should check whether the global function is work-free

martijnbastiaan: Done?

martijnbastiaan:
I'm awfully close to "later this week" now.

PR #849, opened 2019-10-15
Slightly reduce CI load on 1.0

martijnbastiaan:


leonschoorl:
Since the default configuration of the 1.0 release is to have `multiple-hidden` disabled, I think we should keep a `..-singular-hidden` test.

leonschoorl:
Yes

Issue #418, opened 2018-11-19
Generate Verilator-friendly verilog


adamwalker:
Clash generates Verilog that Verilator has difficulty optimising. For example, given the module below:

```
module Mod where

import Clash.Prelude

count :: (HiddenClockReset dom gated sync, Num a) => Signal dom Bool -> Signal dom a
count inc = res
    where
    res = regEn 0 inc $ res + 1

cntrs 
    :: forall dom gated sync. (HiddenClockReset dom gated sync)
    => Signal dom Bool
    -> Vec 2 (Signal dom (BitVector 16))
cntrs v = c1 :> c2 :> Nil
    where
    c1 :: Signal dom (BitVector 16)
    c1 =  count v

    c2 :: Signal dom (BitVector 16)
    c2 =  regEn 0 v $ c2 + 1

----------------------------------------------------------
--Synthesis
----------------------------------------------------------
{-# ANN topEntity
  (Synthesize
    { t_name     = "mod"
    , t_inputs   = []
    , t_output   = PortProduct "" []
    }) #-}
topEntity clk rst = withClockReset clk rst (cntrs @System @Source @Synchronous)
```
Clash generates: 
```
/* AUTOMATICALLY GENERATED VERILOG-2001 SOURCE CODE.
** GENERATED BY CLASH 0.99.3. DO NOT MODIFY.
*/
module Mod_count
    ( // Inputs
      input [1:0] \$d(%,%)  
    , input  inc 

      // Outputs
    , output wire [15:0] res 
    );
  reg [15:0] \#res_rec ;
  wire [1:0] \#res_app_arg ;

  // register begin
  always @(posedge \#res_app_arg [1] ) begin : Mod_count_register
    if (\$d(%,%) [0:0]) begin
      \#res_rec  <= 16'b0000000000000000;
    end else if (\#res_app_arg [0]) begin
      \#res_rec  <= (\#res_rec  + 16'b0000000000000001);
    end
  end
  // register end

  // clockGate begin 
  assign \#res_app_arg  = {\$d(%,%) [1:1],inc};
  // clockGate end

  assign res = \#res_rec ;
endmodule
```
And compilation with Verilator fails with these warnings:
```
%Warning-UNOPTFLAT: verilog/Mod/mod//Mod_count.v:13: Signal unoptimizable: Feedback to clock or circular logic: mod.Mod_count_app_arg.#res_app_arg
[...]

PR #850, opened 2019-10-15
Blackbox scaffolding

blaxill:
[updated]

Similar to #248 I find a top board function with non-simulated blackboxes, which then forwards non-simulated signals to a simulatable core, a productive way to structure projects where writing Haskell implementations of some blackboxes is undesirable. Manually writing them currently is error prone for larger modules as data defintions, haskell function and hdl template must all be kept in sync ( #813). 

This PR adds a template haskell function 'makeScaffold' to create a HDL blackbox with a minimal Haskell function and I/O datatypes. Heres an example Xilinx IBUFDS_GTE2 primitive, with added bogus signals for demonstration (note that  'makeScaffold' becomes more useful the more signals the primitive function has):

~~~Haskell
makeScaffold "xilinxDiffClock" "IBUFDS_GTE2"
  -- A list of parameters
  [ PBool "CLKRCV_TRST" True
  ]

  -- A list of list of ports, corresponding to domains of signals.
  -- Clocks will be lifted to the top of defitions.
  [ [ ClkOut "O"
    , ClkIn "I"
    , In "dummy_signal1" 8
    , ClkIn "IB"
    ]
  , [ In "dummy_signal2" 40
    , Out "dummy_out1" 1
    ]
  ]
~~~

Creates a synthesizable HDL blackbox representation and these Haskell values:

~~~Haskell
data XilinxDiffClockI dom1 dom2
  = XilinxDiffClockI
  { _I :: Clock dom1
  , _IB :: Clock dom1
  , _dummy_signal1 :: Signal dom1 (BitVector 8)
  , _dummy_signal2 :: Signal dom2 (BitVector 40)
  }
data XilinxDiffClockO dom1 dom2
  = XilinxDiffClockI
  { _O :: Clock dom1
  , _dummy_out1 :: Signal dom2 (BitVector 1)
  }

-- Smart constructor taking only the clocks
xilinxDiffClockI arg1 arg2 = XilinxDiffClockI arg1 arg2 (pure def) (pure def)

-- Haskell name tied to HDL instantiation
xilinxDiffClock# arg1 arg2 arg3 arg4
  = XilinxDiffClockO clockGen (pure def)

-- A convenience function taking the input data type and calling the blackbox
xilinxDiffClock (XilinxDiffClockI arg1 arg2 arg3 arg4)
  = xilinxDiffClock# arg1 arg2 arg3 arg4
~~~

PR #815, opened 2019-09-25
Add simulation-level IO

christiaanb:

PR #862, opened 2019-10-18
Seperate out clock arguments from product types

### LINKED ISSUE #477: Unexpected glitch in RTL-level simulation due to clock assignements
With CLaSH 949a05a1d8b1e4b3aabf224248c3e0e124c8fc39 the behavior of the following code is quite strange in at least the Intel/Altera ModelSim:

```
i2cMaster
    :: forall domain gated synchronous. (HiddenClockReset domain gated synchronous)
    => Word32 -> Signal domain Bit
i2cMaster halfDiv = scl
  where
    phase = regEn (0 :: Unsigned 2) (countTo halfDiv) (phase + 1)
    scl = boolToBit . (\phi -> any (== phi) [0, 1]) <$> phase
```

with this version, the RTL-level simulation of `SCL` produces a waveform containing a 1ps glitch at time 360001 (and then repeated at later times): https://gist.github.com/gergoerdi/07d601acbf28d46ae6e9b2d6399a58d1#file-eq_phi-vcd

If I replace the definition of `scl` with

```
    scl = boolToBit . (\phi -> any (phi ==) [0, 1]) <$> phase
```

then there is no glitch visible in the waveform: https://gist.github.com/gergoerdi/07d601acbf28d46ae6e9b2d6399a58d1#file-phi_eq-vcd

christiaanb:
Turns e.g.

```    
f :: (Clock System, Reset System) -> Signal System Int
```

into    

``` 
f :: Clock System -> Reset System -> Signal System Int
```

We need this because HDL synthesis tools don't like it when clocks and resets end up in structs or bitvectors, which product types are usually mapped to. This also solves an annoyance in VHDL where we had to delay register assignment by 1ps to avoid delta-cycle weirdness.
    
Fixes #477
Partially solves #418

### DIFF
--- a/clash-lib/src/Clash/Core/Util.hs
+++ b/clash-lib/src/Clash/Core/Util.hs
@@ -21,7 +21,8 @@ import           Data.Coerce                   (coerce)
 import qualified Data.HashSet                  as HashSet
 import Data.List
   (foldl', mapAccumR, elemIndices, nub)
-import Data.Maybe                              (fromJust, mapMaybe, catMaybes)
+import Data.Maybe
+  (fromJust, isJust, mapMaybe, catMaybes)
 import qualified Data.Text                     as T
 import           Data.Text.Prettyprint.Doc     (line)
 #if !MIN_VERSION_base(4,11,0)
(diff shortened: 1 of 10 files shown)


clash-lib/src/Clash/Core/Util.hs
```diff
@@ -932,3 +933,73 @@ tySym
...
+  = if any (\ty -> isJust (shouldSplit0 tcm ty) || splitTy ty) dcArgVs then
+      Just (mkApps (Data dc) (map Right tyArgs), dcArgs)
+    else
+      Nothing
+ where
+  -- Currently we're only interested in splitting of Clock, Reset, and Enable
+  splitTy (TyConApp tcNm0 _)
+    = nameOcc tcNm0 `elem` [ "Clash.Signal.Internal.Clock"
+                           , "Clash.Signal.Internal.Reset"
+                           , "Clash.Signal.Internal.Enable"
```
martijnbastiaan: Are enable lines special in Verilog/VHDL?

christiaanb: No, just like clock and reset aren't. However, ASIC synthesis tools derive clock gating from enables, so let's make sure they can do their job as easy as possible.

martijnbastiaan: Right, makes sense.

blaxill:
@christiaanb does this automatically work for result product types e.g. https://github.com/clash-lang/clash-compiler/blob/6a822937ca5bac2e78f968d162fd1d068137f217/clash-lib/src/Clash/Primitives/Intel/ClockGen.hs#L74

Where `InstDecl` result assignment is a `DataCon` of potentially Clocks and Signals or is this case unhandled? I'm currently using the `DataCon` method for #850

christiaanb:
@blaxill No... although there's logic in https://github.com/clash-lang/clash-compiler/pull/815 that does that. I can separate it out and add it to this PR.

martijnbastiaan:
@leonschoorl This PR is pending your review. Are you happy with it?

PR #863, opened 2019-10-20
Add scripts to build Debian packages

martijnbastiaan:
Scripts interpret a cabal's plan.json and subsequently fetch and build
Debian packages for clash-ghc and all its dependencies. We can use these
packages to build snaps, replacing the current strategy for building
them. ~~It's not really suitable for hosting our own APT repository, as it
will ignore any dependencies already in the host's repositories -
potentially overriding them with a new, incompatible version.~~

### DIFF
--- a/bindist/linux/debian/Dockerfile
+++ b/bindist/linux/debian/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:20.04
+
+ARG PARALLEL=-j8
+
+ENV DEBIAN_FRONTEND noninteractive
+
+ARG DEPS_GHC="ghc ghc-doc ghc-prof cabal-debian cabal-install happy"
+ARG DEPS_BUILD="build-essential dpkg-dev haskell-devscripts debhelper wget sudo"
+ARG DEPS_CLASH="libtinfo-dev libpthread-stubs0-dev"
+
+RUN apt update \
+  && apt dist-upgrade -y \
+  && apt install $DEPS_GHC $DEPS_BUILD $DEPS_CLASH -y \
+  && apt clean \
+  && echo 1
+
+
+# Yeet all possible GHC dependencies into the global scope, so cabal is sure
+# to find them. This might not be a good idea.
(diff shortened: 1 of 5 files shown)


bindist/linux/debian/build_deps.py
```diff
@@ -0,0 +1,100 @@
...
+    cabal_debian_cmd = [
+        "cabal-debian",
+        "--native",
+        "--disable-tests",
+        "--dep-map", "pthread:libpthread-stubs0-dev"
+    ]
+
+    # clash-prelude work-around
+    if package["pkg-name"] in {"clash-prelude", "clash-lib", "clash-ghc"}:
+        # TODO: Ask Leon
```
martijnbastiaan: @leonschoorl I'm currently disabling haddock for `clash*` due to it failing on 8.6. Can I add something to the `cabal.project` that forces a newer haddock version, or is it tightly coupled with GHC?

christiaanb: Looking at http://hackage.haskell.org/package/haddock-api-2.23.0, it seems pretty tightly bound to the GHC api version. Sadly the changelog doesn't say which version contains the fix for loading plugins.

martijnbastiaan: Alright, that's too bad. I guess we'll have to wait for the docker image of `Ubuntu 20.04 ‘Focal Fossa’` then. They seem to update GHC pretty consistently with every Ubuntu version.

leonschoorl: And yes, the haddock version is tightly coupled with GHC version.

martijnbastiaan: Too bad. Thanks for looking into it anyway!

martijnbastiaan:
For some reason focal (20.04) _still_ include ghc 8.6 with broken Haddock.. but anyway, the scripts are more or less done. Even if we don't end up using them, it's probably useful to keep them around.

Issue #500, opened 2019-02-19
Implementation of Bit leaks to HDL


leonschoorl:
```haskell
topEntity = low
```
```
Main.topEntity8214565720323842134 after normalization:
letrec
  result0 :: Bit
  = BitVector.low
in (result0 :: Bit)
```
VHDL:
```vhdl
  result <= '0';
```
Verilog:
```verilog
  assign \#i  = 64'sd0;
  assign \#i_0  = 64'sd0;
  assign result = 1'b0;
```
SystemVerilog:
```systemverilog
  assign result = 1'b0;
```
------------------------------------
```haskell
topEntity = complement high
```
```
Main.topEntity8214565720323842134 after normalization:
letrec
  result0 :: Bit
  = BitVector.fromInteger## 0
      (-2)
in (result0 :: Bit)
```
VHDL:
```vhdl
  \#i_0\ <= to_signed(0,64);
  \#i\ <= to_signed(-2,64);
  result <= \#i\(0) when '0' = \#i_0\(0) else '-';
```
Verilog:
```verilog
  assign \#i  = 64'sd0;
  assign \#i_0  = -64'sd2;
  assign result = 1'b0;
```
SystemVerilog:
```systemverilog
  assign \#i  = 64'sd0;
  assign \#i_0  = -64'sd2;
  assign result = 1'b0;
```

PR #865, opened 2019-10-22
 Use ~VAR instead of ~ARG before slicing in Verilog 

martijnbastiaan:


martijnbastiaan:
That error message is incredibly weird. It's being tested on 8.4.4, [which shouldn't import `ctEvExpr` at all..](https://github.com/clash-lang/ghc-typelits-knownnat/blob/master/src/GHC/TypeLits/KnownNat/Solver.hs#L146-L150) :S

christiaanb:
It's because of https://gitlab.haskell.org/ghc/head.hackage/blob/master/patches/ghc-typelits-knownnat-0.7.1.patch

I guess we shouldn't be using head.hackage for non-head builds

christiaanb:
@martijnbastiaan does this undo the fix for https://github.com/clash-lang/clash-compiler/issues/500 ?

martijnbastiaan:
@christiaanb Not for the examples in that issue report, but that might just be due to better constant folding?

PR #866, opened 2019-10-23
Don't use head repository for non-head builds

martijnbastiaan:


cabal.project
```diff
@@ -19,15 +19,6 @@ allow-newer:
   , rewrite-inspector:hashable
   , brick:base
 
-repository head.hackage.ghc.haskell.org
```
christiaanb: We sadly need this for building locally with GHC 8.8.1 though

martijnbastiaan: Just for `vty`/`rewrite-inspector` right?

christiaanb: I guess?

martijnbastiaan: Hopefully fixed in https://github.com/omelkonian/rewrite-inspector/pull/3 :-)

martijnbastiaan:
@christiaanb ACK?

PR #868, opened 2019-10-23
makeTopEntity: correctly handle type synonyms

blaxill:
Type synonym support was accidentally dropped as they are not handled by
'th-abstraction'.

- Reenable support for type synonyms
- Add tests for type synonyms
- Merge the two type tree passes into one pass and simplify


martijnbastiaan:
Thanks for all your work on this :)

PR #872, opened 2019-10-25
Tutorial fixes

DigitalBrains1:
Fixes all over. Code was tested to compile, example commands where run
and specified testbenches where run in both clashi and in modelsim. No
tests were done on an FPGA.

While looking at this, I'd suggest using `git diff --word-diff` as well to make parts of this commit a lot more readable!

Some remarks about specific changes:
- JSON-key "type" was included for `blockRam#`, but not for `*#`. Added it for `*#`.

- `blockRam#` missed `KnownDomain`. Changed to match the source. Black boxes where copy-pasted from the real thing, which means `~ACTIVEEDGE` was introduced.

- I had to do three takes to read the sentence

    > Signals of product types and product types (to which tuples belong) of signals are __isomorphic__ [...]

    because of the parenthetical. The sentence is already complicated, better to move that information to a new sentence, IMHO:

    > Signals of product types and product types of signals are __isomorphic__ due to synchronisity principle, but are not (structurally) equal. Tuples are a product type. Use the `bundle` function to convert from a product

    For the record, my brain was parsing that tuples belong to the group defined as "signals of product types and product types", i.e. product types and signals thereof, and after that the sentence broke down semantically.

### DIFF
--- a/clash-prelude/src/Clash/Tutorial.hs
+++ b/clash-prelude/src/Clash/Tutorial.hs
@@ -385,8 +385,8 @@ most contexts.
 The 'register' function is our primary sequential building block to capture
 /state/. It is used internally by one of the "Clash.Prelude" function that we
 will use to describe our MAC circuit. Note that the following paragraphs will
-only show one of many ways to specify a sequential circuit, at the section we
-will show a couple more.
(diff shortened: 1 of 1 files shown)


DigitalBrains1:
New PR #877 created against master branch.

PR #873, opened 2019-10-25
Feature AutoReg

leonschoorl:
`autoReg` is a "smart" version of `register`. It does two things:

  1. It splits product types over their fields. For example, given a 3-tuple,
  the corresponding HDL will end up with three instances of a register (or
  more if the three fields can be split up similarly).

  2. Given a data type where a constructor indicates (parts) of the data will
  (not) be updated a given cycle, it will split the data in two parts. The
  first part will contain the "always interesting" parts (the constructor
  bits). The second holds the "potentially uninteresting" data (the rest).
  Both parts will be stored in separate registers. The register holding the
  "potentially uninteresting" part will only be enabled if the constructor
  bits indicate they're interesting.

  The most important example of this is `Maybe`. Consider `Maybe Byte`;
  when viewed as bits, a `Nothing` would look like:

    >>> pack @(Maybe (Signed 16)) Nothing
    0_...._...._...._....

  and `Just`

    >>> pack @(Maybe (Signed 16)) (Just 3)
    1_0000_0000_0000_0011

  In the first case, Nothing, we don't particularly care about updating the
  register holding the `Signed 16` field, as they'll be unknown anyway. We
  can therefore deassert its enable line.

Making Clash lay it out like this increases the chances of synthesis tools
clock gating the registers, saving energy.

This version of `autoReg` will split the given data type up recursively. For
example, given `a :: Maybe (Maybe Int, Maybe Int)`, a total of five registers
will be rendered. Both the "interesting" and "uninteresting" enable lines of
the inner Maybe types will be controlled by the outer one, in addition to
the inner parts controlling their "uninteresting" parts as described in (2).

The default implementation is just `register`. If you don't need or want the special features of `AutoReg`, you can use that by writing an empty instance.

```haskell
data MyDataType = ...
instance AutoReg MyDataType
```

If you have a product type you can use `deriveAutoReg` to derive an instance.

### DIFF
--- a/clash-prelude/src/Clash/Class/AutoReg/Instances.hs
+++ b/clash-prelude/src/Clash/Class/AutoReg/Instances.hs
@@ -1,0 +1,35 @@
+{-|
+  Copyright   :  (C) 2019, Google Inc.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  Christiaan Baaij <christiaan.baaij@gmail.com>
+-}
+{-# LANGUAGE CPP             #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+module Clash.Class.AutoReg.Instances where
+
+import           Clash.Class.AutoReg.Internal
+import           Clash.CPP                           (maxTupleSize)
+
+#if MIN_VERSION_base(4,12,0)
+import           Data.Complex (Complex)
+import           Data.Ord (Down)
+#endif
+
+import           Data.Ratio (Ratio)
+
+#if MIN_VERSION_base(4,12,0)
+deriveAutoReg ''Complex
+deriveAutoReg ''Down
+#endif
+
+deriveAutoReg ''Ratio
+
+-- TODO: this needs NFDataX instances to maxTupleSize that in turn need
+-- TODO: generic instances for large tuples, but this slows GHC down by a
+-- TODO: lot. See: https://phabricator.haskell.org/D2899. Maybe we could add
+-- TODO: -flarge-tuples to clash-prelude?
+-- deriveAutoRegTuples [2..maxTupleSize]
+deriveAutoRegTuples [2..min maxTupleSize 15]
--- a/tests/shouldwork/AutoReg/AutoReg.hs
+++ b/tests/shouldwork/AutoReg/AutoReg.hs
@@ -1,0 +1,127 @@
+{-|
+  Copyright   :  (C) 2019, Google Inc.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  Christiaan Baaij <christiaan.baaij@gmail.com>
+-}
+{-# LANGUAGE CPP                   #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-# OPTIONS_GHC -Wno-partial-type-signatures #-}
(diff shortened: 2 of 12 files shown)


martijnbastiaan:
@leonschoorl I'm tying the loose ends on this PR, so that we can hopefully merge it tomorrow :)

tests/shouldwork/AutoReg/AutoReg.hs
```diff
@@ -0,0 +1,115 @@
...
+
+test
+  :: forall a dom n rest
+   . ( HiddenClockResetEnable dom
+     , AutoReg a, BitPack a
+     , KnownNat (BitSize a), KnownNat n, KnownNat rest
+     , rest ~ (n-(BitSize a))
+     )
+  => Signal dom (BitVector n) -> Signal dom a
+test = autoReg (unpack 0) . fmap (unpack . fst . split# @rest)
```
martijnbastiaan: Maybe we should add a test for `autoReg1` too, confirming that it only creates to top-level registers.

leonschoorl: We decided to remove `autoReg1` for now

tests/shouldwork/AutoReg/AutoReg.hs
```diff
@@ -0,0 +1,127 @@
...
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
+{-# OPTIONS_GHC -Wno-partial-type-signatures #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
+
+-- Plugins don't load automatically when running the output tests
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.Extra.Solver #-}
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.Normalise #-}
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.KnownNat.Solver #-}
```
martijnbastiaan: Is it possible to use these flags with `runghc`?

clash-prelude/src/Clash/Class/AutoReg/Instances.hs
```diff
@@ -0,0 +1,35 @@
...
+deriveAutoReg ''Complex
+deriveAutoReg ''Down
+#endif
+
+deriveAutoReg ''Ratio
+
+-- TODO: this needs NFDataX instances to maxTupleSize that in turn need
+-- TODO: generic instances for large tuples, but this slows GHC down by a
+-- TODO: lot. See: https://phabricator.haskell.org/D2899. Maybe we could add
+-- TODO: -flarge-tuples to clash-prelude?
```
christiaanb: Or just use TH to generate those instances instead of depending on Generics

martijnbastiaan: Motived me to dig up and freshen up a patch I wrote a while ago: https://github.com/clash-lang/clash-compiler/pull/892

martijnbastiaan:
LGTM. @leonschoorl If you're happy with it, please merge :)

leonschoorl:
Last change: I moved everything around so we don't need orphan instances anymore.

tests/shouldwork/AutoReg/AutoReg.hs
leonschoorl: But because only this test required it, I thought at the time this would be better.
And actually because of other changes these aren't need here anymore. I forgot to take them out.
I'll do another commit to clean them up.

Issue #806, opened 2019-09-17
PortNamesWithRTree test uses a lot memory


leonschoorl:
For some reason compiling (the `testBench` of) the `PortNamesWithRTree` test takes a lot of memory:
```
$ /usr/bin/time cabal run -- clash --vhdl tests/shouldwork/TopEntity/PortNamesWithRTree.hs
GHC: Parsing and optimising modules took: 2.088s
GHC: Loading external modules from interface files took: 0.863s
GHC: Parsing annotations took: 0.001s
Clash: Parsing and compiling primitives took 0.351s
GHC+Clash: Loading modules cumulatively took 12.454s
Clash: Compiling PortNamesWithRTree.topEntity
Clash: Using cached result for: PortNamesWithRTree.topEntity
Clash: Compiling PortNamesWithRTree.testBench
Clash: Using cached result for: PortNamesWithRTree.testBench
Clash: Total compilation took 12.464s
12.23user 0.90system 0:13.16elapsed 99%CPU (0avgtext+0avgdata 3488064maxresident)k
0inputs+2280outputs (0major+860689minor)pagefaults 0swaps

```
> 3488064maxresident)k

3+ GB, which is problematic be to because the CircleCI VMs only have 4GB om memory.
Lately this has been causing test failures specifically with GHC-8.8.1 where memory usage is slightly higher.

Interestingly the problem somewhere pre-normalisation.
The example above isn't even doing normalization, it is using cached result for everything.

alex-mckenna:
Fixed in https://github.com/clash-lang/clash-compiler/pull/944

PR #877, opened 2019-10-26
Tutorial fixes

DigitalBrains1:
Fixes all over. Using CλaSH 1.0.1, code was tested to compile, example
commands where run and specified testbenches where run in both clashi
and in modelsim. No tests were done on an FPGA.

As discussed, this time against master instead of the 1.0 branch.

Some remarks about specific changes:
- JSON-key "type" was included for `blockRam#`, but not for `*#`. Added it for `*#`.

- `blockRam#` missed `KnownDomain`. Changed to match the source. Black boxes where copy-pasted from the real thing, which means `~ACTIVEEDGE` was introduced.

- I had to do three takes to read the sentence

    > Signals of product types and product types (to which tuples belong) of signals are __isomorphic__ [...]

    because of the parenthetical. The sentence is already complicated, better to move that information to a new sentence, IMHO:

    > Signals of product types and product types of signals are __isomorphic__ due to synchronisity principle, but are not (structurally) equal. Tuples are a product type. Use the `bundle` function to convert from a product

    For the record, my brain was parsing that tuples belong to the group defined as "signals of product types and product types", i.e. product types and signals thereof, and after that the sentence broke down semantically.

### DIFF
--- a/clash-prelude/src/Clash/Tutorial.hs
+++ b/clash-prelude/src/Clash/Tutorial.hs
@@ -385,8 +385,8 @@ most contexts.
 The 'register' function is our primary sequential building block to capture
 /state/. It is used internally by one of the "Clash.Prelude" function that we
 will use to describe our MAC circuit. Note that the following paragraphs will
-only show one of many ways to specify a sequential circuit, at the section we
-will show a couple more.
+only show one of many ways to specify a sequential circuit, in the section
+<#mac6 Alternative specifications> we will show a couple more.
 
 A principled way to describe a sequential circuit is to use one of the classic
 machine models, within the Clash prelude library offer standard function to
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
@DigitalBrains1 Your tests should work if you rebase on master. Sorry about that.

martijnbastiaan:
Don't worry about making another PR, just force push! :)

DigitalBrains1:
@martijnbastiaan I'm not sure what I should do. This single commit PR is on top of 2abb2c50, what else should it be?

DigitalBrains1:
Oh, it seems my fetch failed... it's not on top. I'll fix it.

martijnbastiaan:
I see Christiaan added you to the organization too (sorry about any duplicate e-mails you might now get..). If you create a branch on `clash-lang/clash-compiler` you should trigger our own CI which should run quite a bit faster.

martijnbastiaan:
Seems like Circle CI ran out of memory again.. Issue report: https://github.com/clash-lang/clash-compiler/issues/806. 

Thanks for the fixes!

PR #878, opened 2019-10-28
Inline concurrent-supply dependency

martijnbastiaan:
`concurrent-supply` has been blocking a proper 8.8 release for a while now. Judging by the activity on that repo, the project seems pretty much dead. Luckily, the whole project is just a single file, so we can easily house that in clash-lib.


christiaanb:
I dunno about this... can’t you just post an issue on their github asking for an 8.8 release?

martijnbastiaan:
Yeah, that's what Leon did a while ago, but no response.. I'll bump the issue.

martijnbastiaan:
Version bounds on `concurrent-supply` have been fixed.

PR #892, opened 2019-10-31
Generate NFDataX tuple instances with TemplateHaskell

martijnbastiaan:
TODO: 

- [x] Allow users to use `mkNFDataXTupleInstances`, so they can generate more instances if needed
- [x] ~~As a consequence of ^, use fully qualified names~~ Doesn't actually help. Users should just import the right functions/classes.
- [x] Make sure large tuples don't influence Clash developer ergonomics
- [x] ~~Integrate ^ with CI~~ We don't need to. Default should just be different on release branches.

### DIFF
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -212,6 +212,7 @@ Library
                       Clash.Sized.Internal.Unsigned
 
                       Clash.XException
+                      Clash.XException.TH
 
                       Clash.Xilinx.ClockGen
                       Clash.Xilinx.DDR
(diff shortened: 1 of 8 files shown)


martijnbastiaan:
We might want to introduce a way to configure the max tuple size through cabal and set a more sensible default than the absolute max, as we might introduce the same slowdown the GHC developers measured a while ago: https://phabricator.haskell.org/D2899.

christiaanb:
I think the slowdown referred to the `Generic` instances in particular, not instances for large tuples in general.

leonschoorl:
There a significant change in the time it takes to compile `clash-prelude`:
```
-DMAX_TUPLE_SIZE=5      0m37.357s
-DMAX_TUPLE_SIZE=62     2m29.259s
```

But I haven't seen a big difference in compiling with clash.

martijnbastiaan:
Yeah, so what I think we should do is set the default to something reasonable, e.g., 15. Then, for our binary releases, set it to 62.

PR #910, opened 2019-11-06
Refactor testsuite

martijnbastiaan:
"Special" tests remain verbose for now. We probably need to merge
'runTest'/'outputTest'/'runFailingTest' by adding more options to
"TestOptions".


martijnbastiaan:
I think we can do more, but this is a good first step. Agreed @christiaanb ? Could you merge if happy?

PR #864, opened 2019-10-22
Add assertion and coverage functionality

martijnbastiaan:
Proof of concept code for assertion and coverage functionality in Clash. This tries to do the same as #828, but using a ADT keeping track of the combinators used. The ADT is later interpreted and pretty printed by a Haskell blackbox.

Example input:

```haskell
topEntity
  :: Clock System
  -> Signal System Bool
  -> Signal System Bool
  -> Signal System CvResult
topEntity clk a b =
  assert clk "MY_PROP" PSL 
    $ always 
    $ ((a .&&. b) ## b) |-> nextN 2 b
```

results in:

```vhdl
[..]
begin
  my_prop_block : block
    signal s : boolean;
  begin
    s <= a and b;

    -- psl property MY_PROP is
    -- (always {s; b} |-> next[2] b) @ clk;


  end block;
[..]
```

TODO:

- [x] More docs
- [ ] More tests
- [ ] Run it through QuestaSim?

PR #911, opened 2019-11-07
Add 'RenderVoid' option to blackboxes

martijnbastiaan:
Add 'RenderVoid' option to blackboxes
    
    Void data (e.g., `Index 1`, `BitVector 0`, `()`) results in zero-width
    vectors in HDL. HDL tools handle these constructs poorly. To work around
    that 01843d3f1907111c460536ef3e37ccb679d74135 added void-filtering to
    Clash. As a consequence, any functions solely generating a void
    construct aren't rendered at all. Often, this is okay: functions
    yielding voids are generally the result of a generalization and don't
    perform any actually work when instantiated with a void. _Sometimes_
    though, we would like to generate them:
    
     * In the case of `BiSignal`s, one would like to completely remove any
       `BiSignalOut`s from the HDL, as the corresponding `BiSignalIn` will
       already be converted to an `inout` wire. We'd still like to render
       all the blackboxes yielding a `BiSignalOut` though.
    
     * (SystemVerilog) Assertions don't carry any "out" signals. They
       construct a comment instructing the simulator / verification tool to
       check some property. Clash simulation can't support this though:
       functions without outputs simply do not exist. In this case, we'd
       like to mark an "assertion result" as void, so it would get filtered
       in HDL. Similar to the previous use case, we'ld like the comments
       to be rendered still. See #864 for more information.
    
     * Some `IO ()` actions as described in #815 must result in HDL. Like
       the previous two features, `IO ()` would get filtered though.
    
    This commit allows blackboxes to specify whether they would like to be
    rendered even if their result is void. Note that any blackboxes
    specifying this must make sure that they do not assign anything to
    their result - as this is zero bits, Clash won't generate any signal
    declarations for it.
    
    As a happy coincidence, this removes the need for special support for
    BiSignalOut in the compiler.

### DIFF
--- a/clash-lib/src/Clash/Netlist.hs
+++ b/clash-lib/src/Clash/Netlist.hs
@@ -8,12 +8,13 @@
   Create Netlists out of normalized CoreHW Terms
 -}
 
-{-# LANGUAGE MagicHash       #-}
+{-# LANGUAGE MagicHash         #-}
+{-# LANGUAGE MultiWayIf        #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards #-}
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE TupleSections   #-}
-{-# LANGUAGE ViewPatterns    #-}
+{-# LANGUAGE RecordWildCards   #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE ViewPatterns      #-}
 
 module Clash.Netlist where
 
@@ -330,15 +331,7 @@ mkNetDecl (id_,tm) = preserveVarEnv $ do
       go _ _ = return Nothing
 
 
-
-
-isWriteToBiSignalPrimitive :: Term -> Bool
-isWriteToBiSignalPrimitive e = case collectArgs e of
-  (Prim nm _,_) -> nm == StrictText.pack "Clash.Signal.BiSignal.writeToBiSignal#"
-  _             -> False
-
--- | Generate a list of Declarations for a let-binder, return an empty list
--- if the bound expression is represented by 0 bits
+-- | Generate a list of Declarations for a let-binder
 mkDeclarations
   :: HasCallStack
   => Id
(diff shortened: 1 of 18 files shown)


martijnbastiaan:
I was going to add `isVoidType`, but it turned out I didn't need it in the end, so I'll revert the rename in the final version.

clash-lib/src/Clash/Netlist.hs
```diff
@@ -572,6 +570,7 @@ mkFunApp dstId fun args tickDecls = do
           (_,_,_,Component compName compInps co _) <- preserveVarEnv $ genComponent fun
           let argTys = map (termType tcm) args
           argHWTys <- mapM coreTypeToHWTypeM' argTys
+          voidDecls <- renderVoids (zip argHWTys args)
```
christiaanb: Can't we just filter voids later so that `argDecls` contains the these voidDecls?

clash-lib/src/Clash/Netlist.hs
```diff
@@ -769,10 +807,11 @@ mkDcApplication
...
 mkDcApplication dstHType bndr dc args = do
   let dcNm = nameOcc (dcName dc)
-  tcm                 <- Lens.use tcCache
-  let argTys          = map (termType tcm) args
+  tcm <- Lens.use tcCache
+  let argTys = map (termType tcm) args
   argNm <- either return (\b -> extendIdentifier Extended (nameOcc (varName b)) "_dc_arg") bndr
-  argHWTys            <- mapM coreTypeToHWTypeM' argTys
+  argHWTys <- mapM coreTypeToHWTypeM' argTys
+  voidDecls <- renderVoids (zip argHWTys args)
```
christiaanb: Same, can't we filter later so that these `voidDecls` are included in `argDecls`?

clash-lib/src/Clash/Netlist.hs
martijnbastiaan: Hmm, well, we'd later need to filter them again for the `length tysFiltered == length compInps` comparison and the `InstDecl` later on (specifically `inpAssigns`). So I'm not convinced this would improve readability that much.

christiaanb: You'd pull the

```
(argExprs,argDecls)   <- fmap (second concat . unzip) $! mapM (\(e,t) -> mkExpr False (Left dstId) t e) argsFiltered'
```

out of the if-expression, and filter later. You'd still be filtering only once.

martijnbastiaan: Ah right, makes sense.

christiaanb:
@martijnbastiaan When CI passes, this can be merged?

martijnbastiaan:
Yes. Thanks for looking into it.

PR #915, opened 2019-11-14
Add 'WorkerWrapper' to unwanted optimizations list

martijnbastiaan:
The I2C test in our testsuite would trip after some (seemingly
unrelated) internal API changes, after GHC decided to perform a worker
wrapper transformation on a function taking a Clock argument. This
transformation caused Clash to see projections/DC applications of the
Clock type.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -583,6 +583,7 @@ wantedOptimizationFlags df =
                , Opt_Loopification -- STG pass, don't care
                , Opt_CprAnal -- The worker/wrapper introduced by CPR breaks Clash, see [NOTE: CPR breaks Clash]
                , Opt_FullLaziness -- increases sharing, but seems to result in worse circuits (in both area and propagation delay)
+               , Opt_WorkerWrapper -- can cause data constructions of Clocks/Resets to be exposed
                ]
 
     -- Coercions between Integer and Clash' numeric primitives cause Clash to


christiaanb:
@martijnbastiaan What were these changes? Because WW can be beneficial in some cases and that's why it wasn't disabled by default. See e.g. https://github.com/clash-lang/clash-compiler/blob/eca91a0dc4d64cba8f58c760bcd88e8baf431640/clash-prelude/src/Clash/Signal/Internal.hs#L39-L41 to selectively WW for some cases.

martijnbastiaan:
@christiaanb I'm working on removing the most egregious abuses of `unsafePerformIO`. Most notably the one in `Clash.Signal.Trace`, as I'm more or less forced to repeat the same mistakes for `Clash.Validation` if I don't. The basic idea is to make `Signal` (conceptually) a tuple of `MetaData` and the stream itself, and make the application instance of `Signal` merge the meta data of the two signals. This way, meta data from inside a design would "automagically" float to the top - giving users access to internal traces/assertions. 

After a [very basic implementation](https://github.com/clash-lang/clash-compiler/commit/3caa9cca38d3e330eec251ef3bd849131d15c774) Clash would specialize `mealy`, (de)constructing the clock for some reason. I'll see if I can reproduce it sometime this week.

martijnbastiaan:
I should have read the link you sent to start with. I'll investigate whether I can do something similar for my case such that we can leave `WorkerWrapper` enabled.

martijnbastiaan:
We can move the discussion to:

https://github.com/clash-lang/clash-compiler/pull/928

PR #916, opened 2019-11-14
Add acknowledge signal to SPI. Closes #908.

### LINKED ISSUE #908: Send Acknowledge Signals in SPI
Currently when SPI transmits data, it does not send an acknowledge signal to say that the underlying source of the data being sent is free to be changed again. This signal should be added, and the test suite extended to show that data can be changed only after the acknowledge signal has been sent.

alex-mckenna:
The SPI master and slave components now have an acknowledge signal which goes high when data has started to be transferred. This allows devices behind these components which give them data to know when they can change the signal without causing errors in transmission.

While the Lattice reference SPI devices also appear to have an error line, it does not seem necessary here (as bad devices which change the input signal before acknowledgement is sent do not seem to affect the data output buffer for an active device). Dissenting opinions welcome, I am by no means an expert on SPI.

The test suite has also become a lot more flexible, allowing things like SPI mode, slave latching and the sample duration to be arbitrarily chosen using `QuickCheck`. While the tests could still be cleaner (multi-slave tests are still the same as before) it seems to be a reasonable improvement.


alex-mckenna:
@martijnbastiaan, is this more what you wanted to see from commits vis-a-vis rebasing?

Issue #918, opened 2019-11-15
X-Optimization in Clash Core


alex-mckenna:
Currently, Clash leaves all X-optimisation to the logic synthesis tools, including things like case-expressions where all but one branch evaluates to X. The compiler should, either as a separate pass, or as part of an existing pass, eliminate/transform such case-expressions into the non-X branch. Additionally, the WHNF (soon to be NF) evaluator should correctly implement the X-propagation rules for the existing operators.

alex-mckenna:
Update: Offline discussion with @leonschoorl was that there should be a test in `clash-testsuite` that looks at the generated netlist and counts the number of multiplexers that appear (as this is significantly better than just checking that a specific assignment statement appears).

martijnbastiaan:
@alex-mckenna The `Component` in the output of [`genNetlist`](https://github.com/clash-lang/clash-compiler/blob/4eae5faef2b23882f7669206c2ab6761aa8026d8/clash-lib/src/Clash/Netlist.hs#L78) should be what you're looking for. It will [contain `Declaration`s](https://github.com/clash-lang/clash-compiler/blob/4eae5faef2b23882f7669206c2ab6761aa8026d8/clash-lib/src/Clash/Netlist/Types.hs#L154), in turn housing [CondAssignment](https://github.com/clash-lang/clash-compiler/blob/4eae5faef2b23882f7669206c2ab6761aa8026d8/clash-lib/src/Clash/Netlist/Types.hs#L242-L248). 

I'm not entirely convinced this will be less fragile than just checking the textual output, but time will have to tell.

alex-mckenna:
Ja, it might not be _much_ stronger as a test, but at least it doesn't rely on details like internal name generation not changing

PR #923, opened 2019-11-18
Add netlistTest function to Test.Tasty.Clash. Fixes #922

### LINKED ISSUE #922: Netlist Tests in clash-testsuite
Writing tests for #918 are better done by examining the AST of the netlist to be generated, rather than performing string matching on the generated HDL (as we care about the structure, and this is lost when HDL is written to a file).

I propose adding netlist to output tests, where the output is serialized Netlist instead of a particular HDL. It would then be possible to define a function (e.g. `mainNetlist`) for tests which care about looking at this (while still allowing other output tests as before).

alex-mckenna:
It is now possible to write test functions on the netlist AST
generated by clash-lib. The implementation here is slightly ugly,
and the majority has been moved to a separate module because of
this. The reason for this is that the clash executable does not
support stopping and outputting at an arbitrary stage (so clash-lib
is used directly instead).

Merging this PR will go towards allowing #918 to be adequately tested.


testsuite/Test/Tasty/Clash/NetlistTest.hs
```diff
@@ -0,0 +1,108 @@
...
+
+mkClashOpts :: [FilePath] -> ClashOpts
+mkClashOpts idirs = defClashOpts
+  { opt_cachehdl     = False
+  , opt_errorExtra   = True
+  , opt_floatSupport = True
+  , opt_importPaths  = idirs
+  }
+
+backend :: VHDLState
```
christiaanb: Can we test this with the Verilog backend

alex-mckenna: We can, does the choice of backend affect the generated netlist?

martijnbastiaan: Yes, but usually minimally. Some blackboxes will have definitions that can be used inline in one language, but would require an assignment declaration in anther.

martijnbastiaan: @christiaanb Did you mean "test with Verilog _too_", or would you like to discard the VHDL test altogether?

alex-mckenna: Update: The current version allows running netlist tests against any backend, so I would say this is probably resolved.

Q. Can we test this with the Verilog backend?
A. Yes. Yes we can

testsuite/clash-testsuite.cabal
```diff
@@ -1,62 +1,88 @@
...
+    -- it anyway so we're better off doing it beforehand.
+    clash-ghc,
+    clash-lib,
+    clash-prelude
+
   if flag(cosim)
     build-depends:     clash-cosim
 
-  -- hs-source-dirs:
-  default-language:    Haskell2010
```
martijnbastiaan: Why is this removed?

alex-mckenna: It isn't (but I admit the diff could be clearer). It's at the top of the `common` stanza

martijnbastiaan: Woops, thanks.

alex-mckenna:
Rebasing, so stressful. What do you think @martijnbastiaan?

martijnbastiaan:
Looks great!

Don't fear rebasing, [`git reflog` has your back](http://gitready.com/intermediate/2009/02/09/reflog-your-safety-net.html)!

PR #925, opened 2019-11-18
Conditionally stop deduplication

christiaanb:
Normally Clash converts

```haskell
case x of
  A -> f 3 y
  B -> f x x
  C -> h x
```

to

```haskell
let f_arg0 = case x of {A -> 3; _ -> x}
    f_arg1 = case x of {A -> y; _ -> x}
    f_out  = f f_arg0 f_arg1
in  case x of
      A -> f_out
      B -> f_out
      C -> h x
```

i.e. it deduplicates functions (and certain operators) between
case-alternatives to save on area. This comes at the cost of
multiplexers on the arguments of the deduplicated function
(or operator), which add to the propagation delay for the entire
case-expression.

So sometimes we don't want Clash to do the deduplication, i.e.
it adds to the critical path. In that case we can now specifically
tell Clash not to deduplicate:

```haskell
case x of
  A -> noDeDup f 3 y
  B -> f x y
  C -> h x
```

where the application of `f` in the A-alternative is now
explicitly not deduplicated, and the `f` in the B-alternative
is the only remaining application of `f` it also will not be
deduplicated.

If the C-alternative also had an application of `f`, then the
applications of `f` in the B- and C-alternatives would have
been deduplicated; i.e. there would have been two applications
of `f` in total.

### DIFF
--- a/clash-lib/src/Clash/Core/Term.hs
+++ b/clash-lib/src/Clash/Core/Term.hs
@@ -71,6 +71,9 @@ data TickInfo
   | NameMod !NameMod !Type
   -- ^ Modifier for naming module instantiations and registers, are added by
   -- the user by using the functions @Clash.Magic.[prefixName,suffixName,setName]@
+  | NoDeDup
+  -- ^ Do not deduplicate, i.e. /keep/, an expression inside a case-alternative;
+  -- do not try to share expressions between multiple branches.
   deriving (Eq,Show,Generic,NFData,Hashable,Binary)
 
 -- | Tag to indicate which instance/register name modifier was used
--- a/clash-prelude/src/Clash/Magic.hs
+++ b/clash-prelude/src/Clash/Magic.hs
@@ -1,7 +1,7 @@
 {-|
   Copyright   :  (C) 2019, Myrtle Software Ltd
   License     :  BSD2 (see the file LICENSE)
-  Maintainer  :  QBayLogic B.V. <clash@qbaylogic.com>
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 
 Control module instance, and register, names in generated HDL code.
 -}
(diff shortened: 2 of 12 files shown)


clash-prelude/src/Clash/Magic.hs
```diff
@@ -42,3 +42,11 @@ setName
   :: forall (name :: Symbol) a . a -> name ::: a
 setName = id
 {-# NOINLINE setName #-}
+
+-- | Do not deduplicate, i.e. /keep/, an applied function inside a
+-- case-alternative; do not try to share the function between multiple
+-- branches
```
martijnbastiaan: It would be useful to add the "why" to this.

"This is useful for very small functions, where the overhead of sharing would be bigger than simply duplicating it."

martijnbastiaan: PS: How do people make these fancy commitable diff reviews? I can't find it.

clash-lib/src/Clash/Core/Term.hs
```diff
@@ -71,6 +71,7 @@ data TickInfo
   | NameMod !NameMod !Type
   -- ^ Modifier for naming module instantiations and registers, are added by
   -- the user by using the functions @Clash.Magic.[prefixName,suffixName,setName]@
+  | NoDeDup
```
martijnbastiaan:
```
-- ^ Do not deduplicate, i.e. /keep/, an expression inside a
-- case-alternative; do not try to share the expression between multiple
-- branches
```

christiaanb:
@martijnbastiaan I've added some more documentation to the `noDeDup` function to describe its purpose and use.

martijnbastiaan:
Looks great :)!

tests/shouldwork/Netlist/NoDeDup.hs
```diff
@@ -0,0 +1,74 @@
...
+topEntity n abcd = (f n abcd) - (g n abcd)
+
+testPath :: FilePath
+testPath = "tests/shouldwork/Netlist/NoDeDup.hs"
+
+isTwiceInst (InstDecl Entity Nothing "twice" _ _ _) = True
+isTwiceInst _ = False
+
+assertAssignsInOut :: Component -> IO ()
+assertAssignsInOut (Component nm inps outs ds) =
```
alex-mckenna: `assertAssignsInOut` was picked because it was more specific to the example netlist test I defined. I would probably rename this to something more meaningful for this test (like `assertNumTwiceInsts`)

martijnbastiaan: Woops, yes! I just mindlessly copied it.

martijnbastiaan: Done, thanks for catching that.

PR #926, opened 2019-11-19
Fix type of (:<), add COMPLETE pragmas. Fixes #867.

### LINKED ISSUE #867: View pattern definition of :> doesn't play nicely with typechecker
(This is all with Clash 1.0.0)

I think the following program should typecheck:

```
import Clash.Prelude

foo :: (1 <= n, KnownNat n) => Vec n a -> Vec n a
foo = id

bar :: (KnownNat n) => Vec n a -> Vec n a
bar Nil = Nil
bar xs@(_ :> _) = foo xs
```

but it doesn't, because:

```
/tmp/a.hs:16:10: error:
    • Couldn't match type ‘n’ with ‘n0 + 1’
      ‘n’ is a rigid type variable bound by
        the type signature for:
          bar :: forall (n :: Nat) a. KnownNat n => Vec n a -> Vec n a
        at /tmp/a.hs:14:1-42
      Expected type: Vec n a
        Actual type: Vec (n0 + 1) a
    • In the pattern: _ :> _
      In an equation for ‘bar2’: bar2 xs@(_ :> _) = foo xs

/tmp/a.hs:16:20: error:
    • Couldn't match type ‘1 <=? n’ with ‘'True’
        arising from a use of ‘foo’
    • In the expression: foo xs
      In an equation for ‘bar’: bar xs@(_ :> _) = foo xs

```

The same code works when using `Cons` directly:

```
bar1 :: (KnownNat n) => Vec n a -> Vec n a
bar1 Nil = Nil
bar1 xs@Cons{} = foo xs
```

and moreover, it works if I define a `(:>)`-like pattern synonym, but using `Cons` directly instead of `head` and `tail`:

```
pattern x ::> xs = Cons x xs

bar2 :: (KnownNat n) => Vec n a -> Vec n a
bar2 Nil = Nil
bar2 xs@(_ ::> _) = foo xs
```

alex-mckenna:
The previous changes in this issue did not in practice work as
expected in all cases, as it gave some type errors for correct
definitions using (:<), and gave warnings that functions using
the pattern synonyms did not cover all cases. These have been
amended now, and the issue should be solved.

### DIFF
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -321,6 +321,7 @@ test-suite unittests
       clash-prelude,
 
       ghc-typelits-knownnat,
+      ghc-typelits-natnormalise,
 
       base,
       hint          >= 0.7      && < 0.10,
@@ -338,6 +339,7 @@ test-suite unittests
                  Clash.Tests.Signal
                  Clash.Tests.NFDataX
                  Clash.Tests.TopEntityGeneration
+                 Clash.Tests.Vector
 
 
 benchmark benchmark-clash-prelude
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -1,5 +1,5 @@
 {-|
-Copyright  :  (C) 2014-2016, University of Twente,
+Copyright  :  (C) 2013-2016, University of Twente,
                   2017     , Myrtle Software Ltd
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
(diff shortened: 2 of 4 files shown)


martijnbastiaan:
Can we add tests for the type errors in the the unittests of `clash-prelude`?

alex-mckenna:
As in define a test in `shouldwork` that has a bunch of vector functions which should typecheck?
I understand the motivation, are there any special cases that we should check work?

martijnbastiaan:
Not in shouldwork, just in the unittests. So basically a Haskell program that does some vector manipulations. Similar to https://github.com/clash-lang/clash-compiler/blob/master/clash-prelude/tests/Clash/Tests/Signal.hs#L163. I basically want to prevent feature seemingly innocent API changes (like `m <= n + 1` to `m < n`) to break existing usage.

alex-mckenna:
Ah right, I can give that a go :+1:
I'll preemptively add you as a reviewer now so you can check the tests before it goes in

PR #928, opened 2019-11-20
Remove global trace map in `Clash.Signal.Trace`

martijnbastiaan:
This fixes most of the weird behavior of `traceSignal`. We can reuse this infrastructure for assertions and validation in the future.

```
Note [The why and how of "SignalMeta"]
--------------------------------------------------------------
For various features, most notably traces (see "Clash.Signal.Trace") and
assertions (see "Clash.Validation"), we'd like access to some data structure
created inside a design. For example, a user might want to trace a signal
for debugging purposes. At first glance, we've got three basic design options:

  1. Make the user propagate the trace all the way to their top-level function
  2. Use 'unsafePerformIO' to print every value to stdout
  3. Use 'unsafePerformIO' to store the trace in a global map

All three of these options leave a lot to be desired. We either ask too much of
the user (1) or we introduce all the dragons that come with 'unsafePerformIO'
(2, 3).

"SignalMeta" tries to solve these problems by redefining "Signal" to not just
be a stream of values (~ `a :- Signal a`), but a product of meta information and
the stream. 'traceSignal' can now store its trace in the signal's metadata. Any
time two signals are combined, the combinator should make sure the metadata from
both signals is propagated in the combined one. Ignoring blackboxes for now, the
only function combining two signals is '<*>' from the "Applicative" instance of
"Signal". So the next question is, what should the implementation of the
function merging the two metadatas look like? The most obvious implementation
(assuming SignalMeta carries some hashmap) would be:

@
  mergeMeta :: SignalMeta -> SignalMeta -> SignalMeta
  mergeMeta (SignalMeta f1) (SignalMeta f2) = SignalMeta (HashMap.union f1 f2)
@

This simple implementation doesn't really work though. In a feedback loop 'f1'
and 'f2' might (indirectly) be the result of @HashMap.union f1 f2@. Using it
therefore sets off an infinite loop.

We'd like to break the infinite loop triggered by the naive version of
'mergeMeta' written earlier. Instead of adding data to the map directly, we
store a function instead. This function takes a set of _seen_ identifiers. Every
merge-point ('mergeMeta') will generate a unique id and return the aforementioned
function with that id in scope. The function will check whether it has seen it's
own merge-point before. If it hasn't, it will add itself to the set of seen ids
and merge the incoming metadatas. If its own id _is_ in there, it can assume all
the metadata it would add is already in the metadata the caller collected so
far. It can therefore break the loop and simply return an empty map.
```

------------------------------------

Concerning performance: if you don't use the signal metadata, GHC can garbage collect it all, as any functions recurse over _streams_, not signals.

------------------------------------

@christiaanb Concerning `Add 'WorkerWrapper' to unwanted optimizations list`: the PR currently still need it. The I2C testbench fails as soon as we add "SignalMeta" to "Signal", even if "SignalMeta" itself is a data type with a single constructor with zero fields. The behavior is only triggered when using `-O2`. GHC then produces:

```
I2C.ByteMaster.$w$smealy43532 before normalization:

[..]

    λ(ww8286623314361987328 :: Clash.Promoted.Symbol.SSymbol8214565720323789525
                                 "System") ->
    λ(w8286623314361987321 :: Clash.Signal.Internal.Reset8214565720323789571
                                "System") ->
    λ(w18286623314361987322 :: Clash.Signal.Internal.Enable8214565720323789562
                                 "System") ->
    λ(w28286623314361987323 :: I2C.ByteMaster.ByteMasterS8214565720324019851
                           -> i8286623314361987318
                           -> GHC.Tuple.(,)3746994889972252676
                                I2C.ByteMaster.ByteMasterS8214565720324019851
                                o8286623314361987319) ->
    λ(w38286623314361987324 :: I2C.ByteMaster.ByteMasterS8214565720324019851) ->
    λ(w48286623314361987325 :: Clash.Signal.Internal.Signal8214565720323789592
                                 "System"
                                 i8286623314361987318) ->
```

the weird thing is that it's then only used to reconstruct the clock later on..

```
          (Clash.Signal.Internal.register# @"System"
             @I2C.ByteMaster.ByteMasterS8214565720324019851
             Clash.Signal.Internal.$fKnownDomain"System"8214565720323878380[GlobalId]
             I2C.ByteMaster.$fNFDataXByteMasterS8214565720324020842[GlobalId]
             (Clash.Signal.Internal.Clock8214565720323878019
                @"System"
                ww8286623314361987328[LocalId])
```

I don't understand why this is useful, or why it's triggered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
   * Added `Clash.Class.AutoReg`. Improves the chances of synthesis tools inferring clock-gated registers, when used. See [#873](https://github.com/clash-lang/clash-compiler/pull/873).
   * `Clash.Magic.suffixNameP`, `Clash.Magic.suffixNameFromNatP`: enable prefixing of name suffixes
   * Added `Clash.Magic.noDeDup`: can be used to instruct Clash to /not/ share a function between multiple branches
+  * Added `Clash.Annotations.Primitive.idBlackBoxTH`: can be used to easily create blackboxes that simply forward one of its arguments.
 
 * New internal features:
   * [#821](https://github.com/clash-lang/clash-compiler/pull/821): Add `DebugTry`: print name of all tried transformations, even if they didn't succeed
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -584,6 +584,7 @@ wantedOptimizationFlags df =
                , Opt_Loopification -- STG pass, don't care
                , Opt_CprAnal -- The worker/wrapper introduced by CPR breaks Clash, see [NOTE: CPR breaks Clash]
                , Opt_FullLaziness -- increases sharing, but seems to result in worse circuits (in both area and propagation delay)
+               , Opt_WorkerWrapper -- can cause data constructions of Clocks/Resets to be exposed
                ]
 
     -- Coercions between Integer and Clash' numeric primitives cause Clash to
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -265,6 +265,7 @@ Library
                       containers                >= 0.4.0   && < 0.7,
                       data-binary-ieee754       >= 0.4.4   && < 0.6,
                       data-default-class        >= 0.1.2   && < 0.2,
+                      data-default              >= 0.7     && < 0.8,
                       integer-gmp               >= 0.5.1.0 && < 1.1,
                       deepseq                   >= 1.4.1.0 && < 1.5,
                       ghc-prim                  >= 0.5.1.0 && < 0.6,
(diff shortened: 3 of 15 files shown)


christiaanb:
See https://github.com/haskell/containers/blob/master/containers/src/Data/Map/Internal.hs#L813 for more tricks to prevent WW locally

martijnbastiaan:
Good read, thanks. 

I'm afraid it doesn't really help the `Opt_WorkerWrapper` case, as it triggers on seemingly innocent user code. We shouldn't ask our users to jump through hoops. However, this WW transformation is pretty straightforward to undo manually, so maybe we should consider adding a transformations that undoes WW transformations on specific types (`Clock`, `Reset`).

As for the stable-name story: I'm afraid observable sharing is just not reliable (enough) in GHC's Haskell. For it to work, we need a guarantee that we can see it. None of the GHC internals guarantee it, resulting in unpredictable behavior.

martijnbastiaan:
Now that we've enabled `MonomorphismRestriction` this shouldn't be needed anymore. Removing the global map would be nice, but the only proper solution seems to be to move to a monadic approach. Some akin https://gitlab.com/martijnbastiaan/clash-hidden.

PR #933, opened 2019-11-21
Fix instance Data Vec

leonschoorl:
The default implementation of `gfoldl` only "sees" the constructor.

This makes [`Data.Generics.Text.gshow`](https://hackage.haskell.org/package/syb-0.7.1/docs/Data-Generics-Text.html#v:gshow) work properly on Vec's.


martijnbastiaan:
What does it mean to only "see" the constructor?

leonschoorl:
The documentation for [`gfoldl`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Data.html#v:gfoldl) says:
> The default definition for gfoldl is const id, which is suitable for abstract datatypes with no substructures.

Example of using `gshow`:
```
Clash.Prelude> import Data.Generics.Text
Clash.Prelude Data.Generics.Text> gshow $ Just (True :> Nil)
"(Just (Cons))"
```

With this patch we get:
```
Clash.Prelude Data.Generics.Text> gshow $ Just (True :> Nil)
"(Just (Cons (True) (Nil)))"
```

christiaanb:
@leonschoorl Could you add a CHANGELOG entry? Thanks

Issue #792, opened 2019-09-11
Missing blackbox for `GHC.Integer.Type.divModInteger`


martijnbastiaan:
Pretty much the title.

leonschoorl:
And `GHC.Integer.Type.quotRemInteger`

PR #934, opened 2019-11-21
Fix Integral blackboxes

leonschoorl:
Because we weren't actively testing the HDL support for the Integral class methods(`div`,`mod`,`rem`,`quot`,...), it wasn't great.

  * `divModInteger` and `quotRemInteger` were missing #792
  * some blackboxes didn't produces valid HDL
  * the results of `div` on signed numbers were sometimes off by one
  * the results of `mod` on signed numbers were sometimes off by one divisor

This is now all fixed.


christiaanb:
@leonschoorl could you add a CHANGELOG entry? thanks

PR #937, opened 2019-11-22
Fix running clash-testsuite locally

alex-mckenna:
Apparently `clash-testsuite` needs to be built with tests enabled
in order to run tests (as otherwise it complains about the library
modules being in a hidden package). This commit enables tests by
default for `clash-testsuite` so the testsuite can run locally
without problems again.


alex-mckenna:
The reasoning is sound, but it's still getting a thumbs down because it makes me sad.

Is it still the same error?

martijnbastiaan:
Yes, same error.

PR #939, opened 2019-11-25
Error on warnings in clash-cores

martijnbastiaan:
@alex-mckenna If we enable warning checking on the CI for clash-cores, it fails due to unused binders. Did you intend to export these, or should we remove them?

### DIFF
--- a/.ci/cabal.project.local
+++ b/.ci/cabal.project.local
@@ -19,6 +19,10 @@ package clash-cosim
   ghc-options: -Werror
   documentation: True
   tests: True
+package clash-cores
+  ghc-options: -Werror
+  documentation: True
+  tests: True
 
 package clash-testsuite
   ghc-options: -Werror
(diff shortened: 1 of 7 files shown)


alex-mckenna:
Which bindings are unused? I made some top-level definitions to give names to some things. If any of those are not used they could be deleted (as it is obvious enough how to redefine them if needed later).

martijnbastiaan:
According to the CI logs:

```
1971     Defined but not used: ‘idleOnHigh’
1972    |
1973 71 | idleOnHigh = not . idleOnLow
1974    | ^^^^^^^^^^
1975 src/Clash/Cores/SPI.hs:79:1: error: [-Wunused-top-binds, -Werror=unused-top-binds]
1976     Defined but not used: ‘sampleOnFalling’
1977    |
1978 79 | sampleOnFalling = not . sampleOnRising
1979    | ^^^^^^^^^^^^^^^
```

I'll delete them then. Thanks!

PR #943, opened 2019-11-27
Remove :> pattern, rename Cons to :>

martijnbastiaan:
This solves the performance penalty induced by the pattern. We have
supported GADT pattern matching for a while now, so this shouldn't
affect synthesizability.

### DIFF
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -38,7 +38,7 @@ Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 
 module Clash.Sized.Vector
   ( -- * 'Vec'tor data type
-    Vec(Nil,(:>),(:<))
+    Vec(Nil,(:>),Cons,(:<))
     -- * Accessors
     -- ** Length information
   , length, lengthS
@@ -275,6 +275,46 @@ cCons = mkConstr tVec ":>" [] Prefix
 instance NFData a => NFData (Vec n a) where
   rnf = foldl (\() -> rnf) ()
 
+
+{-# COMPLETE Nil, (:>)  #-}
+-- | Add an element to the head of a vector.
+--
+-- >>> 3 :> 4 :> 5 :> Nil
+-- <3,4,5>
+-- >>> let x = 3 :> 4 :> 5 :> Nil
+-- >>> :t x
+-- x :: Num b => Vec 3 b
+--
+-- Can be used as a pattern:
+--
+-- >>> :set -Wno-incomplete-patterns
+-- >>> let f :: Num a => Vec (n + 2) a -> a; f (x :> y :> _) = x + y
+-- >>> :t f
+-- f :: Num a => Vec (n + 2) a -> a
+-- >>> f (3 :> 4 :> 5 :> 6 :> 7 :> Nil)
+-- 7
+--
+-- Also in conjunctions with (':<'):
+--
+-- >>> let g :: Num a => Vec (n + 4) a -> a; g (a :> b :> (_ :< y :< x)) = a + b +  x + y
+-- >>> :t g
+-- g :: Num a => Vec (n + 4) a -> a
+-- >>> g (1 :> 2 :> 3 :> 4 :> 5 :> Nil)
+-- 12
+--
+-- N.B.: Deprecated! Use (:>)
+pattern Cons
+  :: forall m a
+   . ()
+  => forall n
+   . ((n + 1) ~ m)
+  => a -> Vec n a -> Vec m a
+pattern Cons x xs = x :> xs
+{-# DEPRECATED Cons "Cons will be removed in Clash 1.4. Use :> instead." #-}
+
+infixr 5 `Cons`
+
+
 {-# COMPLETE Nil, (:<)  #-}
 -- | Add an element to the tail of a vector.
 --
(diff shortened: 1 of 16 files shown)


martijnbastiaan:
The CI currently fails because GHC warns us about "incomplete" pattern matching. I.e., the following:

```haskell
middleL :: Vec (n + 1) a -> Vec n a
middleL (_ :> xs) = xs
```

Gives us:

```
src/Clash/Sized/Vector.hs:279:1-22: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘middleL’: Patterns not matched: Nil
    |
279 | middleL (_ :> xs) = xs
    | ^^^^^^^^^^^^^^^^^^^^^^
```

This is a known problem (judging by `{-# OPTIONS_GHC -fno-warn-incomplete-patterns -#}` at the top of `Vector.hs`), but if we'd merge this PR we'd expose this behavior to the user. We've got a few ways of fixing this:

1. Add a `{-# COMPLETE (:>) -#}` pragma. Problem: not all case expressions solely matching on `:>` are complete!
2. Add something like `{-# COMPLETE (:>) :: Vec (n+1) -#}`. We would first need to fix GHC to support this though.
3. Write a plugin adding a `Nil` case with `error "not reachable"` if it can prove `n+1`.
4. Have users deal with it and write the not reachable fluff themselves.

martijnbastiaan:
This has apparently been a problem for a very long time now, and we've solved it with a "dangerous" COMPLETE pragma until now. So that's what I'm going to do for this PR too..

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1380,7 +1343,7 @@ replace i y xs = replace_int xs (fromEnum i) y
 -- <interactive>:...
 --     • Couldn't match type ‘4 + n0’ with ‘2’
 --       Expected type: Vec (4 + n0) a
---         Actual type: Vec 2 a
+--         Actual type: Vec (1 + 1) a
```
martijnbastiaan: I can't really explain this change..

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1577,7 +1540,7 @@ transpose = traverse# id
 --
 -- >>> let xs = (1:>2:>3:>4:>5:>6:>Nil)
 -- >>> :t xs
--- xs :: Num a => Vec 6 a
+-- xs :: Num b => Vec 6 b
```
martijnbastiaan: Neither can I explain this one.

martijnbastiaan:
@leonschoorl Could you take a look at the change in AutoReg? Did you intent to write `-Wwarn-[..]` instead of `-fno-warn-[..]`? The CI fails on 8.4.4 if I don't make that change..

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -208,9 +210,10 @@ let populationCount' :: (KnownNat k, KnownNat (2^k)) => BitVector (2^k) -> Index
 --   ending at @'length' - 1@.
 data Vec :: Nat -> Type -> Type where
   Nil  :: Vec 0 a
-  Cons :: a -> Vec n a -> Vec (n + 1) a
+  (:>) :: a -> Vec n a -> Vec (n + 1) a
 
-infixr 5 `Cons`
+infixr 5 :>
+{-# COMPLETE (:>) #-}
```
christiaanb: No. I don't know why we ever had

```
{-# COMPLETE (:>)       #-}	
{-# COMPLETE Nil, (:>)  #-}
```

in there previously, that obviously slipped through the review. I'd rather have the "false positive" behavior of the warnings GHC emits, than the "false negatives" this COMPLETE pragma gives you. Also in GHC 8.12 (maybe even 8.10), the pattern match checker is much improved, hopefully no longer giving the "false positives".

leonschoorl:
@martijnbastiaan `unused-pattern-binds` as actually not causing a build failure here.
What is happening here:
1. `-Wall` enables all warnings
2. `-Werror` turns all warnings into errors
3. https://github.com/clash-lang/clash-compiler/blob/054541e3545e03141a95c27c933d6b3c65157688/clash-prelude/src/Clash/Class/AutoReg/Internal.hs#L19-L21
  Turns `unused-pattern-binds` back into a warning.

But simply disabling the `unused-pattern-binds` warning is much less confusing, so I'm fine with that.

leonschoorl:
Should we keep `Cons` as a pattern for backwards compatibility?
```haskell
infixr 5 `Cons`
pattern x `Cons` xs = x :> xs
pattern Cons :: ...
{-# DEPRECATED Cons "..." #-}
```

clash-prelude/src/Clash/Sized/Vector.hs
martijnbastiaan: I fully agree! Good to know that future GHC's will do it better.

leonschoorl:
I think the slowness of the pattern is a GHC bug.

If you hide the import of `(:>)` and copy the `pattern :>` into the NORX file it compiles in seconds.

clash-prelude/src/Clash/Sized/Vector.hs
martijnbastiaan: Can I read about the improved pattern match checker? I've found [this](https://gitlab.haskell.org/ghc/ghc/wikis/Migration/8.10#improvements-to-pattern-match-checking-with-user-defined-complete-sets), but that doesn't seem relevant for us.

martijnbastiaan:
@leonschoorl I've readded `Cons` with a `DEPRECATED` pragma

@christiaanb I've removed the `COMPLETE` pragma.

clash-prelude/src/Clash/Sized/Vector.hs
christiaanb: SPJ gave a keynote about it at Haskell Exchange this year

PR #946, opened 2019-11-28
Make tuples in tests fit within MAX_TUPLE_SIZE

alex-mckenna:
The tuples in the numbers tests now fit within MAX_TUPLE_SIZE.
Currently this makes one test fail in iverilog (IntegralTB), which
should be examined further.


alex-mckenna:
Apparently, the `iverilog` issue on my machine isn't reproduced by the CI. This PR is being kept as a draft until I understand why...

martijnbastiaan:
If it turns out to be a regression in `iverilog`, I'd be fine with disabling the test for it.

martijnbastiaan:
I really can't reproduce this on four different machines (CI, laptop running 16.04, desktop running 19.10, desktop running 18.04). Judging by the error you were seeing it was either a bug in iverilog, or a botched installation. With that in mind, I'm going to merge this.

PR #947, opened 2019-11-28
Add "Clash.Core.TermLiteral"

martijnbastiaan:
This feature is needed for https://github.com/clash-lang/clash-compiler/pull/864, but might come in handy in other places too.

In the future we might want to add:

* Template Haskell to generate tuple instances.
* `dataToTerm`

### DIFF
--- a/clash-lib/clash-lib.cabal
+++ b/clash-lib/clash-lib.cabal
@@ -171,6 +171,8 @@ Library
                       Clash.Core.Pretty
                       Clash.Core.Subst
                       Clash.Core.Term
+                      Clash.Core.TermLiteral
+                      Clash.Core.TermLiteral.TH
                       Clash.Core.TyCon
                       Clash.Core.Type
                       Clash.Core.TysPrim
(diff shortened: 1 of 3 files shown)


leonschoorl:
Would this also work for types with arguments?
Like `Unsigned n` and friends?

leonschoorl:
For `Value` it seems there are three ways to encode an Integer:
https://github.com/clash-lang/clash-compiler/blob/c860a3924646964ebdcc32d078c803ddaa7c3560/clash-lib/src/Clash/Core/Evaluator.hs#L441-L453
Would the same apply to `Term`?

christiaanb:
@leonschoorl Yes, `termToData` for `Integer` should be handle the other cases.

martijnbastiaan:
@leonschoorl I'm not sure what 'this' refers to, but I guess to the whole approach? I don't see a reason why it wouldn't work.

martijnbastiaan:
And yes, `TermLiteral` should be able to handle those cases. Feel free to add an issue / pr.

leonschoorl:
@martijnbastiaan I meant: shouldn't there be an `instance TermLiteral (Unsigned n)`?
Or is this not possible?

martijnbastiaan:
Discussed offline: yes, it's possible. I just didn't need it for my purposes, so I didn't implement it.

PR #949, opened 2019-12-02
Hide X-Optimization Behind a Flag

alex-mckenna:
The X-optimization feature previously added is now only enabled
when -faggressive-x-optimization is set. As a result, the testsuite
now takes a function to modify the clash options before running
instead of a list of extra include dirs.

### DIFF
--- a/clash-lib/src/Clash/Normalize.hs
+++ b/clash-lib/src/Clash/Normalize.hs
@@ -117,6 +117,7 @@ runNormalization opts supply globals typeTrans reprs tcm tupTcm eval primMap rcs
     rwEnv     = RewriteEnv
                   (opt_dbgLevel opts)
                   (opt_dbgTransformations opts)
+                  (opt_aggressiveXOpt opts)
                   typeTrans
                   tcm
                   tupTcm
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
   * `Vec`'s `Cons` is gone in favor of `:>`. You can now use the actual constructor instead of a pattern to do pattern matches! See [#943](https://github.com/clash-lang/clash-compiler/pull/943).
 
 * New internal features:
-  * [#918](https://github.com/clash-lang/clash-compiler/pull/935): Add X-Optimization to normalization passes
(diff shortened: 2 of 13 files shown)


rowanG077:
Beside the option being mentioned in the changelog could this be documented somewhere as well? It's not really discoverable like this I think.

alex-mckenna:
Sounds like a good idea, I didn't spot a clash-specific flag reference to updat when editing. Does such a thing exist @martijnbastiaan @christiaanb?

christiaanb:
@alex-mckenna No, flags are pretty much un(der)documented.

alex-mckenna:
In that case I would say ignore not having the kind of documentation @rowanG077 wants for now in this PR, but open a new documentation issue for properly documenting the flags in `clash` and `clashi` (which is actually doubly useful because I just realised `clash --show-options` doesn't appear to show any Clash-specific options).

clash-lib/src/Clash/Normalize.hs
```diff
@@ -117,6 +117,7 @@ runNormalization opts supply globals typeTrans reprs tcm tupTcm eval primMap rcs
     rwEnv     = RewriteEnv
                   (opt_dbgLevel opts)
                   (opt_dbgTransformations opts)
+                  (opt_aggressiveXOpt opts)
```
martijnbastiaan: Just a thought: maybe we should add the whole of `ClashOpts` to `RewriteEnv`, instead of adding each flag individually.

alex-mckenna: A small subset of `ClashOpts` is added to `rwEnv` and a different small subset to `normState`, if `rwEnv` is available everywhere `normState` is, then I would say it's okay to just include the options in `rwEnv`.

But equally not every clash option is relevant to rewriting / normalization, so there's a case to be made for the current style too :man_shrugging:

martijnbastiaan: Yeah, it's not for now, that's for sure. I just dislike that we have to change that much for adding a simple flag.

christiaanb: in this case it seems like just 2 extra lines though

PR #961, opened 2019-12-09
Show `-fclash-*` Options in `clash --show-options`

alex-mckenna:
When calling clash with the --show-options flag, it now prints
all options, including common clash options which were previously
not being shown.

~It remains to be seen how to document all the flags which are
added by clash (instead of GHC).~

EDIT: Flag reference is a separate issue, see #987


alex-mckenna:
Closed in favour of #962

PR #966, opened 2019-12-13
Revert vector pattern "fixes"

martijnbastiaan:
This reverts commits:

  * 338558a48fb9ffa0cd6e89af5a262395fdfa81a5
  * e6872e973d04978af1274b29e15e8e5c280fcd50
  * d7f4e48bb8b140ecda69ff1cb0a23fb8f6fc498f
  * 1b1d03c123459846165d66a8b62e26b30d2bfa3d
  * 0db919241ca20f3da4f4afbfe3fe857d93beed09
  * bf453ae0623a80c8c35bbb1d22de02df783a10b6
  * 060ce16b801b24c2e77297d4256cef8943c46225

We found too many issues with exporting the "real" vector constructor:

  1. While it improves type inference in some cases (see issue #867), it
  worsens it in others. (See testsuite changes in #943.)

  2. GADT don't permit lazy patterns, which seems to be widely used in
  existing code bases.

  3. Using the Vec constructor can trigger a bug in GHC stochastically
  making compilation _super_ slow. Having it as a default is therefore
  not ideal for now.


leonschoorl:
We are exporting also exporting the real vector constructor `Cons`.
And thanks to the GADT support, code using it is now synthesizable.
So user can use both `Cons` and `:>`.
Only when deconstructing `Vec`'s,  they have different "type-inferenceability".
`:>` is usually more convenient.
But if you want to write thing like in #867 you must use `Cons`.

PR #970, opened 2019-12-14
Replace odd and even with clash specific Parity type class

### LINKED ISSUE #959: Index: Bug in even and odd
Hi again,

I'd like to report some bugs in 'even' and 'odd'. (Clash, version 1.0.1 (using clash-lib, version: 1.0.1))

    -- even :: Integral a => a -> Bool

    bug1 = even (0 :: Index 1)
    bug2 = even (0 :: Index 2)
    bug3 = even (1 :: Index 2)

Now bug1 gives

*** Exception: X: Clash.Sized.Index: result 2 is out of bounds: [0..0]
  errorX, called at src/Clash/Sized/Internal/Index.hs:255:13 in clash-prelude-1.0.1-BorHjli5pMX3STUvZInnYc:Clash.Sized.Internal.Index
  fromInteger_INLINE, called at src/Clash/Sized/Internal/Index.hs:249:16 in clash-prelude-1.0.1-BorHjli5pMX3STUvZInnYc:Clash.Sized.Internal.Index

You get similar errors for 'odd'.

Here's a possible fix:

    odd', even' :: (KnownNat n) => Index (n+1) -> Bool                         
    odd' = bitToBool . lsb . pack                                     
    even' = not . odd'

  * These errors are __extremely__ difficult to find because the error messages thrown do not refer to sections in the user's source file. If the error is caused by a statement/function call in a deeply nested where clause, you can only find the error by making the deeply nested statements top-level functions and play with these functions until you've found the cause. It can take very long before you know which statement is the cause...
  * I get similar errors like these a lot. They're usually caused by errors on my behalf. Finding them can take very long.
  * Is there any way you can introduce a flag which forces error messages to pin-point the cause of the error to a location in the user's source file?

TIA for your time.

Regards,

Paddy

rowanG077:
Fixes #959

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 1.1.0
 * New features (API):
+  * `Clash.Class.Parity` type class replaces Prelude `odd` and `even` functions due
+    to assumptions that don't hold for Clash specific numerical types, see
+    [#970](https://github.com/clash-lang/clash-compiler/pull/970).
   * `NFDataX.ensureSpine`, see [#748](https://github.com/clash-lang/clash-compiler/pull/803)
   * `makeTopEntity` Template Haskell function for generating TopEntity annotations
     intended to cover the majority of use cases. Generation failures should either
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -150,6 +150,7 @@ Library
                       Clash.Class.HasDomain.CodeGen
                       Clash.Class.HasDomain.Common
                       Clash.Class.Num
+                      Clash.Class.Parity
                       Clash.Class.Resize
 
                       Clash.Clocks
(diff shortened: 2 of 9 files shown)


rowanG077:
There is [Ones' complement](https://en.wikipedia.org/wiki/Ones%27_complement) but I don't think it's used in practice much.

martijnbastiaan:
Hmm, alright. I think a default implementation shouldn't silently return a wrong result, so I think we should stick to the 'mod 2` version.

rowanG077:
Then I would say a default implementation is not possible. It's not possible to tell in a generic way whether `2` fits in the type. I think it's best to just duplicate the code for the instances then.

You are worried about users using the default implementation? If so we could implement `even` and `odd` internally using `lsb` and just not have a default implementation.

martijnbastiaan:
> You are worried about users using the default implementation?

Basically. IMO default implementations should _never_ fail silently. 

Throwing a runtime error is not ideal, but at least users will notice quickly. So in my mind it might be a good compromise between code duplication / maintainabilty and being perfectly correct. This is by no means a hard line though, so if you believe having a bit of code duplication is better, I think we should roll with that :).

martijnbastiaan:
Thanks @rowanG077 !

PR #971, opened 2019-12-15
Fix Index n where n overflows an Int to 0 resulted in wrong behavior 

### LINKED ISSUE #964: SaturatingNum instance for Index has broken behavior when n overflows an Int to 0
The SaturatingNum instance of `Index` internally uses a case statement on the size of the index to choose a different implementation. It converts the size of the index to a `Int` which means if this `n` is large enough it can overflow and cause it to choose the wrong implementation.

example of wrong behavior:

```
satAdd SatWrap (0 :: Index 18446744073709551617) 1 -- results in 0 instead of 1 
```

This overflows because on my system `Int` is 64 bit. 

code is here:

https://github.com/clash-lang/clash-compiler/blob/6e626e462be63662b66e0592a3a6911994a8cfc3/clash-prelude/src/Clash/Sized/Internal/Index.hs#L281-L289

rowanG077:
Fixes #964.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@
   * [#911](https://github.com/clash-lang/clash-compiler/pull/911): Add 'RenderVoid' option to blackboxes
 
 * Fixes issues:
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
Thank you!

PR #989, opened 2020-01-08
 Ensure the result gets a new binder if it's renamed

### LINKED ISSUE #967: Naming registers disconnects their output
```haskell
topEntity x = setName @"myRegister" $ register @System (0::Int) x
```

```vhdl
entity topentity is
  port(-- clock
       \$dIP\  : in main_types.clk_system;
       \$dIP1\ : in boolean;
       -- reset
       \$dIP2\ : in main_types.rst_system;
       eta     : in signed(63 downto 0);
       result  : out signed(63 downto 0));
end;

architecture structural of topentity is
begin
  -- register begin 
  myregister_register : process(\$dIP\,\$dIP2\)
  begin
    if \$dIP2\ =  '1'  then
      myRegister <= to_signed(0,64);
    elsif rising_edge(\$dIP\) then
      if \$dIP1\ then
        myRegister <= eta;
      end if;
    end if;
  end process;
  -- register end
end;
```
The register blackbox uses the name `myRegister`, but there's no signal declared with this name.
Nor is it connected to the `result` out port.

And `delay` has the same issue.

christiaanb:
Fixes #967

### DIFF
--- a/clash-lib/src/Clash/Backend/SystemVerilog.hs
+++ b/clash-lib/src/Clash/Backend/SystemVerilog.hs
@@ -51,7 +51,8 @@ import           Clash.Annotations.BitRepresentation.Util
   (BitOrigin(Lit, Field), bitOrigins, bitRanges)
 import           Clash.Core.Var                       (Attr'(..))
 import           Clash.Backend
-import           Clash.Backend.Verilog                (bits, bit_char, encodingNote, exprLit, include, uselibs)
+import           Clash.Backend.Verilog
+  (bits, bit_char, encodingNote, exprLit, include, noEmptyInit, uselibs)
 import           Clash.Netlist.BlackBox.Types         (HdlSyn (..))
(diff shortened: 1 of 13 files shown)


clash-lib/src/Clash/Backend/SystemVerilog.hs
```diff
@@ -751,14 +753,15 @@ decl (NetDecl' noteM _ id_ tyE iEM) =
...
-    noEmptyInit d = do
-      d1 <- d
-      if isEmpty d1
-         then emptyDoc
-         else (space <> "=" <+> d)
-
 decl _ = return Nothing
 
+noEmptyInit :: SystemVerilogM Doc -> SystemVerilogM Doc
+noEmptyInit d = do
```
leonschoorl: Maybe reuse the one from the Verilog backend

leonschoorl:
The issue doesn't exist in 1.0.1

PR #993, opened 2020-01-09
Add rewrite rules to speed up synthesis of some Vec operations

### LINKED ISSUE #945: Rewrite rules for Vec Applicative Functor
As reported in https://groups.google.com/d/msgid/clash-language/fa0cce6f-662f-4423-acf7-61d15c46cafa%40googlegroups.com?utm_medium=email&utm_source=footer

```haskell
f <$> as <*> bs
```
results in slow synthesis (both Clash and Quartus), compared to:
```haskell
zipWith f as bs
```
That's because the former translates to:
```
zipWith ($) (map f as) bs
```
where we get intermediate `Vec` of type `(b -> c)`, which Clash is forced to unroll.

Perhaps we can use GHC rewrite rules to transform:
```haskell
zipWith ($) (map f as) bs  ==> zipWith f as bs
```
The strictness of both expressions is the same, so no issues there.

christiaanb:
Fixes #945

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Signal.hs
+++ b/clash-prelude/src/Clash/Explicit/Signal.hs
@@ -145,6 +145,7 @@ can potentially introduce situations prone to meta-stability:
 
 {-# OPTIONS_GHC -fplugin=GHC.TypeLits.Normalise #-}
 {-# OPTIONS_GHC -fplugin=GHC.TypeLits.KnownNat.Solver #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 {-# OPTIONS_HADDOCK show-extensions #-}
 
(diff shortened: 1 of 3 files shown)


clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -898,3 +900,5 @@ convertReset clkA clkB (unsafeToHighPolarity -> rstA0) =
       (SAsynchronous, SSynchronous) ->
         delay clkB enableGen True $
           delay clkB enableGen True rstA1
+
+{-# RULES "sequenceAVecSignal" Clash.Sized.Vector.traverse# id = vecBundle# #-}
```
alex-mckenna: Is this actually better? `vecBundle#` is defined as `traverse# id`
Also, should this rule be defined in `Clash.Signal.Bundle`? Then it wouldn't be an orphan

christiaanb: It's better for synthesis because of this magic rule: 

https://github.com/clash-lang/clash-compiler/blob/264e5a4a7e0f818bae986ac20f8ea2b23f37a888/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs#L447

https://github.com/clash-lang/clash-compiler/blob/264e5a4a7e0f818bae986ac20f8ea2b23f37a888/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs#L979-L1004

Also, we can't define it in `Clash.Signal.Bundle`, because then  it gets applied in said module, and we end up with `vecBundle# = vecBundle#` (found this out during implementation of this PR)

PR #1006, opened 2020-01-14
Check flags to invalidate cache

### LINKED ISSUE #1002: Flags that affect the output should invalidate the cache
When I add or remove the `-fclash-no-escaped-identifiers` flag, clash doesn't regenerate the output.
It will just say: `Clash: Using cached result for: Main.topEntity`.

christiaanb:
Fixes #1002

### DIFF
--- a/clash-lib/src/Clash/Driver.hs
+++ b/clash-lib/src/Clash/Driver.hs
@@ -31,6 +31,7 @@ import           Data.IntMap                      (IntMap)
 import           Data.List                        (intercalate)
 import           Data.Maybe                       (fromMaybe)
 import           Data.Semigroup.Monad
+import qualified Data.Set                         as Set
 import qualified Data.Text
 import           Data.Text.Lazy                   (Text)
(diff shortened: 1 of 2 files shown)


leonschoorl:
The comment here is wrong: https://github.com/clash-lang/clash-compiler/blob/4b8cb2f26808025e24d46590da0d9e879bfab584/clash-lib/src/Clash/Driver/Types.hs#L69-L71

But we might still want to ignore `opt_hdlDir`, because it specifies where the manifest and the cache are stored.
If we change it clash won't find the previous cache regards of whether you hash `opt_hdlDir`.
But user might want to move the cached files around and keep reusing them, but if the hdlDir is hashed in there this will invalidate the cache.

christiaanb:
@leonschoorl Both issues should be fixed now:

1. If the user changes the flags in such a way that if affects whether we get a successful compile, the cache is ignored.

2. The hdldir setting is ignored when determining the hash of the flags, so that when you move the directory with generated HDL, you can still use it as the cache.

PR #1008, opened 2020-01-14
Preparation for Partial Evaluator

### LINKED ISSUE #950: Partial evaluation
Clash's existing compilation method of exhaustive rewriting, while significantly improved over the years, turns out to be a large performance bottleneck; especially when forced to unroll loops. Specifically, it's the successive traversal of the expression ADT to basically achieve compile-time evaluation in a very round-about way that is hurting us the most.

The idea is to repurpose Clash's existing WHNF-evaluator to an NF evaluator as a first pass of the compiler. Given that it was specifically designed for compile-time evaluation, it is much faster at this job than the exhaustive rewrite system. An early prototype of this approach did, however, show a lot of circuit duplication. So the final implementation should:
1. Make sure that the new NF-evaluator has techniques to stop duplication
2. Improve the existing Common-Subexpression-Elimination pass to recover more sharing.
3. Not only prevent duplication in the circuit, but also to prevent the NF-evaluator from re-evaluating already evaluated expression (e.g. using some sort of tag)
4. Attempt a parallel implementation, e.g. evaluate case-branches in parallel. This needs investigation with regard to sharing the evaluation of common subexpressions, and sharing/merging evaluator data-structures.

# Various thoughts
* [`Value`](https://github.com/clash-lang/clash-compiler/blob/9624b40f8c79e9b7bbec5f10973c33369c4c93f7/clash-lib/src/Clash/Core/Evaluator.hs#L100) should contain `Tick`ed and `Cast`ed values, this way the evaluator can preserve both. Ticks must be maintained because they contain register and instance names. Adding these values means we have to update the primitive evaluation rules (https://github.com/clash-lang/clash-compiler/blob/master/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs), as they need to be able to look through ticks (ala `coreView`). Additionally, some casts may have to be translated to `Prim`s, such as casts from `Integer` to `Unsigned`

alex-mckenna:
Putting up for review now because the tidying in prep of changing the partial evaluator ended up being more involved than I was expecting. In short

  * the primitive evaluator is no longer so tightly coupled to the generic evaluator. Some coupling still remains, but only in `scrutinise`

  * the evaluator now has more descriptive type names and definitions. This makes it easier to skim a definition and understand how it relates to evaluation

A new PR will be created for the actual feature, and the issue will remain open.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
@@ -105,7 +105,7 @@ runPEM :: PrimEvalMonad a -> Supply -> (a, Supply)
 runPEM (PEM m) = State.runState m
 
 reduceConstant :: PrimEvaluator
-reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
+reduceConstant isSubj tcm h k pInfo tys args = case primName pInfo of
 -----------------
 -- GHC.Prim.Char#
 -----------------
@@ -613,7 +613,7 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
 
 
   "GHC.Prim.newByteArray#"
-    | [iV,PrimVal rwNm rwTy _ _] <- args
+    | [iV,PrimVal rwTy _ _] <- args
     , [i] <- intLiterals' [iV]
     -> let (_,tyView -> TyConApp tupTcNm tyArgs) = splitFunForallTy ty
            (Just tupTc) = lookupUniqMap tupTcNm tcm
@@ -623,17 +623,16 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
            h' = Heap (IntMap.insert p lit gh,p+1) gbl ph ids is0
            mbaTy = mkFunTy intPrimTy (last tyArgs)
            newE = mkApps (Data tupDc) (map Right tyArgs ++
-                    [Left (Prim rwNm rwTy)
-                    ,Left (mkApps (Prim "GHC.Prim.MutableByteArray#"
-                                        (PrimInfo mbaTy WorkNever))
+                    [Left (Prim rwTy)
+                    ,Left (mkApps (Prim (PrimInfo "GHC.Prim.MutableByteArray#" mbaTy WorkNever))
                                   [Left (Literal . IntLiteral $ toInteger p)])
                     ])
        in Just (h',k,newE)
 
   "GHC.Prim.setByteArray#"
-    | [PrimVal _mbaNm _mbaTy _ [baV]
+    | [PrimVal _mbaTy _ [baV]
       ,offV,lenV,cV
-      ,PrimVal rwNm rwTy _ _
+      ,PrimVal rwTy _ _
       ] <- args
     , [ba,off,len,c] <- intLiterals' [baV,offV,lenV,cV]
     -> let Heap (gh,p) gbl ph ids is0 = h
@@ -648,12 +647,12 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                   ByteArray.unsafeFreezeByteArray (ByteArray.MutableByteArray mba)
            ba3 = Literal (ByteArrayLiteral (Vector.Vector voff vlen ba2))
            h'  = Heap (IntMap.insert (fromInteger ba) ba3 gh,p) gbl ph ids is0
-       in Just (h',k,Prim rwNm rwTy)
+       in Just (h',k,Prim rwTy)
 
   "GHC.Prim.writeWordArray#"
-    | [PrimVal _mbaNm _mbaTy _  [baV]
+    | [PrimVal _mbaTy _  [baV]
       ,iV,wV
-      ,PrimVal rwNm rwTy _ _
+      ,PrimVal rwTy _ _
       ] <- args
     , [ba,i] <- intLiterals' [baV,iV]
     , [w] <- wordLiterals' [wV]
@@ -668,11 +667,11 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                   ByteArray.unsafeFreezeByteArray (ByteArray.MutableByteArray mba)
            ba3 = Literal (ByteArrayLiteral (Vector.Vector off len ba2))
            h'  = Heap (IntMap.insert (fromInteger ba) ba3 gh,p) gbl ph ids is0
-       in Just (h',k,Prim rwNm rwTy)
+       in Just (h',k,Prim rwTy)
 
   "GHC.Prim.unsafeFreezeByteArray#"
-    | [PrimVal _mbaNm _mbaTy _ [baV]
-      ,PrimVal rwNm rwTy _ _
+    | [PrimVal _mbaTy _ [baV]
+      ,PrimVal rwTy _ _
       ] <- args
     , [ba] <-  intLiterals' [baV]
     -> let (_,tyView -> TyConApp tupTcNm tyArgs) = splitFunForallTy ty
@@ -681,7 +680,7 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
            Heap (gh,_) _ _ _ _ = h
            Just ba' = IntMap.lookup (fromInteger ba) gh
        in  reduce $ mkApps (Data tupDc) (map Right tyArgs ++
-                      [Left (Prim rwNm rwTy)
+                      [Left (Prim rwTy)
                       ,Left ba'])
 
   "GHC.Prim.sizeofByteArray#"
@@ -696,8 +695,8 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
        in  reduce (Literal (WordLiteral (toInteger (W# w))))
 
   "GHC.Prim.getSizeofMutBigNat#"
-    | [PrimVal _mbaNm _mbaTy _ [baV]
-      ,PrimVal rwNm rwTy _ _
+    | [PrimVal _mbaTy _ [baV]
+      ,PrimVal rwTy _ _
       ] <- args
     , [ba] <- intLiterals' [baV]
     -> let (_,tyView -> TyConApp tupTcNm tyArgs) = splitFunForallTy ty
@@ -707,13 +706,13 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
            Just (Literal (ByteArrayLiteral ba')) = IntMap.lookup (fromInteger ba) gh
            lit = Literal (IntLiteral (toInteger (Vector.length ba')))
        in  reduce $ mkApps (Data tupDc) (map Right tyArgs ++
-                      [Left (Prim rwNm rwTy)
+                      [Left (Prim rwTy)
                       ,Left lit])
 
   "GHC.Prim.resizeMutableByteArray#"
-    | [PrimVal mbaNm mbaTy _ [baV]
+    | [PrimVal mbaTy _ [baV]
       ,iV
-      ,PrimVal rwNm rwTy _ _
+      ,PrimVal rwTy _ _
       ] <- args
     , [ba,i] <- intLiterals' [baV,iV]
     -> let (_,tyView -> TyConApp tupTcNm tyArgs) = splitFunForallTy ty
@@ -731,16 +730,16 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
            ba3 = Literal (ByteArrayLiteral (Vector.Vector 0 (I# i') ba2))
            h'  = Heap (IntMap.insert p ba3 gh,p+1) gbl ph ids is0
            newE = mkApps (Data tupDc) (map Right tyArgs ++
-                    [Left (Prim rwNm rwTy)
-                    ,Left (mkApps (Prim mbaNm mbaTy)
+                    [Left (Prim rwTy)
+                    ,Left (mkApps (Prim mbaTy)
                                   [Left (Literal . IntLiteral $ toInteger p)])
                     ])
        in  Just (h',k,newE)
 
   "GHC.Prim.shrinkMutableByteArray#"
-    | [PrimVal _mbaNm _mbaTy _ [baV]
+    | [PrimVal _mbaTy _ [baV]
       ,lenV
-      ,PrimVal rwNm rwTy _ _
+      ,PrimVal rwTy _ _
       ] <- args
     , [ba,len] <- intLiterals' [baV,lenV]
     -> let Heap (gh,p) gbl ph ids is0 = h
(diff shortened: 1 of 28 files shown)


clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -819,53 +857,64 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
          [ Data lCon, _ ]
            | isTrueDC lCon -> reduce rArgWHNF
            | otherwise     -> reduce (boolToBoolLiteral tcm ty False)
+
          [ _, Data rCon ]
            | isTrueDC rCon -> reduce lArgWHNF
            | otherwise     -> reduce (boolToBoolLiteral tcm ty False)
-         _ -> Nothing
+
+         _ -> trace "reduceConstant: (&&)" Nothing
```
martijnbastiaan: What's up with the `trace`?

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -819,53 +857,64 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
          [ Data lCon, _ ]
            | isFalseDC lCon -> reduce rArgWHNF
            | otherwise      -> reduce (boolToBoolLiteral tcm ty True)
+
          [ _, Data rCon ]
            | isFalseDC rCon -> reduce lArgWHNF
            | otherwise      -> reduce (boolToBoolLiteral tcm ty True)
-         _ -> Nothing
+
+         _ -> trace "reduceConstant: (||)" Nothing
```
martijnbastiaan: What's up with the `trace`?

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2346,24 +2401,24 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
     , _ : xs : i : _ <- args
     , DC intDc [Left (Literal (IntLiteral i'))] <- i
     -> if i' < 0
-          then Nothing
+          then trace "reduceConstant: index_int" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2346,24 +2401,24 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
           else case xs of
                  DC _ vArgs  -> case runExcept (tyNatSize tcm nTy) of
-                    Right 0  -> Nothing
+                    Right 0  -> trace "reduceConstant: index_int" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2346,24 +2401,24 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
                                      ,Right aTy
                                      ,Left (Literal (NaturalLiteral (n'-1)))
                                      ,Left (Either.lefts vArgs !! 2)
                                      ,Left (mkApps (Data intDc)
                                                    [Left (Literal (IntLiteral (i'-1)))])
                                      ]
-                    _ -> Nothing
-                 _ -> Nothing
+                    _ -> trace "reduceConstant: index_int" Nothing
+                 _ -> trace "reduceConstant: index_int" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2538,7 +2593,7 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                  , Left (splitAtSelR (Either.lefts vArgs !! 2) nVecTy [rAlt])
                  ]
          -- v doesn't reduce to a data-constructor
-         _  -> Nothing
+         _  -> trace "reduceConstant: splitAt" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2613,14 +2668,14 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
-                 (mkApps (Prim nm pInfo)
+                 (mkApps (Prim pInfo)
                                       [Right (LitTy (NumTy (n'-1)))
                                       ,Right aTy
                                       ,Right mTy
                                       ,Left (Either.lefts vArgs !! 2)
                                       ,Left (valToTerm (last args))
                                       ])
-         _ -> Nothing
+         _ -> trace "reduceConstant: (++)" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2638,31 +2693,31 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
-                    ,Left $ mkApps (Prim nm pInfo)
+                    ,Left $ mkApps (Prim pInfo)
                       [ Right (LitTy (NumTy (n-1)))
                       , Right mTy
                       , Right aTy
                       , Left t
                       ]
                     ]
-        _ -> Nothing
+        _ -> trace "reduceConstant: concat" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2638,31 +2693,31 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
 
 -- Modifying vectors
   "Clash.Sized.Vector.replace_int" -- :: KnownNat n => Vec n a -> Int -> a -> Vec n a
     | nTy : aTy : _  <- tys
     , _ : xs : i : a : _ <- args
     , DC intDc [Left (Literal (IntLiteral i'))] <- i
     -> if i' < 0
-          then Nothing
+          then trace "reduceConstant: replace_int" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2638,31 +2693,31 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
           else case xs of
                  DC vecTcNm vArgs -> case runExcept (tyNatSize tcm nTy) of
-                    Right 0  -> Nothing
+                    Right 0  -> trace "reduceConstant: replace_int" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2671,8 +2726,8 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                                                       [Left (Literal (IntLiteral (i'-1)))])
                                         ,Left (valToTerm a)
                                         ])
-                    _ -> Nothing
-                 _ -> Nothing
+                    _ -> trace "reduceConstant: replace_int" Nothing
+                 _ -> trace "reduceConstant: replace_int" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2696,14 +2751,14 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
-                ,Left (mkApps (Prim nm pInfo)
+                ,Left (mkApps (Prim pInfo)
                               [Right (LitTy (NumTy (n-1)))
                               ,Right aTy
                               ,Left (Either.lefts vArgs !! 2)
                               ])
                 ,Left (mkVec nilCon consCon aTy 1 [Either.lefts vArgs !! 1])
                 ]
-         _ -> Nothing
+         _ -> trace "reduceConstant: reverse" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2773,7 +2829,8 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                                                 [Right (LitTy (NumTy (d3-1)))
                                                 ,Left  (Literal (NaturalLiteral (d3-1)))])
                                   ]
-         _  -> Nothing
+         _  -> trace "reduceConstant: rotateLeftS" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -2806,7 +2863,7 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                                                 [Right (LitTy (NumTy (d3-1)))
                                                 ,Left  (Literal (NaturalLiteral (d3-1)))])
                                   ]
-         _  -> Nothing
+         _  -> trace "reduceConstant: rotateRightS" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3036,7 +3090,7 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
                  , Left (splitAtSelR (Either.lefts vArgs !! 2) nVecTy [rAlt])
                  ]
          -- v doesn't reduce to a data-constructor
-         _  -> Nothing
+         _  -> trace "reduceConstant: fold_split" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3231,15 +3285,15 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
+                 , Left (mkApps (Prim pInfo)
                                 [ Right (LitTy (NumTy (n'-1)))
                                 , Right mTy
                                 , Left (Literal (NaturalLiteral (n'-1)))
                                 , Left (valToTerm km)
                                 , Left (Either.lefts vArgs !! 2)
                                 ])
                  ]
-         _ -> Nothing
+         _ -> trace "reduceConstant: concatBitVector#" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3281,15 +3335,15 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
...
                         [ Right (LitTy (NumTy (n'-1)))
                         , Right mTy
                         , Left (Literal (NaturalLiteral (n'-1)))
                         , Left (valToTerm km)
                         , Left (Case splitCall n1BVTy [bvAlt])
                         ])
-         _ -> Nothing
-  _ -> Nothing
+         _ -> trace "reduceConstant: unconcatBitVector#" Nothing
+  _ -> trace "reduceConstant: unconcatBitVector#" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3332,165 +3390,179 @@ reduceConstant isSubj tcm h k nm pInfo tys args = case nm of
       where
         resTy = getResultTy tcm ty tys
         selectException e | wantToHandle e = Just e
-                          | otherwise = Nothing
+                          | otherwise = trace "makeUndefinedIf" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3514,13 +3587,13 @@ sizedLitIntLit
...
-    ,PrimVal nm _ _ [_,Lit (IntegerLiteral i)]
+    ,PrimVal p _ [_,Lit (IntegerLiteral i)]
     ,valArgs -> Just [Literal (IntLiteral j)]
     ] <- args
-  , nm == szCon
+  , primName p == szCon
   = Just (nTy,kn,i,j)
   | otherwise
-  = Nothing
+  = trace "sizedlitIntLit" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3534,13 +3607,13 @@ bitVectorLitIntLit
...
-    ,PrimVal nm _ _ [_,Lit (IntegerLiteral m),Lit (IntegerLiteral i)]
+    ,PrimVal p _ [_,Lit (IntegerLiteral m),Lit (IntegerLiteral i)]
     ,valArgs -> Just [Literal (IntLiteral j)]
     ] <- args
-  , nm == "Clash.Sized.Internal.BitVector.fromInteger#"
+  , primName p == "Clash.Sized.Internal.BitVector.fromInteger#"
   = Just (nTy,kn,(m,i),j)
   | otherwise
-  = Nothing
+  = trace "bitVectorLitIntLit" Nothing
```
martijnbastiaan: ^

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -3551,22 +3624,16 @@ extractKnownNat :: TyConMap -> [Type] -> Maybe (Type, Integer)
 extractKnownNat tcm tys = case tys of
   nTy : _ | Right nInt <- runExcept (tyNatSize tcm nTy)
     -> Just (nTy, nInt)
-  _ -> Nothing
-
-extractKnownNatVal :: TyConMap -> [Type] -> Maybe Integer
-extractKnownNatVal tcm tys = fmap snd (extractKnownNat tcm tys)
+  _ -> trace "extractKnownNat" Nothing
```
martijnbastiaan: ^

clash-lib/src/Clash/Core/Evaluator.hs
```diff
@@ -46,519 +44,359 @@ import           Clash.Core.Util
...
--- | Update the Globals with the evaluated term
-gupdate :: Heap -> Stack -> Id -> Value -> State
-gupdate (Heap gh (GPureHeap gbl) h ids is) k x v =
-  (Heap gh (GPureHeap (extendVarEnv x v' gbl)) h ids is,k,v')
+  :: TyConMap
+  -> Machine
+  -> Value
+  -> Maybe Machine
+unwind tcm m v =
+  stackPop m >>= uncurry (flip go)
```
martijnbastiaan: I feel like point-free constructs like this are much harder to read than just writing it out with `do` notation.

clash-lib/src/Clash/Core/Evaluator.hs
```diff
@@ -640,53 +479,39 @@ substAlt dc tvs xs args e = substTm "Evaluator.substAlt" subst e
...
- where
-  (i,ids') = freshId ids
-  x'       = modifyVarName (\nm -> nm {nameUniq = i}) x
-
-wrapUnsigned :: Integer -> Integer -> Integer
-wrapUnsigned n i = i `mod` sz
  where
-  sz = 1 `shiftL` fromInteger n
+  -- | Create a name that's unique in the heap
+  uniqueInHeap :: PureHeap -> Supply -> Id -> (Supply, Id)
```
martijnbastiaan: `mkUniqueHeapId`?

clash-lib/src/Clash/Core/Evaluator.hs
alex-mckenna: Yeah, I figured this might be near the edge of what you would accept
But on the other hand, it saves a whole one line :wink: 

```haskell
do
  (m', sf) <- stackPop m
  go sf m'
```

clash-lib/src/Clash/Core/Evaluator.hs
alex-mckenna: I didn't actually define this function, I just moved it around. I'm actually not happy with the whole `letSubst` function (the type feels a little wacky for what it's doing now there is more abstraction)

alex-mckenna:
Also vis-à-vis documentation, the module doc comment does reference the paper that this is based on. Given that the implementation is close enough, this _should_ be sufficient. In the second PR with the actual new feature, there'll be more documentation on new functions to give an idea what their purpose is, so I would be inclined to leave documenting until then

PR #1016, opened 2020-01-17
`InlinePrimitive` supports multiple HDL backends

### LINKED ISSUE #425: `InlinePrimitive` can only hold a single `HDL` specifier
Consider:

```haskell
{-# ANN myMultiply (InlinePrimitive VHDL ..) #-}
myMultiply
  :: Signal System Int
  -> Signal System Int
  -> Signal System Int
myMultiply a b =
  a * b
{-# NOINLINE myMultiply #-}
```

A blackbox implementation for Verilog, SystemVerilog, and VHDL would be the same. Especially the first two often share implementations. It should be possible to write:

```haskell
{-# ANN myMultiply (InlinePrimitive [SystemVerilog, Verilog] ..) #-}
```

christiaanb:
Fixes #425

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
   * Add `Clash.XException.NFDataX.ensureSpine` [#803](https://github.com/clash-lang/clash-compiler/pull/803)
   * Add `Clash.Class.BitPack.bitCoerceMap` [#798](https://github.com/clash-lang/clash-compiler/pull/798)
   * Add `Clash.Magic.deDup`: instruct Clash to force sharing an operator between multiple branches of a case-expression
+  * `InlinePrimitive` can now support multiple backends simultaneously [#425](https://github.com/clash-lang/clash-compiler/issues/425)
 
 * New features (Compiler):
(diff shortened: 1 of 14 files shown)


christiaanb:
@martijnbastiaan you mean a third constructor for the `Primitive` data type?

martijnbastiaan:
Presumably. Instead of it taking a string, it should basically take a liftable version of whatever the json gets interpreted as. Probably with some good defaults (e.g. `declPrimitive` and `exprPrimitive`). It should probably also use the annotation target as function name.

martijnbastiaan:
As a second step we should probably deprecated the existing `InlinePrimitive` constructor and remove it some time in the future.

basile-henry:
I think the JSON string was more of a hack I did to get the feature working quickly than a well thought through API. Also I think at the time JSON was the only real API exposed to the user to create primitives. Now we have the Haskell API which is more backend agnostic! :smile:

christiaanb:
There is a circular dependency problem though, where the data type for primitives lives in `clash-lib`, but the annotations live in `clash-prelude`. We'd have to move the data types for primitives over to `clash-prelude`, if we want to share data types.

Regardless, I think this PR is an improvement over the current API; and given that it's such a expert feature I don't mind the breakage.

PR #1018, opened 2020-01-17
toInteger function warn when dropping MSBs

### LINKED ISSUE #836: Generate warning when `toInteger` blackbox drops MSBs
Clash "erroneously" translates Haskell's `Integer` to 64-bit signed number; even when the context demands more. Properly translating `Integer` has proven non-trivial, so there's no easy path forward; outright banning `Integer` in Clash designs in intractable because of how ubiquitous it is in Haskell.

This issue mostly shows up when users are converting from one type to the other "through" `Integer`, e.g. using `fromIntegral`. This works when the source type is less than 64 bit, but drops MSBs when the source type is more than 64 bit. This conversion to `Integer` happens in the `toInteger` primitives. 

So Clash should issue a warning when these primitives are used in a "more-than-64-bit" context. The warning will show up late in the synthesis process, which is regrettable; but it's better than no warning at all!! The warning should suggest alternatives, such as using `bitCoerce` to convert between types instead of `fromIntegral`.

christiaanb:
Fixes #836


martijnbastiaan:
Nice! Maybe we should make a Prelude for writing Blackboxes..

PR #1021, opened 2020-01-17
Add `Int` before `Double` in interactive defaulting rules

### LINKED ISSUE #280: Surprising type defaulting
Taking the 3 least significant bits of 5 in clashi:
```
Clash.Prelude> slice d2 d0 5
000
```

Why not `101`?

Because of [type defaulting](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ghci.html#type-defaulting-in-ghci), clashi picks the type of 5 to be Double:
```
Clash.Prelude> slice d2 d0 (5::_)
<interactive>:11:17: error:
    • Found type wildcard ‘_’ standing for ‘Double’
```

Note that this only happens in clashi.
Doing the same in a .hs file, results in an ambiguous type variable error.

But I think clashi should never default to Double.

christiaanb:
Fixes #280

### DIFF
--- a/clash-ghc/src-bin-841/Clash/GHCi/UI.hs
+++ b/clash-ghc/src-bin-841/Clash/GHCi/UI.hs
@@ -625,29 +625,35 @@ runGHCi paths maybe_exprs = do
   case maybe_exprs of
         Nothing ->
           do
+            -- Set different defaulting rules (See #280)
+            runGHCiExpressions ["default ((),[],Integer,Int,Double,String)"]
             -- enter the interactive loop
             runGHCiInput $ runCommands $ nextInputLine show_prompt is_tty
         Just exprs -> do
             -- just evaluate the expression we were given
-            enqueueCommands exprs
-            let hdle e = do st <- getGHCiState
-                            -- flush the interpreter's stdout/stderr on exit (#3890)
-                            flushInterpBuffers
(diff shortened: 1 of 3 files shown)


clash-ghc/src-bin-841/Clash/GHCi/UI.hs
```diff
@@ -625,30 +625,36 @@ runGHCi paths maybe_exprs = do
   case maybe_exprs of
         Nothing ->
           do
+            -- Set different defaulting rules (See #280)
+            runGHCiExpressions ["default ((),[],Integer,Int,Double,String)"]
```
martijnbastiaan: Any reason to prefer `Integer` over `Int`? Should we ever infer `Integer`s anyway?

christiaanb: Yes, otherwise you get weird overflow behavior when you type big literals into the prompt:
```
GHCi, version 8.8.1: https://www.haskell.org/ghc/  :? for help
Loaded package environment from /home/christiaan/devel/clash-compiler/.ghc.environment.x86_64-linux-8.8.1
Prelude> default (Int,Integer)
Prelude> 4444444444444444444444444444444444444444444444444444444444444444444444444444
-4099276460824344804
```

PR #1026, opened 2020-01-22
Don't produce empty letrecs in inlineCleanup

### LINKED ISSUE #1025: `inlineCleanup` can create empty letrecs
Consider the following pseudo-code (due to GHC optimizations I can't write a small example that triggers this behavior):

```haskell
f a = prim (let b = a in (b && False))
```

Thanks to https://github.com/clash-lang/clash-compiler/pull/984 this will be turned into:

```haskell
f a = prim (let b = a in False)
```

`inlineCleanup` will then remove the binder `b`, because it is unused:

```haskell
f a = prim (let {} in False)
```

Leaving us with an `EmptyRec`. If this is forced in a blackbox, Clash will crash with `Forced to evaluate unexpected function argument`. For this particular case, we can patch `inlineCleanup` to omit a let if it has no bindings:

```diff
diff --git a/clash-lib/src/Clash/Normalize/Transformations.hs b/clash-lib/src/Clash/Normalize/Transformations.hs
index 2d24ab3b..5b6a2e3a 100644
--- a/clash-lib/src/Clash/Normalize/Transformations.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations.hs
@@ -2297,8 +2297,10 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
       bodyFVs       = Lens.foldMapOf freeLocalIds unitVarSet body
       (il,keep)     = List.partition (isInteresting allOccs prims bodyFVs) binds
       keep'         = inlineBndrs is1 keep il
-  if null il then return  (Letrec binds body)
-             else changed (Letrec keep' body)
+
+  if | null il -> return  (Letrec binds body)
+     | null keep' -> changed body
+     | otherwise -> changed (Letrec keep' body)
```

In general, I'm not sure what to do. Should we assign the letrec to a var before passing it to a blackbox?

martijnbastiaan:
Fixes #1025.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,7 +70,6 @@
   * [#990](https://github.com/clash-lang/clash-compiler/issues/990): Internal shadowing bug results in incorrect HDL
   * [#945](https://github.com/clash-lang/clash-compiler/issues/945): Rewrite rules for Vec Applicative Functor
   * [#919](https://github.com/clash-lang/clash-compiler/issues/919): Clash generating invalid Verilog after Vec operations #919
-Closed
   * [#996](https://github.com/clash-lang/clash-compiler/issues/996): Ambiguous clock when using `ClearOnReset` and `resetGen` together
   * [#701](https://github.com/clash-lang/clash-compiler/issues/701): Unexpected behaviour with the `Synthesize` annotation
   * [#694](https://github.com/clash-lang/clash-compiler/issues/694): Custom bit representation error only with VHDL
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -50,6 +50,7 @@
   * [#781](https://github.com/clash-lang/clash-compiler/pull/781): Use naming contexts in register names
 
 * Fixes issues:
+  * [#1025](https://github.com/clash-lang/clash-compiler/issues/974): `inlineCleanup` will not produce empty letrecs anymore
```
martijnbastiaan: Wrong link.

PR #1031, opened 2020-01-23
 Consider Enable lines workfreeish. Fix #1030 

### LINKED ISSUE #1030: Enable lines not considered workfreeish
Consider the pseudo-core:

```haskell
let 
  x = ( (x :: Clock dom, $fKnownDomain)
      , (y :: Enable dom, $fKnownDomain) )
in
  f x
```

where `$fKnownDomain` is a constant representing a dictionary belonging to a `KnownDomain` constraint. We need to constant specialize `f` on `$fKnownDomain`, but `constantSpec` will only see a variable - and will therefore not fire. Because `x` does not contain any work, we'd expect `bindConstantVar` to bind the definition. It doesn't do so however, because `y` isn't considered workfreeish.

martijnbastiaan:


clash-lib/src/Clash/Core/Util.hs
```diff
@@ -789,11 +789,14 @@ isClockOrReset
   :: TyConMap
   -> Type
   -> Bool
-isClockOrReset m (coreView1 m -> Just ty)    = isClockOrReset m ty
-isClockOrReset _ (tyView -> TyConApp tcNm _) = case nameOcc tcNm of
+isClockOrReset m ty0@(tyView -> TyConApp tcNm _) = case nameOcc tcNm of
```
alex-mckenna: `isClockishOrResetishOrEnableish` :wink:

But I do think this should be renamed to something more accurate, e.g. `isSpecialSignal`

martijnbastiaan: It does hurt a little, but I think `isClockOrResetOrEnable` is the best name here.

martijnbastiaan: @alex-mckenna You're in luck! I had to split out the function into `isClockOrReset` and `isEnable` anyway ;).

PR #1048, opened 2020-01-29
Don't confuse term/type namespaces in 'lookupIdSubst'. Fix #1046

### LINKED ISSUE #1046: `substTm` mixes up term and type vars
Consider two variables. One is a term variable:

```haskell
termVar :: Var Term
termVar = Id {
    varName = fakeName {nameUniq=unique, nameOcc="term"}
  , varUniq = unique
  , varType = ConstTy (TyCon fakeName)
  , idScope = LocalId
  }
```

and the other a type variable

```haskell
typeVar :: Var Type
typeVar = TyVar {
    varName = fakeName {nameUniq=unique, nameOcc="type"}
  , varUniq = unique
  , varType = ConstTy (TyCon fakeName)
  }
```

We'd expect that deshadowing `Var termVar` with an `inScopeSet` of just `typeVar` should be a no-op, as their uniques live in different name spaces. (In fact, it's not even possible to have both `termVar` and `typeVar` in an InScopeSet, as it only stores uniques!). `deShadowTerm` currently behaves differently though:

```haskell
>>> deShadowTerm (extendInScopeSet emptyInScopeSet typeVar) (Var termVar)
Var (TyVar {varName = Name {nameSort = User, nameOcc = "type", nameUniq = 20, nameLoc = UnhelpfulSpan "<no location info>"}, varUniq = 20, varType = ConstTy (TyCon (Name {nameSort = User, nameOcc = "fake", nameUniq = 0, nameLoc = UnhelpfulSpan "<no location info>"}))})
```

instead of:

```haskell
>>> deShadowTerm (extendInScopeSet emptyInScopeSet typeVar) (Var termVar)
Var (Id {varName = Name {nameSort = User, nameOcc = "term", nameUniq = 20, nameLoc = UnhelpfulSpan "<no location info>"}, varUniq = 20, varType = ConstTy (TyCon (Name {nameSort = User, nameOcc = "fake", nameUniq = 0, nameLoc = UnhelpfulSpan "<no location info>"})), idScope = LocalId})
```

The culprit seems to be anti space-leak measures in `lookupIdSubst`:

https://github.com/clash-lang/clash-compiler/blob/a009fe45a981031d91f58cecaf4af28e7601d416/clash-lib/src/Clash/Core/Subst.hs#L512

It should refuse to substitute `TyVar`s with `Id`s.

martijnbastiaan:


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -503,7 +503,14 @@ lookupIdSubst doc (Subst inScope tmS _ genv) v
...
   -- Vital! See 'IdSubstEnv' Note [Extending the Subst]
-  | Just v' <- lookupInScope inScope v = Var (coerce v')
+  --
+  -- TODO: We match on Id here to workaround an issue where type variables
+  -- TODO: "shadow" term variables. Omitting the check would make 'lookupIdSubst'
+  -- TODO: potentially replace an "Id" with a TyVar. For more information:
+  -- TODO:
+  -- TODO:   https://github.com/clash-lang/clash-compiler/issues/1046
+  -- TODO:
+  | Just v'@(Id {}) <- lookupInScope inScope v = Var (coerce v')
```
christiaanb: We shouldn't need the `coerce` anymore, we already know it's a `Var Term` since we pattern mached on `Id {}`

martijnbastiaan: It's not a GADT, so I don't think we do.

christiaanb: Ah, right

martijnbastiaan: Is that an 'approved' then? :)

christiaanb: Yes

leonschoorl: Why isn't `Var` a GADT?

christiaanb: So we can use "cheap" run-time `coerce` (it's used in quite a few places in the codebase), instead of "expensive" run-time matching on constructors.

martijnbastiaan: The proper solution here is to split up `InScopeSet` into two sets internally, one for term vars and the other for type vars. We wouldn't need coerce or run-time matching on constructors that way.

PR #1053, opened 2020-01-31
Add 'Clash.XException.hwSeqX'

martijnbastiaan:
@JvWesterveld :tada:


christiaanb:
Would be nice to have some example in the documentation as when to use it.

martijnbastiaan:
@christiaanb I've added an example. https://github.com/clash-lang/clash-compiler/pull/864 uses this too, so I'll add a reference to that work once that's merged.

PR #1052, opened 2020-01-31
Filter KnownDomain in 'mkUniqueArguments'. Fix #1033

martijnbastiaan:

PR #1057, opened 2020-02-03
 Update 'TopEntity' annotations in 'splitArguments'. Fix #1033 

### LINKED ISSUE #1033: Synthesize annotations are misinterpreted for functions with hidden constructs
Consider:

```haskell
module Test2 where

import Clash.Prelude

{-# ANN topEntity
  (Synthesize
    { t_name   = "top"
    , t_inputs =
        [ PortName "theHiddenClock"
        , PortName "theHiddenReset"
        , PortName "theInput"
        ]
    , t_output = PortName "theOutput"
    }
  )#-}
topEntity
  :: ( HiddenClock System
     , HiddenReset System )
  => Signal System Int
  -> Signal System Int
topEntity = withEnable enableGen register 0
```

and

```vhdl
entity top is
  port(-- clock
       theHiddenClock : in top_types.clk_system;
       -- reset
       theInput       : in top_types.rst_system;
       eta            : in signed(63 downto 0);
       theOutput      : out signed(63 downto 0));
end;
```

`theHiddenReset` is skipped because due to a combination of https://github.com/clash-lang/clash-compiler/issues/1028 and https://github.com/clash-lang/clash-compiler/pull/862, as Clash will internally work with:

```haskell
topEntity
  :: Clock dom  -- ^ theHiddenClock
  -> %KnownDomain_dom -- ^ theHiddenReset
  -> Reset dom -- ^ theInput
  -> %KnownDomain_dom -- ^ auto generated name
  -> ...
```

Although this is related to https://github.com/clash-lang/clash-compiler/issues/1028, I'm marking it as a separate bug, because we can probably solve it without solving #1028 .

martijnbastiaan:


christiaanb:
I can't agree with this, as this puts the onus on the user to know that downstream synthesis tools dislike clocks in product types. Additionally, this would undo the changes I made to register where I removed the a `after 1 ps`. I think what should happen is that we call (the to be written) `splitArgumentsTopEntity` on `TopEntity` annotations!

martijnbastiaan:
Depending on what you mean with `splitArgumentsTopEntity`, I don't think there's any sensible implementation. Consider the example in the `Note`:

```
   [..]
   , t_inputs = [ PortName "resets", PortName "in"]
   [..]
  f
    :: (Reset dom, Reset dom)
    -> Signal dom Int
    -> Signal dom Int
```

The user clearly intends to give the tuple of resets a single name "resets". Like you said, it's not the user's job to know about this stuff, so it would be at least unexpected if not wrong to split it up from a user's POV.

martijnbastiaan:
In conclusion then, I think the only real "solution" here is to:

1. Be more precise when checking. Topentities without port annotations are fine. Topentities _with_ port annotations are fine, but only if they don't touch ports that are ought to be split up.

2. Make an exception for `KnownDomain`, just always filter that one, similar to what I did in [my last PR](https://github.com/clash-lang/clash-compiler/pull/1052/files).

christiaanb:
Yes, I agree with 1. When a user wants to put a reset, clock, or enable in a product type that cannot be split (by using `PortName`) then we raise an error. However! When the user correctly uses `PortProduct` than the onus is on us to split that into multiple `PortNames` under the hood.

martijnbastiaan:
I'm not sure I'm following. You mean, if the example would have used:

```
t_inputs = [ PortProduct "resets" [], PortName "in"]
```

it should be accepted, but turned into something like:


```
t_inputs = [ PortName "resets0", PortName "resets1", PortName "in"]
```

?

christiaanb:
That's right. And all the variants, so:

```
t_inputs = [ PortProduct "reset" [PortName "a", PortName "b], PortName "in"]
```
would be transformed under the hood to:

```
t_inputs = [PortName "reset_a", PortName "reset_b", PortName "in"]
```

martijnbastiaan:
@christiaanb I'll be adding tests shortly (possibly tomorrow), but the patch is done as far as functionality goes. So if you could review it, that would be great.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -2472,6 +2482,61 @@ separateArguments ctx@(TransformContext is0 _) e@(Lam b eb0) = do
...
+    case portName of
+      PortProduct nm portNames0 ->
+        let
+          (pre, tail -> post) = splitAt (length cctx) t_inputs
+          newPortNames = [PortName (nm ++ "_" ++ show i_) | i_ <- [(0::Int)..]]
+          portNames1 = take n (portNames0 ++ newPortNames)
+          t1 = t0{topAnnotation=Just s{t_inputs=pre ++ portNames1 ++ post}}
+        in do
+          (cf, _) <- Lens.use curFun
+          topEntities Lens.%= extendVarEnv cf t1
```
martijnbastiaan: This fails to update `portNames0`. `nm` + `_` should be prepended to all `PortName`s / `PortProduct`s in `portNames0`.

martijnbastiaan: Fixed.

clash-lib/src/Clash/Driver.hs
```diff
@@ -325,9 +328,10 @@ generateHDL reprs bindingsMap hdlState primMap tcm tupTcm typeTrans eval
       -- See [Note] Create HDL dir before netlist generation
       let dir = hdlDir </> maybe "" t_name annM </> Data.Text.unpack modName'
       prepareDir (opt_cleanhdl opts) (extension hdlState2) dir
+      let topEnts1 = map ((\(Just t) -> t) . flip lookupVarEnv transformedTopEntities . topId) topEnts0
```
alex-mckenna: Don't you mean, `fmap fromJust` :eyes:

clash-lib/src/Clash/Driver.hs
```diff
@@ -682,55 +686,49 @@ normalizeEntity
...
     in  ((varUniq topE,t)
-         ,map ((\(v,_,_) -> (varUniq topE, varUniq v))) (tbRefs ++ topRefs))
+         ,map ((\top -> (varUniq topE, varUniq (topId top)))) (tbRefs ++ topRefs))
 
   goRefs top i =
     let cg = callGraph bindingsMap i
-    in  filter (\(v,_,_) -> v /= top && v /= i && v `elemVarEnv` cg) topEntities
+    in
+      filter
+        (\(TopEntityT v _ _) -> v /= top && v /= i && v `elemVarEnv` cg)
```
alex-mckenna: Even though it's more duplication, I also think using the `topId` field name is clearer here. Every time I see something like this I always get sidelined looking up the data type to see what's actually being used.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -2472,6 +2482,65 @@ separateArguments ctx@(TransformContext is0 _) e@(Lam b eb0) = do
...
+    case portName of
+      PortProduct nm portNames0 ->
+        let
+          (pre, tail -> post) = splitAt (length cctx) t_inputs
+          newPortNames = map (PortName . show) [(0::Int)..]
+          portNames1 = take n (map (prependName nm) (portNames0 ++ newPortNames))
+          t1 = t0{topAnnotation=Just s{t_inputs=pre ++ portNames1 ++ post}}
+        in do
+          (cf, _) <- Lens.use curFun
+          topEntities Lens.%= extendVarEnv cf t1
```
alex-mckenna: This is the real reason I'm here :wink:

I know this kind of thing appears in other places, but I really think qualified operators make things less readable. We also (at least with lenses) are relatively inconsistent with whether we qualify them or not.

(Also by no means are these comments a review in any way, it's just me seeing some of my Haskell bugbears)

clash-lib/src/Clash/Driver.hs
martijnbastiaan: I didn't want to create a situation where we'd see `*** Exception: Maybe.fromJust: Nothing` if this logic ever breaks. At least an incomplete pattern match gives you a line number. (Although HasCallStack might save the day nowadays?)

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: Is this about `Lens.use`, `Lens.%=`, or both?

clash-lib/src/Clash/Driver.hs
alex-mckenna: I would hope it would, but who knows :man_shrugging: 
I tend to prefer using `fromJust` directly because it's (1) a little less noisy and (2) makes `hlint` happier. Pros and cons.

clash-lib/src/Clash/Normalize/Transformations.hs
alex-mckenna: Just `Lens.%=`. I think (especially for `lens` having such generic names) it's nice to qualify functions, but have operators in an explicit unqualified import list. Mostly because qualified operators are weird to read, and sometimes confusing (if the operator starts with a dot for instance)

alex-mckenna: Plus we usually only use about 3 lens operators at most in a particular file, so mentioning them explicitly in an import list doesn't make things too horrible

clash-lib/src/Clash/Driver.hs
```diff
@@ -682,55 +686,49 @@ normalizeEntity
...
   -- ^ TopEntities
   -> ClashOpts
   -- ^ Debug information level for the normalization process
   -> Supply.Supply
   -- ^ Unique supply
   -> Id
   -- ^ root of the hierarchy
-  -> BindingMap
-normalizeEntity reprs bindingsMap primMap tcm tupTcm typeTrans eval topEntities
+  -> (BindingMap, VarEnv TopEntityT)
```
martijnbastiaan: I'm a bit unhappy with out this change turned out. Maybe `TopEntity` should just be part of a `Binding`.

clash-lib/src/Clash/Driver.hs
christiaanb: use `maybe (error “internal error”)`, you’ll get the callstack then. Or use `$(curLoc)` just to be sure.

clash-lib/src/Clash/Driver.hs
christiaanb: But there will only be a single TopEntity in that BindingMap though, since the `normalized` BindingMap only contains the “current” TopEntity, but none of the others (since we don’t renormalize other TopEntities).

martijnbastiaan: `sortTop` uses `bindingsMap` to topsort topentities and uses `callGraph :: BIndingMap -> Id -> ..` to figure out the callgraph. So topentities must be in there, no?

martijnbastiaan: *all topentities

christiaanb: but `runNormalization` returns the `_normalized ` BindingMap found in the `NormalizeState`, not the `_bindings` found in `RewriteState`. So while the latter does contain all topentities, the former does not.

martijnbastiaan: Gotcha. It should probably return the whole thing instead. Netlist doesn't seem to rely on _only_ normalized binders in there. As a bonus, it would allow normalization processes to reuse work of previous normalizations. If we insist on keeping this check (so we don't accidentally try to generate netlist for a non-normalized binder) https://github.com/clash-lang/clash-compiler/blob/44af7dfbef8a04224a31f076c837d0df400565d4/clash-lib/src/Clash/Netlist.hs#L203-L205, we could have wrap the term of a `Binding` in:

```haskell
data NTerm = Normalized Term | NotNormalized Term
```

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -982,11 +983,24 @@ shouldSplit0 tcm (TyConApp tcNm tyArgs)
   , [dc] <- tyConDataCons tc
   , let dcArgs  = substArgTys dc tyArgs
   , let dcArgVs = map (tyView . coreView tcm) dcArgs
-  = if any (\ty -> isJust (shouldSplit0 tcm ty) || splitTy ty) dcArgVs then
+  = if any shouldSplitTy dcArgVs && not (isHidden tcNm tyArgs) then
```
christiaanb: Could you confirm that this does the right thing for:

```haskell
type MyMischievousHidden dom = 
  (  ( Hidden (AppendSymbol dom "_clk" (Clock dom))
     , Hidden (AppendSymbol dom "_rst") (Reset dom) )
  , KnowDomain dom
  )
```

I think it does do the right thing, but I'd like some confirmation. 

However, it wont work for a use who does this I think:

```haskell
type MyMischievousHiddenClk dom
  = ( KnownDomain dom
    , Hidden (AppendSymbol dom "_clk") (Clock dom)
    )

martijnbastiaan: Correct, it won't handle the latter case. (And handling the first case is a happy accident.) It was specifically made to recognize `HiddenClock` & co. I'm happy to make it more general, but I'm wondering whether it's worth the extra engineering time.

christiaanb: I can live with a `-- TODO` comment, and a new issue on the issue tracker.

martijnbastiaan: Ah, it doesn't handle the first case either. It's specifically `(ClockOrResetOrEnable, KnownDomain)`.

martijnbastiaan: Done. (Will add the issue after merging.)

PR #1062, opened 2020-02-07
Sanity check types in 'piResultTys'/'piResultTyMaybe'

martijnbastiaan:


leonschoorl:
Maybe add a `HasCallStack` to help with localizing the errors when they happen.
Something like:
```haskell
piResultTy, piResultTyMaybe, piResultTys
  ::
#if defined(DEBUG)
    HasCallStack =>
#endif
     TyConMap
  -> Type
  -> ...
```

Otherwise LGTM

PR #1064, opened 2020-02-07
Keep all casts, and cast `Signal a ~ a` where appropriate

christiaanb:


leonschoorl:
A more worked out version of this is in the [castSignal-rebased](https://github.com/clash-lang/clash-compiler/tree/castSignal-rebased) branch

PR #1067, opened 2020-02-10
Deshadow in 'caseCase'

martijnbastiaan:


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -350,15 +350,17 @@ elemExistentials _ e = return e
...
     ty1Rep <- representableType <$> Lens.view typeTranslator
                                 <*> Lens.view customReprs
                                 <*> pure False
                                 <*> Lens.view tcCache
                                 <*> pure alts1Ty
     if not ty1Rep
-      then let newAlts = map (second (\altE -> Case altE alts2Ty alts2)) alts1
+      then let newAlts = map
+                           (second (\altE -> Case altE alts2Ty alts2))
+                           (map (deShadowAlt is0) alts1)
```
christiaanb: Add a Note comment on why we need to deshadow, like we do in AppProp

martijnbastiaan: Done

PR #1069, opened 2020-02-10
Add 'Clash.Util.Interpolation'

martijnbastiaan:
It's currently hard to write pretty/consistent error messages in the Clash code base. A lot of errors and warnings use ad-hoc string `concat`s. For example:

```haskell
        trace (concat [ $(curLoc) ++ "InlineNonRep: " ++ showPpr (varName f)
                      ," already inlined " ++ show limit ++ " times in:"
                      , showPpr (varName cf)
                      , "\nType of the subject is: " ++ showPpr scrutTy
                      , "\nFunction " ++ showPpr (varName cf)
                      , " will not reach a normal form, and compilation"
                      , " might fail."
                      , "\nRun with '-fclash-inline-limit=N' to increase"
                      , " the inlining limit to N."
                      ])
```

Recently, we've been trying to do a bit better with the quasiquoter from [neat-interpolation](https://hackage.haskell.org/package/neat-interpolation). This looks a bit better in the source code:

```haskell
          PortName nm ->
            throw (flip (ClashException sp) Nothing $ unindent $ [i|
              Couldn't separate clock, reset, or enable from a product type due
              to a malformed Synthesize annotation. All clocks, resets, and
              enables should be given a unique port name.

              Type to split:

              #{showPpr' (PrettyOptions False True False) (varType b)}

              Given port annotation: #{p}. You might want to use the
              following instead: PortProduct #{show nm} []. This allows Clash to
              autogenerate names based the name #{show nm}.
            |])
```

but it still has two problems:

1. The length of each line in the error message depends on the position of the original source code. The example just presented cuts of at 66, while others might cut off even earlier. 

2. If a multiline string is inserted as an expression, such as `#{showPpr' (PrettyOptions False True False) (varType b)}`, the error ends up being misaligned. 

The new quasiquoter solves (1) by assembling lines and splitting them at 90 chars (on word boundaries) later. It solves (2) by separately handling `#{..}` if it's on a single line, such as in the example.

It's not the prettiest code, but I'm not sure if text processing can ever _be_ beautiful :-).

---------------------------

Before

```
tests/shouldfail/TopEntity/T1033.hs:18:1: error:
                Couldn't separate clock, reset, or enable from a product type due
            to a malformed Synthesize annotation. All clocks, resets, and
            enables should be given a unique port name. Type to split:

            (,,)
(Clock "System")
(Reset "System")
(Enable "System")

            Given port annotation: PortName "wrong". You might want to use the
            following instead: PortProduct "wrong" []. This allows Clash to
            autogenerate names based the name "wrong".

    NB: The source location of the error is not exact, only indicative, as it is acquired after optimisations.
    The actual location of the error can be in a function that is inlined.
    To prevent inlining of those functions, annotate them with a NOINLINE pragma.
   |
18 | topEntity (clk, rst, en) i =
   | ^^^^^^^^^

```

After

```
tests/shouldfail/TopEntity/T1033.hs:18:1: error:
    
    Couldn't separate clock, reset, or enable from a product type due to a malformed 
    Synthesize annotation. All clocks, resets, and enables should be given a unique port 
    name. Type to be split:
    
      (,,)
        (Clock "System")
        (Reset "System")
        (Enable "System")
    
    Given port annotation: PortName "wrong". You might want to use the following instead: 
    PortProduct "wrong" []. This allows Clash to autogenerate names based on the name 
    "wrong".
    
   |
18 | topEntity (clk, rst, en) i =
   | ^^^^^^^^^
```

---------------------------

TODO:
- [x] Add before/after to this PR
- [x] Figure out whether we want line cut off hardcoded at 90 chars. **Fine for now. We'll figure it out later.**
- [x] Write tests
- [x] Figure out how to correctly credit neat-interpolation for their base work **Done**

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Util.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Util.hs
@@ -1,5 +1,8 @@
+{-# LANGUAGE QuasiQuotes #-}
+
 module Clash.GHC.Util where
 
+import Outputable         (SDoc)
 import ErrUtils           (mkPlainErrMsg)
 import GHC                (GhcMonad(..), printException)
 import GhcPlugins         (DynFlags, SourceError, ($$), blankLine, empty, isGoodSrcSpan, liftIO, noSrcSpan, text, throwOneError)
@@ -9,7 +12,15 @@ import GHC.Exception      (SomeException)
 import System.Exit        (ExitCode(ExitFailure), exitWith)
 
 import Clash.Util         (ClashException(..))
+import Clash.Util.Interpolate (i)
 import Clash.Driver.Types (ClashOpts(..))
+
+-- | Like 'lines', but returning a horizontally spaced SDoc instead of a list:
+--
+-- >>> textLines "a\nb"
+-- a $$ b
+textLines :: String -> SDoc
+textLines s = foldl1 ($$) (map text (lines s))
 
 handleClashException
   :: GhcMonad m
@@ -21,7 +32,7 @@ handleClashException df opts e = case fromException e of
   Just (ClashException sp s eM) -> do
     let srcInfo' | isGoodSrcSpan sp = srcInfo
                  | otherwise = empty
-    throwOneError (mkPlainErrMsg df sp (text s $$ srcInfo' $$ showExtra (opt_errorExtra opts) eM))
+    throwOneError (mkPlainErrMsg df sp (blankLine $$ textLines s $$ blankLine $$ srcInfo' $$ showExtra (opt_errorExtra opts) eM))
   _ -> case fromException e of
     Just (ErrorCallWithLocation _ _) ->
       throwOneError (mkPlainErrMsg df noSrcSpan (text "Clash error call:" $$ text (show e)))
@@ -31,9 +42,12 @@ handleClashException df opts e = case fromException e of
         liftIO $ exitWith (ExitFailure 1)
       _ -> throwOneError (mkPlainErrMsg df noSrcSpan (text "Other error:" $$ text (displayException e)))
   where
-    srcInfo = text "NB: The source location of the error is not exact, only indicative, as it is acquired after optimisations." $$
-              text "The actual location of the error can be in a function that is inlined." $$
-              text "To prevent inlining of those functions, annotate them with a NOINLINE pragma."
+    srcInfo = textLines [i|
+      The source location of the error is not exact, only indicative, as it
+      is acquired after optimizations. The actual location of the error can be
+      in a function that is inlined. To prevent inlining of those functions,
+      annotate them with a NOINLINE pragma.
+    |]
 
     showExtra False (Just _)   =
       blankLine $$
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Util/Interpolate.hs
```diff
@@ -0,0 +1,320 @@
...
+i :: QuasiQuoter
+i = QuasiQuoter {
+    quoteExp = (varE 'format `appE`) . toExp . parseNodes . decodeNewlines
+  , quotePat = err "pattern"
+  , quoteType = err "type"
+  , quoteDec = err "declaration"
+  }
+  where
+    err name =
+      error ("Data.String.Interpolate.i: This QuasiQuoter can not be used as a "
```
leonschoorl: Change string

martijnbastiaan: Done

alex-mckenna:
> The length of each line in the error message depends on the position of the original source code. The example just presented cuts of at 66, while others might cut off even earlier.

If the quasiquote produces `Text`, this can be turned into a `Doc ann` from `prettyprinter` and rendered by that. We already use `prettyprinter` in some places (`ClashPretty` is _essentially_ just `Pretty` from this library) so it shouldn't be a big addition, e.g.

```haskell
throwClashError sp = throw
  . flip (ClashException sp)
  . renderStrict
  . layoutPretty defaultLayoutOptions
  . pretty
  . unindent
```

and then at error sites

```haskell
throwClashError sp [i| Oh no, something went wrong! |]
```

_This code might not work. Use at your own risk. Ask your doctor is Haskell is right for you._

clash-ghc/src-ghc/Clash/GHC/Util.hs
```diff
@@ -9,8 +12,12 @@ import GHC.Exception      (SomeException)
 import System.Exit        (ExitCode(ExitFailure), exitWith)
 
 import Clash.Util         (ClashException(..))
+import Clash.Util.Interpolate (i)
 import Clash.Driver.Types (ClashOpts(..))
 
+text' :: String -> SDoc
```
christiaanb:
```suggestion
-- | Like 'lines', but returning a horizontally spaced SDoc instead of a list:
--
-- >>> textLines "a\nb"
-- a $$ b
textLines :: String -> SDoc
```

clash-lib/src/Clash/Util/Interpolate.hs
```diff
@@ -0,0 +1,363 @@
+{-|
+Copyright  :  (C) 2019, QBayLogic B.V.
+                  2013, Nikita Volkov
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+-}
+
+{-
+This is an adaptation of
```
christiaanb: How much effort is to to highlight what's new?

martijnbastiaan: No effort at all! Original code is almost entirely left untouched, I just added some post-processing.

martijnbastiaan: Done

clash-ghc/src-ghc/Clash/GHC/Util.hs
martijnbastiaan: Done

PR #1071, opened 2020-02-10
Deshadow in `caseLet` and `nonRepANF`

christiaanb:
They had the potential to accidentally capture free variables

### DIFF
--- a/clash-lib/src/Clash/Core/Subst.hs
+++ b/clash-lib/src/Clash/Core/Subst.hs
@@ -44,6 +44,7 @@ module Clash.Core.Subst
   , deShadowTerm
   , deShadowAlt
   , freshenTm
+  , deshadowLetExpr
     -- * Alpha equivalence
   , aeqType
   , aeqTerm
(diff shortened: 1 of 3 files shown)


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -622,6 +623,18 @@ deShadowAlt ::
   (Pat, Term)
 deShadowAlt is = substAlt "deShadowAlt" (mkSubst is)
 
+-- | Ensure that non of the binders of a let-expression shadow each other, nor
```
martijnbastiaan: > Ensure that non of the binders of a let-expression shadow each other

Probably true, but `deshadowLetExpr` shouldn't document this behavior as let-bindings with duplicate binders aren't well-defined anyway.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -622,6 +623,18 @@ deShadowAlt ::
...
+-- conflict with the in-scope set
+deshadowLetExpr ::
+  HasCallStack =>
+  InScopeSet ->
+  [LetBinding] ->
+  Term ->
+  ([LetBinding],Term)
```
martijnbastiaan: While nice, I don't think we should use this style just yet, as it makes it impossible to document functions using it.

PR #1072, opened 2020-02-11
Use 'Type' instead of 'Term' in 'splitTopAnn'

martijnbastiaan:
'splitTopAnn' is used on pre-normalized topentities. This means
'etaExpansionTL' has not run yet. The logic in 'separateLambda' relies
on this however so 'splitTopAnn' breaks. Instead, 'splitTopAnn' should
use type information to calculate new port names.


martijnbastiaan:
I'm going to merge this (despite the issues we discovered) to get it into a client's codebase. I'll open an issue report.

PR #1073, opened 2020-02-11
Refactor `caseCon` and add more documentation

christiaanb:
While trying to determine whether `caseCon` has any lingering variable-capture issues I found the code somewhat opaque, so I added some comments, and also shared some of the WHNF logic with `reduceConst`

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations.hs
@@ -61,7 +61,6 @@ module Clash.Normalize.Transformations
   )
 where
 
-import           Control.Concurrent.Supply   (splitSupply)
 import           Control.Exception           (throw)
 import           Control.Lens                (_2)
 import qualified Control.Lens                as Lens
@@ -88,8 +87,6 @@ import           BasicTypes                  (InlineSpec (..))
 
 import           Clash.Annotations.Primitive (extractPrim)
 import           Clash.Core.DataCon          (DataCon (..))
-import           Clash.Core.Evaluator        (whnf')
-import           Clash.Core.Evaluator.Types  (PureHeap)
 import           Clash.Core.Name
   (Name (..), NameSort (..), mkUnsafeSystemName, nameOcc)
 import           Clash.Core.FreeVars
@@ -141,8 +138,7 @@ import           Clash.Primitives.Types
 import           Clash.Rewrite.Combinators
 import           Clash.Rewrite.Types
 import           Clash.Rewrite.Util
-import           Clash.Unique
-  (Unique, lookupUniqMap, toListUniqMap)
+import           Clash.Unique                (lookupUniqMap)
 import           Clash.Util
 
 inlineOrLiftNonRep :: HasCallStack => NormRewrite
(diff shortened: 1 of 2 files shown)


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -494,150 +490,150 @@ inlineNonRep _ e = return e
...
+      -- or a primitive for which the evaluator doesn't have any evaluation
+      -- rules.
+      _ -> do
+        tcm <- Lens.view tcCache
+        let subjTy = termType tcm subj
+        tran <- Lens.view typeTranslator
+        reprs <- Lens.view customReprs
+        case (`evalState` HashMapS.empty) (coreTypeToHWType tran reprs tcm subjTy) of
+          Right (FilteredHWType (Void (Just hty)) _areVoids)
+            | hty `elem` [BitVector 0, Unsigned 0, Signed 0, Index 1]
```
martijnbastiaan: Couldn't we make this a bit more general and use `typeSize`?

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -494,150 +490,150 @@ inlineNonRep _ e = return e
...
+            [] -> altE
+            _  ->
+              -- See Note [CaseCon deshadow]
+              let ((is3,substIds),binds2) =
+                    List.mapAccumL newBinder (is1,[]) binds1
+                  subst = extendIdSubstList (mkSubst is3) substIds
+                  body  = substTm "caseCon0" subst altE
+              in  case Maybe.catMaybes binds2 of
+                    []     -> body
+                    binds3 -> Letrec binds3 body
```
martijnbastiaan: No change requested, but I've found a good way of formatting `let` bindings without indenting "too much" is:

```
let
  a = 123
  b = 456
in
  a + b
```

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -494,150 +490,150 @@ inlineNonRep _ e = return e
...
+      -- Use the original inScopeSet 'is0' here, not the extended inScopeSet
+      -- 'is1', otherwise we'd make the "caseCon1" substitution substitute
+      -- free variables that were shadowed by the pattern!
+      let subst = extendTvSubstList (mkSubst is0)
+                $ zip tvs (drop (length (dcUnivTyVars dc)) (Either.rights args))
+      changed (substTm "caseCon1" subst altE1)
+    _ -> case alts of
+           ((DefaultPat,altE):_) -> changed altE
+           _ -> changed (undefinedTm ty)
```
martijnbastiaan: Why is this true? Is it an invariant to have a `DefaultPat` first? And if it's followed by anything, shouldn't that be an error / thrown away by GHC?

martijnbastiaan: Whatever the answer is, maybe it should be added as a comment.

clash-lib/src/Clash/Normalize/Transformations.hs
christiaanb: No... we really only want to do it for those types; so we'd need to extract the `hty` out of the `Void`.

PR #1080, opened 2020-02-13
Emit message before trying transformation when running with DebugTry

martijnbastiaan:


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -142,6 +143,12 @@ apply
   -- ^ Transformation to be applied
   -> Rewrite extra
 apply = \s rewrite ctx expr0 -> do
+  lvl <- Lens.view dbgLevel
+  dbgTranss <- Lens.view dbgTransformations
+  let isTryLvl = lvl == DebugTry || lvl >= DebugAll
+      isRelevantTrans = s `Set.member` dbgTranss || Set.null dbgTranss
+  Monad.when (isTryLvl && isRelevantTrans) (trace ("Trying: " ++ s) (pure ()))
```
leonschoorl:
```haskell
traceIf (isTryLvl && isRelevantTrans) ("Trying: " ++ s) pure ()
```

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -44,6 +44,7 @@ import qualified Data.Set                    as Set
 import qualified Data.Set.Lens               as Lens
 import           Data.Text                   (Text)
 import qualified Data.Text                   as Text
+import           Debug.Trace
```
leonschoorl: If you use `traceIf` you don't need this

clash-lib/src/Clash/Rewrite/Util.hs
martijnbastiaan: Yeah that's better :). Thanks!

PR #1084, opened 2020-02-14
Change Binding from a type synonym

alex-mckenna:
Use of the `Binding` type (representing a single binding from a
`BindingMap`) is often quite obtuse, as it was a synonym for a
4-tuple. This commit changes it to a type, making it clearer at
use-sites what the data actually represents.

### DIFF
--- a/clash-lib/src/Clash/Normalize.hs
+++ b/clash-lib/src/Clash/Normalize.hs
@@ -59,7 +59,7 @@ import           Clash.Core.VarEnv
    extendVarEnv, lookupVarEnv, mapVarEnv, mapMaybeVarEnv, mkInScopeSet,
    mkVarEnv, mkVarSet, notElemVarEnv, notElemVarSet, nullVarEnv, unionVarEnv)
 import           Clash.Driver.Types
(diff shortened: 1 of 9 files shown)


clash-lib/src/Clash/Normalize.hs
```diff
@@ -289,9 +287,9 @@ flattenNode c@(CLeaf (nm,(_,_,_,e))) = do
                return (Right ((nm,mkApps (mkTicks fun ticks) (reverse remainder)),[]))
           _ -> return (Right ((nm,e),[]))
       _ -> return (Right ((nm,e),[]))
-flattenNode b@(CBranch (_,(_,_,NoInline,_)) _) =
+flattenNode b@(CBranch (_,(Binding _ _ NoInline _)) _) =
```
martijnbastiaan: I'd prefer to use `Binding{bindingSpec=NoInline}` in these cases, but we don't have to do it for this PR.

alex-mckenna: Yeah, the changes to make things compile again were done fairly mechanically. Some of the use sites are already pretty hairy as-is

PR #1096, opened 2020-02-19
Add "Clash.Signal.Bundle.TaggedUnit"

martijnbastiaan:
Helper type to emulate the "old" behavior of Bundle's unit instance. I.e.,
the instance for `Bundle ()` used to be defined as:

```haskell
class Bundle () where
  bundle   :: () -> Signal domain ()
  unbundle :: Signal domain () -> ()
```

In order to have sensible type inference, the `Bundle` class specifies that
the argument type of 'bundle' should uniquely identify the result type, and
vice versa for `unbundle`. The type signatures in the snippet above don't
though, as `()` doesn't uniquely map to a specific domain. In other words,
'domain' should occur in both the argument and result of both functions.

`TaggedUnit` fixes this by carrying the domain in its type. Use `taggedUnit`
to construct a `TaggedUnit`. The actual constructor is hidden to hide some
implementation details. If you need to pattern match on it use `seq` or
BangPatterns.

### DIFF
--- a/clash-prelude/src/Clash/Signal/Bundle.hs
+++ b/clash-prelude/src/Clash/Signal/Bundle.hs
@@ -9,6 +9,7 @@ The Product/Signal isomorphism
 -}
 
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeFamilyDependencies #-}
@@ -23,6 +24,9 @@ The Product/Signal isomorphism
 
 module Clash.Signal.Bundle
   ( Bundle (..)
+  -- ** Tools to emulate pre Clash 1.0 @Bundle ()@ instance
+  , TaggedUnit
+  , TaggedUnit'(..)
   -- ** Internal
   , vecBundle#
   )
@@ -166,3 +170,34 @@ instance Bundle ((f :*: g) a) where
    where
     getL (l :*: _) = l
     getR (_ :*: r) = r
+
+-- | Internal data type, see "TaggedUnit".
+data TaggedUnit' (dom :: k) = TaggedUnit
+
+-- | Helper type to emulate the "old" behavior of Bundle's unit instance. I.e.,
+-- the instance for @Bundle ()@ used to be defined as:
+--
+--     class Bundle () where
+--       bundle   :: () -> Signal domain ()
+--       unbundle :: Signal domain () -> ()
+--
+-- In order to have sensible type inference, the "Bundle" class specifies that
+-- the argument type of 'bundle' should uniquely identify the result type, and
+-- vice versa for 'unbundle'. The type signatures in the snippet above don't
+-- though, as @()@ doesn't uniquely map to a specific domain. In other words,
+-- 'domain' should occur in both the argument and result of both functions.
+--
+-- "TaggedUnit" fixes this by carrying the domain in its type. Use 'TaggedUnit'
+-- to construct a "TaggedUnit" or a "TaggedUnit'".
+type TaggedUnit (dom :: Domain) = TaggedUnit' '(dom, dom)
+
+-- | See https://github.com/clash-lang/clash-compiler/pull/539/commits/94b0bff5770aa4961e04ddce2515130df3fc7863
+-- and documentation for "TaggedUnit".
+instance Bundle (TaggedUnit' dom1) where
+  type Unbundled dom0 (TaggedUnit' dom1) = TaggedUnit' '(dom0, dom1)
+
+  bundle :: TaggedUnit' '(dom0, dom1) -> Signal dom0 (TaggedUnit' dom1)
+  bundle TaggedUnit = pure TaggedUnit
+
+  unbundle :: Signal dom0 (TaggedUnit' dom1) -> TaggedUnit' '(dom0, dom1)
+  unbundle s = seq s TaggedUnit
--- a/clash-prelude/src/Clash/Signal/Bundle.hs
+++ b/clash-prelude/src/Clash/Signal/Bundle.hs
@@ -171,7 +171,7 @@ instance Bundle ((f :*: g) a) where
     getL (l :*: _) = l
     getR (_ :*: r) = r
 
--- | Internal data type, see "TaggedUnit".
+-- | See "TaggedUnit".
 data TaggedUnit' (dom :: k) = TaggedUnit
 
 -- | Helper type to emulate the "old" behavior of Bundle's unit instance. I.e.,
--- a/clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
+++ b/clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
@@ -7,6 +7,7 @@
 -}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE NoImplicitPrelude #-}
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -20,7 +21,6 @@ module Clash.Signal.Delayed.Bundle (
     Bundle(..)
   -- ** Tools to emulate pre Clash 1.0 @Bundle ()@ instance
   , TaggedUnit
-  , TaggedUnit'(..)
   ) where
 
 import           Control.Applicative           (liftA2)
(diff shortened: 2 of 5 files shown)


martijnbastiaan:
@leonschoorl Still marked as WIP as I want to test it on a larger codebase, but should be reviewable :)

clash-prelude/src/Clash/Signal/Bundle.hs
```diff
@@ -9,9 +9,11 @@ The Product/Signal isomorphism
 -}
 
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE TypeSynonymInstances #-}
```
alex-mckenna: is this actually being used?

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
```diff
@@ -11,16 +11,17 @@
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeFamilyDependencies #-}
+{-# LANGUAGE TypeSynonymInstances #-}
```
alex-mckenna: Same question

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
```diff
@@ -219,3 +220,20 @@ instance KnownNat d => Bundle (RTree d a) where
   type Unbundled t delay (RTree d a) = RTree d (DSignal t delay a)
   bundle   = sequenceA
   unbundle = sequenceA . fmap lazyT
+
+-- | Internal data type, see "TaggedUnit".
+data TaggedUnit' (d :: k) (dom :: i) = TaggedUnit
+
+-- | Same as "Clash.Signal.Bundle", but adapted for "DSignal".
+type TaggedUnit (d :: Nat) (dom :: Domain) = TaggedUnit' '(d, d) '(dom, dom)
```
alex-mckenna: Does this need to be `(d :: Nat)`, or can it be `(d :: k)` too?

clash-prelude/src/Clash/Signal/Bundle.hs
martijnbastiaan: it's not anymore, thanks for catching

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
martijnbastiaan: Same answer

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
martijnbastiaan: Could, but I thought this would be a little bit more self-documenting

martijnbastiaan:
@alex-mckenna Otherwise OK?

clash-prelude/src/Clash/Signal/Bundle.hs
```diff
@@ -166,3 +170,34 @@ instance Bundle ((f :*: g) a) where
...
+--
+-- In order to have sensible type inference, the "Bundle" class specifies that
+-- the argument type of 'bundle' should uniquely identify the result type, and
+-- vice versa for 'unbundle'. The type signatures in the snippet above don't
+-- though, as @()@ doesn't uniquely map to a specific domain. In other words,
+-- 'domain' should occur in both the argument and result of both functions.
+--
+-- "TaggedUnit" fixes this by carrying the domain in its type. Use 'TaggedUnit'
+-- to construct a "TaggedUnit" or a "TaggedUnit'".
+type TaggedUnit (dom :: Domain) = TaggedUnit' '(dom, dom)
```
alex-mckenna: It actually occurs to me now that I don't so much like the naming here. I think that `TaggedUnit` should be the name for the general type (instead of `TaggedUnit'`) and the specific tags should have a name which is more separate.

I also have reservations about how conceptually similar this is to `(:::)`, essentially being `(::: ())`. Is there some middle ground where maybe that type is changed / newtyped to provide this behaviour?

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
```diff
@@ -219,3 +219,20 @@ instance KnownNat d => Bundle (RTree d a) where
...
```
alex-mckenna: Also, this being repeated here makes it essentially the same type in three places. At the very least I would expect the types for tags to be shared between normal / delayed `Bundle`, as the implementation should be identical.

martijnbastiaan: As discussed offline: both `TaggedUnit`s have a different number of type arguments (delayed vs non-delayed).

clash-prelude/src/Clash/Signal/Bundle.hs
martijnbastiaan: As discussed offline: I don't want users to write `TaggedUnit' '(dom, dom)`. But I agree it's kinda meh.

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
```diff
@@ -219,3 +219,20 @@ instance KnownNat d => Bundle (RTree d a) where
...
+
+-- | See https://github.com/clash-lang/clash-compiler/pull/539/commits/94b0bff5770aa4961e04ddce2515130df3fc7863
+-- and documentation for "TaggedUnit".
+instance Bundle (TaggedUnit' d1 dom1) where
```
christiaanb: Are you allowed to do:

```haskell
instance Bundle (TaggedUnit' '(d1,dom1)) where
```

And use only one `TaggedUnit'`?

clash-prelude/src/Clash/Signal/Bundle.hs
christiaanb: w.r.t. the `(:::)` becoming a newtype: they’re type-aliases to be very lightweight, making them a newtype negates that. Unless, we create a constraint-solver plugin that automatically (unsafe)Coerces from/to `(:::)`. But I wouldn’t want to spend much time on such a plug-in now.

clash-prelude/src/Clash/Signal/Delayed/Bundle.hs
martijnbastiaan: I'm quite sure that's possible. Is that a "changes requested"?

christiaanb: Yes, since we only need one `TaggedUnit'` then instead of two

christiaanb: And if we remove the `TaggedUnit'/TaggedUnit` distinction, we’d have only a single `TaggedUnit` as well. Which I wouldn’t mind given that its use is quite niche: i.e. I wonder how much it changes user-code outside of type class instances

martijnbastiaan: I don't want users writing / reading `TaggedUnit' '(dom1,dom1)`, so I'd like to keep `TaggedUnit` around. Merging the `TaggedUnit'`s is fine with me.

martijnbastiaan: Tot add to that: it does turn up in code outside of type class instances.

leonschoorl:
Thinks it's rather confusing with the two types of TaggedUnit and the two domains

```haskell
-- for 2-tuples bundle and unbundle like this
sigtup2,sigtup2' :: Signal System (Bool,Bool)
sigtup2 = pure (True,True)
sigtup2' = bundle tup2sig

tup2sig,tup2sig' :: (Signal System Bool, Signal System Bool)
tup2sig = (pure True, pure True)
tup2sig' = unbundle sigtup2

-- then for 0-tuples I'd expect to write:
sigtup0,sigtup0' :: Signal System (TaggedUnit System)
sigtup0 = pure (TaggedUnit)
sigtup0' = bundle tup0sig

tup0sig,tup0sig' :: TaggedUnit System
tup0sig = TaggedUnit
tup0sig' = unbundle sigtup0

-- But that doesn't work :(
-- The types need to be:
sigtup0,sigtup0' :: Signal System (TaggedUnit' System)
tup0sig,tup0sig' :: TaggedUnit' '(System,System)
```

martijnbastiaan:
The implementation in this PR has turned out to be troublesome on multiple fronts. I'll reopen a new, clean PR with an alternative implementation.

PR #1104, opened 2020-02-21
Add '-fstrict-mapSignal' cabal flag

martijnbastiaan:


alex-mckenna:
Can we make this something like `strict-signal` instead of `strict`? There's already a `super-strict` flag and I would expect the two to be related (but they don't seem to be, at least from the descriptions)

martijnbastiaan:
`-fstrict-mapSignal` maybe?

Op za 22 feb. 2020 14:25 schreef Alex McKenna <notifications@github.com>:

clash-prelude/src/Clash/CPP.hs
```diff
@@ -18,7 +18,27 @@ Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
...
+ -- ** Cabal flags
+ , fSuperStrict
+ , fStrictMapSignal
  ) where
 
 maxTupleSize :: Num a => a
 maxTupleSize = MAX_TUPLE_SIZE
+
+-- | Whether clash-prelude was compiled with -fsuper-strict
+fSuperStrict :: Bool
```
christiaanb: Perhaps add
```haskell
{-# INLINE fSuperStrict #-}
```

clash-prelude/src/Clash/CPP.hs
```diff
@@ -18,7 +18,27 @@ Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
...
+#ifdef CLASH_SUPER_STRICT
+fSuperStrict = True
+#else
+fSuperStrict = False
+#endif
+
+-- | Whether clash-prelude was compiled with -fstrict-mapSignal
+fStrictMapSignal :: Bool
```
christiaanb: Perhaps add
```haskell
{-# INLINE fStrictMapSignal #-}
```

PR #1123, opened 2020-02-26
Cache /nix/store on Circle CI

martijnbastiaan:


martijnbastiaan:
@leonschoorl I've implemented the caching strategy Circle CI recommends for "general projects". Cabal / nix / stack now always get cached, and caches fall through on a revision => branch => ci job basis. This seems like the way to go, do you agree?

martijnbastiaan:
First build: https://circleci.com/gh/clash-lang/clash-compiler/14755?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
Second build: https://circleci.com/gh/clash-lang/clash-compiler/14759?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Seems to work!

.circleci/config.yml
```diff
@@ -18,15 +18,24 @@ aliases:
...
+        else
+          echo not running > .circleci_cachekey
+        fi
   - &cache_restore
     restore_cache:
       keys:
-        - cache-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}-{{ .Revision }}
-        - cache-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}-
-        - cache-{{ .Environment.CIRCLE_JOB }}-
+        - cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "/usr/bin/gcc" }}-{{ checksum ".circleci_cachekey" }}-{{ .Branch }}-{{ .Revision }}
```
martijnbastiaan: Why the checksum on `gcc`?

.circleci/config.yml
```diff
@@ -18,15 +18,24 @@ aliases:
...
           .ci/setup.sh
         fi
+  - &cache_calc_key
+    run:
+      name: Calculate cache key
+      command: |
+        if [ -n "$CIRCLE_PR_NUMBER" ]; then
+          date +%Y-%m > .circleci_cachekey
+        else
+          echo not running > .circleci_cachekey
```
martijnbastiaan: It should run for nix-build / stack-build. Does this logic break caching for those?

.circleci/config.yml
```diff
@@ -23,22 +23,23 @@ aliases:
       name: Calculate cache key
       command: |
         if [ -n "$CIRCLE_PR_NUMBER" ]; then
-          md5sum .circleci/config.yml cabal.project .ci/cabal.project.local */*.cabal > .circleci_cachekey
+          date +%Y-%m > .circleci_cachekey
```
martijnbastiaan: I don't think `date` is available on nixos before restoring the cache / running the setup step.

.circleci/config.yml
leonschoorl: Last time we did a big update of the docker base image (moving from ubuntu trusty to bionic) it broke the ghc that stack had installed inside of  `~/.stack`.
My idea was to use the `gcc` binary as a proxy for big changes like that.

.circleci/config.yml
leonschoorl: Good point
But it's there:
```
$ docker run -it nixos/nix
Unable to find image 'nixos/nix:latest' locally
latest: Pulling from nixos/nix
9d48c3bd43c5: Pull complete 
480223332b13: Pull complete 
74bc3bbbef74: Pull complete 
Digest: sha256:5d0a26066dc8e9e1295676203cbc15d0e124f445007f9cb888fde25c3d00a1b2
Status: Downloaded newer image for nixos/nix:latest
362e45a9967c:/# date
Thu Feb 27 14:34:11 UTC 2020
362e45a9967c:/# date +%Y-%m
2020-02
```

.circleci/config.yml
leonschoorl: Yes, you're right

martijnbastiaan:
I've simplified the caching logic:

* The "throw-everything-out" measure is now based on `/proc/version`. This includes kernel/gcc/etc and works across all docker images.
* The "throw-everything-out" measure now actually throws everything out due to it coming first in the cache hierarchy.
* `/var/cache/apt` is now also cached to prevent hitting (slow) PPAs over and over again.

alex-mckenna:
I thought `/proc/version`'s GCC version was the version used to build the kernel instead of the one on the machine?

martijnbastiaan:
Hmm.. I don't know. It seems to match on my system:

```
martijn@enter:~$ gcc -v |& grep "gcc version"
gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04.1) 
martijn@enter:~$ cat /proc/version
Linux version 4.15.0-88-generic (buildd@lgw01-amd64-036) (gcc version 7.4.0 (Ubuntu 7.4.0-1ubuntu1~18.04.1)) #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020
```

It's probably "good enough"?

alex-mckenna:
If ubuntu keeps the build gcc for the kernel and the one in the package repo the same, then no problem. If not, there exists the potential for the dockerfile to be updated and these two versions to be out of sync. Should this invalidate the cache? Why did the ghc for stack break last time?

martijnbastiaan:
@alex-mckenna no clue

```
if [ -e /usr/bin/gcc ]; then /usr/bin/gcc --version >> ./circleci_cachekey; fi
```

it is.

leonschoorl:
But `/proc/version` is the kernel version of the host.
It has nothing to do with what is inside the container.

martijnbastiaan:
Hmm right, so _just_ the if statement should be enough then.

PR #1124, opened 2020-02-26
Lift recursive group

christiaanb:
In order to deal with a set of "to-lift" let-binders that form a recursive group, e.g.

```haskell
topEntity x y z
  = let g p q k v = k (p + q) * v
        h = ((3 :: Integer) *) . (g x y f)
        f = ((4 :: Integer) *) . (g x y h)
    in  f (h z) + h (f z)
```

The above definition remains recursive after normalization, at the global level, but it contains no self-recursive let-bindings. The previous implementation of the algorith would result in a normalized set of expressions with self-recursive, non-representable, let-bindings; which is not allowed.

### DIFF
--- a/clash-lib/src/Clash/Rewrite/Util.hs
+++ b/clash-lib/src/Clash/Rewrite/Util.hs
@@ -449,7 +449,51 @@ substituteBinders inScope toInline toKeep body =
     in  if x `localIdOccursIn` e1 then
           go subst ((x,e1):inlRec) toInl
         else
-          go subst2 inlRec (map (second (substTm "substToInl" substE)) toInl)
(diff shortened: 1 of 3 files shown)


martijnbastiaan:
Seems like a good testcase to add to the testsuite:

```haskell
topEntity x y z
  = let g p q k v = k (p + q) * v
        h = ((3 :: Integer) *) . (g x y f)
        f = ((4 :: Integer) *) . (g x y h)
    in  f (h z) + h (f z)
```

christiaanb:
I could make it a failing test: just that it now fails at the end of normalization, and not during.

PR #1140, opened 2020-03-02
Use the correct prefix when naming top-level components

### LINKED ISSUE #1139: Manifest location incorrect for implicit topEntity in (System)Verilog
```haskell
module TwoTops where
import Clash.Prelude

topEntity, otherTopEntity :: Bool -> Bool
topEntity = not
{-# NOINLINE topEntity #-}

otherTopEntity = topEntity
{-# ANN otherTopEntity (defSyn "otherTopEntity") #-}
```
When rendering the body of `otherTopEntity` to (System)Verilog, clash will try to load the manifest file of `topEntity` from the the wrong location:
```
$ cabal run -- clash --verilog TwoTops.hs 
[...]
Clash: Compiling TwoTops.topEntity
Clash: Normalisation took 0.000s
Clash: Compiling TwoTops.otherTopEntity
Clash: Normalisation took 0.000s

<no location info>: error:
    Other error:
    ./verilog/TwoTops/otherTopEntity_topEntity.manifest: openFile: does not exist (No such file or directory)
```
It wrote the manifest to: `./verilog/TwoTops/topEntity.manifest`

This only happens when compiling to (System)Verilog and `topEntity` is an implicit topentity (it has no Synthesize annotation).

leonschoorl:
Fixes #1139


clash-lib/src/Clash/Netlist.hs
```diff
@@ -160,8 +160,8 @@ runNetlistMonad
   -- ^ Seen components
   -> FilePath
   -- ^ HDL dir
-  -> (Maybe Identifier,Maybe Identifier)
   -- ^ Component name prefix
+  -> ComponentPrefix
```
alex-mckenna: Haddock comment accidentally no longer refers to the right argument

tests/shouldwork/TopEntity/TwoTops.hs
```diff
@@ -0,0 +1,10 @@
+module TwoTops where
+import Clash.Prelude
+
+-- https://github.com/clash-lang/clash-compiler/issues/1139
```
alex-mckenna: I actually really like the way Christiaan's being naming tests like this recently, i.e. `T1139.hs` It makes it more traceable why a test case exists in the first place IMO

PR #1149, opened 2020-03-04
Single-dc, single field types cannot have a PortProduct annotation

### LINKED ISSUE #1063: Port naming error detection logic not always working
```haskell
module TestNaming where
import Clash.Prelude

type T = Bool -- error as expected
-- type T = Int  -- no error :(

{-# ANN topEntity (defSyn "top")
  {t_inputs = [PortProduct "wrong" [PortName "one", PortName "two"]]}
  #-}
topEntity :: T -> T
topEntity = id
```
Returns an error as expected:
```
*** Exception: Clash.Netlist.Util(635): Ports were annotated as product, but type wasn't one: 

   Filtered was: FilteredHWType Bool []

   Ports was: PortProduct "wrong" [PortName "one",PortName "two"]
```

But if I change `T` to be `Int` clash will happily name the resulting in port `wrong`.

leonschoorl:
@christiaanb: Changes the logic in 'filterVoidPorts' to stop recognizing these "wrapper" types as one-products

Fixes #1063

### DIFF
--- a/clash-lib/src/Clash/Netlist/Util.hs
+++ b/clash-lib/src/Clash/Netlist/Util.hs
@@ -630,10 +630,12 @@ filterVoidPorts
   -> PortName
 filterVoidPorts _hwty (PortName s) =
   PortName s
-filterVoidPorts (FilteredHWType _hwty [filtered]) (PortProduct s ps) =
-  PortProduct s [filterVoidPorts f p | (p, (void, f)) <- zip ps filtered, not void]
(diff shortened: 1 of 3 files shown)


christiaanb:
Sorry @leonschoorl for force-pushing to your branch, but now we have a fix for all "wrappers", not just `Int` and `Word`

christiaanb:
@martijnbastiaan done

PR #1151, opened 2020-03-04
Simplify ByteArrayLiteral

alex-mckenna:
The ByteArrayLiteral type was a primitive vector (a bytearray with
offset and size fields included), and the only field ever used in
Clash was the ByteArray field. Consequently, the type is simply
changed to a ByteArray with no problems.

This change will also make it easier to implement ByteArray
operations in the new partial evalautor, as it means ByteArray can
be marshalled to and from Value easier.


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 * Fixes issues:
   * No fixes yet!
 * Fixes without issue reports:
-  * No fixes yet!
+  *  Use `ByteArray` instead of `Vector Word8` in the `ByteArrayLiteral` type.
```
martijnbastiaan: Link to the PR

PR #1162, opened 2020-03-06
Auto change grade in snapcraft.yaml

martijnbastiaan:
Please merge https://github.com/clash-lang/clash-compiler/pull/1163 first.

### DIFF
--- a/.ci/snap_publish.sh
+++ b/.ci/snap_publish.sh
@@ -6,16 +6,35 @@ export SNAPCRAFT_BUILD_ENVIRONMENT_CPU
 SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=16G
 export SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY
 
-# Prevent running if we've already published this revision
+# Install git to detect branch names / revisions
 apt update
 apt install git -y
-touch snap-last-run-hash
-if [ "$(cat snap-last-run-hash)" == "$(git rev-parse HEAD)-${RELEASE_CHANNEL}" ]; then
-  echo "Already built and published $(git rev-parse HEAD) on ${RELEASE_CHANNEL}. Nothing to do!";
+
+# Set release channel based on branch:
+#
+# A build with RELEASE_CHANNEL=beta_or_edge only runs at night as a schedule
+# (or if a schedule is manually triggered). The schedules target either
+# master or a release branch (1.0, 1.2, ..). If it's run on master it should
+# go into Snap's 'edge' channel, otherwise it's a pre-release dot-version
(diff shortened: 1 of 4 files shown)


.ci/snap_publish.sh
```diff
@@ -0,0 +1,33 @@
...
+set -euo pipefail
+
+SNAPCRAFT_BUILD_ENVIRONMENT_CPU=$(nproc)
+export SNAPCRAFT_BUILD_ENVIRONMENT_CPU
+SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=16G
+export SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY
+
+# Prevent running if we've already published this revision
+apt update
+apt install git -y
```
alex-mckenna: From the manual,

> All features of apt(8) are available in dedicated APT tools like apt-get(8) and apt-
       cache(8) as well.  apt(8) just changes the default value of some options (see apt.conf(5)
       and specifically the Binary scope). So you should prefer using these commands (potentially
       with some additional options enabled) in your scripts as they keep backward compatibility
       as much as possible.

alex-mckenna: I wouldn't expect usage / configuration of these commands to change in a way that would break things though. Just good to consider.

martijnbastiaan: I don't expect it to change either, but if I'm doing something more complicated I'll make sure to use the proper commands.

.ci/snap_publish.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+set -euo pipefail
+
+SNAPCRAFT_BUILD_ENVIRONMENT_CPU=$(nproc)
+export SNAPCRAFT_BUILD_ENVIRONMENT_CPU
+SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=16G
+export SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY
```
leonschoorl:
```
export SNAPCRAFT_BUILD_ENVIRONMENT_CPU=$(nproc)
export SNAPCRAFT_BUILD_ENVIRONMENT_MEMORY=16G
```
?

leonschoorl: And the same for `THREADS` and `ROOT`

martijnbastiaan: Shell check suggestion. Export+assign+command makes it fail silently.

martijnbastiaan: Woops.. that means memory could have gone on a single line.

leonschoorl: I had no idea.

martijnbastiaan: Yeah, me neither!

PR #1164, opened 2020-03-06
Move to CHANGELOG files

martijnbastiaan:


martijnbastiaan:
@leonschoorl @alex-mckenna @christiaanb Agreed?

martijnbastiaan:
@christiaanb proposes going all the way https://docs.gitlab.com/ee/development/changelog.html. I think I agree.

christiaanb:
I think should go for a manual approach now, and automate later.

PR #1165, opened 2020-03-06
Add release instructions

martijnbastiaan:


bindist/README.md
```diff
@@ -0,0 +1,50 @@
...
+  * `clash-ghc/clash-ghc.cabal`
+  * `bindist/linux/snap/snap/snapcraft.yaml`
+  * `docs/conf.py`
+  * `README.md`
+  * `content/install/index.md` @ `clash-lang/clash-lang.org`
+2. Change the defaults of cabal flags in `clash-prelude/clash-prelude.cabal`
+   where necessary. At the time of writing this applies to `-fmultiple-hidden`
+   and `-flarge-tuples`.
+3. Update the CHANGELOG (see changelog/README.md).
+4. Create a branch of the called ~`1.2`.
```
alex-mckenna: Point 4 isn't very clearly worded

bindist/README.md
```diff
@@ -0,0 +1,50 @@
...
+5. Ask someone with admin permissions on GitLab to create a new nightly schedule
```
alex-mckenna: Point 5 also doesn't say anything on how to do this. Maybe it's obvious, maybe it isn't?

martijnbastiaan: I've elaborated a bit, but it should be obvious what to do at https://gitlab.com/clash-lang/clash-compiler/pipeline_schedules :)

alex-mckenna: I get a 404 at that link, but I'll trust when you say it's obvious enough

martijnbastiaan: Yeah, if you're not logged in with the right account you'll get a 404 (because they don't want to leak whether you activated that feature or not)

PR #1172, opened 2020-03-06
Synthesize precompiled modules

martijnbastiaan:
Allows users to specify modules in existing package databases. For
example, if "Data.List" would contain a synthesizable 'topEntity', you
could simply invoke "clash Data.List --vhdl" and expect it to work.

Motivation: first step to adding Shake support.

PR contains a refactor that (hopefully?) makes loading logic easier to read and modify.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -97,7 +97,8 @@ import qualified Var
 -- Internal Modules
 import           Clash.GHC.GHC2Core                           (modNameM, qualifiedNameString')
 import           Clash.GHC.LoadInterfaceFiles
-  (loadExternalExprs, getUnresolvedPrimitives, loadExternalBinders)
+  (loadExternalExprs, getUnresolvedPrimitives, loadExternalBinders,
+   LoadedBinders(..))
 import           Clash.GHCi.Common                            (checkMonoLocalBindsMod)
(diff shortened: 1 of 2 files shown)


clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -154,147 +327,43 @@ loadModules useColor hdl modName dflagsM idirs = do
...
-    let (binders,modFamInstEnvs) = unzip tidiedMods
-        bindersC                 = concat binders
-        binderIds                = map fst (CoreSyn.flattenBinds bindersC)
-        plusFamInst f1 f2        = FamInstEnv.extendFamInstEnvList f1 (FamInstEnv.famInstEnvElts f2)
-        modFamInstEnvs'          = foldl' plusFamInst FamInstEnv.emptyFamInstEnv modFamInstEnvs
-
-    modTime <- startTime `deepseq` length binderIds `deepseq` MonadUtils.liftIO Clock.getCurrentTime
-    let modStartDiff = reportTimeDiff modTime startTime
-    MonadUtils.liftIO $ putStrLn $ "GHC: Parsing and optimising modules took: " ++ modStartDiff
+    () <- setupGhc useColor dflagsM idirs
```
christiaanb: why the pattern match on unit?

martijnbastiaan: Good question, I'll take it out.

PR #1193, opened 2020-03-10
Switch around Vector.fold arguments

christiaanb:
This makes it easer to use `fold` in a `1 <= n` context so you can "simply" do `fold @(n-1)`. The plugins won't be able to do this automatically because we end up with constraints:

```
[G] 1 <= m
[W] (n + 1) ~ m
```

where there's simply no strategy to discharge that `[W] (n+1) ~ m` in a deterministic and sound way.


martijnbastiaan:
I don't want to sound like a broken record but.. I don't think we should backport API changes. `fold @_ @(n-1)` would still work in combination with older plugins right?

PR #1202, opened 2020-03-11
Treat `Signed 0`, `Unsigned 0`, `Index 1`, `BitVector 0` as unit

martijnbastiaan:


clash-prelude/src/Clash/Promoted/Nat.hs
```diff
@@ -115,15 +117,39 @@ instance ShowX (SNat n) where
 withSNat :: KnownNat n => (SNat n -> a) -> a
 withSNat f = f SNat
 
+-- | Same as 'snatToInteger', but doesn't take term arguments. Example usage:
+--
+-- >>> natToInteger @5
+-- 5
+natToInteger :: forall n . KnownNat n => Integer
```
christiaanb: But we already have `GHC.TypeLits.natVal`... perhaps the comment should be that this is the ambiguous version of that

clash-prelude/src/Clash/Promoted/Nat.hs
```diff
@@ -115,15 +117,39 @@ instance ShowX (SNat n) where
...
 -- | Reify the type-level 'Nat' @n@ to it's term-level 'Integer' representation.
 snatToInteger :: SNat n -> Integer
 snatToInteger p@SNat = natVal p
 {-# INLINE snatToInteger #-}
 
+-- | Same as 'snatToNatural', but doesn't take term arguments. Example usage:
+--
+-- >>> natToNatural @5
+-- 5
+natToNatural :: forall n . KnownNat n => Natural
```
christiaanb: This is `GHC.TypeNats.natVal` (which is distinct from `GHC.TypeLits.natVal`)

clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
```diff
@@ -259,14 +260,16 @@ instance KnownNat n => Bounded (Unsigned n) where
...
-{-# NOINLINE minBound# #-}
 minBound# :: Unsigned n
 minBound# = U 0
+{-# NOINLINE minBound# #-}
 
+maxBound# :: forall n. KnownNat n => Unsigned n
+maxBound# =
+  case natToNum @n of
+    0 -> 0
+    n -> let m = 1 `shiftL` n in  U (m - 1)
```
christiaanb: Why the case? isn't `shiftL 1 0` equal to `1`

martijnbastiaan: Woops, forgot to revert that after having implemented `errorX ....`

clash-prelude/src/Clash/Promoted/Nat.hs
martijnbastiaan: I completely missed the difference between `GHC.TypeNats.natVal` and `GHC.TypeLits.natVal`. It's not the same right? This function is ambiguous. I.e.,

```haskell
natVal :: forall n proxy . KnownNat n => proxy n -> Natural
```

vs

```haskell
natToNatural :: forall n. KnownNat n => Natural
```

alex-mckenna: Is there any reason to prefer `natToNatural @5` over `natVal (Proxy @5)`?

martijnbastiaan: No `Proxy` clutter, clear difference between `GHC.TypeNats.natVal` and `GHC.TypeLits.natVal`. But it doesn't make a big difference.

christiaanb: in the places where you've used these new functions, they can be rewritten using `GHC.TypeNats.natVal` like so:

```
xyz :: KnownNat n => Unsigned n
xyz = let r = case GHC.TypeNats.natVal r of { ... }  in r
```

since our numeric types are all proxies for `n`

alex-mckenna: Personally I would say using `Proxy` is less clutter than defining a (technically) redundant function, although granted this doesn't matter too much because it doesn't leak into user-facing APIs.

:man_shrugging:

martijnbastiaan: Actually, that's why I wrote the functions in the first place. 

```haskell
xyz :: forall n. KnownNat n => Unsigned n
xyz = case natToNatural @n of { ... }
```

seems more obvious to me.

I'd be happy to revert it though if there are strong opinions on this.

christiaanb:
Added backport, because I don't think anyone was depending on `Signed 0` being the odd one out and having `maxBound` throw an exception.

PR #1207, opened 2020-03-11
 Add '-main-is' support 

martijnbastiaan:
```
-main-is
  When using one of ``--vhdl``, ``--verilog``, or ``--systemverilog``, this
  flag refers to synthesis target. For example, running Clash with
  ``clash My.Module -main-is top --vhdl`` would synthesize ``My.Module.top``.
  Note that this will make Clash NOT compile any other binders with Synthesize
  annotations.
```

### DIFF
--- a/clash-ghc/src-bin-841/Clash/GHCi/UI.hs
+++ b/clash-ghc/src-bin-841/Clash/GHCi/UI.hs
@@ -1998,6 +1998,8 @@ makeHDL backend optsRef srcs = do
                 let dbs = reverse [p | PackageDB (PkgConfFile p) <- packageDBFlags dflags]
                 (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs) <-
                   generateBindings color primDirs idirs dbs hdl src (Just dflags)
+                let getMain = getMainTopEntity src bindingsMap topEntities
+                mainTopEntity <- traverse getMain (GHC.mainFunIs dflags)
                 prepTime <- startTime `deepseq` bindingsMap `deepseq` tcm `deepseq` Clock.getCurrentTime
                 let prepStartDiff = reportTimeDiff prepTime startTime
                 putStrLn $ "GHC+Clash: Loading modules cumulatively took " ++ prepStartDiff
@@ -2013,6 +2015,7 @@ makeHDL backend optsRef srcs = do
                   (ghcTypeToHWType iw fp)
                   primEvaluator
                   topEntities
+                  mainTopEntity
                   opts2
                   (startTime,prepTime)
 
--- a/clash-ghc/src-bin-861/Clash/GHCi/UI.hs
+++ b/clash-ghc/src-bin-861/Clash/GHCi/UI.hs
@@ -2047,6 +2047,8 @@ makeHDL backend optsRef srcs = do
                 let dbs = reverse [p | PackageDB (PkgConfFile p) <- packageDBFlags dflags]
                 (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs) <-
                   generateBindings color primDirs idirs dbs hdl src (Just dflags)
+                let getMain = getMainTopEntity src bindingsMap topEntities
+                mainTopEntity <- traverse getMain (GHC.mainFunIs dflags)
                 prepTime <- startTime `deepseq` bindingsMap `deepseq` tcm `deepseq` Clock.getCurrentTime
                 let prepStartDiff = reportTimeDiff prepTime startTime
                 putStrLn $ "GHC+Clash: Loading modules cumulatively took " ++ prepStartDiff
@@ -2062,6 +2064,7 @@ makeHDL backend optsRef srcs = do
                   (ghcTypeToHWType iw fp)
                   primEvaluator
                   topEntities
+                  mainTopEntity
                   opts2
                   (startTime,prepTime)
 
--- a/clash-ghc/src-bin-881/Clash/GHCi/UI.hs
+++ b/clash-ghc/src-bin-881/Clash/GHCi/UI.hs
@@ -2139,6 +2139,9 @@ makeHDL backend optsRef srcs = do
                 let dbs = reverse [p | PackageDB (PkgConfFile p) <- packageDBFlags dflags]
                 (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs) <-
                   generateBindings color primDirs idirs dbs hdl src (Just dflags)
+
+                let getMain = getMainTopEntity src bindingsMap topEntities
+                mainTopEntity <- traverse getMain (GHC.mainFunIs dflags)
                 prepTime <- startTime `deepseq` bindingsMap `deepseq` tcm `deepseq` Clock.getCurrentTime
                 let prepStartDiff = reportTimeDiff prepTime startTime
                 putStrLn $ "GHC+Clash: Loading modules cumulatively took " ++ prepStartDiff
@@ -2154,6 +2157,7 @@ makeHDL backend optsRef srcs = do
                   (ghcTypeToHWType iw fp)
                   primEvaluator
                   topEntities
+                  mainTopEntity
                   opts2
                   (startTime,prepTime)
 
--- a/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
+++ b/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
@@ -1,15 +1,19 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE QuasiQuotes #-}
 
 module Clash.GHCi.Common
   ( checkImportDirs
   , checkMonoLocalBinds
   , checkMonoLocalBindsMod
   , checkClashDynamic
+  , getMainTopEntity
   ) where
 
 -- Clash
-import           Clash.Driver.Types     (ClashOpts (..))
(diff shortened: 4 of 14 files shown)


gergoerdi:
Would it be possible to make this also have the effect of `makeTopEntity`. i.e. to get port names from annotations when present?

martijnbastiaan:
I think that's something we'd like to do in general. I.e., when there's no `Synthesize` annotation present or there's a `Synthesize` annotation but it doesn't specify port names, Clash should try and recover port names from the type annotations.

clash-ghc/src-bin-841/Clash/GHCi/UI.hs
```diff
@@ -1998,6 +1998,8 @@ makeHDL backend optsRef srcs = do
                 let dbs = reverse [p | PackageDB (PkgConfFile p) <- packageDBFlags dflags]
                 (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs) <-
                   generateBindings color primDirs idirs dbs hdl src (Just dflags)
+                mainTopEntity <-
+                  sequence (getMainTopEntity src topEntities <$> GHC.mainFunIs dflags)
```
christiaanb:
```suggestion
                  traverse (getMainTopEntity src topEntities) (GHC.mainFunIs dflags)
```

clash-ghc/src-bin-861/Clash/GHCi/UI.hs
```diff
@@ -2047,6 +2047,8 @@ makeHDL backend optsRef srcs = do
                 let dbs = reverse [p | PackageDB (PkgConfFile p) <- packageDBFlags dflags]
                 (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs) <-
                   generateBindings color primDirs idirs dbs hdl src (Just dflags)
+                mainTopEntity <-
+                  sequence (getMainTopEntity src topEntities <$> GHC.mainFunIs dflags)
```
christiaanb:
```suggestion
                  traverse (getMainTopEntity src topEntities) (GHC.mainFunIs dflags)
```

clash-ghc/src-bin-881/Clash/GHCi/UI.hs
```diff
@@ -2139,6 +2139,8 @@ makeHDL backend optsRef srcs = do
                 let dbs = reverse [p | PackageDB (PkgConfFile p) <- packageDBFlags dflags]
                 (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs) <-
                   generateBindings color primDirs idirs dbs hdl src (Just dflags)
+                mainTopEntity <-
+                  sequence (getMainTopEntity src topEntities <$> GHC.mainFunIs dflags)
```
christiaanb:
```suggestion
                  traverse (getMainTopEntity src topEntities) (GHC.mainFunIs dflags)
```

clash-lib/src/Clash/Driver.hs
```diff
@@ -225,14 +225,23 @@ generateHDL
...
   -> (Clock.UTCTime,Clock.UTCTime)
   -> IO ()
 generateHDL reprs bindingsMap hdlState primMap tcm tupTcm typeTrans eval
-  topEntities0 opts (startTime,prepTime) =
-    go prepTime HashMap.empty (sortTop bindingsMap topEntities2)
+  topEntities0 mainTopEntity opts (startTime,prepTime) =
+    go
+      prepTime
+      HashMap.empty
+      -- TODO: When given a main top entity (by -main-is) we don't currently
```
christiaanb: This TODO is for this PR? or a next?

clash-ghc/src-bin-841/Clash/GHCi/UI.hs
martijnbastiaan: `traverse` because it's more general, or is there something else going on?

clash-lib/src/Clash/Driver.hs
martijnbastiaan: I want the current behavior for `clash-shake`, so it's enough for me for now. Unless you think this is behavior is ridiculous, then I'll implement the TODO for this PR.

clash-ghc/src-bin-841/Clash/GHCi/UI.hs
christiaanb: Because that's what traverse is for, to map an effectful computation over a Functor-like thing.

clash-lib/src/Clash/Driver.hs
christiaanb: As long as we don't release a version with the TODO still in, i.e. add an issue to the 1.4 milestone to implement the TODO.

martijnbastiaan: In that case I'll add it to this PR. I'll request a review when it's done.

clash-prelude/src/Clash/Signal/BiSignal.hs
```diff
@@ -167,19 +167,15 @@ newtype BiSignalOut (ds :: BiSignalDefault) (dom :: Domain) (n :: Nat)
...
   (BiSignalOut b1) <> (BiSignalOut b2) = BiSignalOut (b1 ++ b2)
-#endif
 
 -- | Monoid instance to support concatenating
 --
 -- __NB__ Not synthesizable
 instance Monoid (BiSignalOut defaultState dom n) where
   mempty                                    = BiSignalOut []
-#if !MIN_VERSION_base(4,11,0)
   mappend (BiSignalOut b1) (BiSignalOut b2) = BiSignalOut $ b1 ++ b2
```
leonschoorl: > NOTE: This method is redundant and has the default implementation mappend = '(<>)' since base-4.11.0.0.

clash-ghc/src-bin-common/Clash/GHCi/Common.hs
```diff
@@ -17,11 +16,7 @@ import           Clash.Netlist.Types    (TopEntityT(..))
 
 -- The GHC interface
 import qualified DynFlags
-#if MIN_VERSION_base(4,11,0)
```
leonschoorl: You should update the lower bound of the base dependency in clash-ghc.cabal

martijnbastiaan:
@christiaanb Turns out it's impossible to implement compiling a single `topEntity` anyway :(. In order to generate HDL we need the manifest files of other top entities. We can't deduce this based on their types / synthesize annotations as they might use custom types and we have no idea what they'll be before normalization / netlist generation.

I guess it's not that important for now, but ideally I'd like to either add a config option to `Synthesize` that allows users to specify whether they'd like to use custom types. Clash could then tell users to disable it if they want parallel / single-main builds. (Alternatively: throw out custom types on top entities entirely. Might be too much, but it is easier to maintain.)

Thoughts?

leonschoorl:
Assuming these other top entities are already compiled (maybe in other runs using `-main-is`) can't you just read their manifests from disk?

martijnbastiaan:
@leonschoorl I had hoped this would "trivially" enable parallelization in clash-shake, but alas, it looks like I just have to have Clash compile everything in order (which I've got a patch for :-)).

christiaanb:
I think this calls for a "database" that can be shared between different batch runs of the Clash compiler, in addition to these manifest residing in other directories. Then you can just specify that you want to use a "database" that has been used to build the other top entities.

clash-prelude/src/Clash/Signal/BiSignal.hs
alex-mckenna: It _should_ also be an error (pragmatically speaking). Do we not use `-Wcompat`?

clash-ghc/src-bin-common/Clash/GHCi/Common.hs
martijnbastiaan: Good idea

clash-prelude/src/Clash/Signal/BiSignal.hs
martijnbastiaan: Now we do! https://github.com/clash-lang/clash-compiler/pull/1210

clash-lib/src/Clash/Driver.hs
martijnbastiaan: I've removed support for compiling singular top entities altogether. The TODO is now implemented.

martijnbastiaan:
@christiaanb I've removed the ability to compile single top entities. Instead, Clash will always compile all dependencies. Should be good to go now.

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -330,6 +330,10 @@ loadModules useColor hdl modName dflagsM idirs = do
   startTime <- Clock.getCurrentTime
   GHC.runGhc (Just libDir) $ do
     setupGhc useColor dflagsM idirs
+    -- TODO: We currently load the transitive closure of _all_ bindings found
+    -- TODO: in the top module. This is wasteful if one or more binders don't
+    -- TODO: contribute to any top entities. This effect is worsened when using
+    -- TODO: -main-is, which only synthesizes a single top entity.
```
christiaanb: Not true any longer? the bit about the single top entity

martijnbastiaan: We still load _all_ potential top entities and their dependency closure, even though in the end we might only pick out one (and _its_ dependencies) to synthesize. The last sentence is a bit misleading: maybe it should read "which only synthesizes a single top entity (and any top entities it depends on)".

martijnbastiaan: Changed the last sentence.

PR #1210, opened 2020-03-12
Remove <=8.2 CPP and compile with '-Wcompat'

martijnbastiaan:


christiaanb:
Added backport, because this changes so many things, there's bound to be conflicts if we fix a but around one of those lines.

martijnbastiaan:
OK! We don't support 8.2 in 1.2 anyway, so I think this is fair to backport anyway.

PR #1212, opened 2020-03-12
Use Microstache instead of Mustache

gergoerdi:
Switch over to Microstache, which is already a transitive dependency of Clash (via Criterion).

Note that [Microstache has no TH API](https://github.com/phadej/microstache/issues/18) and I am waiting to hear back from the maintainer if he'd be happy accepting a patch to add it back in. I think a "real" production version of `clash-shake` should include its templates compiled in instead of relying on fragile external files.


gergoerdi:
Actually I take it back. I think I have mis-read some dependency output -- Clash doesn't depend on any of the Mustache libraries on Hackage, so we should simply choose one which has good TH support and integrates nicely into everything else.

PR #1214, opened 2020-03-12
Prune utils modules

alex-mckenna:
Pruning `Clash.Utils` and `Clash.Core.Utils` as they both contain definitions which are redundant or can be moved to a more focused module. This is just a quick-fire improvement, intended to make it easier to improve the `Clash.Core` namespace later.


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -239,7 +239,7 @@ zipTvSubst
   -> Subst
 zipTvSubst tvs tys
   | debugIsOn
-  , neLength tvs tys
+  , length tvs /= length tys
```
martijnbastiaan: Depending on what magic GHC applies, the new version does a lot more work. (Might be a microoptimization though.)

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -610,7 +610,7 @@ substTyWith
   -> Type
   -> Type
 substTyWith tvs tys =
-  ASSERT( tvs `equalLength` tys )
+  ASSERT( length tvs == length tys )
```
martijnbastiaan: Same

clash-lib/src/Clash/Util.hs
```diff
@@ -295,24 +254,6 @@ indexNote
...
 
--- | Safe version of 'head'
-headMaybe :: [a] -> Maybe a
-headMaybe (a:_) = Just a
-headMaybe _ = Nothing
-
--- | Safe version of 'tail'
-tailMaybe :: [a] -> Maybe [a]
-tailMaybe (_:as) = Just as
-tailMaybe _ = Nothing
```
martijnbastiaan: IMO keep these around, they're useful no?

clash-lib/src/Clash/Util.hs
```diff
@@ -447,12 +366,6 @@ allM p (x:xs) = do
   else
     return False
 
-traceWith :: (a -> String) -> a -> a
-traceWith f a = trace (f a) a
-
-traceShowWith :: Show b => (a -> b) -> a -> a
-traceShowWith f a = trace (show (f a)) a
-
```
martijnbastiaan: These are super useful when debugging!

clash-lib/src/Clash/Util.hs
```diff
@@ -295,24 +254,6 @@ indexNote
...
-headMaybe (a:_) = Just a
-headMaybe _ = Nothing
-
--- | Safe version of 'tail'
-tailMaybe :: [a] -> Maybe [a]
-tailMaybe (_:as) = Just as
-tailMaybe _ = Nothing
-
--- | Split the second list at the length of the first list
-splitAtList :: [b] -> [a] -> ([a], [a])
```
christiaanb: Why remove this? `splitAtList xs ys` is faster than `splitAt (lenght xs) ys` when `xs` is bigger than `ys`, as you don't have to fully traverse `xs`

clash-lib/src/Clash/Util.hs
```diff
@@ -348,20 +289,6 @@ clogBase x y | x > 1 && y > 0 =
                 else Just (I# z1)
 clogBase _ _ = Nothing
 
--- | Determine whether two lists are of equal length
-equalLength
```
christiaanb: Same here, why remove this when `equalLength xs ys` finishes earlier than `lenght xs == length ys`

clash-lib/src/Clash/Util.hs
```diff
@@ -348,20 +289,6 @@ clogBase x y | x > 1 && y > 0 =
...
-  :: [a] -> [b] -> Bool
-equalLength [] [] = True
-equalLength (_:as) (_:bs) = equalLength as bs
-equalLength _ _ = False
-
--- | Determine whether two lists are not of equal length
-neLength
```
christiaanb: Same question as `splitAtList` and `equalLength`

clash-lib/src/Clash/Util.hs
alex-mckenna: They were only used in one place and both were used (equivalent to uncons).
I also don't think `fmap fst . uncons` etc. is any more work to write :man_shrugging:

martijnbastiaan: I wasn't aware of `uncons` (and it certainly a  better fit for where it's used). Having said that, `headMaybe` is definitely easier to write, but more importantly read, than `fmap fst . uncons`.

clash-lib/src/Clash/Util.hs
alex-mckenna: Fair point, what about a debugging module that exports these (e.g. `Clash.Internal.Debug`)?

clash-lib/src/Clash/Util.hs
alex-mckenna: True, but there's also the extra where do I import that from / where is that imported from which is avoided with `uncons`. Given it was only used in one place I don't think this is much of a loss anyway

alex-mckenna:
With the removed functions that @christiaanb mentioned, I think removing them from utils is still a good idea as

  * each was only used one
  * `equalLength` is only used in an `ASSERT`, I don't think the increase in speed is important given it only occurs in debug builds anyway
 
With `splitAtList`, it might be a good idea to add it back, as the function that uses it is used in `coreView1`. Although I think it should be added back locally to the function that uses it's `where` block if this is what we want to do

alex-mckenna:
Discussion offline, the list functions should be kept for now, but will be moved to a more appropriate module (i.e. `Data.List.Extra`, but feel free to bikeshed :wink:).

clash-lib/src/Clash/Util.hs
martijnbastiaan: The nice thing about these is that they're usually imported by default as many modules import `Clash.Util` unqualified (eww..).

clash-lib/src/Clash/Core/Term.hs
```diff
@@ -126,6 +139,57 @@ data Pat
...
+patIds _                    = ([],[])
+
+patVars :: Pat -> [Var a]
+patVars (DataPat _ tvs ids) = coerce tvs ++ coerce ids
+patVars _ = []
+
+-- | Abstract a term over a list of term and type variables
+mkAbstraction :: Term
+              -> [Either Id TyVar]
+              -> Term
```
martijnbastiaan: If you feel like it: reformat these functions to adhere to the styleguide (i.e., type signature on a single line OR the `::` on the next line.

alex-mckenna: I always feel like making the code prettier, I just didn't want to make all the diffs confusing
If this gets approved, how about I squash the two pruning commits and then do the stylistic changes on the squashed commit?

martijnbastiaan: Works for me, no need to squash IMO.

alex-mckenna: Fair, in that case I'll leave a style fix commit at the end of the PR so the history doesn't become more tedious to read

PR #1221, opened 2020-03-14
`hexImage`: track dependency on input file

gergoerdi:


martijnbastiaan:
Thanks! Just a heads-up: don't depend on that branch being stable, I'm probably going to shift a lot of things around after I've refactored Clash to have it interact nicely with external things :)

PR #1227, opened 2020-03-18
Don't use manifest files when instantiating top entity

martijnbastiaan:
Before this patch top entities would contain ports with types from that
topentity's _types module. In order to use a top entity the caller would
need to convert to those types using toBitVector/fromBitVector calls. In
order to use a top entity in a larger design Clash would therefore need
to compile _that_ top entity first, in order to know what conversion
functions to use. With this patch types Clash exposes only builtin HDL
types on top entities, making them predictable without having to run
normalization/netlist for a used top entity.

The motivation for this patch is two-fold:

  1. Not relying on a Manifest file is a first step towards compiling
     designs in parallel.

  2. Exposing "custom" types at a top entity makes it difficult to
     work with simulation/external tooling.

There's still a big TODO: see the comment in 'mkFunApp' for more
information.

------------------------

TODO for this PR:

 - [x] Documentation
 - [x] Reivew comments


clash-lib/src/Clash/Backend.hs
```diff
@@ -61,6 +61,19 @@ data Usage
...
 
+-- | Kind of a HDL type. Used to determine whether types need conversions in
+-- order to cross top entity boundaries.
+data HWKind
+  = PrimitiveType
+  -- ^ A type defined in an HDL spec. Usually types such as: bool, bit, ..
+  | SynonymType
+  -- ^ A user defined type that's simply a synonym for another type, very much
+  -- like a type synonym in Haskell. As long as two synonym types refer to the
+  -- same type, they can be used interchangeably. E.g., a subtype in VHDL.
```
christiaanb: I would really like to see a test case for this.

That is, given:
```vhdl
subtype SubA is std_logic_vector(4 downto 0);
subtype SubB is std_logic_vector(4 downto 0):
signal a: SubA;
signal b: SubB;
signal s: std_logic_vector(4 downto 0);
```
I believe the following assignments would work:
```vhdl
a <= s;
s <= a;
b <= s;
s <= b;
```
but under the normal understanding of subtyping I'm unconvinced the following assignments are allowed:
```vhdl
a <= b;
b <= a;
```

But that's what the comment, and the rest of the code seem to suggest!

martijnbastiaan: Actually I stumbled on this by accident, I was surprised too! I'll add an explicit test if you remain unconvinced, but consider our I2C example; you'll see the following in `i2c.vhdl`:

```vhdl
  bitmaster_ds1 : entity bitmaster.bitmaster
    port map
      ( [..]
      , cmd    => cmd
      , [..]
      );
```

In this case, `cmd` is declared as:

```vhdl
signal cmd              : i2c_types.i2ccommand;
```

when you hop over to `bitmaster.vhdl`, you'll see the following instead:

```vhdl
entity bitmaster is
  port(-- clock
       [..]
       cmd    : in bitmaster_types.i2ccommand;
       [..]
end;
```

The VHDL Language Reference mentions the following about subtypes:

*  A subtype declaration does not define a new type.
* A subtype is the same type as its base type; thus, no type conversion is needed when objects of a subtype and its base type are assigned (in either direction). Also, the set of operations allowed on operands of a subtype is the same as the set of operations on its base type.

martijnbastiaan: Two notes:

* Although the VHDL Language Reference doesn't _explicitly_ say it, I think it implies that you can assign two subtypes as long as their base types are the same.
* Although GHDL doesn't complain, it might be worthwhile checking this in modelsim.

christiaanb: Okay, it seems to work in Quartus as well. So we can leave this as is.

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1000,6 +1004,41 @@ prefixParent parent (PortName p)        = PortName (parent <> "_" <> p)
...
+    UserType -> action
+    SynonymType -> pure a
+    PrimitiveType -> pure a
+
+toPrimitiveType
+  :: Identifier
+  -> HWType
+  -> NetlistMonad ([Declaration], Identifier, Expr, HWType)
+toPrimitiveType id0 hwty0 = convPrimitiveType hwty0 dflt $ do
+  id1 <- mkUniqueIdentifier Basic id0
```
christiaanb: Any reason to use `Basic` instead of `Extended` identifiers here? Extended identifiers are better at preserving original Haskell names, and should be preferred. We only use `Basic` identifiers in specific cases, such as: label names, and GenSym template names.

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -111,6 +113,37 @@ instance Backend VHDLState where
 
   genHDL          = genVHDL
   mkTyPackage     = mkTyPackage_
+  hdlHWTypeKind = \case
```
martijnbastiaan: Reuse in `tyDec`

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1000,6 +1004,41 @@ prefixParent parent (PortName p)        = PortName (parent <> "_" <> p)
 prefixParent parent (PortProduct "" ps) = PortProduct parent ps
 prefixParent parent (PortProduct p ps)  = PortProduct (parent <> "_" <> p) ps
 
+mkAssign :: Identifier -> HWType -> Expr -> [Declaration]
+mkAssign id_ hwty expr = [NetDecl Nothing id_ hwty, Assignment id_ expr]
+
+convPrimitiveType :: HWType -> a -> NetlistMonad a -> NetlistMonad a
```
martijnbastiaan: Documentation please

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1000,6 +1004,41 @@ prefixParent parent (PortName p)        = PortName (parent <> "_" <> p)
...
+convPrimitiveType hwty a action = do
+  b <- Lens.use backend
+  let kind = case b of {SomeBackend s -> State.evalState (hdlHWTypeKind hwty) s}
+  case kind of
+    UserType -> action
+    SynonymType -> pure a
+    PrimitiveType -> pure a
+
+toPrimitiveType
```
martijnbastiaan: ^

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1000,6 +1004,41 @@ prefixParent parent (PortName p)        = PortName (parent <> "_" <> p)
...
+  -> NetlistMonad ([Declaration], Identifier, Expr, HWType)
+toPrimitiveType id0 hwty0 = convPrimitiveType hwty0 dflt $ do
+  id1 <- mkUniqueIdentifier Basic id0
+  pure (mkAssign id1 hwty1 expr, id1, expr, hwty1)
+ where
+  dflt = ([], id0, Identifier id0 Nothing, hwty0)
+  hwty1 = BitVector (typeSize hwty0)
+  expr = ToBv Nothing hwty0 (Identifier id0 Nothing)
+
+fromPrimitiveType
```
martijnbastiaan: ^

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1139,12 +1181,27 @@ mkInput pM = case pM of
...
                   netassgn = Assignment pN dcExpr
               return (concat ports,[netdecl,netassgn],dcExpr,pN)
             _ -> error "Unexpected error for PortProduct"
 
-        _ ->  return ([(pN,hwty)],[],Identifier pN Nothing,pN)
+        _ ->
+          portProductError $(curLoc) hwty' (PortProduct p ps)
+
+portProductError :: String -> HWType -> PortName -> a
+portProductError loc hwty portProduct = error $ loc ++ [I.i|
```
martijnbastiaan: throw `ClashException` with curLoc set instead

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1344,62 +1424,73 @@ mkOutput' pM = case pM of
...
             _ -> error "Unexpected error for PortProduct"
 
-        _ -> return ([(pN,hwty)],[],pN)
+        _ ->
+          portProductError $(curLoc) hwty' (PortProduct p ps)
 
     assignId p hwty con i n =
       Assignment i (Identifier p (Just (Indexed (hwty,con,n))))
 
+mkTopCompDecl
```
martijnbastiaan: Documentation please

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1000,6 +1004,41 @@ prefixParent parent (PortName p)        = PortName (parent <> "_" <> p)
...
+  dflt = ([], id0, Identifier id0 Nothing, hwty0)
+  hwty1 = BitVector (typeSize hwty0)
+  expr = ToBv Nothing hwty0 (Identifier id0 Nothing)
+
+fromPrimitiveType
+  :: Identifier
+  -> HWType
+  -> NetlistMonad ([Declaration], Identifier, Expr, HWType)
+fromPrimitiveType id0 hwty0 = convPrimitiveType hwty0 dflt $ do
+  id1 <- mkUniqueIdentifier Basic id0
```
christiaanb: Same, why `Basic` instead of `Extended`?

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1063,28 +1105,28 @@ mkInput pM = case pM of
...
     -- PortName specified by user
-    go' (PortName p) (i,hwty) = do
-      pN <- uniquePortName p i
-      return ([(pN,hwty)],[],Identifier pN Nothing,pN)
+    go'
+      :: PortName
+      -> (Identifier, HWType)
+      -> NetlistMonad ([(Identifier, HWType)], [Declaration], Expr, Identifier)
+    go' (PortName p) (i0, hwty0) = do
+      pN <- uniquePortName p i0 -- << Seems silly, remove?
```
christiaanb: No, don't remove. When `p` is not the empty string, it ensures that we don't pick another basic identifier with a differently cased letters in a case-insensitive language. So it doesn't actually change `p` when `p` is not empty, but does ensure it's unique w.r.t. generated identifiers.

clash-lib/src/Clash/Netlist/Util.hs
martijnbastiaan: No! I'd been meaning to ask what to do, but I forgot.

clash-lib/src/Clash/Netlist/Util.hs
martijnbastiaan: Got it, forgot to remove that comment after our discussion yesterday.

clash-lib/src/Clash/Netlist.hs
```diff
@@ -673,16 +667,54 @@ mkFunApp dstId fun args tickDecls = do
...
+        --       between two top entities, defeating the ability to compile in
+        --       parallel.
+        --
+        --       One option is to split the normalization process into two
+        --       chunks: preprocessing (e.g., 'splitArguments') and actually
+        --       normalizing. This would ensure only minimal work is being done
+        --       serially.
+        --
+        --       The current workaround is to not rely on named arguments, using
+        --       positional ones instead when instantiating a top entity.
```
alex-mckenna, line 688: Is the current workaround preferable to this, or is it just that this option is worth a PR by itself?

tests/shouldwork/TopEntity/T1182B.hs
```diff
@@ -3,12 +3,14 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 
+-- {-# OPTIONS_GHC -ddump-splices #-}
+
```
alex-mckenna, line 7: Left in from debugging?

clash-lib/src/Clash/Netlist.hs
martijnbastiaan: It would be nice to use named arguments again, but it's going to take quite an annoying restructure of how we normalize. In other words, other PR seems like a good idea to me!

clash-lib/src/Clash/Driver.hs
```diff
@@ -258,20 +258,25 @@ generateHDL
...
-  topEntities2 = map (\(TopEntityT var annM tbM) -> TopEntityT var{varType=tvSubstWithTyEq (varType var)} annM tbM) topEntities1
+  topEntities1 =
+    -- Remove forall's used in type equality constraints
+      map (\(TopEntityT var annM tbM) ->
+              TopEntityT var{varType=tvSubstWithTyEq (varType var)} annM tbM)
+    -- Split clocks and resets to separate port
+    $ map (splitTopEntityT tcm bindingsMap)
+    -- Compile all top entities, or just top entities (+dependencies) of
+    -- designs passed in using '-main-is'.
+    $ maybe topEntities0 (uncurry (:)) mainTopEntity
```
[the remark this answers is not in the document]
martijnbastiaan: Agreed

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -212,6 +213,10 @@ data Identifier
       (Maybe Identifier)
       -- FIELD Parsed version of raw identifier. Will not be populated if this
       -- identifier was created with an unsafe function.
+      !CallStack
+      -- ^ Stores where this identifier was generated. Tracking is only enabled
+      -- is 'debugIsOn', otherwise this field will be populated by an empty
+      -- callstack.
```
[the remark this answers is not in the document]
martijnbastiaan: It's not used yet, but I've used it for debugging (by adding traces here/there). My plan is to eventually print these stack traces to HDL, but I'd like to do that in a different PR.

tests/shouldwork/TopEntity/T1182B.hs
martijnbastiaan: yes!

clash-lib/src/Clash/Netlist/Types.hs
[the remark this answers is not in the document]
alex-mckenna: What's the use, to find out why something has the identifier that it's rendered with?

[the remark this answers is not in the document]
martijnbastiaan: I don't quite get the edit, but the use is purely debugging. Identifiers are created using `makeBasic` or `make`. If you see two conflicting identifiers (or weird names), it should be pretty easy to find out where in our codebase they're created now. This should in turn be a pretty big hint as to why it generated conflicting names. i've used this to find one issue and figured I'd leave it in.

It's something that can be done by manually adding traces in the various functions in `Id` too, but the end goal is to render these callstacks to HDL so we can just instruct a client / issue reporter to enable `+debug`. (Ideally through a runtime flag, so we don't have to wait for a recompile.)

PR #1233, opened 2020-03-19
Add `Vector.unfoldr`

### LINKED ISSUE #1220: `unfoldr` for Vec
I'd like to propose the following addition to Clash.Sized.Vector:

```
unfoldr :: SNat n -> (s -> (a, s)) -> s -> Vec n a
```

I have an implementation in terms of `generate`, but it (safely) uses a partial function internally. A recursive, direct implementation of course wouldn't have this shortfall; which is why I think this functionality should be exposed by `Clash.Sized.Vector`.

```
unfoldr :: SNat n -> (s -> (a, s)) -> s -> Vec n a
unfoldr n f s0 = map unpack $ generate n f' (Nothing, s0)
  where
    f' (_, s) = let (x, s') = f s in (Just x, s')
    unpack (Just x, _) = x

unfoldrI :: (KnownNat n) => (s -> (a, s)) -> s -> Vec n a
unfoldrI = unfoldr SNat
```

If anyone's curious, I needed it to write the following function:

```
loadVec :: (KnownNat n) => [a] -> a -> Vec n a
loadVec xs x0 = unfoldrI uncons xs
  where
    uncons (x:xs) = (x, xs)
    uncons [] = (x0, [])
```

What do you think?

christiaanb:
Fixes #1220

### DIFF
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -45,6 +45,7 @@ module Clash.Sized.Vector
   , singleton
   , replicate, repeat
   , iterate, iterateI, generate, generateI
+  , unfoldr, unfoldrI
     -- *** Initialisation from a list
   , listToVecTH
     -- ** Concatenation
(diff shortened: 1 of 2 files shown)


clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1495,6 +1496,36 @@ iterateI f a = xs
...
+-- a simple use of 'unfoldrI':
+--
+-- >>> unfoldrI (\s -> (s,s-1)) 10 :: Vec 10 Int
+-- <10,9,8,7,6,5,4,3,2,1>
+unfoldrI :: KnownNat n => (s -> (a,s)) -> s -> Vec n a
+unfoldrI f s0 = map fst xs
+ where
+  xs = init (f s0 `Cons` ws)
+  ws = map (f . snd) (lazyV xs)
+{-# INLINE unfoldrI #-}
```
martijnbastiaan: Any reason for adding `INLINE` here? Is that something Clash specific?

christiaanb: It's not a primitive, so the more inlining GHC can do, the better.

PR #1235, opened 2020-03-19
Decouple evaluators

alex-mckenna:
In the implementation of the evaluator in Clash.Core.Evaluator, there
are details about the GHC frontend which leak through. As clash-lib
is supposed to be frontend-agnostic, this is undesirable to say the
least.

The evaluator has been moved to Clash.GHC, and the evaluator for
primitives moved to Clash.GHC.Evaluator.Primitive. This frees clash-lib
to only expose a more general API. This is quite limited now, but in
the new partial evaluator will be able to contain more combinators
for default implementations of evaluators. This is not the case now,
as an evaluator expressed with small-step semantics makes this harder
(it is harder to pick apart a structure from what is being evaluated,
as each evaluation function only represents a single step in
isolation).

PR #1239, opened 2020-03-20
Partial evaluator Models

alex-mckenna:
This PR introduces the model that new partial evaluator will use. Note that this model is currently not used, and the old evaluator is still in place. The new model differs from the old model by

  * representing normalisation as translations between similar languages instead of a process evaluated by an abstract machine. This allows common building blocks to be more easily abstracted away in the semantics as there is no "machine" being mutated

  * making arbitrary terms, WHNF terms and NF terms distinct in the type system. This also makes it easier to enforce restrictions (e.g. WHNF and NF terms must not have let bindings anywhere)

## TODO

  - [x] Wait for #1235 to be merged into master, then rebase
  - [x] Check that `asSubst` is implemented correctly
  - [x] Document `Clash.Core.Evaluator.Models`

### DIFF
--- a/clash-lib/src/Clash/Core/Evaluator/Models.hs
+++ b/clash-lib/src/Clash/Core/Evaluator/Models.hs
@@ -1,0 +1,307 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Clash.Core.Evaluator.Models where
+
+import Control.Concurrent.Supply (Supply)
+import Control.Monad ((>=>))
+import Control.Monad.State.Strict (State)
+import Control.DeepSeq (NFData(..))
+import Data.Bifunctor (first, second)
+import Data.Foldable (foldl')
+import Data.IntMap.Strict (IntMap)
+import qualified Data.IntMap.Strict as IntMap
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import GHC.Generics (Generic)
+
+import Clash.Core.DataCon (DataCon)
+import Clash.Core.FreeVars (localFVsOfTerms, tyFVsOfTypes)
+import Clash.Core.Literal (Literal)
+import Clash.Core.Subst (extendTvSubstList, mkSubst, substTm)
+import Clash.Core.Term (Term(..), Pat, PrimInfo, TickInfo, mkApps)
+import Clash.Core.TyCon (TyConMap)
+import Clash.Core.Type (Type)
+import Clash.Core.Var (Id, IdScope(..), TyVar)
+import Clash.Core.VarEnv
+  ( InScopeSet, extendInScopeSet, mkInScopeSet
+  , VarEnv, delVarEnv, extendVarEnv, lookupVarEnv
+  , unionVarSet, uniqAway
+  )
+
+import Clash.Driver.Types (BindingMap, Binding(..))
+
+-- | An evaluator contains the basic functions that define partial evaluation.
+-- This consists of:
+--
+--   * evaluateWhnf, which evaluates terms to WHNF
+--   * quote, which recursively evaluates sub-terms and eta-expands to NF
+--
+-- Users of an evalautor can choose whether they convert terms to WHNF or NF,
+-- as the 'AsTerm' class (which converts back to Term) has instances for both.
+--
+data Evaluator = Evaluator
+  { evaluateWhnf :: Term -> State Env Value
+  , quote        :: Value -> State Env Nf
+  }
+
+-- | Evaluate a term to normal form without stopping at WHNF.
+--
+evaluateNf :: Evaluator -> Term -> State Env Nf
+evaluateNf (Evaluator e q) = e >=> q
+
+-- Local bindings are not stored in a VarEnv, as we still want to be
+-- able to access the keys (VarEnv simply indexes by unique).
+
+type EnvTyMap = Map TyVar Type
+type EnvTmMap = Map Id (Either Term Value)
+
+-- | EnvGlobals refers to global bindings in scope during evalaution, i.e.
+-- other top-level functions used in a program.
+--
+type EnvGlobals = VarEnv (Binding (Either Term Value))
+
+-- | EnvPrimsIO refers to the result of IO operations performed in primitives
+-- during evalaution e.g. creating a new ByteArray in the GHC frontend.
+--
+type EnvPrimsIO = (IntMap Value, Int)
+
+-- | An Environment contains all in scope terms and types while evaluating.
+-- This consists of
+--
+--   * local types and terms from the term being evaluated
+--   * top-level definitions in scope (i.e. global bindings)
+--   * IO results of primitive operations
+--   * in scope uniques and a means of generating uniques
+--
+-- When running the partial evaluator at different times, it may be desirable
+-- to keep the globals (some of which may have been partially evaluated) and
+-- the results of primitives in IO for future runs.
+--
+data Env = Env
+  { envTypes   :: !EnvTyMap
+  , envLocals  :: !EnvTmMap
+  , envGlobals :: !EnvGlobals
+  , envPrimsIO :: !EnvPrimsIO
+  , envTcMap   :: !TyConMap
+  , envInScope :: !InScopeSet
+  , envSupply  :: Supply
+  }
+
+instance Show Env where
+  show env = show (envTypes env, envLocals env, envPrimsIO env)
+
+-- This instance cannot be derived with Generic, as Supply does not have an
+-- NFData instance. Instead, we use this instance which forces all fields
+-- which can be forced.
+--
+instance NFData Env where
+  rnf env = rnf (envTypes env)
+    `seq` rnf (envLocals env)
+    `seq` rnf (envGlobals env)
+    `seq` rnf (envPrimsIO env)
+    `seq` rnf (envTcMap env)
+    `seq` rnf (envInScope env)
+
+mkEnv :: BindingMap -> EnvPrimsIO -> TyConMap -> InScopeSet -> Supply -> Env
+mkEnv bs = Env mempty mempty gs
+ where
+  gs = fmap Left <$> bs
+
+-- | Add a new term / value to the local environment. The unique for the new
+-- element is guaranteed to be unique within the environment.
+--
+insertEnv :: Id -> Either Term Value -> Env -> Env
+insertEnv i etv env = env
+  { envLocals  = Map.insert i' etv (envLocals env)
+  , envInScope = extendInScopeSet (envInScope env) i'
+  }
+ where
+  i' = uniqAway (envInScope env) i
+
+insertAllEnv :: [(Id, Either Term Value)] -> Env -> Env
+insertAllEnv xs env = foldl' (flip $ uncurry insertEnv) env xs
+
+-- | Update a local or global pure binding in the environment. If the Id is not
+-- in the environment, the original environment is returned.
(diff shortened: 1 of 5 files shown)


alex-mckenna:
Checking `asSubst` was done over Slack, but if you could double check it @christiaanb that would be helpful

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+
+-- EnvGlobals refers to global bindings in scope during evalaution (i.e. other
+-- top-level functions used in a program).
+
+type EnvGlobals = VarEnv (Binding (Either Term Value))
+
+-- EnvPrimsIO refers to the result of IO operations performed in primitives
+-- during evalaution e.g. creating a new ByteArray in the GHC frontend.
+
+type EnvPrimsIO = (IntMap Value, Int)
```
martijnbastiaan: You can Haddock type aliases too AFAIK. Just use `-- |` as usual.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+instance Show Env where
+  show env = show (envTypes env, envLocals env, envPrimsIO env)
+
+instance NFData Env where
+  rnf env = rnf (envTypes env)
+    `seq` rnf (envLocals env)
+    `seq` rnf (envGlobals env)
+    `seq` rnf (envPrimsIO env)
```
martijnbastiaan: I think this either:

1. Needs a comment on why it breaks NFData's laws.
2. Should be removed and be auto-derived using Generics.

martijnbastiaan: Also: it looks like `Env` could benefit from strictness annotations on its fields.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+import Clash.Core.Type
+import Clash.Core.Var
+import Clash.Core.VarEnv
+import Clash.Driver.Types
+
+-- | An evaluator contains the basic functions that define partial evaluation.
+-- This consists of:
+--
+--   * evaluate, which evaluates terms to WHNF
+--   * quote, which recursively evaluates sub-terms and eta-expands to NF
```
martijnbastiaan: Why not: `evalWhnf` / `evalNf` (or with "To" interspersed)? Seems a lot more self-descriptive.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Clash.Core.Evaluator.Models where
```
martijnbastiaan: Could you fill me in on the name "Models"? I suspect that this is jargon I should know, but I don't..

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+--
+data Neutral a
+  = NeVar   (Var Term)
+  | NePrim  PrimInfo [Either Value Type]
+  | NeApp   (Neutral a) a
+  | NeTyApp (Neutral a) Type
+  | NeCase  a Type [(Pat, a)]
+  deriving (Show, Generic, NFData)
+
+-- A term which has been normalised to weak head normal form (WHNF). This has
```
martijnbastiaan: `-- |`

clash-lib/src/Clash/Core/Evaluator/Models.hs
alex-mckenna: You can't derive `NFData Env` generically, as `Supply` doesn't have an instance IIRC.

But yes, I'll look at which fields made it fail and mention it

clash-lib/src/Clash/Core/Evaluator/Models.hs
alex-mckenna: Normally in literature it would be eval / quote, although I see your point.

What about changing `evaluate` to `evaluateWhnf` and adding another function `evaluateNf x = evaluateWhnf x >>= quote` (I'm hesitant to change `quote` to `evaluateNf` because I would expect all evalaution functions to start from the same point - an unevaluated `Term`)

clash-lib/src/Clash/Core/Evaluator/Models.hs
alex-mckenna: Discussed offline: For now the module is called `Models` (because it models partial evaluation), but when the old evaluator is gone the module can be called `Types`

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+    LocalId -> env
+      { envLocals = Map.insert i' etv (envLocals env)
+      , envInScope = extendInScopeSet (envInScope env) i'
+      }
+
+    GlobalId ->
+      let b = Binding i' noSrcSpan Inline etv
+       in env
+            { envGlobals = extendVarEnv i' b (envGlobals env)
+            , envInScope = extendInScopeSet (envInScope env) i'
```
christiaanb: Global variable references are not free variables, but more akin to constants; i.e `\x[LocalId] -> x[GlobalId]` is the constant function, not identity. You therefor won't need to extend the `InScopeSet` for the regular reason: generating non-shadowing local identifiers.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+  asTerm (NeApp x y)      = App (asTerm x) (asTerm y)
+  asTerm (NeTyApp x ty)   = TyApp (asTerm x) ty
+  asTerm (NeCase x ty as) = Case (asTerm x) ty (second asTerm <$> as)
+
+instance AsTerm Value where
+  asTerm (VNeu n)         = asTerm n
+  asTerm (VData dc args)  = mkApps (Data dc) (first asTerm <$> args)
+  asTerm (VLit l)         = Literal l
+  asTerm (VPrim p args)   = mkApps (Prim p) (first asTerm <$> args)
+  asTerm (VLam x e env)   = substTm "asTerm.VLam" (asSubst env) (Lam x e)
```
christiaanb: Why not substitute the types of the environment, and let-bind the terms? instead substituting both.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,258 @@
...
+
+-- A term which has been normalised to weak head normal form (WHNF). This has
+-- no redexes at the head of the term, but subterms may still contain redexes.
+--
+data Value
+  = VNeu    (Neutral Value)
+  | VData   DataCon [Either Value Type]
+  | VLit    Literal
+  | VPrim   PrimInfo [Either Value Type]
+  | VLam    Id Term Env
```
christiaanb: Why do only lambdas have an environment? why not all (with the exception of Literals)?

clash-lib/src/Clash/Core/Evaluator/Models.hs
alex-mckenna: That's actually something I thought about yesterday, and one of the things me and @martijnbastiaan talked about on our call this morning (reaching the same conclusion) :+1:

One extra thing I've noticed though: The old evaluator did eagerly substitute terms when unwinding an `Apply` strack frame. Should these substitutions still occur, and if so, how do you decide between terms that should be in a `let` binding, or terms that should be substituted?

christiaanb: That's because we only have variables on the `Apply` stack frame, so no sharing is lost.

clash-lib/src/Clash/Core/Evaluator/Models.hs
martijnbastiaan: `seq`ing Supply should be enough I think!

alex-mckenna: Ah, so should I add

```haskell
 `seq` envInScope env `seq` envSupply env `seq ()
```

or something to the end?

martijnbastiaan: If you were planning on making it a strict data type anyway, just the `!` on the field should suffice. (Worth a comment though, might be confusing..)

I.e.,

```haskell

data Env = Env
  { envTypes   :: EnvTyMap
  , envLocals  :: EnvTmMap
  , envGlobals :: EnvGlobals
  , envPrimsIO :: EnvPrimsIO
  , envTcMap   :: TyConMap
  , envInScope :: InScopeSet
  , envSupply  :: !Supply
  }
```

(and for all the other fields you want to make strict, if any)

alex-mckenna: The semantics in my sandbox branch are pretty strict anyway, so I see no reason to not make the data type strict too (at least in the heaps). I'm not sure if there would be any real benefit to strictifying the Supply / InScopeSet / TyConMap, that would need some profiling to investigate

clash-lib/src/Clash/Core/Evaluator/Models.hs
alex-mckenna: Added `TODO` comment for explaining this when the semantics are in (for cross-referencing)

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,295 @@
...
+insertEnv scope i etv env =
+  case scope of
+    LocalId -> env
+      { envLocals = Map.insert i' etv (envLocals env)
+      , envInScope = extendInScopeSet (envInScope env) i'
+      }
+
+    GlobalId ->
+      let b = Binding i' noSrcSpan Inline etv
+       in env { envGlobals = extendVarEnv i' b (envGlobals env) }
```
christiaanb: doesn't the `envInScope` only contain local ids? if so, then this can accidentally overwrite global binders. So perhaps find a different way to generate a unique global id.

alex-mckenna: The `InScopeSet` is given as an argument to the evaluator (I think from normalize or rewrite state). So I'm genuinely not sure if it contains the global bindings or not...

christiaanb: Then it won't

christiaanb: But if this `insertEnv` is to be seen an "update action" for globals, then you can just `extendVarEnv` at `i`.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -0,0 +1,305 @@
...
+  , envInScope :: InScopeSet
+  , envSupply  :: Supply
+  }
+
+instance Show Env where
+  show env = show (envTypes env, envLocals env, envPrimsIO env)
+
+-- This instance cannot be derived with Generic, as neither InScopeSet
+-- or Supply have instances for NFData. Instead, we use this instance
+-- which forces all fields which can be forced.
```
martijnbastiaan: `Supply` has no `NFData` instance as it's an infinitely big construct, but maybe add it for `InScopeSet`?

alex-mckenna: As in add the instance to the type?

PR #1241, opened 2020-03-20
Ensure the register# primitive is inlined

### LINKED ISSUE #1240: `iterate` cannot be used to describe a reset value
The following fails to compile:

```haskell
module Test where

import Clash.Prelude

topEntity = register @System (iterate d1 not True)
```

giving the following error:

```
Test.hs:5:1: error:

    Clash.Netlist.BlackBox(188): Couldn't instantiate blackbox for Clash.Signal.Internal.register#. Verification procedure reported:

    Argument 5 should be literal, as blackbox used ~CONST[5], but was:

    Identifier "xs" Nothing

  |
5 | topEntity = register @System (iterate d1 not True)
  | ^^^^^^^^^
```

looking at the debug output we see:

```
  xs[8286623314361788996] :: Clash.Sized.Vector.Vec[8214565720323789569]
                               1
  = Clash.Sized.Vector.init @1
      @GHC.Types.Bool[3674937295934324744]
      (Clash.Sized.Vector.Cons[8214565720323789572]
         @(GHC.TypeNats.+[3674937295934325068] 1 1)
         @GHC.Types.Bool[3674937295934324744]
         @1
         (_CO_
            @(GHC.Prim.~#[3674937295934324842]
                GHC.Types.Nat[3674937295934325064]
                GHC.Types.Nat[3674937295934325064]
                (GHC.TypeNats.+[3674937295934325068] 1 1)
                (GHC.TypeNats.+[3674937295934325068] 1 1)))
         GHC.Types.True[3891110078048108589]
         (Clash.Sized.Vector.Cons[8214565720323789572] @1
            @GHC.Types.Bool[3674937295934324744]
            @0
            _CO_
            (GHC.Classes.not
               (Clash.Sized.Vector.head @0
                  @GHC.Types.Bool[3674937295934324744]
                  xs[8286623314361788996][LocalId]))
            (Clash.Sized.Vector.Nil[8214565720323789573] @0
               @GHC.Types.Bool[3674937295934324744]
               _CO_)))
```

i.e.

```haskell
let xs = init (Cons True (Cons (not (head xs)) Nil))
```

which is semantically a constant, because the second element of the list is never used, but not syntactically; and so Clash doesn't want to inline it.

christiaanb:
But use `GHC.Magic.inline` to make sure it doesn't want to share the value for powerup and reset.

~Fixes #1240~

Sadly, it only fixes the example in #1240, but not the general case. e.g. changing the example to:
```haskell
topEntity = register @System (iterate d10 not True)
```
still results in failure

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Signal.hs
+++ b/clash-prelude/src/Clash/Explicit/Signal.hs
@@ -282,6 +282,8 @@ import           Clash.Sized.Index              (Index)
 import qualified Clash.Sized.Vector
 import           Clash.XException               (NFDataX, deepErrorX, fromJustX)
 
+import GHC.Magic (inline)
+
 {- $setup
 >>> :set -XDataKinds -XTypeApplications -XFlexibleInstances -XMultiParamTypeClasses -XTypeFamilies
 >>> :set -fno-warn-deprecations
@@ -661,8 +663,8 @@ register
   -- will also be the initial value.
   -> Signal dom a
   -> Signal dom a
-register clk rst gen initial i =
-  register# clk rst gen initial initial i
+register clk rst gen initial =
+  register# clk rst gen (inline initial) (inline initial)
 {-# INLINE register #-}
 
 -- | Version of 'register' that only updates its content when its fourth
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
@christiaanb Given that this PR doesn't fundamentally fix anything, and it's a bit of a weird hack anyway - can we close this PR?

PR #1251, opened 2020-04-02
Add CoreTest to testsuite

alex-mckenna:
Similar to netlist tests added in ee5d000, core tests are now
available. These run the clash-ghc frontend as far as generating
bindings for top level definitions, giving access to the bindings
map and the TyConMap.

Core tests allow tests to be defined which inspect the result of
operating on Clash's core IR. This makes it easier to define tests
that inspect the behaviour of the partial evaluator, or
transformations in the normalizer.

### DIFF
--- a/testsuite/src/Test/Tasty/Clash.hs
+++ b/testsuite/src/Test/Tasty/Clash.hs
@@ -724,7 +724,7 @@ outputTest env targets extraClashArgs extraGhcArgs modName funcName path =
   testGroup testName
     [outputTest' env target extraClashArgs extraGhcArgs modName funcName path' | target <- targets]
 
-netlistTest'
+clashLibTest'
   :: FilePath
   -- ^ Work directory
   -> BuildTarget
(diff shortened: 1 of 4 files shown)


testsuite/src/Test/Tasty/Clash.hs
```diff
@@ -724,7 +724,7 @@ outputTest env targets extraClashArgs extraGhcArgs modName funcName path =
   testGroup testName
     [outputTest' env target extraClashArgs extraGhcArgs modName funcName path' | target <- targets]
 
-netlistTest'
+clashLibTest'
```
martijnbastiaan, line 727: This would still operate on Netlist right? Or are you planning on extending this?

alex-mckenna: Still operates on netlist, but can be reused for core. I changed the name because it really applies to any test that needs to run parts of `clash-lib`

testsuite/src/Test/Tasty/Clash.hs
```diff
@@ -754,7 +754,7 @@ netlistTest' env target extraGhcArgs modName funcName path =
              , "runghc"
              ]
              ++ langExts ++
-             [ "-DNETLISTTEST"
+             [ "-DCLASHLIBTEST"
```
martijnbastiaan, line 757: is this used somewhere?

alex-mckenna: Output tests define a CPP symbol `OUTPUTTEST`, so when netlist tests were created they defined a similar symbol. I'm not actually sure if either are used, I don't think I've ever seen 

```c
#if OUTPUTTEST
...
#endif
```

in any files

PR #1258, opened 2020-04-13
Infix use of `defaultSeqX` and `deepSeqX` righ-assoc

### LINKED ISSUE #1256: Space leak with memory simulators
I've asked around on the mailing list a couple weeks ago about tracking down a space leak in simulation, and I think now I have managed to show that there is some inherent problem with the simulation of memory primitives.

To set the scene, let's suppose I have the following Clash circuit:

```
module Circuit where

import Clash.Prelude
import Data.Word
import Control.Monad.State

type PC = Unsigned 12

logicBoard
    :: (HiddenClockResetEnable dom)
    => Signal dom ()
    -> Signal dom ()
logicBoard x = dummy <$> cpuOut
  where
    cpuOut = cpu cpuIn
    romRead = withStart 0 $ blockRam1 NoClearOnReset (SNat @10000) 0 (progAddr <$> cpuOut) (pure Nothing)
    cpuIn = CPUIn <$> romRead

withStart x = mux (register True $ pure False) (pure x)

data CPUIn = CPUIn
    { instr :: Word8
    }
    deriving (Generic, NFDataX)

data CPUOut = CPUOut
    { progAddr :: PC
    , dummy :: ()
    }
    deriving (Generic, NFDataX)

defaultOutput :: PC -> CPUOut
defaultOutput pc = CPUOut
    { progAddr = pc
    , dummy = ()
    }

cpu :: (HiddenClockResetEnable dom) => Signal dom CPUIn -> Signal dom CPUOut
cpu = mealy step 0
  where
    step s CPUIn{} = let (y, s') = (defaultOutput s, s + 1) in (s', y)
```

I then use the following function to simulate interactively:

```
module Sim where

import Clash.Prelude
import Control.Concurrent

driveIO_ :: ([i] -> [o]) -> i -> IO ((o -> IO i) -> IO ())
driveIO_ circuit input0 = do
    inChan <- newChan
    writeChan inChan input0

    ins <- getChanContents inChan
    outs <- newMVar $ circuit ins

    return $ \world -> do
        out <- modifyMVar outs $ \(out:outs) -> return (outs, out)
        input <- world out
        writeChan inChan input
```

Now, I will simulate my circuit at two levels: at `cpu` and at `logicBoard`. Note that in `cpu`, we are lazy in the fields of `CPUOut`, and `cpu` is lazy in the fields of `CPUIn`, so there is no "hidden" forcing of anything anywhere. 

```
module Main where

import Clash.Prelude
import Circuit
import Sim
import Control.Monad

main :: IO ()
main = do
    sim <- driveIO_ (simulate @System cpu) (CPUIn 0)
    replicateM_ 100000 $ sim $ \CPUOut{} -> return (CPUIn 0)
```

```
module Main where

import Clash.Prelude
import Circuit
import Sim
import Control.Monad

main :: IO ()
main = do
    sim <- driveIO_ (simulate @System logicBoard) ()
    replicateM_ 100000 $ sim $ \() -> return ()
```

I think that depending on how `simulate` is implemented, either both of these functions should leak, or neither should. However, testing with `+RTS -M2M`, i.e. with a heap limit of 2 Mb, I find that 

* `sim-cpu` works as expected, without leaking
* `sim-logicboard` quickly runs out of heap space
* `sim-logicboard`, with the type of the `instr` field of `CPUIn` changed to be strict, doesn't leak.

Note that if `CPUIn` has multiple fields, some connected to RAM or ROM, others connected to the outside world, then only making the RAM/ROM-connected fields strict is enough to fix the space leak.

All this is suggesting to me that the problem is with the implementation of the memory primitives.

christiaanb:
In the register and delay primitive we were writing:

```
q `defaultSeqX` q :- rest
```

which was parsed as:

```
(q `defaultSeqX` q) :- rest
```

because of the missing fixity declaration, our intention was the following however: !!!

```
q `defaultSeqX` (q :- rest)
```

so that the currunt value in the register was forced to WHNF (or NF) whenever the tail of the Signal was requested in order to prevent space-leaks. Without the fixity declaration on `defaultSeqX`, this space-leak prevention was of course not happening! And we were leaking space as demonstrated by
https://github.com/clash-lang/clash-compiler/issues/1256

Fixes #1256

### DIFF
--- a/changelog/2020-04-13T17:44:01+02:00_fix1256
+++ b/changelog/2020-04-13T17:44:01+02:00_fix1256
@@ -1,0 +1,2 @@
+FIXED: Memory leak in register primitives [#1256](https://github.com/clash-lang/clash-compiler/issues/1256)
+CHANGED: Infix use of `deepseqX` is now right-associative
--- a/clash-prelude/src/Clash/XException.hs
+++ b/clash-prelude/src/Clash/XException.hs
@@ -89,6 +89,7 @@ instance Exception XException
 defaultSeqX :: NFDataX a => a -> b -> b
 defaultSeqX = if fSuperStrict then deepseqX else seqX
 {-# INLINE defaultSeqX #-}
+infixr 0 `defaultSeqX`
 
 -- | Like 'error', but throwing an 'XException' instead of an 'ErrorCall'
 --
@@ -453,6 +454,7 @@ forceX x = x `deepseqX` x
 deepseqX :: NFDataX a => a -> b -> b
 deepseqX a b = rnfX a `seq` b
 {-# NOINLINE deepseqX #-}
+infixr 0 `deepseqX`
 
 -- | Reduce to weak head normal form
 --


martijnbastiaan:
Any clue to why `seq` and `deepseq` don't seem to have the same fixities?

https://hackage.haskell.org/package/ghc-prim-0.5.3/docs/src/GHC.Prim.html#seq
https://hackage.haskell.org/package/deepseq-1.4.4.0/docs/src/Control.DeepSeq.html#deepseq

christiaanb:
I have no idea... I was hoping my initial commit for `seqX` would explain its fixity annotation, but it doesn’t. It is the fixity you would want though... I think.

christiaanb:
@martijnbastiaan So there is actually a builtin 
```
infixr `seq` 0 
```
but haddock doesn't render it. You can see it though when you do `:i seq` in `ghci`. Now as to why `deepseq` doesn't have a fixity declaration, that might be an oversight, which I'm gonna ask around about.

christiaanb:
I've also asked on the Haskell libraries mailing list whether the omission of the fixity declaration for `deepseq` was intentional or unintended: https://mail.haskell.org/pipermail/libraries/2020-April/030361.html

basile-henry:
Any idea if this could help with some of the memory leaks we've seen on big simulations? :smile: 

That would be great, I might give it try.

martijnbastiaan:
@basile-henry Yes! Please do, let us know :)

PR #1260, opened 2020-04-15
Build with GHC 8.10

leonschoorl:


cabal.project
```diff
@@ -41,3 +41,19 @@ package clash-lib
...
+    type: git
+    location: https://github.com/haskell-hint/hint.git
+    tag: dbb878955528d9125a601773b25a3c45b9fcef06
+
+source-repository-package
+    type: git
+    location: https://github.com/leonschoorl/doctest.git
+    tag: 8aeee0a89b077d7dcb401fa1b380d9a51f94e634
+package doctest
+  tests: False
```
alex-mckenna, line 55: Can we leave comments by these saying (1) why they're needed and (2) when they can be removed
Otherwise I think there's a risk they might outstay their welcome

PR #1261, opened 2020-04-15
Build an 'Automaton' from a function over 'Signal's

### LINKED ISSUE #1255: `simulate` &c. should be `Stream a -> Stream b` ?
The various non-`N` family of `simulate` functions are supposed to be applied on infinite lists, and in response return infinite lists. Would it make sense to change their type to consume and return [`Stream`](https://hackage.haskell.org/package/Stream)s?

I'm asking because I'd like to write something like this:

```
driveIO :: ([i] -> [o]) -> i -> IO ((o -> IO i) -> IO ())
driveIO f input0 = do
    inChan <- newChan
    writeChan inChan input0

    ins <- getChanContents inChan
    outs <- newMVar $ f ins

    return $ \world -> do
        modifyMVar_ outs $ \(out:outs) -> do
            world out >>= writeChan inChan
            return outs
```

and this crucially depends on `outs` never becoming `[]`.

christiaanb:
Fixes #1255


gergoerdi:
What is still TODO on this?

christiaanb:
I guess a CHANGELOG entry

PR #1265, opened 2020-04-17
Fold identifier logic into 'Clash.Netlist.Id'. Fix #1230 

### LINKED ISSUE #1230: Clash produces name conflicts
~~According to HDL specs a basic identifier `x` and the "same" identifier but extended `\x\` should be considered separate. Some HDL tools seem to have missed the memo though. We should make sure Clash doesn't generate these apparent name conflicts.~~ `x` and `\x\` (basic vs. extended) are the same variable according to (System)Verilog spec and Quartus VHDL implementation, we should account for that during unique id generation.

edit: clarified which spec says what

martijnbastiaan:
```
    Fold identifier logic into 'Clash.Netlist.Id'. Fix #1230
    
    This megacommit moves all identifier (uniqueness) logic to a single
    module "Clash.Netlist.Id". "Identifier" is no longer a type alias for
    Text, but a data type allowing cheap new unique identifier generation.
    
    It fixes a number of (unreported) bugs:
    
     * #1230: Clash no longer produces "name conflicts" between basic and
       extended identifiers. I.e., \x\ and x are now considered the same
       variable in VHDL (likewise for other HDLs). Although the VHDL spec
       considers them distinct variables, some HDL tools - like Quartus -
       don't.
    
     * Capitalization of Haskell names are now preserved in VHDL. Note that
       VHDL is a case insensitive languages, so there are measures in place
       to prevent Clash from generating both Foo and fOO. This used to be
       handled by promoting every capitalized identifier to an extended one,
       and wasn't handled for basic ones.
    
     * Names generated for testbenches can no longer cause collisions with
       previously generated entities.
    
     * Names generated for components can no longer cause collisions with
       user specified top entity names.
    
     * For (System)Verilog, variables can no longer cause collisions with
       (to be) generated entity names.
    
     * HO blackboxes can no longer cause collisions with identifiers
       declared in their surrounding architecture block.

```

--------------------------------------------

TODO:
 - [x] Fully expand top entity annotations (even if they don't exist!) _before_ running `mkInput`, `mkOutput`, `mkTopInput`, `mkTopOutput`. This will remove half of the (duplicated) code base for these functions, and will solve the test failure in `PortNamesWithVector`.
 - [x] Check docs for legacy comments
 - [x] Change misleading documentation/type of second ~~NetDecl~~ InstDecl field
 - [x] Think about whether I'm happy with the "abuse" of the IdentifierSet in the VHDL/SystemVerilog backend
 - [x] ~~Change `make#`, `deepen#`, etc. to work on lenses, allowing for a cheap way of using multiple identifiersets in a single monadic context. Would probably solve previous TODO.~~ meh
 - [x] Add tests for (unreported) issues
 - [x] ~~Test on production codebase~~
 - [x] Check whether use of `Id.unsafeMake` usage is actually safe
 - [x] Current code doesn't clean up between components, therefore generating more and more "ugly" identifier names as it goes on 
 - [x] Rebase on master
 - [x] In cases where this PR drops `Identifier` in favor of `Text`, it should be replaced by `IdentifierText` instead. (Unless it's really not an identifier)
 - [x] Add `Id.toLText`
 - [x] Remove `Clash.Primitives.DSL.newName` (it's the same as `Id.makeBasic`)

### DIFF
--- a/clash-lib/src/Clash/Backend/SystemVerilog.hs
+++ b/clash-lib/src/Clash/Backend/SystemVerilog.hs
@@ -19,7 +19,7 @@ module Clash.Backend.SystemVerilog (SystemVerilogState) where
 import qualified Control.Applicative                  as A
 import           Control.Lens                         hiding (Indexed)
 import           Control.Monad                        (forM,liftM,zipWithM)
-import           Control.Monad.State                  (State, StateT)
+import           Control.Monad.State                  (State)
 import           Data.Bits                            (Bits, testBit)
 import           Data.HashMap.Lazy                    (HashMap)
 import qualified Data.HashMap.Lazy                    as HashMap
@@ -179,13 +179,12 @@ genSystemVerilog
   -> Component
   -> SystemVerilogM ((String, Doc), [(String, Doc)])
 genSystemVerilog _ sp seen c = do
-    -- Don't have type names conflict with module names
-    Mon $ idSeen .= seen
-
-    -- Don't have type names conflict with type names generated in previous
-    -- genVHDL
-    typNames <- HashMap.elems <$> use nameCache
-    mapM_ (Id.addRaw . Id.toText) typNames
+    -- Don't have type names conflict with module names or with previously
+    -- generated type names.
+    --
+    -- TODO: Collect all type names up front, to prevent relatively costly union.
+    -- TODO: Investigate whether type names / signal names collide in the first place
+    Mon $ idSeen %= Id.union seen
 
     Mon $ setSrcSpan sp
     v    <- verilog
@@ -593,7 +592,7 @@ tyName t@(Product nm _ _)      = do
   tN <- normaliseType t
   PP.pretty =<< Mon (makeCached tN nameCache prodName)
  where
-  prodName :: StateT SystemVerilogState Identity Identifier
+  prodName :: State SystemVerilogState Identifier
   prodName = Id.makeBasicOr (last (TextS.splitOn "." nm)) "product"
 
 tyName t@(SP _ _) = "logic_vector_" <> int (typeSize t)
--- a/clash-lib/src/Clash/Backend/SystemVerilog.hs
+++ b/clash-lib/src/Clash/Backend/SystemVerilog.hs
@@ -19,11 +19,10 @@ module Clash.Backend.SystemVerilog (SystemVerilogState) where
 import qualified Control.Applicative                  as A
 import           Control.Lens                         hiding (Indexed)
 import           Control.Monad                        (forM,liftM,zipWithM)
-import           Control.Monad.State                  (State)
+import           Control.Monad.State                  (State, StateT)
 import           Data.Bits                            (Bits, testBit)
 import           Data.HashMap.Lazy                    (HashMap)
 import qualified Data.HashMap.Lazy                    as HashMap
-import qualified Data.HashMap.Strict                  as HashMapS
 import           Data.HashSet                         (HashSet)
 import qualified Data.HashSet                         as HashSet
 import           Data.List                            (nub, nubBy)
@@ -36,6 +35,7 @@ import           Data.Semigroup.Monad
 import qualified Data.Text.Lazy                       as Text
 import qualified Data.Text                            as TextS
 import           Data.Text.Prettyprint.Doc.Extra
+import qualified Data.Text.Prettyprint.Doc.Extra      as PP
 import qualified System.FilePath
 
 import           Clash.Annotations.Primitive          (HDL (..))
@@ -53,9 +53,9 @@ import           Clash.Backend.Verilog
 import           Clash.Netlist.BlackBox.Types         (HdlSyn (..))
 import           Clash.Netlist.BlackBox.Util
   (extractLiterals, renderBlackBox, renderFilePath)
-import           Clash.Netlist.Id                     (IdType (..), mkBasicId')
+import qualified Clash.Netlist.Id                     as Id
 import           Clash.Netlist.Types                  hiding (_intWidth, intWidth)
-import           Clash.Netlist.Util                   hiding (mkIdentifier, extendIdentifier)
+import           Clash.Netlist.Util
 import           Clash.Signal.Internal                (ActiveEdge (..))
 import           Clash.Util
   (SrcSpan, noSrcSpan, curLoc, makeCached, first, on, indexNote)
@@ -65,11 +65,10 @@ import           Clash.Util.Graph                     (reverseTopSort)
 data SystemVerilogState =
   SystemVerilogState
     { _tyCache   :: HashSet HWType -- ^ Previously encountered  HWTypes
-    , _tySeen    :: HashMapS.HashMap Identifier Word -- ^ Product type counter
-    , _nameCache :: HashMap HWType Doc -- ^ Cache for previously generated product type names
+    , _nameCache :: HashMap HWType Identifier -- ^ Cache for previously generated product type names
     , _genDepth  :: Int -- ^ Depth of current generative block
-    , _modNm     :: Identifier
-    , _idSeen    :: HashMapS.HashMap Identifier Word
+    , _modNm     :: ModName
+    , _idSeen    :: IdentifierSet
     , _oports    :: [Identifier]
     , _srcSpan   :: SrcSpan
     , _includes  :: [(String,Doc)]
@@ -84,17 +83,35 @@ data SystemVerilogState =
     -- ^ Are we in the context of generating the @_types@  package?
     , _intWidth  :: Int -- ^ Int/Word/Integer bit-width
     , _hdlsyn    :: HdlSyn
-    , _escapedIds :: Bool
     , _undefValue :: Maybe (Maybe Int)
     , _aggressiveXOptBB_ :: AggressiveXOptBB
     }
 
 makeLenses ''SystemVerilogState
 
+instance HasIdentifierSet SystemVerilogState where
+  identifierSet = idSeen
+
 instance Backend SystemVerilogState where
-  initBackend     = SystemVerilogState HashSet.empty HashMapS.empty HashMap.empty
-                                       0 "" HashMapS.empty [] noSrcSpan [] [] []
-                                       [] [] False
+  initBackend w hdlsyn_ esc undefVal xOpt = SystemVerilogState {
+      _tyCache=HashSet.empty
+    , _nameCache=HashMap.empty
+    , _genDepth=0
+    , _modNm=""
+    , _idSeen=Id.emptyIdentifierSet esc SystemVerilog
+    , _oports=[]
+    , _srcSpan=noSrcSpan
+    , _includes=[]
+    , _imports=[]
+    , _libraries=[]
+    , _dataFiles=[]
+    , _memoryDataFiles=[]
+    , _tyPkgCtx=False
+    , _intWidth=w
+    , _hdlsyn=hdlsyn_
+    , _undefValue=undefVal
+    , _aggressiveXOptBB_=xOpt
+    }
   hdlKind         = const SystemVerilog
   primDirs        = const $ do root <- primsRoot
                                return [ root System.FilePath.</> "common"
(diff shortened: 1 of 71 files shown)


christiaanb:
So issues with the commit message:

* #1230 is a Quartus VHDL bug, and a Clash Verilog bug
* Clash always preserved capitalization, but used extended identifiers to do so.

martijnbastiaan:
> #1230 is a Quartus VHDL bug, and a Clash Verilog bug

I don't think I claimed anything, but I'll clarify nonetheless.

> Clash always preserved capitalization, but used extended identifiers to do so.

It now does it across the board. Basic identifiers will also preserve capitalization.

clash-lib/src/Clash/Driver.hs
```diff
@@ -236,57 +242,62 @@ generateHDL
...
         _ -> case annM of
           Just ann -> case hdlKind (undefined :: backend) of
-            VHDL -> (t_name ann, ComponentPrefix Nothing Nothing)
+            VHDL -> (Data.Text.pack (t_name ann), Nothing)
             -- Prefix other with 't_name'
-            _    -> (t_name ann, ComponentPrefix Nothing (Just (Data.Text.pack (t_name ann))))
-          _ -> (modName1, ComponentPrefix Nothing Nothing)
+            _ -> (Data.Text.pack (t_name ann), Just modName)
+          _ -> (modName1, Nothing)
+      modNameS  = Data.Text.unpack modName
```
christiaanb, line 289: Why not `pack` (variations on) `t_name ann` once, and make this:
```
modNameT = Data.Text.pack modName
```

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -145,53 +145,55 @@ extractLiterals = map (\case (e,_,_) -> e)
 setSym
   :: forall m
    . Monad m
-  => (IdType -> Identifier -> m Identifier)
+  => (IdentifierType -> Data.Text.Text -> m Identifier)
```
christiaanb, line 148: Perhaps make a type alias:
```
type IdentifierText = Text
```
and use that where the old code used to use `Identifier`

clash-lib/src/Clash/Netlist/BlackBox.hs
```diff
@@ -118,7 +119,9 @@ mkBlackBoxContext
   -> NetlistMonad (BlackBoxContext,[Declaration])
 mkBlackBoxContext bbName resId args@(lefts -> termArgs) = do
     -- Make context inputs
-    let resNm = nameOcc (varName resId)
+    -- 'unsafeMake' is safe here: 'mkUniqueArguments' made sure these names
```
christiaanb, line 122: This comment will be removed?

clash-lib/src/Clash/Netlist/BlackBox.hs
```diff
@@ -1034,24 +1043,29 @@ mkFunInput resId e =
...
+      tcm <- Lens.use tcCache
+      resTy <- unsafeCoreTypeToHWTypeM' $(curLoc) (termType tcm app)
       let (tmArgs,tyArgs) = partitionEithers args
       if null tyArgs
         then
           withTicks ticks $ \tickDecls -> do
-            appDecls <- mkFunApp "~RESULT" fun tmArgs tickDecls
-            nm <- mkUniqueIdentifier Basic "block"
-            return (Right ((nm,appDecls),Wire))
+            resNm <- Id.make "resultbar"
```
christiaanb, line 1052: resultbar?

clash-lib/src/Clash/Netlist/Id.hs
```diff
@@ -7,111 +7,406 @@
...
+  -> IdentifierSet
+emptyIdentifierSet esc hdl = makeSet esc hdl []
+
+-- | Union of two identifier sets. Errors if given sets have been made with
+-- different options enabled.
+union :: HasCallStack => IdentifierSet -> IdentifierSet -> IdentifierSet
+union is1@(IdentifierSet esc1 hdl1 _ _) is2@(IdentifierSet esc2 hdl2 _ _)
+  | esc1 /= esc2 = error $ "Internal error: esc1 /= esc2, " <> show (esc1, esc2)
+  | hdl1 /= hdl2 = error $ "Internal error: hdl1 /= hdl2, " <> show (hdl1, hdl2)
+  | otherwise = makeSet esc1 hdl1 (toList is1 ++ toList is2)
```
christiaanb, line 100: I don't think I'm happy with the conversion through list, especially when these `IdentifierSet`s get large. `HashMap` also has a `foldl'`

clash-lib/src/Clash/Netlist/Id.hs
```diff
@@ -7,111 +7,406 @@
...
+  -> IdentifierSet
+makeSet esc hdl ids = IdentifierSet esc hdl fresh store
+ where
+  fresh = List.foldl' updateFreshCache# mempty ids
+  store = HashSet.fromList ids
+
+toList :: IdentifierSet -> [Identifier]
+toList (IdentifierSet _ _ _ idStore) = HashSet.toList idStore
+
+lookupFreshCache# :: FreshCache -> Identifier -> Maybe Word
```
christiaanb, line 119: Some top-level descriptions of these `xyz#` functions would be appreciated

martijnbastiaan:
Answered on Slack, but because I was almost done typing the reply.. for prosperity!:

> How are you handing `-fclash-no-escaped-identifiers` now that `mkIdentifier` and `extendIdentifier` have been moved out of the `Backend` class?

It's specified when constructing an `IdentifierSet`:

https://github.com/clash-lang/clash-compiler/blob/a60200031841450da0a6f74571649748ee453ac9/clash-lib/src/Clash/Netlist/Id.hs#L85-L92

And then it's used in `make#` to force creating basic identifiers:

https://github.com/clash-lang/clash-compiler/blob/a60200031841450da0a6f74571649748ee453ac9/clash-lib/src/Clash/Netlist/Id.hs#L179

clash-lib/src/Clash/Netlist/BlackBox.hs
martijnbastiaan: Woops... debugging attempt!

clash-lib/src/Clash/Netlist/Id.hs
martijnbastiaan: Me neither, but this function should be removed soon(tm). Implementing parallel compile should get rid of the need for it.

clash-lib/src/Clash/Netlist/Id.hs
martijnbastiaan: yes!

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -77,25 +86,145 @@ data TopEntityT = TopEntityT
...
+-- HDL.
+--
+-- __NB__: use the functions in Clash.Netlist.Id. Don't use these constructors
+-- directly.
+data Identifier
+  -- | Unparsed identifier. Used for things such as port names, which should
+  -- appear in the HDL exactly as the user specified.
+  = RawIdentifier
+      !Text
+      -- ^ An identifier exactly as given by the user
```
leonschoorl, line 192: The haddock that comes with 8.4 doesn't commenting on bare fields.
You can use this hack if you want: https://github.com/clash-lang/clash-compiler/blob/master/clash-lib/src/Clash/Netlist/BlackBox/Types.hs#L12-L15

clash-lib/src/Clash/Netlist/Id.hs
martijnbastiaan: Functions should still be removed, but `union` is implemented more efficiently now.

clash-lib/src/Clash/Backend/SystemVerilog.hs
```diff
@@ -176,68 +166,31 @@ instance Backend SystemVerilogState where
...
+  -> SystemVerilogM ((String, Doc), [(String, Doc)])
+genSystemVerilog _ sp seen c = do
     -- Don't have type names conflict with module names
-    Mon $ tySeen %= HashMapS.unionWith max seen
+    Mon $ idSeen .= seen
+
+    -- Don't have type names conflict with type names generated in previous
+    -- genVHDL
+    typNames <- HashMap.elems <$> use nameCache
+    mapM_ (Id.addRaw . Id.toText) typNames
```
christiaanb, line 188: I don't get this part... we're still in the same `SystemVerilogM` session, so why wouldn't previous calls to `genSystemVerilog` have already added the type names to the `identifierSet`?

martijnbastiaan: The goal of this PR was to move all logic concerning unique name generation to `Clash.Netlist.Id`. To that end, type names are generated using the `IdentifierSet` (stored at `idSeen` for `SystemVerilogState`). `nameCache` is just used to create a mapping between `HWType` and an `Identifier`. This has a side effect: when `idSeen` is reset duplicate names might be generated, so I re-add them here. Alternatively, `Mon $ idSeen .= seen` (just a couple of lines above the quote code) could do a union instead.

martijnbastiaan: Having said that.. I feel this is the "wrong" approach in the first place. Having this dependency between `genHDL` runs means we can't have parallel compilation of top entities (even after getting rid of the dependence of manifest files https://github.com/clash-lang/clash-compiler/pull/1227!). Instead, `IdentifierSet` should distinguish between global and local ids. Then, global ids should only be generated in their own namespace, e.g. prefixed by the module name. This would get rid of this^ quirk too.

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -81,23 +95,170 @@ data TopEntityT = TopEntityT
...
-  NetlistMonad { runNetlist :: StateT NetlistState (ReaderT NetlistEnv IO) a }
+  NetlistMonad { runNetlist :: Strict.StateT NetlistState (ReaderT NetlistEnv IO) a }
   deriving newtype (Functor, Monad, Applicative, MonadReader NetlistEnv,
-                    MonadState NetlistState, MonadIO, MonadFail)
+                    Strict.MonadState NetlistState, Strict.MonadIO, MonadFail)
 
 type HWMap = HashMap Type (Either String FilteredHWType)
 
+-- | See 'is_freshCache'
+type FreshCache = HashMap Text (IntMap Word)
```
martijnbastiaan, line 132: This can be simplified: we can just store a single extension here and store additional extensions in the basename of the identifier. This would make the fresh cache just `HashMap Text Word`.

clash-lib/src/Clash/Netlist/Id.hs
```diff
@@ -90,35 +90,40 @@ import qualified Clash.Netlist.Id.Common as Common
...
   | hdl1 /= hdl2 = error $ "Internal error: hdl1 /= hdl2, " <> show (hdl1, hdl2)
-  | otherwise = makeSet esc1 hdl1 (toList is1 ++ toList is2)
+  | lw1 /= lw2 = error $ "Internal error: lw1 /= lw2 , " <> show (lw1, lw2)
+  | otherwise = makeSet esc1 lw1 hdl1 (toList is1 ++ toList is2)
 
 -- | Make a identifier set filled with given identifiers
 makeSet
   :: Bool
   -- ^ Allow escaped identifiers?
+  -> Bool
```
christiaanb, line 100: Perhaps use a custom enum here:
```haskell
data PreserveCase
  = PreserveCase
  | ToLower
```

And of course propagate this throughout the rest of the design. Just so that we have fewer "random" Booleans in our codebase.

clash-lib/src/Clash/Netlist/Types.hs
martijnbastiaan: Welp, this cannot be simplified. It's really hard to set it up such that you recognize that `foo` + extension `0` is the same as  `foo_0`, without blowing up the complexity class.

PR #1278, opened 2020-04-23
When looking for port name annotations in a closed type family

### LINKED ISSUE #1271: Port name annotations in type familiy applications aren't handled correctly
I'd like to use Barbie-types for IO. Unfortunately, it seems the types generated by `barbies-th` don't work, because the annotations are on the signal contents.

To illustrate, first of all here is a complete source file:

```
{-# LANGUAGE TypeFamilies, DataKinds #-}
module Hello where

import Clash.Prelude
import Clash.Annotations.TH
import Data.Kind (Type)

-- Setting the scene: mini-Barbies
data Cover = Bare | Covered
type family Wear (c :: Cover) (f :: Type -> Type) a :: Type where
    Wear Bare f a = a
    Wear Covered f a = f a

-- This is what `barbies-th` emits...
data Dressed c f = Dressed
    { fd1' :: Wear c f ("FIELD1" ::: Bool)
    , fd2' :: Wear c f ("FIELD2" ::: Bool)
    }

-- This one works
data Inlined f = Inlined
    { fd1 :: f ("FIELD1" ::: Bool)
    , fd2 :: f ("FIELD2" ::: Bool)
    }

topEntity
    :: ( "FOO" ::: Dressed Covered (Signal System)
       , "BAR" ::: Inlined (Signal System)
       )
topEntity =
    ( Dressed (pure True) (pure False)
    , Inlined (pure True) (pure False)
    )

makeTopEntity 'topEntity
```

And now some explanation: In `Dressed`, I have replicated exactly the data type that `barbies-th` emits for the simple non-HKD declaration of

```
data Dressed0 = Dressed0{ bar2 :: "FIELD1" ::: Bool, baz2 :: "FIELD2" ::: Bool }
```

Since my goal would be to use `barbies-th` instead of hand-writing these HKD definitions, I'd like to be able to use `Dressed Covered (Signal System)` as a `topEntity` IO type.  

Because I have absolutely no idea how the TH annotation system works (and frankly, I am too scared to look), I am hoping that there might be a way to support this. 

I don't see why `fd1 :: Signal dom ("FIELD1" ::: Bool)` and `fd1' :: Wear Covered (Signal dom) ("FIELD1" ::: Bool)` would behave differently, when the latter reduces to the former.

gergoerdi:
application, do proper unification to find and apply the right type family clause
(fixes #1271)

### DIFF
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -342,6 +342,7 @@ Library
                       time                      >= 1.8     && < 1.11,
                       transformers              >= 0.5.2.0 && < 0.6,
                       type-errors               >= 0.2.0.0 && < 0.3,
+                      uniplate                  >= 1.6.12  && < 1.7,
                       vector                    >= 0.11    && < 1.0
 
 test-suite doctests
--- a/clash-prelude/src/Clash/Annotations/TH.hs
+++ b/clash-prelude/src/Clash/Annotations/TH.hs
@@ -73,6 +73,7 @@ import qualified Data.Map                      as Map
 #if !(MIN_VERSION_base(4,11,0))
 import           Data.Semigroup                as Semigroup
 #endif
+import           Data.Maybe                     ( catMaybes )
 import           Language.Haskell.TH
 
 import           Data.Functor.Foldable          ( para )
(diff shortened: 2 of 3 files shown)


martijnbastiaan:
Could you add a test exploiting this new behavior in [clash-prelude/tests/Clash/Tests/TopEntityGeneration.hs](https://github.com/clash-lang/clash-compiler/blob/4f685db373b88b5bc4f179b769fc9fd04a22a553/clash-prelude/tests/Clash/Tests/TopEntityGeneration.hs)?

martijnbastiaan:
You can run the tests using:

```bash
cabal new-run clash-prelude:unittests
```

devving with ghcid should work fine too:

```
ghcid -c "cabal new-repl clash-prelude:unittests"
```

gergoerdi:
This new version works with Barbies as well!

gergoerdi:
> Could you add a test exploiting this new behavior in [clash-prelude/tests/Clash/Tests/TopEntityGeneration.hs](https://github.com/clash-lang/clash-compiler/blob/4f685db373b88b5bc4f179b769fc9fd04a22a553/clash-prelude/tests/Clash/Tests/TopEntityGeneration.hs)?

Test is included in the latest version.

PR #1288, opened 2020-04-24
Partial Evaluator Semantics

alex-mckenna:
The second PR for #950 

This PR defines the common semantics of partial-evaluation in `clash-lib`, and uses these to partially define GHC specific semantics in `clash-ghc`. Primitive operations are not evaluated currently: adding these is a large task unto itself and is left for later PRs.

There are still some issues to iron out before this can be merged. Currently these are:

  - [x] Eta-expand data and primitives when evaluating
  - [x] Document the semantics and changes to models
  - [x] Move common parts of data and primitive eval from `clash-ghc` to `clash-lib`
  - [x] Disable tests which fail on GHC 8.4 (disable for under 8.6.5 ONLY)
  - [ ] Continue expanding the testsuite in `tests/shouldwork/PartialEvaluation`


clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,372 @@
...
+  :: (Term -> State Env Value)
+  -> DataCon
+  -> State Env Value
+evaluateDataWith eval dc
+  | isFullyApplied (dcType dc) [] = pure (VData dc [])
+  | otherwise = etaExpand (Data dc) >>= eval
+
+-- | Default implementation for evaluating primitive operations. If the primop
+-- is not fully applied, then it is eta-expanded and its eta-expanded form is
+-- delta-reduced to obtain a result.
```
leonschoorl, line 120: Everywhere else you talk about evaluating, not delta-reducing.

tests/shouldwork/PartialEvaluation/EtaExpansion.hs
```diff
@@ -0,0 +1,145 @@
...
+mainCommon hdl = do
+  entities <- runToCoreStage hdl id testPath
+
+  -- Eta Expansion of Data Constructors
+  let data1 = findBinding "EtaExpansion.etaReducedData" entities
+      data2 = findBinding "EtaExpansion.etaExpandedData" entities
+
+  if aeqTerm (asTerm data1) (asTerm data2)
+     then pure ()
+     else error ("Not alpha equivalent: " <> show data1 <> "\n\n" <> show data2)
```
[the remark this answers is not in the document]
leonschoorl: same in KnownCase.hs

testsuite/clash-testsuite.cabal
```diff
@@ -87,3 +96,8 @@ executable clash-testsuite
   build-depends:
     clash-testsuite
 
+executable partial-eval
+  import: basic-config
+  main-is: partial-eval.hs
+  build-depends: concurrent-supply
+
```
leonschoorl, line 103: Is this something we should (optionally) be running on the CI?

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -109,90 +183,126 @@ mkEnv bs = Env mempty mempty gs
...
+-- | Add a new type to the environment. The unique for the new
+-- element is guaranteed to be unique within the environment.
+--
+insertType :: TyVar -> Type -> Env -> Env
+insertType i ty env = env
+  { envTypes = Map.insert i' ty (envTypes env)
+  , envInScope = extendInScopeSet (envInScope env) i'
+  }
+ where
+  i' = uniqAway (envInScope env) i
```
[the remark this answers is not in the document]
leonschoorl: already fixed

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -28,10 +30,15 @@ import Clash.Rewrite.Util
...
-normalization = rmDeadcode >-> constantPropagation >-> rmUnusedExpr >-!-> anf >-!-> rmDeadcode >->
-                bindConst >-> letTL >-> evalConst >-!-> cse >-!-> cleanup >->
-                xOptim >-> rmDeadcode >->
-                cleanup >-> recLetRec >-> splitArgs
+normalization =
+  rmDeadcode >-> constantPropagation >-> rmUnusedExpr >-!-> anf >-!-> rmDeadcode >->
+  bindConst >-> letTL >-!->
+#if !EXPERIMENTAL_EVALUATOR
+  evalConst >->
+#endif
```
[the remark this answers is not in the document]
alex-mckenna: No, I had removed the whole thing and then in a later commit / amend put it back manually

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -67,6 +124,23 @@ type EnvGlobals = VarEnv (Binding (Either Term Value))
...
+evaluated with fewer traversals of the term. Consider a small-step reduction
+machine: on finding a redex it would apply beta-reduction, and then have to
+re-traverse the resulting term from the point where it changed in case there
+are new redexes to reduce.
+
+In contrast, a big-step machine allows fewer traversals. Constructors of the
+Value type do not contain redexes that need removing to obtain a WHNF term.
+This means when it is possible to create a value (by evaluating a term and all
+subterms to values), the resulting term does not need traversing again. Further
+evaluation to NF is achieved by introducting another function, quote, which
```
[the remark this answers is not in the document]
alex-mckenna: Yes, the function changed name but the docs didn't

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,372 @@
...
+     -> either (State.withState (deleteLocal i) . eval) pure etv
+
+     |  Just b <- lookupGlobal i env
+     ,  canInline (bindingSpec b)
+     -> either (State.withState (deleteGlobal i) . eval) pure (bindingTerm b)
+
+     |  otherwise
+     -> pure (VNeu (NeVar i))
+ where
+  canInline spec = spec == Inline || spec == Inlinable
```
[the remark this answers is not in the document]
alex-mckenna: Yes! I actually changed this just before the call today, but I haven't commited

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,372 @@
...
+  (f, args) = collectArgs term
+  evalArg   = bitraverse eval pure
+  evalArgs  = traverse evalArg
+
+-- | Default implementation for evaluating a letrec expression. This adds all
+-- bindings to the heap without eagerly evaluating them, then evaluates the
+-- body of the expression.
+--
+-- Bindings are not evaluated eagerly, as they may not be needed depending how
+-- the body of the expression evaluates (i.e. bindings may be in an elided
```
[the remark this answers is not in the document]
alex-mckenna: Elide, as in omitted. Eliminate isn't _quite_ the word I'd want to use for this because it means something in terms of reduction (whereas this is more of a pruning)

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+         S# _  -> dcTag c == 1
+         Jp# _ -> dcTag c == 2
+         Jn# _ -> dcTag c == 3
+
+    |  NaturalLiteral i <- l
+    -> case i of
+         S# _  -> dcTag c == 1 && i >= 0
+         Jp# _ -> dcTag c == 2
+         Jn# _ -> False
+
```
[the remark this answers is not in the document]
alex-mckenna: I'm not entirely sure on this one. The old evaluator didn't take these types into account here, but I suspect that something like, e.g. `Int`, would be turned into `Literal (IntLiteral n)` by the core2core pass, but maybe `Integer` and `Natural` might not be?

@christiaanb, do you know anything that might help?

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,372 @@
...
+
+-- | Default implementation for evaluating a literal.
+-- This simply wraps the literal up into a Value.
+--
+evaluateLiteral :: Literal -> State Env Value
+evaluateLiteral = pure . VLit
+
+-- | Default implementaion for evaluating data constructors. If the constructor
+-- is not fully applied, then it is eta-expanded and its eta-expanded form is
+-- evaluated to obtain a result.
```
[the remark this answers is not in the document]
alex-mckenna: A nullary data constructor / primitive is fully applied, otherwise it isn't. I should probably update the documentation with a note to make clear these are kind of a degenerate case though (compared to the evaluation of data / primtives in `evaluateAppWith`

testsuite/clash-testsuite.cabal
alex-mckenna: No, this program just runs up to the evaluator and then spits out the unevaluated and evaluated versions of each binding in a module. In terms of CI, the core tests are very similar as they run entities from the module under test through the evaluator to compare. This is more for debugging the evaluator.

testsuite/Main.hs
```diff
@@ -293,31 +338,54 @@ runClashTest = defaultMain $ clashTestRoot
...
+      , clashTestGroup "BitVector"
+        [ runTest "Box" def
+        , runTest "BoxGrow" def
+        , runTest "CLZ" def
+        , runTest "RePack" def{hdlSim=False}
+        , runTest "ReduceZero" def
+        , runTest "ReduceOne" def
+        , runTest "ExtendingNumZero" def
+        , runTest "AppendZero" def
+        ]
```
leonschoorl, line 356: Most of these test are also defined above.
This looks like the start of a maintenance nightmare to me.
I think it would be easier to maintain when every testgroup and every test are only at one place in the file.
With some CPP `#if`s around them where required.

```haskell
      , clashTestGroup "BitVector"
        [ runTest "test1" def
        , runTest "test2" def

#if some_condition
        , runTest "test3" def
#endif

#if some_other_condition
        , runTest "test4" def
#endif
        ...
        ]
```

alex-mckenna: Yeah, I'm not so happy with how much chopping and changing has happened in this file. It's also harder to exclude just some tests in a block because of the brackets / commas starting on the beginning of lines :cry: 

I'll try and improve it a bit now (it's gotten worse since I've rebased twice onto master since making this branch) and reduce the number of CPP directives. Fortunately most of these are due to the primitives for things like Natural and Enum not being implemented, so a lot of them can be removed quickly one the primitive evaluator work starts.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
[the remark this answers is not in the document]
alex-mckenna: > A nullary data constructor / primitive is fully applied, otherwise it isn't.

In hindsight this is a terrible sentence. What I meant to express is that if you go to evaluate an expression and it's simply a constructor / prim, then it's only fully applied if it's nullary. If not, it's eta expanded.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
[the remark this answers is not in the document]
christiaanb: `IntLiteral i` corresponds to something of type `Int#` in Haskell. While both `IntegerLiteral i` and `App (Data S#) (IntLiteral i)` correspond to something of type `Integer` in Haskell. To expand further:

The following Haskell:
```haskell
case (x :: Int) of
  1  -> True
  _  -> False
```
is desugared to:
```haskell
case x of
  DataPat I# i# -> case i# of
   DefaultPat -> False
   LitPat 1# -> True
```
while the following Haskell:
```haskell
case (x :: Integer) of
  1 -> True
  _ -> False
```
is desugared to:
```haskell
case (x :: Integer) of
  DefaultPat -> False
  LitPat 1 -> True
```
but we can in Haskell also write:
```haskell
case (x :: Integer) of
  S# i# -> case i# of
    1# -> True
    _ -> False
  _ -> False
```
which is desugared to:
```haskell
case (x :: Integer) of
  DefaultPat -> False
  DataPat S# i# -> case i# of
    DefaultPat -> False
    LitPat 1# -> True
```
So only for `Integer` (and `Natural`) do we need to be able to match against `LitPat`and `DataPat`. While for the other data types the LitPat are all the unboxed versions.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+    Case x ty xs -> evaluateCaseWith ghcEvaluate ghcIsLiteral ghcIsData x ty xs
+    Cast x a b -> evaluateCastWith ghcEvaluate x a b
+    Tick ti x -> evaluateTickWith ghcEvaluate ti x
+
+-- TODO Implement evaluation for primitives and call here.
+--
+ghcEvaluatePrim :: PrimInfo -> [Either Value Type] -> State Env Value
+ghcEvaluatePrim p args = pure (VNeu $ NePrim p args)
+
+ghcIsLiteral :: Literal -> Pat -> Bool
```
christiaanb, line 44: Perhaps rename this to `ghcMatchLiteral`? with some comments why it's written the way it is.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+         Jn# _ -> False
+
+    |  otherwise
+    -> False
+
+  LitPat m -> l == m
+  DefaultPat -> True
+  _ -> False
+
+ghcIsData :: DataCon -> Pat -> Bool
```
christiaanb, line 66: Rename to `ghcMatchData` ?

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+ghcIsData dc = \case
+  DataPat c _ _ -> dc == c
+  LitPat _ -> False
+  DefaultPat -> True
+
+ghcApply :: Value -> Either Value Type -> State Env Value
```
christiaanb, line 72: Add top-level comment

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+  , addBinder <- insertType i argTy
+  = State.put (addBinder env) >> ghcEvaluate x
+
+  | otherwise
+  = error ("ghcApply: Cannot apply " <> show arg <> " to " <> show val)
+ where
+  -- TODO Where to add ticks back?
+  (f, _) = collectValueTicks val
+
+ghcQuote :: Value -> State Env Nf
```
christiaanb, line 93: Add toplevel comment. Explain why this couldn't be made part of the "front-end agnostic" part of PE. I assume it is to do with `ghcApply` which does this call-by-need/sharing stuff.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+  VNeu n -> fmap NNeu (ghcQuoteNeutral n)
+  VLit l -> quoteLiteral l
+  VData dc args -> quoteDataWith ghcQuote dc args
+  VPrim p args -> quotePrimWith ghcQuote p args
+  VLam i x env -> quoteLamWith ghcQuote ghcApply (Left i) x env
+  VTyLam i x env -> quoteLamWith ghcQuote ghcApply (Right i) x env
+  VCast x a b -> quoteCastWith ghcQuote x a b
+  VTick x t -> quoteTickWith ghcQuote x t
+
+ghcQuoteNeutral :: Neutral Value -> State Env (Neutral Nf)
```
christiaanb, line 104: Add toplevel comment

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+import GHC.Integer.GMP.Internals (Integer(..))
+
+import Clash.Core.DataCon
+import Clash.Core.Evaluator.Models
+import Clash.Core.Evaluator.Semantics
+import Clash.Core.Literal
+import Clash.Core.Term
+import Clash.Core.Type
+
+ghcEvaluator :: Evaluator
```
christiaanb, line 20: Add top-level comment

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+ghcEvaluator = Evaluator ghcEvaluate ghcQuote
+
+ghcEvaluate :: Term -> State Env Value
```
christiaanb, line 23: Add top-level comment. Say something about the architecture; i.e. why `ghcEvaluate` is passed to all of those other `evaluateXYZ` functions.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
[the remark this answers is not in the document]
leonschoorl: I'm not sure either.
I agreed that you should get the `IntLiteral` for 
```
case x of
  5 -> ....
```

But if you have:
```
y = case x of
  2 -> 3#
  I# x -> x
  _    -> error "should be unreachable"

x :: Int
x = fromIntegral (4 :: Signed 4) 
```
I'm afraid this code might get to the `| otherwise -> False`, conclude that the second alternative doesn't match and evaluate to the third alternative.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+  DefaultPat -> True
+
+ghcApply :: Value -> Either Value Type -> State Env Value
+ghcApply val arg
+  | VNeu n <- f
+  = pure $ VNeu (either (NeApp n) (NeTyApp n) arg)
+
+  | VLam i x env <- f
+  , Left argV <- arg
+  , addBinder <- insertLocal i (Right argV)
```
christiaanb, line 79: Perhaps add a comment here saying that we heap/let-bind the argument for call-by-need/sharing purposes

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
```diff
@@ -0,0 +1,111 @@
...
+  = State.put (addBinder env) >> ghcEvaluate x
+
+  | VTyLam i x env <- f
+  , Right argTy <- arg
+  , addBinder <- insertType i argTy
```
christiaanb, line 84: What's the purpose of adding types to the environment? Mostly curious, and probably worth documenting in a comment.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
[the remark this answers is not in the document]
christiaanb: That unreachable is removed by desugaring

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
alex-mckenna: Looking back at it, I might just make a `quoteWith` function in semantics that takes `apply` as an argument. There's nothing else in quoting which is front-end specific.

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -290,7 +415,7 @@ instance AsTerm Value where
       | null bs = x
       | otherwise = Letrec bs x
      where
-      bs = Map.toList $ fmap asTerm (envLocals env)
+      bs = Map.toList . fmap asTerm $ Map.filter isRight (envLocals env)
```
christiaanb, line 418: Could you add a comment why you only want the Values from the local environment? Or explain that/why the un-evaluated Terms cannot/should not be referenced by the body of the (ty)lambda.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,405 @@
...
+of a primitive may potentially be a function (e.g. primtives that return an
+action in IO). This is handled in evaluateAppWith, by checking if more
+arguments are given than the primitive requires, and partitioning the
+arguments. The arguments for the primitive are applied and the primtive is
+evaluated, then the remaining arguments are applied.
+-}
+
+isFullyApplied :: Type -> [Either Term Type] -> Bool
+isFullyApplied ty args =
+  length args == length (fst $ splitFunForallTy ty)
```
christiaanb, line 84:
```suggestion
  args `Data.List.Extra.equalLength` (fst $ splitFunForallTy ty)
```

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,405 @@
...
+-- | Default implementation for looking up a variable in the environment.
+-- This checks both the local and global environments, and inlines global
+-- binders which are not marked as NOINLINE.
+--
+evaluateVarWith :: (Term -> State Env Value) -> Id -> State Env Value
+evaluateVarWith eval i = do
+  env <- State.get
+
+  if |  Just etv <- lookupLocal i env
+     -> either (State.withState (deleteLocal i) . eval) pure etv
```
christiaanb, line 122: Perhaps add a comment why we need to remove the variable from the environment.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,405 @@
...
+  (f, args) = collectArgs term
+  evalArg   = bitraverse eval pure
+  evalArgs  = traverse evalArg
+
+-- | Default implementation for evaluating a letrec expression. This adds all
+-- bindings to the heap without eagerly evaluating them, then evaluates the
+-- body of the expression.
+--
+-- Bindings are not evaluated eagerly, as they may not be needed depending how
+-- the body of the expression evaluates (i.e. bindings may be in an elided
```
christiaanb, line 222:
```suggestion
-- the body of the expression evaluates (i.e. bindings may only be referenced in an elided
```

clash-lib/src/Clash/Normalize/Strategy.hs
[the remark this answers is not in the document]
martijnbastiaan: Would it be possible to keep the CPP in one place (`Clash.CPP`) and just do the rest with `if then else`? Inline CPP is hard to read / easy to mess up.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,405 @@
...
+      case findMatchingAlt (matchLit l) ys of
+        Just (_, y) -> eval y
+        Nothing -> error ("evaluateCaseWith: No pattern for literal " <> show l)
+
+    VData dc _ ->
+      case findMatchingAlt (matchData dc) ys of
+        Just (_, y) -> eval y
+        Nothing -> error ("evaluateCaseWith: No pattern for data " <> show dc)
+
+    v -> do
```
christiaanb, line 268: What about 
```haskell
case x of
  _ -> y
```
If `x` evaluates to a neutral, we can still elide the case, and only evaluate `y`. (If `x` evaluates to _|_, the case should evaluate to _|_ as well).

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,405 @@
...
+  -> State Env Value
+evaluateLetrecWith eval bs x =
+  State.withState addTerms (eval x)
+ where
+  addTerms env =
+    let terms = fmap (fmap Left) bs
+     in foldl' (flip $ uncurry insertLocal) env terms
+
+-- | Default implementation for evaluating a case expression. This elides the
+-- case expression if the branch taken can be determined statically, otherwise
```
christiaanb, line 238:
```suggestion
-- case expression, and only evaluates the alternative if the branch taken can be determined statically, otherwise
```

clash-lib/src/Clash/Normalize/Strategy.hs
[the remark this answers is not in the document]
alex-mckenna: Hmm, I wouldn't want to start adding stuff to `Clash.CPP` in `clash-prelude` for this. I don't think it's too bad here, maybe at the start of `testsuite/Main.hs` I could define some macros to tidy up the mess there

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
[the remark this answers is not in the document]
leonschoorl: I had to look up elide.
I could guess based on context, but it wasn't immediately. 
Any of omitted,pruned,removed,eliminated seem much clearer to me.

clash-lib/src/Clash/Normalize/Strategy.hs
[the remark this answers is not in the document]
leonschoorl: Using CPP directly makes it clear that this is a compile time choice.
You don't have to wonder where the condition of the `if` comes from and if it could change.
And you can conditionally import modules which don't have to even exist when certain flags are set.

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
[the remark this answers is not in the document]
leonschoorl: Ok, the third case doesn't exists.
But the point still stands, if it doesn't match the `I# x` pattern correctly it will run into this [error](https://github.com/clash-lang/clash-compiler/blob/cb90adaf8ece84d1c1d7951ca6e3647a90a4fd10/clash-lib/src/Clash/Core/Evaluator/Semantics.hs#L261)

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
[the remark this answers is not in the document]
alex-mckenna: Fair enough, in that case I'll change things to say pruned. It has the strongest "I'm cutting this out of the term" connotation

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
alex-mckenna: Another thing to correct I've just noticed: identifiers / tyvars bound in `DataPat` need to be added to the environment. Currently those bindings are being lost and these things are becoming free vars

clash-ghc/src-ghc/Clash/GHC/PartialEval.hs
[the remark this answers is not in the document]
christiaanb: No, because Integer and Natural are the only special types. Only they can have a DataPat and LitPat match. All boxed types only have a DataPat match; and unboxed types only have a LitPat match.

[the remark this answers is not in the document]
christiaanb: i.e. GHC will never generate LitPat matches for Boxed types, nor generate DataPat matches for unboxed types. If it does it’s an actual error.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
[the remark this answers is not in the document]
christiaanb: If *elide* captures the intention and nuance better than any of its "synonyms", I think I'd rather people grab a dictionary to understand the comments, than to change the comment.

testsuite/Main.hs
```diff
@@ -303,31 +348,54 @@ runClashTest = defaultMain $ clashTestRoot
...
         ]
       , clashTestGroup "BoxedFunctions"
         [ runTest "DeadRecursiveBoxed" def{hdlSim=False}
         ]
+#if !EXPERIMENTAL_EVALUATOR || __GLASGOW_HASKELL__ >= 865
       , clashTestGroup "CSignal"
-        [ runTest "CBlockRamTest" def{hdlSim=False}
-        , runTest "MAC" def{hdlSim=False}
+        [ runTest "MAC" def{hdlSim=False}
+        , runTest "CBlockRamTest" def{hdlSim=False}
```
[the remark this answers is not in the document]
alex-mckenna: Yeah...

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,548 @@
...
+-- currently performed in an environment without the given identifier, to
+-- prevent expressions like
+--
+--   let x = x in ...
+--
+-- looping if x is looked up, as it exists in the environment as
+--
+--   Var x |-> Var x
+--
+-- TODO: In the future this should not delete identifiers from the environment
```
christiaanb, line 164: Ah, I guess this is were the small-step and big-step differ. In the small step, evaluating a variable (global or local) would put an Update frame on the stack, so once the variable was evaluated to a value (e.g. a Lambda), the heap/environment would be expanded with the variable-value combo again. And so, deleting functions from the heap would no prevent e.g. functions like fmap to be unrolled. Could something similar be done for the big-step version?

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,548 @@
...
+--
+-- and
+--
+--   case (x :: Integer) of
+--     S# 1 -> ...
+--
+-- match the same value, but do so in different ways. This means the function
+-- to check if a literal is matched would have to check both literal patterns
+-- and data patterns for the constructors of Integer.
+--
```
christiaanb, line 308: I think this needs a TODO on what to do when the subject of the case expression evaluates to bottom. That is, if the subject evaluates to bottom, then the entire case expression evaluates to bottom.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
alex-mckenna: The small step evaluator (as implemented in Clash) actually behaves in a similar way, unrolling things like `fmap` once. Consider the topentity

```haskell
topEntity :: [Int]
topEntity = fmap (*2) [1..10]
```

using the `partial-eval` debugging program in this branch (with the `experimental-evaluator` flag off it uses the small-step evaluator), you can see that this doesn't unroll all the way to

```haskell
topEntity :: [Int]
topEntity = [2,4,6,8,10,12,14,16,18,20]
```

but only unrolls one step giving something similar to [1]

```haskell
topEntity 2 : let wild = 1 in case wild of
                                10 -> []
                                _  -> topEntity_go (wild + 1)
```

where `topEntity_go` is a generated worker. For it to unroll all the way, it would need the `split` function from the paper it's based on to supercompile subterms, and ideally the `terminates` function from the same paper to prevent divergence.

As far as big-step / small-step differences go: the `quote` function is akin to their `split` function, so nothing to be done there. As already mentioned in #950, I want a termination checker for the evaluator too (but _not_ one using the tag bags / well-quasi-orderings from the small-step paper).

---

[1] I actually noticed while double checking this that the old evaluator gives a slightly incorrect result too, the binder `x` from `topEntity_go = \x -> ...` isn't fully substituted in the unrolled `topEntity`, appearing twice as a free variable, more like:

```haskell
topEntity = (I# (x *# 2)) : let wild = x in case wild of
                                10 -> []
                                _  -> topEntity_go (x + 1)
```

while the new evaluator substitutes in the value of `x` correctly.

christiaanb: Right, it only unrolls once, because it only evaluates to WHNF; not NF. That has nothing to do with removing values form the heap while evaluating them to WHNF.

alex-mckenna:
Re-requesting reviews again as the last commit is quite a change.
Although this should be ready for prime time now @christiaanb @leonschoorl

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,581 @@
...
+--   Var x |-> Var x
+--
+-- TODO: In the future this should not delete identifiers from the environment
+-- if they are terminating. This will allow functions like fmap to be fully
+-- unrolled where possible.
+--
+evaluateVarWith :: (Term -> Eval Value) -> Id -> Eval Value
+evaluateVarWith eval i =
+  go (VNeu (NeVar i)) [evalLocal, evalGlobal]
```
christiaanb, line 162: The `Id` contains a field telling us whether it refers to a global or local variable. So we should predicate on that, instead of trying local first, followed by trying on global.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -0,0 +1,581 @@
...
+  = case compare (length args) nArgs of
+      LT -> etaExpand term >>= eval
+      EQ -> evalPrim p args
+      GT -> do
+        let (pArgs, rArgs) = splitAt nArgs args
+        primRes <- evalPrim p pArgs
+        foldM apply primRes (first Left <$> rArgs)
+
+  | otherwise
+  = withIgnored $ do
```
christiaanb, line 259: Add a comment why `withIgnored` is only needed here, and not for `Data` or `Prim`

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -1,202 +1,274 @@
...
 {-# LANGUAGE OverloadedStrings #-}
 
 module Clash.Core.Evaluator.Models where
 
 import Control.Concurrent.Supply (Supply)
-import Control.Monad ((>=>))
-import Control.Monad.State.Strict (State)
-import Control.DeepSeq (NFData(..))
+import Control.Monad.RWS.Strict (MonadReader, MonadState, RWS)
+import qualified Control.Monad.RWS.Strict as RWS
```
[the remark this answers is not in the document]
alex-mckenna: I'm not sure about this, the CPS variant is only to make the Writer part constant space (which we don't use) and would incur extra boilerplate (`mtl` doesn't provide instances for the CPS variant yet).

Would you be happy with a comment saying that if the Writer part is used in the future to change to the CPS module instead?

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
alex-mckenna: Yes! I kept feeling as though I was missing something obvious that would make this function nicer to implement

clash-lib/src/Clash/Core/Evaluator/Models.hs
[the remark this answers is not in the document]
christiaanb: But couldn't you use the `transformers` functions instead of the `mtl` one?:
* http://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-RWS-CPS.html#g:3
* http://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-RWS-CPS.html#g:5

Unless we're using some code that has a `MonadState` and `MonadReader` constraint of course and we have to use `mtl`.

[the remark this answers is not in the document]
alex-mckenna: Currently the code uses the `mtl` instances derived for `Eval`, changing to the `transformers` ones directly would only work if `Eval` was changed from `newtype` to `type` or the `MonadState` and `MonadReader` classes were implemented for `Eval` using the functions in the `CPS` module.

Given we don't use the `Writer`, I would rather keep it as-is, and change the `mtl` import when the next version of `mtl` is released (which includes [a new CPS module](https://github.com/haskell/mtl/blob/master/Control/Monad/RWS/CPS.hs))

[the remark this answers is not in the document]
christiaanb: Alright, sounds good.

benchmark/benchmark-normalization.hs
```diff
@@ -1,21 +1,30 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE PartialTypeSignatures #-}
 
-{-# OPTIONS_GHC -Wno-orphans                 #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
martijnbastiaan, line 4: Do we still need this?

alex-mckenna: Yes, this module has an orphan `NFData` instance for `Supply` which I took out at one point (and added back later)

alex-mckenna:
Regarding the recommendation for the CI: I don't think that would be particularly effective. Running with `+experimental-evaluator` can change the results of tests in `clash-testsuite`, and does so differently depending on GHC version. Fortunately, with the recent changes to the CI, it _feels_ quicker anyway so I'm not so fussed about it for this PR (although maybe in one of the next PE PRs)

PR #1299, opened 2020-04-28
Consider primitives undefined if one of their arguments is

### LINKED ISSUE #1297: Prim evaluation fails on undefined arguments
The following fails to compile:

```haskell
import Clash.Prelude
import Numeric.Natural (Natural)

topEntity :: Int
topEntity =
  case compare (toInteger @Natural (0 - 1)) 0 of
    LT -> 2
    _ -> 3
```

It first emits a warning:

```
Irreducible constant as case subject: GHC.Integer.Type.compareInteger
  (GHC.Natural.naturalToInteger
     (GHC.Natural.minusNatural 0 1))
  0
Can be reduced to: GHC.Integer.Type.compareInteger
  (GHC.Natural.naturalToInteger
     (Clash.Transformations.undefined
        @GHC.Natural.Natural[3674937295934324782]))
  0
```

And after a few transformations it crashes:

```
Test.hs:5:1: error:
    
    Clash.Netlist.BlackBox(284): No blackbox found for: GHC.Integer.Type.compareInteger. Did you forget to include directories containing primitives? You can use '-i/my/prim/dir' to achieve this.
    
  |
5 | topEntity =
  | ^^^^^^^^^
```

The expression should be reduced, to undefined. The issue is that the evaluator bails as soon as it sees an undefined value ("Clash.Transformations.undefined") instead of making the result of the primitive undefined too.

martijnbastiaan:
Fixes #1297


martijnbastiaan:
@christiaanb Should be good now. I'll backport it manually.

PR #1311, opened 2020-05-05
Types with unconstrained existentials are non-representable

### LINKED ISSUE #1310: Clash fails to eliminate all sources of polymorphism
The following:

```haskell
{-# LANGUAGE GADTs #-}
module Test where

import Clash.Prelude

data Ex where
  Ex :: forall n . SNat n -> Bool -> Ex

f :: Ex -> Bool
f (Ex n y) = h (replicate n y)
{-# NOINLINE f #-}

h :: Vec n Bool -> Bool
h xs = foldr (||) True xs
{-# NOINLINE h #-}

g :: Int -> Bool -> Ex
g 0 b = Ex (SNat @3) b
g n b = g (n-1) b

topEntity :: Bool -> Bool
topEntity b = f (g 1 b)
```

fails with:

```
Test.hs:14:1: error:
    
    Clash.Normalize(179): Clash can only normalize monomorphic functions, but this is polymorphic:
    Test.h :: forall (n :: GHC.Types.Nat).
          Clash.Sized.Vector.Vec n GHC.Types.Bool
          -> GHC.Types.Bool
    
   |
14 | h xs = foldr (||) True xs
   | ^
```

Which is happening because Clash considered

```
data Ex where
  Ex :: forall n . SNat n -> Bool -> Ex
```

to be "representable" (can have a "trivial" bit-encoding), which in a way it true because the underlying representation of `SNat n` is a `Natural`; but does cause certain transformations not to fire leaving `h` not being specialized on the vector length `n`. To remedy the situation, Clash should simply consider all types with unconstrained existential types in their data constructors to be "non-representable" as they can induce lingering polymorphism.

christiaanb:
Fix #1310

### DIFF
--- a/changelog/2020-05-05T11_54_37+02_00_fix1310
+++ b/changelog/2020-05-05T11_54_37+02_00_fix1310
@@ -1,0 +1,1 @@
+FIXES: Fails to propagate type information of existential type [#1310](https://github.com/clash-lang/clash-compiler/issues/1310)
--- a/clash-lib/src/Clash/Netlist/Util.hs
+++ b/clash-lib/src/Clash/Netlist/Util.hs
@@ -20,6 +20,7 @@
 
 module Clash.Netlist.Util where
 
+import           Data.Coerce             (coerce)
 import           Control.Error           (hush)
 import           Control.Exception       (throw)
 import           Control.Lens            ((.=),(%=))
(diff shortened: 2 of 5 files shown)


martijnbastiaan:
Oh and maybe the non-happy path should be tested too.

christiaanb:
@martijnbastiaan The non-happy part? you mean when there is an unconstrained existential? that's what we test with `T1310`. If you mean constrained existentials, we have a lot of those in the form of our GADT tests.

martijnbastiaan:
I was thinking of a test that will throw the `Can't translate data types with existentials` error. That's not tested by `GADTs/T1310.hs` right?

PR #1315, opened 2020-05-08
Add '-fclash-debug-transformations-{from,limit}'

martijnbastiaan:
Clash now emits transformation numbers when running in debug mode. For
example, running with -fclash-debug DebugName could output:

    caseCase {0}
    caseCon {1}
    bindOrLiftNonRep {2}
    [..]

Running with -fclash-debug-transformations-from=1 would omit the first
debug line. Running with -fclash-debug-transformations-limit=2 would
only print the first two and stop Clash after that. These flags can be
combined to get only a slice of transactions.

Useful when a "cheap" debug level (e.g., DebugName) identified a number
of troublesome transformations, while running with an "expensive" debug
level (e.g., DebugApplied) would produce massive logs.

--------------------------------------

Note: purposely didn't add a changelog entry, as this isn't really a user feature nor bug.

### DIFF
--- a/clash-lib/src/Clash/Rewrite/Util.hs
+++ b/clash-lib/src/Clash/Rewrite/Util.hs
@@ -11,6 +11,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE NondecreasingIndentation #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -241,7 +264,7 @@ applyDebug lvl _transformations name exprOld hasChanged exprNew =
     error $ $(curLoc) ++ "Expression changed without notice(" ++ name ++  "): before"
                       ++ before ++ "\nafter:\n" ++ after
 
-  traceIf (lvl >= DebugName && hasChanged) name $
+  traceIf (lvl >= DebugName && hasChanged) (name <> " {" <> show nTrans <> "}") $
```
martijnbastiaan, line 267: This purposely printed as `{n}` as to not collide with uniques (printed as `[n]`)

PR #1317, opened 2020-05-08
DEC: Correctly apply self-substition in lifted expression

### LINKED ISSUE #1316: Recursive let-expressions throw DEC  transformation into a loop
The following reduced test-case:

```haskell
import Clash.Prelude

incr :: Index 2 -> Index 2
incr i = if i == maxBound then 0 else i + 1
{-# NOINLINE incr #-}

topEntity :: Index 10 -> Index 2
topEntity j = case j < 1 of
  False ->
    let xs = init (Cons 1 (Cons (incr (head xs)) Nil)) in last xs
  True ->
    let ys = init (Cons 2 (Cons (incr (last ys)) Nil)) in head ys
```

throws the DEC transformation into a loop. Specifically, DEC rewrites:

```haskell
case j < 1 of
  False ->
    let xs = init (Cons 1 (Cons (incr (head xs)) Nil)) in last xs
  True ->
    let ys = init (Cons 2 (Cons (incr (last ys)) Nil)) in head ys
```

to 

```haskell
let incrOut = incr (case j < 1 of
  False ->
    let xs = init (Cons 1 (Cons (incr (head xs)) Nil)) in head xs
  True ->
    let ys = init (Cons 2 (Cons (incr (last ys)) Nil)) in last ys)
in case j < 1 of
  False ->
    let xs = init (Cons 1 (Cons incrOut Nil)) in last xs
  True ->
    let ys = init (Cons 2 (Cons incrOut Nil)) in head ys
```

trying to share uses of `incr` between the two case-alternatives; only to create a new case-expression on which DEC can match again. What DEC should be doing is rewrite to:

```haskell
let incrOut = incr (case j < 1 of
  False ->
    let xs = init (Cons 1 (Cons incrOut Nil)) in head xs
  True ->
    let ys = init (Cons 2 (Cons incrOut Nil)) in last ys)
in case j < 1 of
  False ->
    let xs = init (Cons 1 (Cons incrOut Nil)) in last xs
  True ->
    let ys = init (Cons 2 (Cons incrOut Nil)) in head ys
```

christiaanb:
Fixes #1316

### DIFF
--- a/changelog/2020-05-08T15_29_02+02_00_fix1316
+++ b/changelog/2020-05-08T15_29_02+02_00_fix1316
@@ -1,0 +1,1 @@
+FIXES: Certain recursive let-expressions in side a alternatives of a case-expression throw the Clash compiler into an infinite loop [#1316](https://github.com/clash-lang/clash-compiler/issues/1316)
--- a/clash-lib/src/Clash/Normalize/DEC.hs
+++ b/clash-lib/src/Clash/Normalize/DEC.hs
@@ -66,7 +66,7 @@ import Clash.Core.TyCon      (tyConDataCons)
 import Clash.Core.Type       (Type, isPolyFunTy, mkTyConApp, splitFunForallTy)
 import Clash.Core.Var        (isGlobalId)
 import Clash.Core.VarEnv
-  (InScopeSet, elemInScopeSet, notElemInScopeSet)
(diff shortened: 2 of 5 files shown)


christiaanb:
@martijnbastiaan The branch now has a commit that should avoid the accidental variable capture.

martijnbastiaan:
Looks good!

PR #1349, opened 2020-05-25
Do not disable the monomorphism restriction

### LINKED ISSUE #1270: Enabling the monomorphism restriction
Clash, unlike GHC, currently _disables_ the monomorphism restriction (see https://wiki.haskell.org/Monomorphism_restriction if you want to know more about the monomorphism restriction). I did some git blame code archeology, and the commit that introduced it (https://github.com/clash-lang/clash-compiler/commit/295a79889c64d5b7d94e644fc1c4aba043cc2318) just wanted clash and clashi to be on par. This disabling of the monomorphism restriction was actually somewhat misguided, as the monomorphism restriction is actually only turned off for the interactive session of ghci, not for the files that are loaded by the session.

This weekend, I was debugging some code that triggered `Clash.Signal.Trace` "Already tracing signal with name" exception (https://github.com/clash-lang/clash-compiler/blob/eccd263baafb4478a8f30c11fc70d5652f648798/clash-prelude/src/Clash/Signal/Trace.hs#L171) even though there was only on `traceSignal` with that name in the code. After a whole day of searching, this turned out to be GHC evaluating that expression two (or even more) times; exactly what the monomorphism restriction normally prevents!

This is bad for two reasons:
1. This disabling of the monomorphism restriction breaks our `Clash.Signal.Trace` API. You can work around it by locally turning it back on by `{-# LANGUAGE MonomorphismRestriction #-}`
2. Although not measured, I'm think this is affecting our simulation performance!

Now, enabling the monomorphism restriction by default, would come at a price. That is, in a function that wasn't annotated with a top-level type signature, with the monomorphism restriction enabled, I got:
```
error:
    • Unbound implicit parameters
        (GHC.Classes.IP
           (Clash.Signal.HiddenClockName System) (Clock System),
         GHC.Classes.IP
           (Clash.Signal.HiddenEnableName System) (Enable System),
         GHC.Classes.IP
           (Clash.Signal.HiddenResetName System) (Reset System))
```
The work-around/solution was simple though: add a top-level type signature. Using `{-# LANGUAGE PartialTypeSignatures #-}` , I could even reduce it down to:

```
circuit :: _ => Signal dom _
```

All in all, I'm in favor is enabling the monomorphism restriction by default in Clash; and just document the fallout. But am interested in hearing arguments in favor of leaving it turned off by default.

### LINKED ISSUE #1345: Seemingly small change drastically impairs circuit complexity
I have a small circuit that generates VGA signal from a frame buffer. A tiny refactoring that really just moves code around turns the generated Verilog from 1052 lines to 2948 lines (so ~3 times as much Verilog), and the Verilator performance also goes down from 

> 60 frames at 3938 ms, 15.24 fps

to 

> 60 frames at 7155 ms, 8.39 fps

The commit in question is [this one](https://github.com/gergoerdi/clash-draw-toy/commit/58bd6df890b41f713a827c6c07afd25d2b02c220); as you can see, all it does is it moves the VGA sync signal generation from the `topEntity` to another function.

The full code is available [here](https://github.com/gergoerdi/clash-draw-toy/tree/clash-issue-1345), with the last commit being the one that causes this sudden explosion in circuit complexity.

christiaanb:
Warn when it is turned off when converting Hs to HDL. Also enable -Wmonomorphism-restriction by default so you know when you might want to add a type signature to prevent unexpected monomorphism.

This commit also demonstrates the fallout of leaving the `MonomorphismRestriction` enabled by default, and can be used as a guide for fixing Clash code suffering from the same
issues once this commit lands in a released version of the compiler.

Fixes #1270
Fixes #1345

### DIFF
--- a/tests/shouldwork/BitVector/UnpackUndefined.hs
+++ b/tests/shouldwork/BitVector/UnpackUndefined.hs
@@ -4,4 +4,4 @@ import Clash.Sized.Internal.BitVector
 
 -- https://github.com/clash-lang/clash-compiler/issues/804
 state = unpack undefined#    :: Vec 3 Bit
-topEntity = register @System state
+topEntity a = register @System state a
--- a/changelog/2020-05-28T16_27_54+02_00_no_disable_monomorphism
+++ b/changelog/2020-05-28T16_27_54+02_00_no_disable_monomorphism
@@ -1,0 +1,4 @@
+CHANGED: Clash no longer disables the monomorphism restriction.
+See https://github.com/clash-lang/clash-compiler/issues/1270, and mentioned issues, as to why.
+This can cause, among other things, certain eta-reduced descriptions of sequential circuits to no longer type-check.
+See https://github.com/clash-lang/clash-compiler/pull/1349 for code hints on what kind of changes to make to your own code in case it no longer type-checks due to this change.
--- a/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
+++ b/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
@@ -71,29 +71,36 @@ getMainTopEntity modName bindingMap topEnts nm =
     let topIdNm = Text.unpack (nameOcc (varName topId)) in
     topIdNm == nm || ('.':nm) `isSuffixOf` topIdNm
 
--- | Checks whether MonoLocalBinds language extension is enabled or not in
--- modules.
+-- | Checks whether MonoLocalBinds and MonomorphismRestricton language extensions
+-- are enabled or not in modules.
 checkMonoLocalBindsMod :: GHC.ModSummary -> IO ()
-checkMonoLocalBindsMod x =
-  unless (active . GHC.ms_hspp_opts $ x) (hPutStrLn stderr $ msg x)
+checkMonoLocalBindsMod x = do
+  unless (active LangExt.MonoLocalBinds . GHC.ms_hspp_opts $ x)
+         (hPutStrLn stderr $ msg LangExt.MonoLocalBinds x)
+  unless (active LangExt.MonomorphismRestriction . GHC.ms_hspp_opts $ x)
+         (hPutStrLn stderr $ msg LangExt.MonomorphismRestriction x)
   where
-    msg = messageWith . GHC.moduleNameString . GHC.moduleName . GHC.ms_mod
+    msg ext = messageWith ext . GHC.moduleNameString . GHC.moduleName . GHC.ms_mod
 
--- | Checks whether MonoLocalBinds language extension is enabled when generating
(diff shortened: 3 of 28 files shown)


alex-mckenna:
Just for my own understanding:

To relate this to issues like #1345, would that mean the current behaviour (slowdown) is the default, _but_ you would get a warning that a particular thing may be the cause of the slowdown?

christiaanb:
No, this change will, at places where the most polymorphic type used to be inferred (leading to the slowdown), will now do one of two:
1. Result in type-errors, see the failing CI
2. Give a warning that the monomorphism restriction is in effect

christiaanb:
Failing CI leads me to believe this should not be backported to 1.2

christiaanb:
So for #1345 I see the warning:

```
/home/christiaan/clash-draw-toy/src/DrawToy.hs:81:5: warning: [-Wmonomorphism-restriction]
    • The Monomorphism Restriction applies to the bindings
      for ‘vgaSync’, ‘vgaX’, ‘vgaY’
        Consider giving them a type signature
    • In an equation for ‘drawToy’:
          drawToy input
            = (frameEnd, delayVGA vgaSync rgb)
            where
                VGADriver {..} = vgaDriver vga640x480at60
                frameEnd = isFalling False (isJust <$> vgaY)
                x = center @128 . scale (SNat @5) $ vgaX
                y = center @64 . scale (SNat @5) $ vgaY
                ....
   |
81 |     VGADriver{..} = vgaDriver vga640x480at60
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tests/shouldwork/BitVector/UnpackUndefined.hs
```diff
@@ -4,4 +4,4 @@ import Clash.Sized.Internal.BitVector
 
 -- https://github.com/clash-lang/clash-compiler/issues/804
 state = unpack undefined#    :: Vec 3 Bit
-topEntity = register @System state
+topEntity a = register @System state a
```
martijnbastiaan, line 7: I don't understand why this needs to change. Is this explained anywhere?

christiaanb: It's explained somewhere on the internet. The monomorphism restriction is annoying.

martijnbastiaan: https://wiki.haskell.org/Monomorphism_restriction

martijnbastiaan:
I think we should include this in 1.4, but not backport it to 1.2. Looks fine otherwise :).

Could you write a changelog entry? I think this warrants a more verbose entry than usual.

christiaanb:
@martijnbastiaan I've added a changelog entry

PR #1352, opened 2020-05-27
Add a DSL for creating primitives

cchalmers:

PR #1354, opened 2020-05-28
Add blackbox for 'Clash.Sized.Vector.iterateI'

martijnbastiaan:
Adds blackbox (and associated reduceNonRepPrim/evaluator rule) for `Clash.Sized.Vector.iterateI`. Works around https://github.com/clash-lang/clash-compiler/issues/1240. 

--------------------

| Name              | Before (ms) | After (ms) | B/A  |
|-------------------|-------------|------------|------|
| FIR               | 35.73       | 34.61      | 1.03 |
| Reducer           | 771.6       | 732        | 1.05 |
| Queens            | 16.89       | 16.30      | 1.03 |
| BundleMapRepeat   | 21.70       | 16.88      | 1.28 |
| PipelinesViaFolds | 1083        | 208.1      | 5.20 |
| AES               | 326.6       | 281.6      | 1.07 |
| T1354B            | ∞           | 508        |      |


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -2241,6 +2241,19 @@ reduceNonRepPrim c@(TransformContext is0 ctx) e@(App _ _) | (Prim p, args, ticks
                     in  (`mkTicks` ticks) <$> reduceImap c n argElTy resElTy fun arg
                else return e
           _ -> return e
+      "Clash.Sized.Vector.iterateI" | argLen == 5 -> do
```
alex-mckenna, line 2244: You don't need a `do` here, it's just `let ... in case ... of ...`

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -2241,6 +2241,19 @@ reduceNonRepPrim c@(TransformContext is0 ctx) e@(App _ _) | (Prim p, args, ticks
...
+        case runExcept (tyNatSize tcm nTy) of
+          Right n -> do
+            shouldReduce1 <- List.orM [
+                pure (ultra || n < 2)
+              , shouldReduce ctx
+              , isUntranslatableType_not_poly aTy ]
+
+            if shouldReduce1
+            then (`mkTicks` ticks) <$> reduceIterateI c n aTy eTy f a
+            else return e
```
alex-mckenna, line 2255: A small nitpick, both the list and the if expression are formatted slightly untidily. I know this module isn't the neatest, but it's a small thing to change

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -159,40 +193,80 @@ newName :: Backend backend => Text -> State (BlockState backend) Identifier
...
   => Text
-  -- ^ guide name to be assigned
+  -- ^ Name hint for assignment
   -> TExpr
   -- ^ expression to be assigned to
   -> State (BlockState backend) TExpr
   -- ^ the identifier of the expression that actually got assigned
 assign aName (TExpr ty aExpr) = do
-  texp@(~(TExpr _ (Identifier uniqueName Nothing))) <- declare aName ty
+  texp@(~(TExpr _ (Identifier uniqueName Nothing))) <- declare aName Wire ty
```
alex-mckenna, line 237: For this and line 253, why are these patterns lazy? Not saying they shouldn't be, just that I haven't really used lazy patterns that much and am curious why they're needed here

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -268,28 +268,32 @@ boolFromBitVector
...
 -- entity.
+--
+-- TODO: Implement for (System)Verilog
 boolFromBits
-  :: Backend backend
-  => [Identifier]
+  :: [Identifier]
   -> TExpr
-  -> State (BlockState backend) [TExpr]
+  -> State (BlockState VHDLState) [TExpr]
```
alex-mckenna, line 373: Why was this changed to VHDL only?

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -352,6 +371,26 @@ outputFn _ outType _ _ texpr =
...
+-- | Create a vector of 'TExpr's
+vec
+  :: (HasCallStack, Backend backend)
+  => [TExpr]
+  -- ^ Elements of vector
+  -> State (BlockState backend) TExpr
+  -- ^ Vector elements
+vec els@(el:_)
+  | () <- foldl checkType () els
+  = pure (TExpr (Vector (length els) (ety el)) theVec)
```
alex-mckenna, line 437: Does `ghc` not complain about exhaustiveness here? What if the list is non-empty but not all checks pass?

Would this be clearer as `all (\x -> ety x == ety el) els` with `checkType` being removed? In the `otherwise` case you could print a list of all unique types.

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -391,12 +391,15 @@ vec els@(el:_)
...
--- | Create a tuple of 'TExpr'
-tuple :: TExpr -> TExpr -> TExpr
-tuple a b =
-  let
-    ty = Product "GHC.Tuple.(,)" Nothing [ety a, ety b]
-  in TExpr ty (DataCon ty (DC (ty,0)) [eex a, eex b])
+-- | Create an n-tuple of 'TExpr'
+tuple :: [TExpr] -> TExpr
+tuple [] = error $ "nTuple: Cannot create empty tuple"
+tuple [_] = error $ "nTuple: Cannot create a 1-tuple"
```
alex-mckenna, line 451: Is an error here preferable to a 0-tuple being `()` and a 1-tuple just being the value?

clash-lib/src/Clash/Primitives/Sized/Vector.hs
```diff
@@ -45,9 +47,53 @@ import           Clash.Netlist.Types
...
+  | [ (fromMaybe (error "n") . tExprToInteger -> n, _)
+    , _hoFunction
+    , (a, aType)
+    ] <- tInputs bbCtx
+  , let aTemplateType = [TypElem (Typ (Just 2))]
+  , let inst arg = instHO bbCtx 1 (aType, aTemplateType) [(arg, aTemplateType)]
+  = declarationReturn bbCtx "iterateI" (Prim.vec =<< iterateNM (fromInteger n) inst a)
+
+iterateTF' args =
+  error $ "Unexpected number of arguments: " ++ show (length (bbInputs args))
```
alex-mckenna, line 95: This could just be `| otherwise =`

clash-lib/src/Clash/Primitives/DSL.hs
martijnbastiaan: GHC will throw `MonadFail` warnings otherwise, even though we know this can never fail :(

clash-lib/src/Clash/Primitives/DSL.hs
martijnbastiaan: The implementation mentions `and`, which is VHDL only AFAIK. Verilog uses `&&`.

clash-lib/src/Clash/Primitives/DSL.hs
martijnbastiaan: Funny, no it doesn't, but it should.

martijnbastiaan: Yes that'd be clearer, thanks :)

clash-lib/src/Clash/Primitives/DSL.hs
martijnbastiaan: > Is an error here preferable to a 0-tuple being ()

Hmm. I'd like to be a bit cautious with `()`. Zero-width paths should almost always be treated with care in blackboxes. Let's add it when we have use case.

> a 1-tuple just being the value?

That seems useful, will add.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -3016,6 +3016,31 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
...
+  "Clash.Sized.Vector.iterateI"
+    | isSubj
+    , [nTy, aTy] <- tys
+    , [_n, f, a] <- args
+    , Right n <- runExcept (tyNatSize tcm nTy)
+    ->
+      let
+        TyConApp vecTcNm _ = tyView (getResultTy tcm ty tys)
+        Just vecTc = lookupUniqMap vecTcNm tcm
+        [nilCon, consCon] = tyConDataCons vecTc
```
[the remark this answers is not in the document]
martijnbastiaan: I think the current conditions in the pattern guard are expected to fail sometimes, while the ones in the let-bindings aren't. If the patterns in the let-binding fail there's something really weird going on with the tycon map / function signature. I'd like it to not fail silently in that case. Is that fair?

martijnbastiaan:
> One question that didn't make sense being tagged in a particular area: does this work in Verilog / SystemVerilog? Some parts are explicitly VHDL instead of any backend, but other parts seem to be over every backend.

Yes this works for all HDLs. The changes regarding the this are just an improvement on the DSL that was donated/upstreamed by Chris over at https://github.com/clash-lang/clash-compiler/pull/1352. I haven't used any of those function for the `iterateI` blackbox though :)

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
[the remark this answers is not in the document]
alex-mckenna: Yeah, that's fair. I know on the whole that evaluator allows most things to fail silently but that's no reason to disallow noisier errors

clash-lib/src/Clash/Primitives/DSL.hs
alex-mckenna: Probably because there's only one _actual_ value for type `()` so you've covered all cases, it was just slightly strange to read

clash-lib/src/Clash/Primitives/DSL.hs
alex-mckenna: Fair point about the zero width path

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1492,7 +1492,7 @@ iterate SNat = iterateI
 -- \"'iterateI' @f z@\" corresponds to the following circuit layout:
 --
 -- <<doc/iterate.svg>>
-iterateI :: KnownNat n => (a -> a) -> a -> Vec n a
+iterateI :: forall n a. KnownNat n => (a -> a) -> a -> Vec n a
```
christiaanb, line 1497: This should be logged in the Changelog, as I think you had to type-apply them in reverse order before this change. I assume that's why you changed the order in the first place.

clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -1244,25 +1244,15 @@ mkInput pM = case pM of
...
-                                [e]
-mkVectorChain sz elTy (e:es) = DataCon (Vector sz elTy) VecAppend
-                                [ e
-                                , mkVectorChain (sz-1) elTy es
-                                ]
+mkVectorChain :: Int -> HWType -> [Expr] -> Expr
+mkVectorChain _ elTy [] = DataCon (Vector 0 elTy) VecAppend []
+mkVectorChain _ elTy [e] = DataCon (Vector 1 elTy) VecAppend [e]
+mkVectorChain sz elTy (e:es) =
+  DataCon (Vector sz elTy) VecAppend [e, mkVectorChain (sz-1) elTy es]
```
christiaanb, line 1251: Not a fan of these changes, they neither clarify (e.g. more comments), nor change the behavior; so why not leave as is?

martijnbastiaan:
Thanks for the review @alex-mckenna! I've implemented your suggestions.

clash-lib/src/Clash/Netlist/Util.hs
martijnbastiaan: Fair enough. I just thought they were formatted very far from what the styleguide says.

clash-prelude/src/Clash/Sized/Vector.hs
martijnbastiaan: Oh, no I just added it to make the order explicit. The order shouldn't have changed, but I'll test!

martijnbastiaan: This works in Clash 1.2.1:

```
Clash.Prelude> iterateI @5 @Int succ 3
<3,4,5,6,7>
```

I'd like to keep the explicit order in though.

clash-lib/src/Clash/Netlist/Util.hs
martijnbastiaan: And I usually don't mind cleanup-commits, my IDE lets me click through them easily. I'll revert the change though.

clash-prelude/src/Clash/Sized/Vector.hs
christiaanb: Ah, my bad. I thought it was type variables on the RHS of `=>` are traverse before they're traversed on the `LHS`, and then you'd encounter `a` before `n`.

martijnbastiaan:
Converting to draft PR. The following example makes Clash get stuck in an infinite loop:

```haskell
module Test where

import Clash.Prelude

f :: forall n. KnownNat n => Index (n + 1) -> Int -> Int
f n = foldl (.) id lanes
 where
  lanes :: Vec (n + 1) (Int -> Int)
  lanes = map (\i -> if i < n then id else id) (iterateI succ 0) 

topEntity :: Int -> Int
topEntity = f (1 :: Index 5) 
```

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
```diff
@@ -73,6 +79,172 @@ import           Clash.Rewrite.Types
...
+  Prim (PrimInfo "Clash.Sized.Vector.head" (vecHeadTy vecTcNm) WorkNever)
+
+vecLastPrim
+  :: TyConName
+  -- ^ Vec TyCon name
+  -> Term
+vecLastPrim vecTcNm =
+  Prim (PrimInfo "Clash.Sized.Vector.last" (vecHeadTy vecTcNm) WorkNever)
+
+vecHeadTy
```
alex-mckenna, line 110: Comment with the type signature above this so it's more obvious that this is correct for `vecLastPrim`. There's a slight dissonance seeing `primA` have type `primBType`

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
```diff
@@ -73,6 +79,172 @@ import           Clash.Rewrite.Types
...
+  Prim (PrimInfo "Clash.Sized.Vector.tail" (vecTailTy vecTcNm) WorkNever)
+
+vecInitPrim
+  :: TyConName
+  -- ^ Vec TyCon name
+  -> Term
+vecInitPrim vecTcNm =
+  Prim (PrimInfo "Clash.Sized.Vector.init" (vecTailTy vecTcNm) WorkNever)
+
+vecTailTy
```
alex-mckenna, line 138: Comment this with the type signature too

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
```diff
@@ -73,6 +79,172 @@ import           Clash.Rewrite.Types
...
+  -> Type
+  -- ^ Element type
+  -> Integer
+  -- ^ Length of the vector
+  -> Term
+  -- ^ head of the vector
+  -> Term
+  -- ^ tail of the vector
+  -> Term
+mkVecCons consCon resTy n h t =
```
alex-mckenna, line 222: Should this check if `n > 0`?

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
```diff
@@ -829,34 +829,37 @@ reduceLast inScope n aTy vArg = do
...
+ where
+  go tcm (coreView1 tcm -> Just ty') = go tcm ty'
+  go tcm (tyView -> TyConApp vecTcNm _)
+    | (Just vecTc) <- lookupUniqMap vecTcNm tcm
+    , nameOcc vecTcNm == "Clash.Sized.Vector.Vec"
+    , [nilCon, consCon]  <- tyConDataCons vecTc
+    = case n of
+        0 -> mkVecNil nilCon aTy
+        _ ->
+          let
```
alex-mckenna, line 850: `if n == 0 then mkVecNil nilCon aTy else let ...` ?

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -520,31 +521,48 @@ liftAndSubsituteBinders inScope toLift toKeep body = do
...
+    -- Things like clock or reset generator always perform work
+    WorkAlways -> pure False
+  Lam _ e -> andM [isWorkFree e, allM isWorkFreeArg args]
+  TyLam _ e -> andM [isWorkFree e, allM isWorkFreeArg args]
   Letrec bs e ->
-    isWorkFree e && all (isWorkFree . snd) bs && all isWorkFreeArg args
-  Case s _ [(_,a)] -> isWorkFree s && isWorkFree a && all isWorkFreeArg args
-  Cast e _ _       -> isWorkFree e && all isWorkFreeArg args
-  _                -> False
+    andM [isWorkFree e, allM (isWorkFree . snd) bs, allM isWorkFreeArg args]
```
alex-mckenna, line 557: I suspect this might loop on recursive let, have you tested it?

martijnbastiaan: I recursive let should end up in `Var` and then exit on the `isLocalId` branch. But! I think it can loop with self-recursive global binders. A even put a TODO in..:

```
-- TODO: Will this get stuck for self-recursive binders?
```

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
martijnbastiaan: yeah that's probably a good idea!

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -520,31 +521,48 @@ liftAndSubsituteBinders inScope toLift toKeep body = do
...
-  -> Bool
+  -> RewriteMonad extra Bool
 isWorkFree (collectArgs -> (fun,args)) = case fun of
-  Var i            -> isLocalId i && not (isPolyFunTy (varType i))
-  Data {}          -> all isWorkFreeArg args
-  Literal {}       -> True
+  Var i ->
+    if | isPolyFunTy (varType i) -> pure False
+       | isLocalId i -> pure True
+       | otherwise -> isWorkFreeBinder i
```
christiaanb, line 543: Wait... the arguments applied to a global function should be work-free as well. And actually, it's more complicated than that. For example, if the global functions is (isomorphic to) `const`, then we don't have to look at the work of the arguments; but for global functions (isomorphic to) `id` we do need to look at the arguments. I guess the best we can do is a heuristic, and just say the arguments have to be work-free. Given we already do something similar Lambda.

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -1136,6 +1183,13 @@ bindPureHeap tcm heap rw (TransformContext is0 hist) e = do
         ty = termType tcm term
         nm = mkLocalId ty (mkUnsafeSystemName "x" uniq) -- See [Note: Name re-creation]
 
+    inlineTest _ (i, stripTicks -> e_) =
+      if isLocalVar e_ then
+        -- Don't inline `let x = x in x`, it throws  us in an infinite loop
```
christiaanb, line 1188: inlineBinders already does this check: https://github.com/clash-lang/clash-compiler/blob/3762801bcfe4be851639fadbc46170a493f3bee4/clash-lib/src/Clash/Rewrite/Util.hs#L474

clash-lib/src/Clash/Rewrite/Util.hs
martijnbastiaan: Is that true for local ids too then?

christiaanb: Yes, but for local function we just always assume they perform work, so we don't need to look at their arguments. That's why there's the `isPolyFunTy` guard.

clash-lib/src/Clash/Rewrite/Util.hs
martijnbastiaan: Fixed.

clash-lib/src/Clash/Rewrite/Util.hs
martijnbastiaan: Alright, implemented the heuristic. Thanks.

martijnbastiaan:
Testing on a private, bigger codebase revealed a 20% speedup @christiaanb :). I've applied the review suggestions, so I think this is good to go. (If merging, it's probably best to create a merge commit for this one, as there's quite a lot of info in this thread.)

christiaanb:
Any reason not to backport to 1.2?

martijnbastiaan:
It's a bit of work, most of which I've done in `backport-workaround1240`, but it needs updating to include the latest force pushes.

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
```diff
@@ -71,6 +79,183 @@ import           Clash.Rewrite.Types
...
+  -- ^ Element type
+  -> Integer
+  -- ^ Length of the vector
+  -> Term
+  -- ^ Vector to extract head from
+  -> (Term, Term)
+  -- ^ (head of vector, tail of vector)
+extractHeadTail consCon elTy n vec =
+  ( Case vec elTy [(pat, Var el)]
+  , Case vec restTy [(pat, Var rest)] )
```
christiaanb, line 175: @martijnbastiaan  Doesn't this duplicate (at least compiler) work for the `vec` term?

martijnbastiaan: Possibly / probably. How would you write it?

There should also be a "quick" path that just extract the element if it sees a data constructor.

PR #1365, opened 2020-06-09
VHDL: automatically generate a type qualification inside ~TOBV

### LINKED ISSUE #1360: Problem synthesising toSLV for sum types in VHDL with Vivado with -fclash-hdlsyn Vivado
Please refer to the following gist for code:
https://gist.github.com/xavieran/f7aa0409bbb42ed4550b553b276e41ac

When compiling this clash 
```
data StatusA = InactiveA | ActiveA

topEntity :: HiddenClockResetEnable System =>
  Signal System (Vec 2 StatusA) -> Signal System (Index 2) -> Signal System (Vec 2 StatusA)
topEntity input idx = out
  where
    out = replace <$> idx <*> pure InactiveA <*> input
```
with -fclash-hdlsyn Vivado, the "replace" function sets the value of the Vec to 
`ivec(vec_index) := test_types.toSLV("0");`
instead of 
`ivec(vec_index) := "0";`
without -fclash-hdlsyn Vivado

This causes an error during synthesis with Vivado (2019.2)
```
ERROR: [Synth 8-2380] near toslv ; 3 visible identifiers match here [~/vhdl/Test/topentity.vhdl:53]
INFO: [Synth 8-1047] first match for 'toslv' found here [~/vhdl/Test/test_types.vhdl:22]
INFO: [Synth 8-1047] another match for 'toslv' found here [~/vhdl/Test/test_types.vhdl:26]
INFO: [Synth 8-1047] another match for 'toslv' found here [~/vhdl/Test/test_types.vhdl:30]
INFO: [Synth 8-2810] unit structural ignored due to previous errors [~/vhdl/Test/topentity.vhdl:22]
```

It seems that toSLV is ambiguous when applied to the "0".

leonschoorl:
This makes sure VHDL always knows which `toSLV `to call.
Fixes #1360

This also leave out `to`- and `fromSLV` calls when the type is already `std_logic_vector`.
And fixes the layout of the entity declaration `end` keyword.

### DIFF
--- a/clash-lib/src/Clash/Backend/VHDL.hs
+++ b/clash-lib/src/Clash/Backend/VHDL.hs
@@ -135,8 +135,13 @@ instance Backend VHDLState where
     | isBV t = pretty id_
     | otherwise = do
       nm <- Mon $ use modNm
-      -- TODO It would be nice to leave out the type qualification if we know id_ is a var
-      pretty (TextS.toLower nm) <> "_types.toSLV" <> parens (hdlTypeMark t <> squote <> parens (pretty id_))
+      seen <- use seenIdentifiers
+      -- This is a bit hacky, as id_ is just a rendered expression.
+      -- But if it's a bare identifier that we've seen before,
+      -- then this identifier has a defined type and we can skip the explicit type qualification.
+      let e | T.toStrict id_ `HashMapS.member` seen = pretty id_
+            | otherwise = hdlTypeMark t <> squote <> parens (pretty id_)
+      pretty (TextS.toLower nm) <> "_types.toSLV" <> parens e
   fromBV t id_
     | isBV t = pretty id_
     | otherwise = do
--- a/clash-lib/src/Clash/Backend/VHDL.hs
+++ b/clash-lib/src/Clash/Backend/VHDL.hs
@@ -131,12 +131,17 @@ instance Backend VHDLState where
   inst            = inst_
   expr            = expr_
   iwWidth         = use intWidth
-  toBV _ id_      = do
-    nm <- Mon $ use modNm
-    pretty (TextS.toLower nm) <> "_types.toSLV" <> parens (pretty id_)
-  fromBV t id_  = do
-    nm <- Mon $ use modNm
-    qualTyName t <> "'" <> parens (pretty (TextS.toLower nm) <> "_types.fromSLV" <> parens (pretty id_))
+  toBV t id_
+    | isBV t = pretty id_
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -131,12 +131,17 @@ instance Backend VHDLState where
...
-    nm <- Mon $ use modNm
-    pretty (TextS.toLower nm) <> "_types.toSLV" <> parens (pretty id_)
-  fromBV t id_  = do
-    nm <- Mon $ use modNm
-    qualTyName t <> "'" <> parens (pretty (TextS.toLower nm) <> "_types.fromSLV" <> parens (pretty id_))
+  toBV t id_
+    | isBV t = pretty id_
+    | otherwise = do
+      nm <- Mon $ use modNm
+      -- TODO It would be nice to leave out the type qualification if we know id_ is a var
```
alex-mckenna, line 138: Is checking whether `id_` is in the `seenIdentifiers` map a way to check if `id_` is a var?

testsuite/Main.hs
```diff
@@ -618,6 +618,7 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "VecOfSum" def{hdlSim=False}
         , runTest "T452" def{hdlSim=False}
         , runTest "T895" def{hdlSim=False,hdlTargets=[VHDL]}
+        , runTest "T1360" def{hdlSim=False, hdlTargets=[VHDL], clashFlags=["-fclash-hdlsyn", "Vivado"]}
```
alex-mckenna, line 621: Should this be tested with other flags than `-fclash-hdlsyn Vivado`? I know the issue was only mentioning Vivado, but should this hold for `Quartus` or `Other`?

clash-lib/src/Clash/Backend/VHDL.hs
leonschoorl: `id_ :: Text`
So we can't really do this without changing the API of `toBV`, which I didn't want to do for this PR.

alex-mckenna: `type Identifier = Text`

christiaanb: I think @alex-mckenna is right, and we can just check the `id_` against `seenIdentifier`. If the expressions rendered to a string corresponds to a `seenIdentifier` while it shouldn't have then that's a whole different can of worms anyway.

PR #1383, opened 2020-06-15
Don't warn about statically linked Clash on Windows

martijnbastiaan:
Dynamic linking not supported on Windows

### DIFF
--- a/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
+++ b/clash-ghc/src-bin-common/Clash/GHCi/Common.hs
@@ -33,6 +33,7 @@ import qualified GHC.LanguageExtensions as LangExt (Extension (..))
 import           Panic                  (GhcException (..), throwGhcException)
 
 import           Control.Monad          (forM_, unless, when)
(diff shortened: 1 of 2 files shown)


clash-ghc/src-bin-common/Clash/GHCi/Common.hs
```diff
@@ -114,7 +115,7 @@ checkClashDynamic dflags = do
   let isStatic = case lookup "GHC Dynamic" (DynFlags.compilerInfo dflags) of
         Just "YES" -> False
         _          -> True
-  when isStatic
```
christiaanb, line 117: This seems wrong... shouldn't we check for the os (`System.Info.os`) on which we run, not the one we build with.

christiaanb: Especially if we ever to decide to crosscompile.

martijnbastiaan: Sorry for merging. Yes you're right.

PR #1400, opened 2020-06-22
Generate SDC files for top entity

### LINKED ISSUE #1387: Generate SDC file with clock specifications
Clash should generate an SDC file with `create_clock` commands for all `Clock` arguments of functions that are a synthesis boundary (`topEntity` or functions with a `Synthesize` annotation). For example, for https://github.com/clash-lang/clash-compiler/blob/master/examples/Blinker.hs  Clash should generate a file called `blinker.sdc` with the following content:
```tcl
create_clock -name {CLOCK_50} -period 20.000 -waveform { 0.000 10.000 } [get_ports {CLOCK_50}]
```

alex-mckenna:
Fixes #1387

This PR uses createHDL to generate SDC files for each component
with a clock as an input. If a component has multiple clocks as an
input, each clock will appear in the SDC file.

### DIFF
--- a/clash-lib/src/Clash/Driver.hs
+++ b/clash-lib/src/Clash/Driver.hs
@@ -2,8 +2,10 @@
   Copyright   :  (C) 2012-2016, University of Twente,
                      2016-2017, Myrtle Software Ltd,
                      2017     , QBayLogic, Google Inc.
+                     2020       QBayLogic
+
   License     :  BSD2 (see the file LICENSE)
-  Maintainer  :  Christiaan Baaij <christiaan.baaij@gmail.com>
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 
   Module that connects all the parts of the Clash compiler library
 -}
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Driver.hs
```diff
@@ -757,7 +760,22 @@ createHDL backend modName seen components _domainConfs top (topName,manifestE) =
     ) manifestE
   let manDoc = ( Data.Text.unpack topName <.> "manifest"
                , pretty (Text.pack (show manifest)))
-  return (manDoc:topFiles,manifest,dataFiles,memFiles)
+
+  let sdcDocs = flip mapMaybe components $ \c ->
```
christiaanb, line 764: You don't want it for all the components, just for the TopEntity.

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -269,5 +274,33 @@ data Manifest
...
+--
+pprSDC :: SdcInfo -> Doc ()
+pprSDC = vcat . fmap go . sdcClock
+ where
+  go (i, dom) =
+        -- VDomainConfiguration stores period in ps, SDC expects it in ns.
+    let p        = vPeriod dom `div` 1000
+        name     = braces (pretty i)
+        period   = pretty p
+        waveform = braces ("0" <+> pretty (p `div` 2))
```
christiaanb, line 297: I might have 5000ps period, i.e 5ns, I then want the the half period to be 2.5ns, so `div 2` won't do. We want to pretty print a `Real` (or `Float`) with 3 decimals of precision.

PR #1409, opened 2020-06-26
Use readMaybe to unpack domain configuration

alex-mckenna:
For @martijnbastiaan

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -55,6 +55,7 @@ import           Debug.Trace
 import           Language.Haskell.TH.Syntax      (lift)
 import           GHC.Natural                     (naturalFromInteger)
 import           GHC.Stack                       (HasCallStack)
+import           Text.Read                       (readMaybe)
 
 #ifdef USE_GHC_PATHS
 import           GHC.Paths                       (libdir)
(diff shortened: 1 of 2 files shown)


clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -521,21 +522,10 @@ unpackKnownConf ty
...
-  asInitBehaviour "Unknown" = Unknown
-  asInitBehaviour "Defined" = Defined
-  asInitBehaviour x = error $ $(curLoc) ++ "Unknown init behaviour: " ++ show x
-
-  asResetPolarity "ActiveHigh" = ActiveHigh
-  asResetPolarity "ActiveLow"  = ActiveLow
-  asResetPolarity x = error $ $(curLoc) ++ "Unknown reset polarity: " ++ show x
+  asActiveEdge x = fromMaybe (error $ $(curLoc) ++ "Unknown active edge: " ++ show x) (readMaybe x)
+  asResetKind x = fromMaybe (error $ $(curLoc) ++ "Unknown reset kind: " ++ show x) (readMaybe x)
+  asInitBehaviour x = fromMaybe (error $ $(curLoc) ++ "Unknown init behaviour: " ++ show x) (readMaybe x)
```
martijnbastiaan, line 527: Ofc it shouldn't happen, but this might end up at a user's screen. Hence, we should use :us: spelling.

martijnbastiaan: As this function is partial, I think it should have a `HasCallStack` constraint.

PR #1416, opened 2020-07-01
Run clashLibTests serially

martijnbastiaan:
'runghc' puts temporary files next to the source files, with no apparent
    way to change this behavior. This causes race conditions in our CI for
    netlist tests.

### DIFF
--- a/testsuite/src/Test/Tasty/Clash.hs
+++ b/testsuite/src/Test/Tasty/Clash.hs
@@ -281,7 +281,7 @@ sequenceTests path (unzip -> (testNames, testTrees)) =
     where
       -- Make pattern for a single test
       pat :: TestName -> String
-      pat nm = "$0 == \"" ++ intercalate "." (reverse (nm:path)) ++ "\""
+      pat nm = "$0 ~ /" ++ intercalate "." (reverse (nm:path)) ++ "/"
 
       -- Test patterns for all given tests such that each executes sequentially
       testPatterns = init (map (fmap pat) (Nothing : map Just testNames))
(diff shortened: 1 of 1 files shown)


alex-mckenna:
Does `outputTest` not race for the same reason? That uses `runghc` as well

martijnbastiaan:
You'd think so, but I have never seen it error on those!

alex-mckenna:
_Twilight Zone theme plays_

martijnbastiaan:
Ah so maybe it's Clash, not runghc! Tnx.

testsuite/src/Test/Tasty/Clash.hs
```diff
@@ -796,8 +796,11 @@ clashLibTest
   -- one closest to the test.
   -> TestTree
 clashLibTest env targets extraGhcArgs modName funcName path =
-  let testName = modName ++ " [clash-lib test]" in
-  let path' = testName : path in
-  testGroup testName
-    [clashLibTest' env target extraGhcArgs modName funcName path' | target <- targets]
-
+  -- HACK: clashLibTests are run sequencially to prevent race issues. See:
```
alex-mckenna, line 799: `s/sequencially/sequentially/`

martijnbastiaan: done tnx

testsuite/src/Test/Tasty/Clash.hs
```diff
@@ -281,7 +281,7 @@ sequenceTests path (unzip -> (testNames, testTrees)) =
     where
       -- Make pattern for a single test
       pat :: TestName -> String
-      pat nm = "$0 == \"" ++ intercalate "." (reverse (nm:path)) ++ "\""
+      pat nm = "$0 ~ /" ++ intercalate "." (reverse (nm:path)) ++ "/"
```
alex-mckenna, line 284: `sequenceTests` is used in multiple places, does this mean all types of test are now run sequentially?

martijnbastiaan: No, just the ones that were being run sequentially before (`clash` -> `build hdl` -> `simulate hdl`) and where I added `sequenceTest` in this PR. We could add it to `outputTest` too, but I haven't seen it fail ever.

martijnbastiaan: You can test it by running the testsuite and see all your cores fire up :cat:

alex-mckenna: Ah, so tests are in parallel, but steps within a test (run clash, run X, run Y, ...) are sequential?

martijnbastiaan: Yes. And for the clashLibTests it has the same behavior at the `VHDL`/`Verilog`/`SystemVerilog` level.

alex-mckenna: Sounds good to me

PR #1419, opened 2020-07-03
Basic Termination Analysis and Fuel

alex-mckenna:
The partial evaluator now performs incredibly limited termination
analysis, by taking into consideration whether a global binding is
non-recursive when determining whether to inline. This means that
any non-recursive binding can always be unfolded.

This commit also introduces fuel as a way to limit the number of
times a potentially non-terminating function can be unfolded, with
global bindings being ignored if there is no fuel available.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
@@ -151,6 +151,11 @@ generateBindings useColor primDirs importDirs dbs hdl modName dflagsM = do
          , domainConfs
          )
 
+-- TODO This function should be changed to provide the information that
+-- Clash.Core.Termination.mkRecInfo provides. To achieve this, it should also
+-- be changed to no longer flatten recursive groups (see the documentation for
+-- mkRecInfo for an explanation of these).
+--
 mkBindings
   :: CompiledPrimMap
   -> [GHC.CoreBind]
--- a/clash-lib/clash-lib.cabal
+++ b/clash-lib/clash-lib.cabal
@@ -143,6 +143,7 @@ Library
 
   Build-depends:      aeson                   >= 0.6.2.0  && < 1.5,
                       ansi-terminal           >= 0.8.0.0  && < 0.11,
+                      array,
                       attoparsec              >= 0.10.4.0 && < 0.14,
                       base                    >= 4.11     && < 5,
                       binary                  >= 0.8.5    && < 0.11,
@@ -209,6 +210,7 @@ Library
                       Clash.Core.TermInfo
                       Clash.Core.TermLiteral
                       Clash.Core.TermLiteral.TH
+                      Clash.Core.Termination
                       Clash.Core.TyCon
                       Clash.Core.Type
                       Clash.Core.TysPrim
--- a/clash-lib/src/Clash/Core/Evaluator/Models.hs
+++ b/clash-lib/src/Clash/Core/Evaluator/Models.hs
@@ -22,8 +22,6 @@ import Data.Bifunctor (first, second)
 import Data.IntMap.Strict (IntMap)
 import Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
-import Data.Set (Set)
-import qualified Data.Set as Set
 import qualified Data.Set.Lens as Set
 
 import Clash.Core.DataCon
@@ -32,6 +30,7 @@ import Clash.Core.Literal
 import Clash.Core.Name (OccName)
 import Clash.Core.Subst (extendTvSubstList, mkSubst, substTm)
 import Clash.Core.Term
+import Clash.Core.Termination
 import Clash.Core.TyCon (TyConMap)
 import Clash.Core.Type
 import Clash.Core.Util (mkUniqSystemId, mkUniqSystemTyVar)
@@ -69,7 +68,7 @@ newtype Eval a = Eval { unEval :: RWS LocalEnv () GlobalEnv a }
 --
 runEval :: GlobalEnv -> Eval a -> (a, GlobalEnv)
 runEval genv x =
-  let lenv = LocalEnv mempty mempty mempty
+  let lenv = LocalEnv mempty mempty
       (x', genv', _) = RWS.runRWS (unEval x) lenv genv
    in (x', genv')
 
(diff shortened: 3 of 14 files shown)


clash-lib/clash-lib.cabal
```diff
@@ -158,6 +159,7 @@ Library
                       errors                  >= 1.4.2    && < 2.4,
                       exceptions              >= 0.8.3    && < 0.11.0,
                       extra                   >= 1.6.18   && < 1.8,
+                      fgl,
```
martijnbastiaan, line 162: `dd` :-)

alex-mckenna: Gone!

clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
```diff
@@ -142,6 +143,9 @@ generateBindings useColor primDirs importDirs dbs hdl modName dflagsM = do
   let prepStartDiff = reportTimeDiff prepTime startTime
   putStrLn $ "Clash: Parsing and compiling primitives took " ++ prepStartDiff
 
+  -- TODO: Just a temporary measure to see the RecInfo
+  print (mkRecInfo allBindings)
+
```
martijnbastiaan, line 148: ^

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -173,10 +159,12 @@ data GlobalEnv = GlobalEnv
   { genvGlobals :: VarEnv (Binding TermOrValue)
     -- ^ Global term environment. These are functions in global scope which
     -- are evaluated on lookup, and updated after evaluation.
-  , genvIgnore  :: Set Id
-    -- ^ Global ignore list. These are global identifiers which are not
-    -- inlined while evaluating. Items are added / removed during sections
-    -- where inlining would lead to non-termination.
+  , genvRecInfo :: RecInfo
+    -- ^ Information about which global bindings are recursive, used to decide
+    -- whether or not to inline a global binding.
```
martijnbastiaan, line 164: Should we store this on the binder itself? Does your discussion with @christiaanb change any of this?

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -173,10 +159,12 @@ data GlobalEnv = GlobalEnv
...
+  , genvFuel    :: Word
```
martijnbastiaan, line 165: No need to align `::`

clash-lib/src/Clash/Core/Evaluator/Models.hs
```diff
@@ -232,27 +213,26 @@ updateGlobal i x =
...
-  res <- x
-  env <- RWS.get
+preserveState :: Eval a -> Eval a
+preserveState x = do
+  fuel <- RWS.gets genvFuel
+  res  <- x
+  env  <- RWS.get
 
-  RWS.put (env { genvIgnore = is })
+  RWS.put (env { genvFuel = fuel })
```
martijnbastiaan, line 235: Is it a good idea to call this `preserveState` when it only preservers fuel?

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -167,20 +168,40 @@ evaluateVarWith eval i
     Just x  -> forceLocal x
     Nothing -> pure (VNeu (NeVar i))
    where
+    -- TODO Does this need to change?
```
martijnbastiaan, line 171: I don't know :-). Do you still need to look at this?

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
```diff
@@ -167,20 +168,40 @@ evaluateVarWith eval i
...
+
+        -- The binding can be inlined if there is enough fuel.
+        |  isRecursive i gRecInfo
+        -> if gFuel == 0
+             then pure (VNeu (NeVar i))
+             else do
+               v <- forceGlobal (bindingTerm b)
+               putFuel (gFuel - 1)
+               updateGlobal i v
+               pure v
```
martijnbastiaan, line 193: Should fuel be kept per binding instead? I don't have strong feelings either way, but I'm wondering about the reasoning (easy to implement being a valid reason :-)).

clash-lib/src/Clash/Core/Termination.hs
```diff
@@ -0,0 +1,95 @@
...
+import Clash.Driver.Types
+
+-- Quick lookup for whether a binding is recursive or non-recursive. If a
+-- binding is non-recursive, we can assume that it terminates and skip
+-- analysing it.
+--
+data RecInfo = RecInfo
+  { recBindings    :: [VarSet]
+    -- ^ Recursive bindings, organized into groups of strongly connected
+    -- components.
```
martijnbastiaan, line 27: Clash Core only has self-recursive binders, not mutually recursive ones. I believe that happens over here: 

https://github.com/clash-lang/clash-compiler/blob/305427f3e2bb44669ae971c4d0e34da6a7818ed0/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs#L154-L197

christiaanb: To say that Clash Core only has self-recursive bindings is a bit of a stretch. We ensure that the list of bindings given to `generateHDL` only has self-recursive bindings by turning any recursive group:
```haskell
f = <fBody>
g = <gBody>
h = <hBody>
```
into:
```haskell
f = let g = <gBody>; h = <hBody> in <fBody>
g = let f = <fBody>; h = <hBody> in <gBody>
h = let f = <fBody>; g = <gBody> in <hBody>
```
and then try to take great care that we never introduce mutually recursive global bindings as we perform all the normalization transformations. I would be preferable if we actually dealt with mutually recursive bindings properly; so this is actually a step in the right direction.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
christiaanb: Or at least per recursive group. Although it becomes problematic when we evaluate a binding from one recursive group inside the binding of another recursive group.

clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs
christiaanb: I think this should have a TODO that we eventually want to get this information from:
https://github.com/clash-lang/clash-compiler/blob/305427f3e2bb44669ae971c4d0e34da6a7818ed0/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs#L154-L197
And that we stop flattening recursive groups.

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
alex-mckenna: I did initially think of per-binder / per-group but went with this because (1) it was more straightforward to implement and (2) it felt slightly easier to reason about the result, as all paths from root to leaf of the final AST use at most n fuel

A consequence is that programs with wide ASTs may incur a performance penalty because of this, but I would hope that can be spotted / tuned later

clash-lib/src/Clash/Core/Evaluator/Models.hs
alex-mckenna: I changed it to `preserveState` when it had both fuel and the ignore list, I guess `preserveFuel` is a better name for the end result though. I don't think anything else would need to be saved / restored later...

clash-lib/src/Clash/Core/Evaluator/Semantics.hs
alex-mckenna: I don't think this needs to change, so I'm going to take out the TODO

alex-mckenna:
See #1424 for identifying recursive groups in `mkBindings`.
@martijnbastiaan, are you happy with this PR?

martijnbastiaan:
@alex-mckenna jep :)

PR #1433, opened 2020-07-08
Put the edalize output in a new directory

christiaanb:
Polluting the directory that has the generated HDL files is not ideal

### DIFF
--- a/clash-lib/src/Clash/Edalize/Edam.hs
+++ b/clash-lib/src/Clash/Edalize/Edam.hs
@@ -65,7 +65,7 @@ data EdamFile = EdamFile
 pprFile :: EdamFile -> Doc ann
 pprFile (EdamFile n ty ln) =
   pyRecord
-    [ pyField "name" $ relPath n
+    [ pyField "name" $ joinPath n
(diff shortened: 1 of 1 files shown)


clash-lib/src/Clash/Edalize/Edam.hs
```diff
@@ -303,10 +305,10 @@ pyField n x = hcat [squotes (pretty n), colon, space, x]
 pyDocField :: Text -> Text -> Doc ann -> Doc ann
 pyDocField n d x = vsep [pyComment d, pyField n x]
 
-relPath :: FilePath -> Doc ann
-relPath x = hcat
-  [ "os.path.relpath"
-  , parens (hsep $ punctuate comma [squotes (pretty x), "work_root"])
+joinPath :: FilePath -> Doc ann
+joinPath x = hcat
+  [ "os.path.joinpath"
```
alex-mckenna, line 310: This function doesn't exist (at least in the python version on my work computer). Did you mean `os.path.join` or `joinpath` from `pathlib`?

christiaanb: I meant `join`

PR #1437, opened 2020-07-10
Add `PrintfArg` instance (going via `Integer`) for sized integral types

gergoerdi:
This is useful for simulation, especially with hexadecimal formats like `"%04x"` to print a 0-padded 4-hex-digit value.


martijnbastiaan:
@gergoerdi If you rebase on `master` the CI should pass!

PR #1440, opened 2020-07-10
Prevent name clashes with VHDL imports

alex-mckenna:
When generating VHDL, clash no longer generates component names
which clash with identifiers from imported modules.


martijnbastiaan:
Cool. Do we have the same issues for (System)Verilog?

alex-mckenna:
I don't think so, the only reason it was an issue for VHDL is because we import things from modules and didn't consider how that changed the namespace. Verilog / SystemVerilog don't have any extra imports in the generated code to worry about

christiaanb:
In verilog namespacing is solved *externally* 🤪🙈🙉

https://www.cc.gatech.edu/~hadi/teaching/cs3220/doc/verilog/verilog_2001_ref_guide.pdf#page=47

PR #1452, opened 2020-07-13
Add benchmarking to CI

martijnbastiaan:
Publishes results as GitLab CI artificats

-------------------------

The plan is to push the benchmark results to https://github.com/clash-lang/clash-benchmark-results, but I haven't found a safe way of doing so (read: without exposing secrets to random people). Maybe by running a nightly job pointed to master, gathering and pushing these results?

### DIFF
--- a/.ci/gitlab/benchmark.yml
+++ b/.ci/gitlab/benchmark.yml
@@ -1,0 +1,42 @@
+.benchmark:
+  image: clashlang/clash-ci:2020-06-29
+  stage: build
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
@leonschoorl Let's sit down after the scrum :)

leonschoorl:
We should try to make something like: https://github.com/matchai/criterion-compare-action

A bot that comments on PRs with what changes that PR has on the benchmark results

PR #1456, opened 2020-07-16
Store seen PrimInfo in HashMap for lookup

alex-mckenna:
When translating from GHC core to Clash core, any PrimInfo which is
constructed is stored in a map indexed by the name of the primitive.
This can be used later in the evaluator to lookup PrimInfo rather
than creating it ad-hoc where it is needed. The benefits being:

  * no chance to construct a primitive with the wrong type / workinfo
  * more succinct code in the evaluator when building prims

### DIFF
--- a/Clash.hs
+++ b/Clash.hs
@@ -49,12 +49,12 @@ doHDL
 doHDL b src = do
   startTime <- Clock.getCurrentTime
   pd      <- primDirs b
-  (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs,domainConfs) <- generateBindings Auto pd ["."] [] (hdlKind b) src Nothing
+  (bindingsMap,tcm,tupTcm,topEntities,primMap,reprs,domainConfs,primInfos) <- generateBindings Auto pd ["."] [] (hdlKind b) src Nothing
   prepTime <- startTime `deepseq` bindingsMap `deepseq` tcm `deepseq` reprs `deepseq` Clock.getCurrentTime
   let prepStartDiff = reportTimeDiff prepTime startTime
   putStrLn $ "Loading dependencies took " ++ prepStartDiff
 
-  generateHDL (buildCustomReprs reprs) domainConfs bindingsMap (Just b) primMap tcm tupTcm
+  generateHDL (buildCustomReprs reprs) domainConfs primInfos bindingsMap (Just b) primMap tcm tupTcm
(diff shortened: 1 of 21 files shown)


martijnbastiaan:
Is it possible / wouldn't it be easier to store this on the binder instead?

alex-mckenna:
As in, store it in the `BindingMap`? It makes more sense to me to store it once and just say _here's the info for all primitives in this design_. For my use case in the partial evaluator it's definitely easier (as a caller) to just get the map in one go than have to keep the union of every `PrimInfo` used in every inlined binding.

christiaanb:
But now you only get the primitives you see in a compile run. You will still need to construct it manually when you didn’t get to see it

alex-mckenna:
Are there any places we construct a primitive we haven't seen before? (excluding things like `GHCCore2Core`)

alex-mckenna:
@christiaanb, for reference my main use case in this is not wanting to have to manually construct `PrimInfo` in the evaluator. IMO it leads to a lot of line noise and potential mistakes which would be easily avoidable if we could look up `PrimInfo` during evaluation

christiaanb:
I get the intention of the PR, but then we create only a couple of PrimInfo's out of thin air:

- GHC.Prim.MutableByteArray#
- Clash.Sized.Internal.Index.+#
- Clash.Sized.Internal.BitVector.fromInteger##
- Clash.Sized.Internal.BitVector.fromInteger#
- Clash.Sized.Internal.Index.fromInteger#
- Clash.Sized.Internal.Signed.fromInteger#
- Clash.Sized.Internal.Unsigned.fromInteger#
- Clash.Sized.Vector.splitAt
- Clash.Sized.Vector.++
- Clash.Sized.Vector.zipWith
- Clash.Sized.Internal.BitVector.++#
- Clash.Sized.Internal.BitVector.split#

And two special ones:

- Clash.Sized.Vector.imap_go
- Clash.Sized.Vector.fold_split

Where those special two will never be found through a traversal of the current program, as they're internal normally to `Vector.imap` and `Vector.fold`.

The above aside, I think it would make more sense to lookup the info directly, instead of hoping there will never be a case where you wouldn't find them in a traversal of the program but would still need them for evaluation. You'll have to use the GHC API to lookup the GHC types of the primitive (and then translate them to Clash types), and the `WorkInfo` is already stored in the `PrimMap`.

PR #1462, opened 2020-07-27
Preprocess JSON files to be compatible with aeson-1.5

martijnbastiaan:
Aeson versions <1.5 accept unescaped newline characters in JSON strings.
For example, it would accept this non-RFC7159-compliant JSON:

    { "a": "foo
            bar" }

The primitive definitions in `clash-lib/prims/` (ab)use this fact. Aeson
1.5 fixed this oversight though, hence Clash breaking on this version.
This patch preprocesss the JSON files such that they're standards-
compliant.

### DIFF
--- a/clash-lib/src/Data/Aeson/Extra.hs
+++ b/clash-lib/src/Data/Aeson/Extra.hs
@@ -4,13 +4,22 @@
   Maintainer  :  Christiaan Baaij <christiaan.baaij@gmail.com>
 -}
 
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE OverloadedStrings #-}
+
 module Data.Aeson.Extra where
 
 import           Control.Exception    (throw)
+import qualified Data.IntMap          as IntMap
+import           Data.IntMap          (IntMap)
 import qualified Data.Ix              as Ix
 import qualified Data.Text            as T
 import           Data.Text            (Text,pack,unpack)
-import           Data.List            (intercalate)
+import qualified Data.Text.Lazy       as LT
+import qualified Data.Text.Lazy.Encoding as LT
+import           Data.Text.Encoding.Error (UnicodeException(..))
+import           Data.List            (intercalate, groupBy)
+import           Data.Tuple.Extra     (second)
 import           Data.Aeson           (FromJSON, Result (..), fromJSON, json)
 import           Data.Attoparsec.Lazy (Result (..), parse)
 import           Data.ByteString.Lazy (ByteString)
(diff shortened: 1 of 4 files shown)


clash-lib/src/Data/Aeson/Extra.hs
```diff
@@ -4,47 +4,118 @@
...
+    $ groupBy (\a b -> fst a == fst b)
+    $ zip virtuals [(0::Int)..]
+
+genLineErr' :: [Text] -> (Int, Int) -> [Int] -> Text
+genLineErr' allLines range errorLines =
+  T.unlines [ T.concat [ if elem i errorLines then pack ">> " else  pack "   "
+            , pack $ show (i + 1)
+            , pack ". "
+            , allLines !! i
+            ] | i <- Ix.range range]
```
leonschoorl, line 101: I'd suspect doing `drop` and `take` on `allLines`  would be slightly faster and more "proper" FP.
But since this code isn't on any critical path it doesn't matter

clash-lib/src/Data/Aeson/Extra.hs
```diff
@@ -4,47 +4,118 @@
...
+  -> Either UnicodeException (LineMap, ByteString)
+toSpecNewlines bs = do
+ s0 <- LT.unpack <$> LT.decodeUtf8' bs
+ Right ( toLineMap (go2 0 False s0)
+       , LT.encodeUtf8 (LT.pack (go False s0)))
+ where
+  -- replace newlines with escaped ones
+  go :: Bool -> String -> String
+  go _ [] = []
+  go True ('\n':rest) = '\\' : 'n' : go True rest
```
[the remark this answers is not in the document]
martijnbastiaan: Good point!

clash-lib/src/Data/Aeson/Extra.hs
```diff
@@ -4,47 +4,118 @@
...
 import           GHC.Stack            (HasCallStack)
 
--- Quick and dirty way of replacing fake escapes in naively converted bytestring
-replaceCommonEscapes :: Text -> Text
-replaceCommonEscapes = ( T.replace (pack "\\n") (pack "\n") ) .
-                       ( T.replace (pack "\\\\") (pack "\\") ) .
-                       ( T.replace (pack "\\\"") (pack "\"") )
+-- | See 'toSpecNewlines'. A line map maps "virtual" lines to a range of
+-- "real" lines.
+type LineMap = IntMap (Int, Int)
```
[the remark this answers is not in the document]
martijnbastiaan: It maps a virtual line to a range of real lines, I'll rework the documentation :)

clash-lib/src/Data/Aeson/Extra.hs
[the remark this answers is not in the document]
martijnbastiaan: Good point!

clash-lib/src/Data/Aeson/Extra.hs
martijnbastiaan: Yeah I thought about changing it, but wasn't worth the effort for the reason you mention.

clash-lib/src/Data/Aeson/Extra.hs
```diff
@@ -4,47 +4,122 @@
...
-                                                             , pack $ show i
-                                                             , pack ". "
-                                                             , allLines !! i
-                                                             ] | i <- Ix.range range]
+  toLineMap :: [Int] -> LineMap
+  toLineMap virtuals =
+      IntMap.fromList
+    $ map (second (\reals -> (minimum reals, maximum reals)))
+    $ map (\(unzip -> (virts, reals)) -> (head virts, reals))
+    $ groupBy (\a b -> fst a == fst b)
```
alex-mckenna, line 96: Few suggestions:

  * `\(unzip -> (virts, reals)) -> (head virts, reals)` -> `first head. unzip`
  * `groupBy (\a b -> fst a == fst b)` -> `groupOn fst` (from `Data.List.Extra`)

martijnbastiaan: ooh `groupOn`, never seen that before

alex-mckenna: If you really want to invoke your inner perl programmer, you can also just do one `map` instead of two and make it a pointfree one-liner:

```haskell
IntMap.fromList . fmap (bimap head (minimum &&& maximum) . unzip) . groupOn fst
```

martijnbastiaan: I'm going to have to respectfully decline, I don't think that'd make it easier to read! (But I guess that's why you alluded to it being Perl-like like ;))

PR #1473, opened 2020-08-11
Fix SNat logarithm primitives

alex-mckenna:
~The naturalToInteger function has an implementation in the evaluator but was not listed as a blackbox in the Json primitives.~

The real cause of the issue: the logarithm primitives on SNat were assuming the wrong number of type arguments, leading to these prims not being reduced.

### DIFF
--- a/changelog/2020-08-11T15_28_24+02_00_fix-snat-log-prims.md
+++ b/changelog/2020-08-11T15_28_24+02_00_fix-snat-log-prims.md
@@ -1,0 +1,3 @@
+FIXED: flogBaseSNat, clogBaseSNat and logBaseSNat primitives are now implemented correctly.
+Previously these primitives would be left unevaluated causing issues like #1473.
+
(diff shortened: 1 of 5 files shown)


changelog/2020-08-11T15_28_24+02_00_fix-snat-log-prims.md
```diff
@@ -0,0 +1,3 @@
+FIXED: flogBaseSNat, clogBaseSNat and logBaseSNat primitives are now implemented correctly.
```
martijnbastiaan, line 1: Jinxing it ;-)

alex-mckenna: It's only true if you believe :pray:

alex-mckenna:
Should the `comment` field be `warning` instead as with blackboxes?
(And yes, I'll do this if @christiaanb agrees it's the right way to do it)

christiaanb:
Yeah, comment, warning, both are fine. Just have it say something along the lines of: "BlackBox explicitly omitted until we have size-inference for Natural and Integer"

PR #1483, opened 2020-08-20
Use let when using type signatures

martijnbastiaan:
Doctests fail in some configurations if we don't


clash-prelude/src/Clash/XException.hs
```diff
@@ -172,7 +172,7 @@ xToErrorCtx ctx a = unsafeDupablePerformIO
 --
 -- Unlike 'xToErrorCtx', where we have an extra String argument to distinguish
 -- one call to 'xoToError' to the other, 'xToError' will use the 'GHC.CallStack'
--- mechanism to aid the user in distinquishing different call to 'xToError'.
+-- mechanism to aid the user in distinguishing different call to 'xToError'.
```
alex-mckenna, line 175: Good spot :eyes:

alex-mckenna (approved):
LGTM

PR #1484, opened 2020-08-23
Add `nameHint` to `Clash.Magic`

JvWesterveld:
Gives a hint for how to name e.g. a `Signal` in the generated HDL.

My use case is the Vivado integrated logic analyzer GUI. It uses the signal names used in the ILA's port map in VHDL as probe names, so it is nice to be able to give these signals sensible identifiers.

Code is courtesy of @christiaanb.

Should something like this have a test?

### DIFF
--- a/changelog/2020-08-23T19_14_45+02_00_add_nameHint
+++ b/changelog/2020-08-23T19_14_45+02_00_add_nameHint
@@ -1,0 +1,1 @@
+FEATURE: Added `nameHint` to allow explicitly naming terms, e.g. `Signal`s.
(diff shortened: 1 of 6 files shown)


martijnbastiaan (approved):
Looks great, thanks!

martijnbastiaan:
> Should something like this have a test?

You could use a clashLibTest:

https://github.com/clash-lang/clash-compiler/blob/ccebebb4fb2ccf70098b6b75a717baf1775c3a3e/testsuite/Main.hs#L456

JvWesterveld:
Something like this? It does tack on another 10 seconds or so to the testsuite runtime on my machine.

martijnbastiaan (approved):
Perfect!

Note you can run a specific tests using: `cabal run testsuite -- -p MyTest`

JvWesterveld:
Great! Yes, not having to run the whole testsuite over and over again made my testing of the test a whole lot faster. :smiley:

alex-mckenna:
Isn't it a bit late to be working? :moon:

martijnbastiaan:
Clicking merge doesn't feel like work!

JvWesterveld:
Yes, yes it is a bit late :night_with_stars:

PR #1490, opened 2020-08-25
Deprecate 'Clash.Prelude.DataFlow'

martijnbastiaan:
Current API makes it impossible to use for designs on multiple domains.
We should probably design something closer to cchalmers's circuit as
defined over at github.com/cchalmers/circuit-notation.

Deprecating due to namespace clashes for users designing dataflow
libraries that are better suited for real world designs.

### DIFF
--- a/clash-prelude/src/Clash/Prelude/DataFlow.hs
+++ b/clash-prelude/src/Clash/Prelude/DataFlow.hs
@@ -18,7 +18,7 @@ Self-synchronizing circuits based on data-flow principles.
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise       #-}
 {-# OPTIONS_HADDOCK show-extensions #-}
 
-module Clash.Prelude.DataFlow
+module Clash.Prelude.DataFlow {-# DEPRECATED "Module will be removed in future versions of clash-prelude." #-}
   ( -- * Data types
     DataFlow (..)
     -- * Creating DataFlow circuits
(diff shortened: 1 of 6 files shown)


clash-prelude/src/Clash/Prelude/DataFlow.hs
```diff
@@ -18,7 +18,7 @@ Self-synchronizing circuits based on data-flow principles.
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise       #-}
 {-# OPTIONS_HADDOCK show-extensions #-}
 
-module Clash.Prelude.DataFlow
+module Clash.Prelude.DataFlow {-# DEPRECATED "Module will be removed in future versions of clash-prelude." #-}
```
alex-mckenna, line 21: May or may not be the time for this, but I feel like we should commit to some fixed offset between deprecating something and removing it. The reason being

  * users would know how quickly they need to change their code before it breaks
  * we could better ensure deprecated code isn't removed too quickly / slowly

Thoughts?

martijnbastiaan: Agreed. Ideally we should point users to some alternative.

martijnbastiaan: Guess I'll work a bit harder on `martijn-circuits` this weekend :-)

alex-mckenna: Well now that you know a circuit is a _symmetric monoidal category_, what's stopping you? :wink:

PR #1491, opened 2020-08-27
Add checked versions of 'resize', 'truncateB', and 'fromIntegral'

martijnbastiaan:
Depending on the type 'resize', 'truncateB', and 'fromIntegral' either
yield an XException or silently perform wrap-around if its argument does
not fit in the resulting type's bounds. The added functions check the
bound condition and fail with an error call if the condition is
violated. They do not affect HDL generation.

Useful in cases where runtime behavior should ensure that and out of
bound or wrap around should not occur and users want their code to fail
hard if this invariant is ever violated.

### DIFF
--- a/clash-prelude/src/Clash/Class/Resize.hs
+++ b/clash-prelude/src/Clash/Class/Resize.hs
@@ -1,7 +1,8 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente
+                  2020,      Myrtle Software Ltd
 License    :  BSD2 (see the file LICENSE)
-Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
 {-# LANGUAGE Safe #-}
@@ -9,9 +10,18 @@ Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.KnownNat.Solver #-}
 {-# OPTIONS_HADDOCK show-extensions #-}
 
-module Clash.Class.Resize where
+module Clash.Class.Resize
+ ( Resize(..)
+
+ -- * Resize helpers
+ , checkedResize
+ , checkedFromIntegral
+ , checkedTruncateB
+ ) where
 
 import Data.Kind (Type)
+import Data.Proxy (Proxy(Proxy))
+import GHC.Stack (HasCallStack)
 import GHC.TypeLits (Nat, KnownNat, type (+))
 
 -- | Coerce a value to be represented by a different number of bits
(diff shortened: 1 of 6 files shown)


clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -37,3 +47,61 @@ class Resize (f :: Nat -> Type) where
...
+  (Show a, Show b, Integral a, Integral b, Bounded b) =>
+  a -> b
+checkedFromIntegral a =
+  checkIntegral (Proxy @b) a `seq` fromIntegral a
+
+-- | Like 'resize', but errors if /f a/ is out of bounds for /f b/. Useful when
+-- you "know" /f a/ can't be out of bounds, but would like to have your
+-- assumptions checked.
+--
+-- __N.B.__: Check only affects simulation. I.e., it will not affect generated HDL.
```
martijnbastiaan, line 84: Hmm, perhaps: 

> No checks will be inserted into the generated HDL

to make it a bit more obvious

martijnbastiaan:
@christiaanb Concerning the offline chat: I do agree that parsing is better than validating, but:

* The `resize` / `truncateB` functions are already here and "unsafe".
* Unless we add type level information to `Bounded` (or another class) we can't write type safe conversion functions
* This would have, if used, made my debug run over at our client a matter of minutes instead of hours.

In short, I'm very much for adding this to clash-prelude

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -37,3 +47,65 @@ class Resize (f :: Nat -> Type) where
   signExtend = resize
   -- | Remove bits from the MSB
   truncateB :: KnownNat a => f (a + b) -> f a
+
+-- | Helper function of 'fromIntegralChecked' and 'resizeChecked'
```
alex-mckenna, line 51: I presume you mean `checkedFromIntegral` and `checkedResize`?

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -37,3 +47,65 @@ class Resize (f :: Nat -> Type) where
...
+  else
+    ()
+
+-- | Like 'fromIntegral', but errors if /a/ is out of bounds for /b/. Useful when
+-- you "know" /a/ can't be out of bounds, but would like to have your assumptions
+-- checked.
+--
+-- __N.B.__: Check only affects simulation. I.e., no checks will be inserted
+-- into the generated HDL
+checkedFromIntegral::
```
alex-mckenna, line 74: Missing space :eyes:

clash-prelude/src/Clash/Class/Resize.hs
martijnbastiaan: I do!

clash-prelude/src/Clash/Class/Resize.hs
martijnbastiaan: Blasphemy..

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -37,3 +47,65 @@ class Resize (f :: Nat -> Type) where
...
+-- __N.B.__: Check only affects simulation. I.e., no checks will be inserted
+-- into the generated HDL
+checkedTruncateB ::
+  forall a b f.
+  ( HasCallStack
+  , Resize f
+  , KnownNat b, Integral (f (a + b))
+  , KnownNat a, Integral (f a), Bounded (f a) ) =>
+  f (a + b) -> f a
+checkedTruncateB = checkedResize
```
alex-mckenna, line 111: This calls `resize` indirectly, shouldn't it be calling `truncateB`?

martijnbastiaan: It _should_ be the same thing. IIRC `resize` calls either `extend` or `truncateB` depending on its type arguments.

martijnbastiaan: I'll double check

martijnbastiaan: I changed it to `truncateB` so I don't have to double check :-)

christiaanb (approved):

alex-mckenna (approved):

PR #1492, opened 2020-08-28
Fix satSucc/satPred for small instances

DigitalBrains1:
The previous definition of `satSucc`/`satPred` only worked for types that can represent the number 1. But, e.g., `Index 1` and `Signed 1` cannot.

So `satSucc` and `satPred` are moved to the `SaturatingNum` class, but with their current implementation as the default method, such that existing code will keep working. For types where small instances of the type cannot represent 1, a new definition is provided that takes care of the issue.

A unit test for Fixed was added. The definitions `saturateToBounded` and `satSuccPropList` (in `Clash.Tests.Fixed`) work for `Unsigned`, `Signed` and `Fixed`. They could be factored out to some common module, but they are needlessly complicated for integer types.

### DIFF
--- a/clash-prelude/src/Clash/Class/Num.hs
+++ b/clash-prelude/src/Clash/Class/Num.hs
@@ -52,7 +52,7 @@ data SaturationMode
   | SatSymmetric -- ^ Become 'maxBound' on overflow, and (@'minBound' + 1@) on
                  -- underflow for signed numbers, and 'minBound' for unsigned
                  -- numbers.
-  deriving Eq
+  deriving (Show, Eq, Enum, Bounded)
 
 -- | 'Num' operators in which overflow and underflow behavior can be specified
 -- using 'SaturationMode'.
--- a/clash-prelude/src/Clash/Class/Num.hs
+++ b/clash-prelude/src/Clash/Class/Num.hs
@@ -20,8 +20,6 @@ module Clash.Class.Num
   , boundedAdd
   , boundedSub
   , boundedMul
-  , satSucc
-  , satPred
   )
 where
 
@@ -65,16 +63,16 @@ class (Bounded a, Num a) => SaturatingNum a where
   satSub  :: SaturationMode -> a -> a -> a
   -- | Multiplication with parameterizable over- and underflow behavior
   satMul :: SaturationMode -> a -> a -> a
-
--- | Get successor of (or in other words, add 1 to) given number
-satSucc :: SaturatingNum a => SaturationMode -> a -> a
-satSucc s n = satAdd s n 1
-{-# INLINE satSucc #-}
-
--- | Get predecessor of (or in other words, subtract 1 from) given number
-satPred :: SaturatingNum a => SaturationMode -> a -> a
-satPred s n = satSub s n 1
-{-# INLINE satPred #-}
+  -- | Get successor of (or in other words, add 1 to) given number
+  satSucc :: SaturationMode -> a -> a
+  -- Default method suitable for types that can represent the number 1
+  satSucc s n = satAdd s n 1
+  {-# INLINE satSucc #-}
+  -- | Get predecessor of (or in other words, subtract 1 from) given number
+  satPred :: SaturationMode -> a -> a
+  -- Default method suitable for types that can represent the number 1
+  satPred s n = satSub s n 1
+  {-# INLINE satPred #-}
 
 -- | Addition that clips to 'maxBound' on overflow, and 'minBound' on underflow
 boundedAdd :: SaturatingNum a => a -> a -> a
--- a/clash-prelude/src/Clash/Sized/Fixed.hs
+++ b/clash-prelude/src/Clash/Sized/Fixed.hs
@@ -134,7 +134,7 @@ import Clash.Class.Num            (ExtendingNum (..), SaturatingNum (..),
                                    SaturationMode (..), boundedAdd, boundedSub,
                                    boundedMul)
 import Clash.Class.Resize         (Resize (..))
-import Clash.Promoted.Nat         (SNat)
+import Clash.Promoted.Nat         (SNat, natToNum, natToInteger)
 import Clash.Prelude.BitIndex     (msb, split)
 import Clash.Prelude.BitReduction (reduceAnd, reduceOr)
 import Clash.Sized.BitVector      (BitVector, (++#))
@@ -453,15 +453,15 @@ instance ENumFixedC rep int1 frac1 int2 frac2 =>
   type AResult (Fixed rep int1 frac1) (Fixed rep int2 frac2) =
                Fixed rep (1 + Max int1 int2) (Max frac1 frac2)
   add (Fixed f1) (Fixed f2) =
-    let sh1 = fromInteger (natVal (Proxy @(Max frac1 frac2)) - natVal (Proxy @frac1)) :: Int
+    let sh1 = natToNum @(Max frac1 frac2) - natToNum @frac1 :: Int
         f1R = shiftL (resize f1) sh1 :: rep ((1 + Max int1 int2) + (Max frac1 frac2))
-        sh2 = fromInteger (natVal (Proxy @(Max frac1 frac2)) - natVal (Proxy @frac2)) :: Int
+        sh2 = natToNum @(Max frac1 frac2) - natToNum @frac2 :: Int
         f2R = shiftL (resize f2) sh2 :: rep ((1 + Max int1 int2) + (Max frac1 frac2))
     in  Fixed (f1R + f2R)
   sub (Fixed f1) (Fixed f2) =
-    let sh1 = fromInteger (natVal (Proxy @(Max frac1 frac2)) - natVal (Proxy @frac1)) :: Int
+    let sh1 = natToNum @(Max frac1 frac2) - natToNum @frac1 :: Int
         f1R = shiftL (resize f1) sh1 :: rep ((1 + Max int1 int2) + (Max frac1 frac2))
-        sh2 = fromInteger (natVal (Proxy @(Max frac1 frac2)) - natVal (Proxy @frac2)) :: Int
+        sh2 = natToNum @(Max frac1 frac2) - natToNum @frac2 :: Int
         f2R = shiftL (resize f2) sh2 :: rep ((1 + Max int1 int2) + (Max frac1 frac2))
     in  Fixed (f1R - f2R)
   type MResult (Fixed rep int1 frac1) (Fixed rep int2 frac2) =
@@ -522,7 +522,7 @@ instance (NumFixedC rep int frac) => Num (Fixed rep int frac) where
     | a == 0       = 0
     | a <  0       = -1
     | otherwise    = 1
-  fromInteger i    = let fSH = fromInteger (natVal (Proxy @frac))
+  fromInteger i    = let fSH = natToNum @frac
                          res = i `shiftL` fSH
                          rMax = toInteger (maxBound :: rep (int + frac))
                          rMin = toInteger (minBound :: rep (int + frac))
@@ -615,11 +615,11 @@ resizeF (Fixed fRep) = Fixed sat
   where
     fMin  = minBound :: rep (int2 + frac2)
     fMax  = maxBound :: rep (int2 + frac2)
-    argSZ = natVal (Proxy @(int1 + frac1))
-    resSZ = natVal (Proxy @(int2 + frac2))
+    argSZ = natToInteger @(int1 + frac1)
+    resSZ = natToInteger @(int2 + frac2)
 
-    argFracSZ = fromInteger (natVal (Proxy @frac1))
-    resFracSZ = fromInteger (natVal (Proxy @frac2))
+    argFracSZ = natToNum @frac1
+    resFracSZ = natToNum @frac2
 
     -- All size and frac comparisons and related if-then-else statements should
     -- be optimized away by the compiler
@@ -706,7 +706,7 @@ fLit a = [|| Fixed (fromInteger sat) ||]
                            then rMin
                            else truncated
     truncated = truncate shifted :: Integer
-    shifted   = a * (2 ^ (natVal (Proxy @frac)))
+    shifted   = a * (2 ^ (natToInteger @frac))
 
 -- | Convert, at run-time, a 'Double' to a 'Fixed'-point.
 --
@@ -878,7 +878,7 @@ fLitR a = Fixed (fromInteger sat)
                            then rMin
                            else truncated
     truncated = truncate shifted :: Integer
-    shifted   = a * (2 ^ (natVal (Proxy @frac)))
+    shifted   = a * (2 ^ (natToInteger @frac))
 
 instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
   satAdd w (Fixed a) (Fixed b) = Fixed (satAdd w a b)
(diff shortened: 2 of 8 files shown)


clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -340,6 +340,18 @@ instance (KnownNat n, 1 <= n) => SaturatingNum (Index n) where
...
         z -> resize# z
 
+  satSucc w !a =
+    case snatToNum @Integer (SNat @n) of
+      1 -> fromInteger# 0
+      _ -> satAdd w a $ fromInteger# 1
+  {-# INLINE satSucc #-}
+
+  satPred w !a =
+    case snatToNum @Integer (SNat @n) of
```
martijnbastiaan, line 350: `natToInteger @n` :-)

clash-prelude/src/Clash/Class/Num.hs
```diff
@@ -65,16 +63,16 @@ class (Bounded a, Num a) => SaturatingNum a where
...
-{-# INLINE satPred #-}
+  -- | Get successor of (or in other words, add 1 to) given number
+  satSucc :: SaturationMode -> a -> a
+  -- Default method suitable for types that can represent the number 1
+  satSucc s n = satAdd s n 1
+  {-# INLINE satSucc #-}
+  -- | Get predecessor of (or in other words, subtract 1 from) given number
+  satPred :: SaturationMode -> a -> a
+  -- Default method suitable for types that can represent the number 1
+  satPred s n = satSub s n 1
```
martijnbastiaan, line 74: How about:

```haskell
  -- Instances should override this default implementation if their interval
  -- can be entirely negative. I.e., if they can be instantiated such that
  -- both 'minBound' and 'maxBound' are less than zero.
  satPred s n = 
    if maxBound @a > 0 then
      satSub s n 1
    else
      0
```

and

```haskell
  -- Instances should override this default implementation if their interval
  -- can be entirely negative. I.e., if they can be instantiated such that
  -- both 'minBound' and 'maxBound' are less than zero.
  satSucc s n = 
    if maxBound @a > 0 then
      satSucc s n 1
    else
      0
```

martijnbastiaan: This would remove the need for a custom implementation for all our types AFAIK.

martijnbastiaan (commented):
This kind of test would work I think:

```haskell
module Clash.Tests.Num (tests) where

import Data.Proxy
import GHC.TypeNats (KnownNat, SomeNat (..), natVal, someNatVal)
import Test.Tasty
import Test.Tasty.HUnit
import Test.Tasty.QuickCheck

import Clash.Sized.Internal.Signed
import Clash.Tests.SizedNum

-- | Test whether `satSucc` yield no error
noErrorSatSuccProp :: SaturatingNum a => a -> Property
noErrorSatSuccProp n = satSucc n === satSucc n

tests :: TestTree
tests = testGroup "SaturatingNum"
  [ testGroup "satSucc"
    [ testProperty "Index 1" (noErrorSatSucc @(Index 1))
    , testProperty "Index 2" (noErrorSatSucc @(Index 2))
    , testProperty "Unsigned 0" (noErrorSatSucc @(Unsigned 0))
    , testProperty "Unsigned 1" (noErrorSatSucc @(Unsigned 1))
    ...
    ]
  ]
```

I haven't compiled it, so there's probably some errors in there.

clash-prelude/src/Clash/Sized/Internal/Index.hs
DigitalBrains1: That's what I thought, but I thought it better to maintain symmetry with the other definitions of `SaturatingNum`. It's a bit odd, though, essentially doing `fromInteger @Integer`. I can change all occurrences if you'd prefer that, but they have nothing to do with the purpose of the patch set.

clash-prelude/src/Clash/Class/Num.hs
DigitalBrains1: What about `SFixed 0 n` or `SFixed 1 n`? Unfortunately, it's not that simple.

clash-prelude/src/Clash/Sized/Internal/Index.hs
martijnbastiaan: Changing it sounds good to me!

clash-prelude/src/Clash/Class/Num.hs
martijnbastiaan: You're right!

martijnbastiaan:
@DigitalBrains1 Don't tell the AIVD, but we're breaking laws left and right..

In [this commit](https://github.com/clash-lang/clash-compiler/commit/74fdd82483ce916dfd4022def9e00b31eafa36eb) I add a bunch of law tests. You can run them with:

```bash
cabal run clash-prelude:unittests -- --hide-successes
```

I've added two very simple tests, which should make adding your own tests easier:

https://github.com/clash-lang/clash-compiler/commit/74fdd82483ce916dfd4022def9e00b31eafa36eb#diff-f7bedcdb375093e275907ac9a64ba060R139-R143

Let me know if I can clarify anything!

martijnbastiaan:
To clarify: I don't expect this PR to fix all the problems those tests find..

DigitalBrains1:
The original PR introduced a regression bug for `Fixed`. Our `succ` and `pred` for `Fixed` have always had this particular bug: they enumerate all values a `Fixed` can take, but the `Enum` laws say they should add or subtract the number 1. `satSucc`/`satPred` actually did add/subtract 1 before this PR.

The new PR fixes that and correctly adds/subtracts 1 for `satSucc`/`satPred`.

Additionally, a unit test for Fixed was added. The definitions `saturateToBounded` and `satSuccPropList` (in `Clash.Tests.Fixed`) work for `Unsigned`, `Signed` and `Fixed`. They could be factored out to some common module, but they are needlessly complicated for integer types.

clash-prelude/tests/Clash/Tests/Fixed.hs
```diff
@@ -0,0 +1,115 @@
...
+
+-- A list of Fixed numbers close to the bounds of the type.
+--
+-- Warning: this enumerates all fractional values, so it should only be used
+-- with types Fixed rep int frac where frac is reasonably small!
+closeToBounds
+  :: forall f rep int frac
+   . ( NumFixedC rep int frac
+     , f ~ Fixed rep int frac)
+  => [ f ]
```
DigitalBrains1, line 77: @martijnbastiaan , should I rewrite my tests to use `Hedgehog.Gen` instead of enumerating all interesting cases?

martijnbastiaan: Yes, if would have the same effect for small types, while making it possible to test larger ones too :).

clash-prelude/tests/Clash/Tests/Fixed.hs
```diff
@@ -0,0 +1,115 @@
...
+  | x > maxB  = maxB
+  | otherwise = x
+
+-- Saturate to bounds of type b.
+--
+-- It assumes the following values for the span (the range of numbers in the
+-- bounded type):
+--   - 2*(-minBound) when minBound < 0
+--   - floor maxBound + 1 otherwise
+-- This is correct for Unsigned, Signed and Fixed.
```
martijnbastiaan, line 46: This is confusing to me. Would span be "the number of numbers in the range"? If so, I would rename `span` to something like `nInhabitants` to make it a bit more obvious. Otherwise, if it is the "length" of the type's interval, wouldn't it simply be `abs (maxBound - minBound)`?

clash-prelude/tests/Clash/Tests/Fixed.hs
```diff
@@ -0,0 +1,115 @@
...
+        [ maxBound, pred maxBound .. ])
+
+-- Verify satSucc for interesting values of the Fixed type.
+fixedSatSuccProp
+  :: forall f rep int frac
+   . ( NumFixedC rep int frac
+     , f ~ Fixed rep int frac)
+  => SaturationMode
+  -> Property
+fixedSatSuccProp w =
```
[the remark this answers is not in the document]
DigitalBrains1: Okay! I mirrored the module it is testing w.r.t. arg names, but it's rather cryptic.

clash-prelude/tests/Clash/Tests/Fixed.hs
```diff
@@ -0,0 +1,115 @@
...
+  let fs = closeToBounds @f
+  in satSuccPropList w fs
+
+tests :: TestTree
+tests =
+  testGroup "SaturatingNum"
+    [ testGroup "satSucc"
+      $ P.map (\(ws, w) ->
```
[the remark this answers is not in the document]
DigitalBrains1: Only if we first merge PR #1495 or more to the point, lift the `deriving` for `SaturationMode` from that PR and incorporate it here :-)

clash-prelude/tests/Clash/Tests/Fixed.hs
```diff
@@ -0,0 +1,115 @@
...
+               True  -> 2 * negate (toRational (minBound @b))
+               False -> 1 + toRational (floor @Double @Integer $ fromRational
+                              $ toRational $ maxBound @b)
+  in saturate (toRational (minBound @b)) (toRational (maxBound @b)) span w x
+
+-- Verify satSucc for a list of values.
+satSuccPropList
+  :: forall a
+   . (Real a, SaturatingNum a)
+  => SaturationMode -> [a] -> Property
```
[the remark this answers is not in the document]
DigitalBrains1: I'll change this to use `Hedgehog.Gen` anyway. Then I should use `Property`?

clash-prelude/tests/Clash/Tests/Fixed.hs
```diff
@@ -0,0 +1,115 @@
...
+
+-- Saturate a number to be within certain bounds according to SaturationMode.
+saturate
+  :: RealFrac a
+  => a
+  -- minBound
+  -> a
+  -- maxBound
+  -> a
+  -- Span: the repetition interval of wrapping behaviour.
```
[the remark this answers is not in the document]
DigitalBrains1: Sure.

clash-prelude/tests/Clash/Tests/Fixed.hs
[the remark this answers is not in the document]
martijnbastiaan: Yeah just add it here, I'll deal with the merge conflicts later :)

clash-prelude/tests/Clash/Tests/Fixed.hs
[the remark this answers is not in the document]
martijnbastiaan: Yes!

clash-prelude/tests/Clash/Tests/Fixed.hs
DigitalBrains1: No, it's the repetition interval of the type. For `Unsigned` we have:

```
minBound == 0
maxBound == 15
```
but wrapping behaves as modulo 16, not modulo 15.

clash-prelude/tests/Clash/Tests/Fixed.hs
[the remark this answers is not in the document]
martijnbastiaan: Never lower your standards to implementations in clash-prelude ;)

clash-prelude/tests/Clash/Tests/Fixed.hs
DigitalBrains1: I'll rephrase it to use the phrase "repetition interval".

martijnbastiaan: Oh gotcha

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -936,6 +936,37 @@ instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
...
+    in case natVal (Proxy @int) of
+         0 -> case w of
+                SatWrap -> f
+                SatZero -> 0
+                _       -> maxBound
+         _ -> succ'
+  {-# INLINE satSucc #-}
+
+  satPred w f@(Fixed fRep) =
+    let sh       = fromInteger (natVal (Proxy @frac))
```
martijnbastiaan, line 953: Maybe `natToNum @frac` instead

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -936,6 +936,37 @@ instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
...
+        succ'    =
+          case isSigned fRep of
+            True  -> satAdd w f $ Fixed $ fromInteger $ (-1) `shiftL` sh
+            False -> satSub w f $ Fixed $ fromInteger $ 1 `shiftL` sh
```
martijnbastiaan, line 957: Maybe

```haskell
if isSigned fRep 
then satAdd w f $ Fixed $ fromInteger $ (-1) `shiftL` sh
else satSub w f $ Fixed $ fromInteger $ 1 `shiftL` sh
```

for a bit less syntactic overhead

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -936,6 +936,37 @@ instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
...
+        symBound = case isSigned fRep of
+                     True  -> Fixed $ succ minBound
+                     False -> minBound
+    in case natVal (Proxy @int) of
```
martijnbastiaan, line 961: `natToInteger @int`? :)

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -936,6 +936,37 @@ instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
                      0 -> unpack (resize (shiftR rR sh))
                      _ -> maxBound
 
+  satSucc w f@(Fixed fRep) =
```
martijnbastiaan, line 939: `w` -> `satMode` like in the tests would make this a bit more obvious

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -936,6 +936,37 @@ instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
...
+    let sh    = fromInteger (natVal (Proxy @frac))
```
martijnbastiaan, line 940: `natToNum @frac`

martijnbastiaan (commented):
Bunch of nitpicks coming in! :-)

martijnbastiaan (approved):
Perfect! Could you add a CHANGELOG entry? I can't remember if you've done it before, so here's the [changelog/README.md](https://github.com/clash-lang/clash-compiler/blob/master/changelog/README.md) on how to do so.

Edit: there's also some warnings that fail the CI, but I'm sure you've seen that already.

DigitalBrains1:
> Perfect! Could you add a CHANGELOG entry? I can't remember if you've done it before, so here's the [changelog/README.md](https://github.com/clash-lang/clash-compiler/blob/master/changelog/README.md) on how to do so.

I'll read up on it and add a Changelog.

> Edit: there's also some warnings that fail the CI, but I'm sure you've seen that already.

Yes, there were some stray imports from mucking about earlier, I should first check why they didn't give _me_ warnings before I pushed the PR, and after that it's an easy fix.

Thanks a lot for all your comments!

PR #1499, opened 2020-09-03
Store whether bindings are primitive unfoldings

alex-mckenna:
In the partial evaluator, it is important to be able to fallback
to the core unfolding for a primitive without an explicit rule.
However, these primitives are marked as NOINLINE in source, meaning
the evaluator will refuse to inline them.

Rather than ignoring INLINE pragmas, we track whether a binding
refers to a primitive with a known implementation. If this is true,
the evaluator can use the core unfolding. The change to the
evaluator is not made in this commit.

### DIFF
--- a/clash-lib/src/Clash/Driver/Types.hs
+++ b/clash-lib/src/Clash/Driver/Types.hs
@@ -40,14 +40,30 @@ import           Clash.Core.Var                 (Id)
 import           Clash.Core.VarEnv              (VarEnv)
 import           Clash.Netlist.BlackBox.Types   (HdlSyn (..))
 
+data IsPrim
+  = IsPrim
+    -- ^ The binding is the unfolding for a primitive.
+  | IsFun
+    -- ^ The binding is an ordinary function.
+  deriving (Binary, Eq, Generic, NFData, Show)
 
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -45,9 +45,19 @@ import           Clash.Netlist.BlackBox.Types   (HdlSyn (..))
 --
 data Binding a = Binding
   { bindingId :: Id
+    -- ^ The core identifier for this binding.
   , bindingLoc :: SrcSpan
+    -- ^ The source location of this binding in the original source code.
   , bindingSpec :: InlineSpec
+    -- ^ the inline specification for this binding, in the original source code.
+  , bindingIsPrim :: Bool
```
martijnbastiaan, line 53: Would you be opposed to make this `IsPrimitive` to avoid boolean blindness?

martijnbastiaan (approved):
LGTM.

clash-lib/src/Clash/Driver/Types.hs
alex-mckenna: I would not, one moment...

PR #1509, opened 2020-09-09
Fix msb for large types

DigitalBrains1:
`msb` would misbehave when operating on smallish values in large types:
for types with more than 64 bits, it would do `uncheckedShiftRL#` with a
shift length more than 63, which gives undefined, and indeed
undesired results.

### DIFF
--- a/clash-prelude/tests/Clash/Tests/BitVector.hs
+++ b/clash-prelude/tests/Clash/Tests/BitVector.hs
@@ -1,18 +1,50 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
 
-module Clash.Tests.BitVector where
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.Extra.Solver #-}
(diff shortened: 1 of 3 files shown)


martijnbastiaan (approved):

clash-prelude/tests/Clash/Tests/BitVector.hs
```diff
@@ -1,19 +1,51 @@
...
 import Clash.Sized.Internal.BitVector (BitVector (..))
 
 import Clash.Tests.SizedNum
 
+-- | Generates a BitVector either with the MSB set or not, and checks whether
+-- 'msb' agrees with it.
+msbTest :: forall n. (1 <= n, KnownNat n) => H.Property
+msbTest = H.property $ do
+  (bv, b) <- H.forAll msbGen
+  b H.=== msb bv
```
DigitalBrains1, line 36: This catches the problem by the virtue of the definition of `Eq Bit`, which indeed compares full values in the datatype. For instance `Show Bit` would not catch the problem because a malformed `Bit` might print correctly.

I think the way of testing used here is fine. I do think it's important to realise that it works because `Eq Bit` correctly compares malformed `Bit`s, which it is obviously not required to do as no malformed `Bit`s should ever occur in the first place.

DigitalBrains1: On the other hand, even if `Eq Bit` did a less thorough comparison, I suppose the chances that all randomly generated values happen to coincide requires a really odd bug.

PR #1517, opened 2020-09-10
Simplify evaluator and add case-of-case

alex-mckenna:
This PR simplifies the implementation of the partial evaluator, and in the process fixes some issues with `partial-evaluator-primitives`. The main changes are: 

  * global bindings and arguments are now always `Value` instead of `Either Term Value`. The inclusion of the `VThunk` constructor meant that there were two ways to express terms which had not been evaluated: now there is only one way. This also solves the subtle fault in the previous implentation of thunks not being forced when necessary
  * the `Eval` monad is now `RWST` with `IO` as the base monad, rather than `RWS`. This means `Eval` can implement the exception handling monad classes from `exceptions`, allowing exceptions thrown from primitive evaluation to be handled properly (the old implementation made it _very_ difficult to reliably catch exceptions)
  * Functions that call the evaluator now make use of the flag which specifies if the term under evaluation is the subject of a case expression. This is a correction of the implementation in both `master` and `partial-evaluator-primitives`
  * The case-of-case transformation is now applied when evaluating case expressions. The order of evaluation is changed so the transformation is applied before any part of the case expression is evaluated.
  * all evaluator logic is in `clash-ghc`: having backed-agnostic parts in `clash-lib` serves no real purpose, as there are no immediate plans for clash to have another frontend. Moving it all to `clash-ghc` removes a lot of unnecessary noise
  * both `clash-lib` and `clash-ghc` now use a `PartialEval` module (instead of `clash-lib` using `Evaluator` and clash-ghc using `PartialEval`). This is just housekeeping; nothing to see here

Still TODO before this can be reviewed / merged

  - [x] Add module level documentation for new modules
  - [x] Copy relevant documentation from the `partial-evaluator-primitives` branch to here
  - [x] Rebase `partial-evaluator-primtives` onto this branch and check for issues


alex-mckenna:
Update: I'm somewhat on the fence about whether this has made another defect visible. Should the evaluator be substituting type variables for types it encounters during evaluation? :thinking:

christiaanb:
You'll have to if you want to do alternative-elimination for GADT pattern-matching down the line. i.e. when you get like:

```
(/\(n : Nat) -> \x : Vec n Int -> case x of
  Cons (m : Nat) (m+1 ~ n) y ys -> ...
  Nil (m : Nat) (m~0) -> ..) 3
```

you want to eliminate that `Nil` branch based on the fact that `3 ~ 0` is absurd.

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -1123,25 +1124,32 @@ whnfRW
...
+  case unsafePerformIO (nf eval genv isSubj i e) of
+    (!e', !genv') -> do
+      ioHeap Lens..= genvHeap genv'
+      ioAddr Lens..= genvAddr genv'
 
-  case runEval (mkGlobalEnv bndrs ri fuel gh tcm is0 ids1) (evaluateNf eval e) of
-    (!e', !env') -> do
-      globalHeap Lens..= genvPrimsIO env'
-      rw ctx (asTerm e')
+      -- TODO I remain unsure about this. Do I want to use bindPureHeap?
```
martijnbastiaan, line 1147: Is this still TODO?

alex-mckenna: I'm still unsure about it, so probably.

alex-mckenna:
Update: there are a few extra issues in the evaluator where I want to add the fixes to this branch. Mostly to do with free variables and preserving work-doing let bindings. Going back to draft

alex-mckenna:
Update: The partial evaluator primitives branch is starting to show some promising results (`Blinker` as an example now compiles with no work duplication and full PE running). The relevant changes (excluding those related to primitives) are contained in this branch. I don't want to add more to this PR, it was already quite large before this last set of fixes...

clash-ghc/src-ghc/Clash/GHC/PartialEval/Eval.hs
```diff
@@ -0,0 +1,603 @@
...
+
+    CyclicSCC ys -> do
+      eYs <- traverse evalBind ys
+      pure (eYs <> k, i)
+
+evalCase :: Term -> Type -> [Alt] -> Eval Value
+evalCase term ty as
+  -- There is only one option, the subject need not be evaluated.
+  | [(DefaultPat, tm)] <- as
+  = eval tm
```
martijnbastiaan, line 275: What if the subject is `_|_`?

martijnbastiaan: To clarify; I'm not worried that this is incorrect, but we might end up generating larger designs this way.

martijnbastiaan (approved):

clash-ghc/src-ghc/Clash/GHC/PartialEval/Eval.hs
alex-mckenna: Yes, reasonable point. The new evaluator currently considers `_|_` but only after checking if there is only `_` as a pattern. It would probably better to check subject first, because there might be a long chain of `case ... of _ -> ...` in core which can be completely skipped if the first subject is `_|_`.

alex-mckenna: Okay, changed to check for `_|_` first in case expressions. It also occurred to me that there isn't really a reason to check for just `DefaultPat` first (since that was just to avoid evaluating the subject) so I removed that special path and tidied it up a little.

PR #1519, opened 2020-09-11
Make msb agnostic regarding machine word size

DigitalBrains1:
Previously, it would assume a 64-bit machine word size, and do the wrong
thing on both 32-bit as well as hypothetical >64-bit machines.


martijnbastiaan (approved):
Nice one. Aren't a lot of ARM chips still 32-bit? It might not be as an obscure fix as you might think!

DigitalBrains1:
Yes, 32-bit ARM is common. 64-bit ARM is already fairly common in the server market and is just making its entry into the consumer market.

alex-mckenna:
As long as @JvWesterveld can run it on his phone, that's all that matters :wink:

PR #1525, opened 2020-09-16
Ensure wrappers for `delay#` and `register#` are always inlined

christiaanb:
I noticed that Clash was specialising `Explicit.register` too many times during a project, which meant that GHC didn't inline it. Moving formal arguments to lamda's should mean GHC always inlines the definitions of `Explicit.register` and friends.


martijnbastiaan (approved):
OK!

This is just a normalization performance optimization, right? I.e., shifting the work to GHC. (And it would reduce the chance of users having to add `-fclash-inline-foo=333`.)

christiaanb:
Right, this is so that users don't hit the specialisation limit.

PR #1526, opened 2020-09-16
Add a "high-speed" version of `autoReg`

christiaanb:
This still splits registers of products into products of registers, but it no longer uses the tag bit to control the enable lines of the registers holding the value bits. This decreases the logic depth, thus shortening the potential critical path, at the cost of power, since the fields bit can freely toggle.

### DIFF
--- a/clash-prelude/src/Clash/Class/AutoReg/Internal.hs
+++ b/clash-prelude/src/Clash/Class/AutoReg/Internal.hs
@@ -113,6 +113,19 @@ class NFDataX a => AutoReg a where
   autoReg = register
   {-# INLINE autoReg #-}
 
+  -- | A "High-Speed" version of 'autoReg', that still splits product types
+  -- into multiple registers, but no longer uses the tag bit to control the
+  -- enable of the field bits. This reduces the potential logic depth, at the
+  -- cost of power due to toggling field bits.
+  autoRegHs
+    :: (HasCallStack, KnownDomain dom)
+    => Clock dom -> Reset dom -> Enable dom
+    -> a  -- ^ Reset value
+    -> Signal dom a
+    -> Signal dom a
+  autoRegHs = autoReg
+  {-# INLINE autoRegHs #-}
+
 instance AutoReg ()
 instance AutoReg Bool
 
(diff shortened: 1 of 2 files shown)


clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -113,6 +113,19 @@ class NFDataX a => AutoReg a where
   autoReg = register
   {-# INLINE autoReg #-}
 
+  -- | A "High-Speed" version of 'autoReg', that still splits product types
+  -- into multiple registers, but no longer uses the tag bit to control the
+  -- enable of the field bits. This reduces the potential logic depth, at the
+  -- cost of power due to toggling field bits.
+  autoRegHS
```
martijnbastiaan, line 120: Drive-by nitpick: `autoRegHs` maybe? Makes it easier to extend (`autoRegHsFoo` vs `autoRegHSFoo`) and a bit easier to type.

christiaanb: Done

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -160,6 +173,22 @@ instance AutoReg a => AutoReg (Maybe a) where
...
   {-# INLINE autoReg #-}
 
+  autoRegHs clk rst en initVal input = createMaybe <$> tagR <*> valR
+   where
+     tag = isJust <$> input
+     tagInit = isJust initVal
+     tagR = suffixName @"tag" register clk rst en tagInit tag
+
+     val = fromMaybe (deepErrorX "autoReg'.val") <$> input
+     valInit = fromMaybe (deepErrorX "autoReg'.valInit") initVal
```
leonschoorl, line 183:
```suggestion
     val = fromMaybe (deepErrorX "autoRegHs.val") <$> input
     valInit = fromMaybe (deepErrorX "autoRegHs.valInit") initVal
```

leonschoorl: And could you also remove the `'`s from the error strings in the original `autoReg`, which I accidentally left in there?

leonschoorl (changes requested):

leonschoorl:
Other then those two comments, LGTM

leonschoorl (changes requested):
You should also add `autoRegHs` to the RTree instance and to the [TH code that generates instance for product types](https://github.com/clash-lang/clash-compiler/blob/b0a239e1a6b0e0d0f7cf63a0c85f2c1448a0a13e/clash-prelude/src/Clash/Class/AutoReg/Internal.hs#L336)

christiaanb:
Yeah... need to fix that hard-to-fix bug in Clash first as well...

PR #1542, opened 2020-09-28
Use the exact component corresponding to the topEntity

### LINKED ISSUE #1533: Clash generates manifest and sdc file for the wrong function
Because of https://github.com/clash-lang/clash-compiler/blob/b0a239e1a6b0e0d0f7cf63a0c85f2c1448a0a13e/clash-lib/src/Clash/Driver.hs#L439 and that component names are made all lower case.

For something akin to:

```
topEntity x = myTopEntity x x

myTopEntity x y = not x || y
{-# NOINLINE myTopEntity #-}
```

Clash can/will generate a `.manifest` file with info for `myTopEntity` instead of `topEntity`.

christiaanb:
Code archeology didn't really explain why we ever used `isSuffixOf` to find the component. I think it's some relic of when we had less control over naming, or perhaps at some point we included the full hierarchy...
Regardless, I think this is the proper solution.

Fixes #1533

### DIFF
--- a/changelog/2020-09-28T12_23_22+02_00_fix1533
+++ b/changelog/2020-09-28T12_23_22+02_00_fix1533
@@ -1,0 +1,1 @@
+FIXED: Clash generates manifest and sdc file for the wrong function [#1533](https://github.com/clash-lang/clash-compiler/issues/1533)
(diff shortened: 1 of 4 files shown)


martijnbastiaan (approved):
LGTM. Changelog?

christiaanb:
Added the changelog

PR #1545, opened 2020-10-03
Grouping and binding in bitPattern

### LINKED ISSUE #1540: Idea: digit grouping in `bitPattern`
I've just now realized that I would sometimes like to group the digits of a `bitPattern`, similar to [the `NumericUnderscores` language extension](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/numeric_underscores.html); for example, I'd like to rewrite this:

```
    $(bitPattern "01110110") -> HLT
    $(bitPattern "01......") -> MOV dest src
    $(bitPattern "00...110") -> MVI dest
    $(bitPattern "00..0001") -> LXI rr
```

into this:

```
    $(bitPattern "01_110_110") -> HLT
    $(bitPattern "01_..._...") -> MOV dest src
    $(bitPattern "00_..._110") -> MVI dest
    $(bitPattern "00_..0_001") -> LXI rr
```

I am of course volunteering to implement this, if the consensus is that it's a good idea. Let me know what you think.

basile-henry:
This PR extends the functionality of the `bitPattern` TH pattern function.

It adds the possibility to use `_` as a spacer in a similar fashion to the `NumericUnderscores` language extension. There is no restriction as to where this spacer can be used (which is a bit looser than the language extension which doesn't allow the spacer on either end of a number `_0011` and `64_` for example).

It also adds the possibility to bind arbitrary bits to a variable of type `BitVector n` where `n` is the number of bits bound using the same character. Examples of this mechanism have been added to the docs.
Note this feature does not allow any given bit to be bound by multiple variables.

Was there anything else you wanted for this feature @martijnbastiaan @gergoerdi ?
Should I add tests/more docs/a changelog entry?

This was pretty fun to implement ! Hopefully it'll be useful for someone :sweat_smile: 

Closes #1540

### DIFF
--- a/clash-prelude/src/Clash/Sized/Internal/BitVector.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/BitVector.hs
@@ -1140,8 +1140,8 @@ bitPattern s = [p| ((\_x -> $preprocess) -> $tuple) |]
       | C.isAlpha c && C.isLower c =
         ( succ i
         , Nothing:b
-        , M.alter (Just . maybe [i] (i:)) c n
+        , M.alter (Just . (i:) . fromMaybe []) c n
         )
       | otherwise = error $
         "Invalid bit pattern: " ++ show c ++
-        ", expecting one of '0', '1', '.', '_', or a lower case alphabetic character"
+        ", expecting one of '0', '1', '.', '_', or a lowercase alphabetic character"
--- a/clash-prelude/src/Clash/Sized/Internal/BitVector.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/BitVector.hs
@@ -1091,22 +1091,28 @@ fromBits = L.foldl (\v b -> v `shiftL` 1 .|. fromIntegral b) 0
 -- pattern. The scrutinee can be any type that is an instance of the
 -- 'Num', 'Bits' and 'Eq' typeclasses.
 --
--- The bit pattern is specified by a string which contains @\'0\'@ or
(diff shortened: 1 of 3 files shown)


clash-prelude/src/Clash/Sized/Internal/BitVector.hs
```diff
@@ -1091,27 +1093,55 @@ fromBits = L.foldl (\v b -> v `shiftL` 1 .|. fromIntegral b) 0
...
+    tuple = tupP $ target : (uncurry var <$> ns)
+
+    parse '.' (i, b, n) = (succ i, Nothing:b, n)
+    parse '0' (i, b, n) = (succ i, Just 0:b, n)
+    parse '1' (i, b, n) = (succ i, Just 1:b, n)
+    parse c (i, b, n)
+      | C.isAlpha c && C.isLower c =
+        ( succ i
+        , Nothing:b
+        , M.alter (Just . maybe [i] (i:)) c n
```
gergoerdi, line 1143: `Maybe` it's just me, but wouldn't `(i:) . fromMaybe []` be more readable?

martijnbastiaan:
Nice one! Looking good.

>  Should I add tests/more docs/a changelog entry?

The docs are fine :), but a changelog entry would be nice. See [this README](https://github.com/clash-lang/clash-compiler/blob/master/changelog/README.md). A few simple test cases would be great too! You can add them [over here](https://github.com/clash-lang/clash-compiler/tree/master/clash-prelude/tests/Clash/Tests).

martijnbastiaan (approved):
Thanks!

PR #1546, opened 2020-10-07
Move isWorkFree out of RewriteMonad

alex-mckenna:
The isWorkFree function is needed by the partial evaluator, but
cannot be used as it lives inside RewriteMonad. As the only
part of RewriteMonad it needs is access to a BindingMap, this can
be changed to be a parameter to the function, and a pure isWorkFree
can be provided which can be used anywhere.


christiaanb (commented):
But this looses caching, what's the performance impact?

alex-mckenna:
@christiaanb, it keeps caching now (and as a bonus has a smaller diffstat)

christiaanb (approved):

PR #1560, opened 2020-10-29
Allow primitives to assign to multiple results

martijnbastiaan:
Approach:

```
-- A multi result primitive assigns its results to multiple result variables
-- instead of one. Besides producing nicer HDL it works around issues with
-- synthesis tooling described in:
--
--   https://github.com/clash-lang/clash-compiler/issues/1555
--
-- This transformation rewrites primitives indicating they can assign their
-- results to multiple signals, such that netlist can easily render it.
--
-- Example:
--
-- @
-- prim :: forall a. a -> (a, a)
-- @
--
-- will be rewritten to:
--
-- @
--   \a0 -> let
--            r  = prim @t0 a0 r0 r1     -- With 'Clash.Core.Term.MultiPrim'
--            r0 = multiPrimSelect r0 r
--            r1 = multiPrimSelect r1 r
--          in
--            (x, y)
-- @
--
-- Netlist will not render any @multiPrimSelect@ primitives. Similar to
-- primitives having a /void/ return type, /r/ is not rendered either.
--
-- This transformation is currently hardcoded to recognize tuples as return
-- types, not any product type. It will error if it sees a multi result primitive
-- with a non-tuple return type.
--
```

### DIFF
--- a/clash-lib/src/Clash/Netlist/Util.hs
+++ b/clash-lib/src/Clash/Netlist/Util.hs
@@ -26,6 +26,7 @@ import           Control.Exception       (throw)
 import           Control.Lens            ((.=),(%=))
 import qualified Control.Lens            as Lens
 import           Control.Monad           (unless, when, zipWithM, join)
+import           Control.Monad.Extra     (concatMapM)
 import           Control.Monad.Reader    (ask, local)
 import qualified Control.Monad.State as State
 import           Control.Monad.State.Strict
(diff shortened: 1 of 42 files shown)


clash-lib/src/Clash/Netlist/Util.hs
```diff
@@ -788,90 +802,122 @@ mkUniqueNormalized is0 topMM (args,binds,res) = do
...
+  case k of
+    Prim p ->
+      case primMultiResult p of
+        SingleResult -> extractPrimWarnOrFail (primName p) >>= goSingle p
+        MultiResult -> extractPrimWarnOrFail (primName p) >>= goMulti p
+    _ -> pure []
  where
-  go nm (BlackBox {resultName = Just (BBTemplate nmD)}) = withTicks ticks $ \_ -> do
-    (bbCtx,_) <- preserveVarEnv (mkBlackBoxContext nm i args)
+  -- Routine for multi result primitives (the default kind of primitive). For
```
christiaanb, line 871: What do you mean by "the default kind of primitive"?

christiaanb (approved):

clash-lib/src/Clash/Netlist/Util.hs
martijnbastiaan: It should have said that for 'single result primitives', not multi result ones!

PR #1563, opened 2020-10-31
Use ghcup on CI so we can use GHC 9.0

martijnbastiaan:


martijnbastiaan:
@alex-mckenna Seems like compiling GHDL with LLVM 9 works on 20.04. The Docker images are getting pretty large though..

alex-mckenna:
Yeah, we can get back a little space using multi state builds in docker, but the real problem is all the GHC / Cabal installations. Maybe it's time to have separate images built and tagged per GHC, i.e. `clash-ci:2020-11-03-ghc8-10`?

martijnbastiaan:
I've compressed the GHC installations individually with `pixz`. This gives us multithreaded compression _and_ decompression. It takes `diepenheim` 6 seconds to extract a single installation this way. The image size is reduced to 1GB from a whopping 10 GB! That's well worth the 6 seconds I think :)

PR #1569, opened 2020-11-06
Set topentities as NOINLINE in generateBindings

### LINKED ISSUE #1568: Set all bindings for topentities as NoInline
The Clash tutorial says that

> The Clash compiler is based on the GHC Haskell compiler, and the GHC machinery does not understand Synthesize annotations and it might subsequently decide to inline those functions. You should therefore also add a `{-# NOINLINE f #-}` pragma to the functions which you give a Synthesize functions.

but this is not enforced in user code. It would be nice if during compile-time any global binding which is a known topentity is marked as being `NOINLINE`, so that if the pragma is forgotten the component can never be accidentally inlined. The most direct way of doing this would be to add a pass over the `BindingMap` in `Clash.Driver`.

alex-mckenna:
When generating bindings, any identified top entity is changed to
be NOINLINE in the bindings map. This prevents transformations /
the partial evaluator from accidentally inlining a topentity.

Ideally this should be a source plugin, as GHC may still inline
the top entity before it reaches clash.

Closes #1568

### DIFF
--- a/changelog/2020-11-06T11_04_40+01_00_noinline_topentities
+++ b/changelog/2020-11-06T11_04_40+01_00_noinline_topentities
@@ -1,0 +1,1 @@
+CHANGED: generateBindings now sets all topentities as NOINLINE [#1568](https://github.com/clash-lang/clash-compiler/issues/1568)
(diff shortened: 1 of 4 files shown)


christiaanb (approved):

martijnbastiaan (approved):
> Ideally this should be a source plugin

Maybe I can hack something together next week :)

Issue #1575, opened 2020-11-09
Allow th-abstraction-0.4


jchia:
Can the clash-prelude upper bound on th-abstraction be increased to allow th-abstraction-0.4? I've run "stack test clash-prelude" under nightly-2020-11-08 with the upper bound increased to 0.4.1 and the unittests still all passed. There are some preexisting failures for doctests that also happened with nightly-2020-11-06.

https://github.com/commercialhaskell/stackage/issues/5668

PR #1576, opened 2020-11-09
Bump th-abstraction version

jchia:
#1575


jchia:
I see this is a duplicate of #1576

PR #1589, opened 2020-11-10
Created stack template.

### LINKED ISSUE #1582: Support Stack templates
It would be nice if, instead of making users download / unpack the Clash starter project, to add a [Stack template repository](https://docs.haskellstack.org/en/stable/GUIDE/#templates). Setting up a new Clash project would be as easy as running:

```
stack new my-clash-project clash-lang/simple
```

This would eliminate the need for users to go through all the files in the starter project and replace them with appropriate values. We should also start promoting this on clash-lang.org.

rowanG077:
Fixes #1582

I have left the license exactly as the starter project but maybe it's better to remove it completely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,3 +117,18 @@ Or [use Nix](https://nixos.org/nix/download.html) to get a shell with the `clash
 nix-shell
 ```
 
+## Start projects
+
+### Example project
+An example project is available under `clash-starters/clash-example-project`. Refer
+to the readme in that project for further instruction.
+
+### Stack template
+It's also possible to create a Clash project from a [stack template](https://docs.haskellstack.org/en/stable/GUIDE/#templates). It will create all necessary files for you. You have to have installed [Stack](https://docs.haskellstack.org/en/stable/install_and_upgrade/). To create a new project from the template run the following command:
+
+```bash
+stack new my-clash-project https://raw.githubusercontent.com/clash-lang/clash-compiler/master/clash-starters/clash-starter.hsfiles
+```
+
+Now a new folder called `my-clash-project` is created with everything setup.
+Refer to the readme in that project on how to further use it.
--- a/changelog/2020-11-10T11_16_50+01_00_stack_template
+++ b/changelog/2020-11-10T11_16_50+01_00_stack_template
@@ -1,0 +1,1 @@
+ADDED: Created Clash stack template [#1582](https://github.com/clash-lang/clash-compiler/issues/1582)
--- a/clash-starters/clash-starter.hsfiles
+++ b/clash-starters/clash-starter.hsfiles
@@ -1,0 +1,422 @@
+{-# START_FILE {{name}}.cabal #-}
+cabal-version:       2.4
+name:                {{name}}
+version:             0.1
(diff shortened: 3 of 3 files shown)


alex-mckenna:
IMO the license for this shouldn't be the same as the starter project (CC0). This isn't a project of ours left out as an example, the main use would be people creating their own projects which likely want less permissive licences. I would lean towards using some variant of the BSD license as this is _probably_ the most common for Haskell projects anyway.

martijnbastiaan:
Ooh this is great, thanks @rowanG077! 

There's a number of organizational issues we need to address I think:

* It would be neat if we could use a simpler `stack new my-clash-project clash-lang/simple` instead. This would require us to make a separate repository `stack-templates`.

* We currently risk the starter project and the stack template going out of sync. Ideally, we would generate one from the other.

I think this should look like:

 * A repository `clash-lang/stack-templates`, which has a readme redirecting to `clash-lang/clash-starters`.
 * A push to stack templates would trigger a CI that builds all the templates and pushes them to `clash-lang/clash-starters` with some default config
 * `clash-lang/clash-starters` contain a number of subdirectories. One for each rendered Stack template. It would instuct Stack users to use `stack new foo/bar`, while instructing cabal users to use `curl foo.tar.gz | untar bar`.

I'm pushing for multiple templates, because it would be cool if we could point users to user `stack new clash-lang/aws` for a starter project containing all the setup foo for running Clash on AWS (for example)!

@alex-mckenna @rowanG077 what do you think?

alex-mckenna:
It sounds like `clash-lang/clash-starters` would just be a repository of build artefacts. Why not have something more like releases for `clash-lang/stack-templates` with ready to extract and use archives? If it is a single repo with releases, you would also probably want the `clash-lang/clash-starters` name instead of `stack-templates` for discoverability.

Multiple templates is a :+1:  from me, the needs of someone deploying to something different like F1 are suitably different, as is the _hello world_ example to include by default in the project.

martijnbastiaan:
> name instead of `stack-templates` for discoverability.

To motivate that name a bit: it's the default repository Stack looks for when downloading templates, so it keeps the `new` command nice and short.

alex-mckenna:
Makes sense, slightly annoying but I guess that comes with using `stack`. Would it be possible to edit the most recent release of `clash-lang/clash-compiler` to include the new starter project builds? I know you can edit releases but there may be limitations which stop this

martijnbastiaan:
I don't believe there are any limitations. We'd just have to figure out how to use the GitHub API.

alex-mckenna:
Hopefully that's not too much of a limitation for _Meneer DevOps_

martijnbastiaan:
:unamused:

rowanG077:
If you make the repos I can try to get this working. I have done something like this before :). @martijnbastiaan

martijnbastiaan:
Yay! I've given you write access to:

* https://github.com/clash-lang/clash-starters
* https://github.com/clash-lang/stack-templates

martijnbastiaan:
Closing in favor of https://github.com/clash-lang/stack-templates/pull/2. Thanks :)

PR #1595, opened 2020-11-17
Fix various issues in the starter project

martijnbastiaan:


alex-mckenna (approved):
Tentative approval: does this mean we're intending to drop snap for good?

martijnbastiaan:
I think it's fine for a classroom environment for now. In case you don't need:

- dependencies from Hackage
- fast executables
- the starter project
- strict versioning

I.e., you just want to synthesize a single file / run clashi. But we should stop promoting it for real:tm: projects.

alex-mckenna:
I wonder how well tools like `cabal-rpm` / `cabal-debian` / `cabal-macosx` work in practice. Perhaps these and a light bit of CI scripting could work?

martijnbastiaan:
Yeah `cabal-debian` works fine. There's a Python script in `bindist/`, but it's a big hackjob. There's lots of annoying details to get right (non-hackage deps, failing doc builds, ..).

PR #1597, opened 2020-11-18
Add scripts to build Debian packages

martijnbastiaan:
Known "issues":

- [x] The package containing clash is called `haskell-clash-ghc-utils`. There's probably a way to change it, but I couldn't get `cabal-debian` to cooperate. **edit: created metapackages called `clash`, `clash-dev`, `clash-prof`, `clash-doc`.**
- [x] GHCs installed using ghcup don't interfere, but GHCs using HVRs PPA do. Need to investigate more. **edit: can't reproduce**
- [x] Clash emits the warning (in a VM, can't copy for some reason):

Selection_216

While it seems slightly slower than usual, this doesn't seem to apply anymore? We can probably remove the warning, maybe after some more testing. **edit: removed warning**


clash-prelude/Setup.hs
```diff
@@ -9,14 +9,6 @@ import Distribution.Extra.Doctest ( defaultMainWithDoctests )
...
 main = defaultMainWithDoctests "doctests"
 #else
-#ifdef MIN_VERSION_Cabal
--- If the macro is defined, we have new cabal-install,
--- but for some reason we don't have cabal-doctest in package-db
---
--- Probably we are running cabal sdist, when otherwise using new-build
--- workflow
-import Warning ()
-#endif
```
martijnbastiaan, line 19: Building a package fails with this in (module `Warning` can't be found). No idea if it's useful to keep?

alex-mckenna: I encountered this when I was looking at `cabal-rpm` and was going to ask you. Is there anything useful if you `git blame` it?

bindist/linux/debian/scripts/build.py
```diff
@@ -0,0 +1,121 @@
...
+  return subprocess.run(*args, **kwargs)
+
+def get_buildinfo():
+  """"Parse and return 'buildinfo.json'"""
+  path = os.path.join(SCRIPT_DIR, "..", DISTRIBUTION, "buildinfo.json")
+  return json.load(open(path, "rb"))
+
+def apt_install(pkgs):
+  """Install a package using system's 'apt'"""
+  return run("apt install {} -y".format(" ".join(pkgs)).split())
```
alex-mckenna, line 47: We should use `apt-get` instead of `apt` for scripts, `apt` isn't necessarily stable

bindist/linux/debian/scripts/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Bootstrap Python installation and run build.py
+set -x -euf -o pipefail
+
+cd $(dirname $0)
+apt update
+apt install python3 -y
```
alex-mckenna, line 7: Isn't this run inside the docker container? If so why are you installing python3 here again?

bindist/linux/debian/focal/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:focal
+
+ENV DEBIAN_FRONTEND=noninteractive LANG=C.UTF-8 LC_ALL=C.UTF-8
+
+RUN apt update \
+ && apt dist-upgrade -y \
```
alex-mckenna, line 6: I'm still uneasy about the use of `dist-upgrade` in dockerfiles. The official Ubuntu docker repo keeps builds of different versions by date, e.g. `ubuntu:groovy-20201022.1` so I believe always pulling the newest `ubuntu:groovy` before building would give the most recent kernel updates anyway. If it is genuinely needed in our dockerfiles that's a bad sign IMO

bindist/linux/debian/Dockerfile
```diff
@@ -1,24 +0,0 @@
...
-ARG DEPS_CLASH="libtinfo-dev libpthread-stubs0-dev"
-
-RUN apt update \
-  && apt dist-upgrade -y \
-  && apt install $DEPS_GHC $DEPS_BUILD $DEPS_CLASH -y \
-  && apt clean \
-  && echo 1
-
-
-# Yeet all possible GHC dependencies into the global scope, so cabal is sure
```
alex-mckenna, line 18: _Yeets no more_

alex-mckenna (commented):
I like the look of this, definitely an improvement on the old situation. Some thoughts:

  * Should this live in `.ci/` instead of at the top level? I feel like building OS packages _should_ be a CI task by default instead of one which is manually triggered. If we manually need to build / deploy to a particular repo, that should be a special case

  * Can we also try this on Debian stable instead of just Ubuntu? I'm curious whether it would work, or whether the Debian packages are too out of date

  * Do we / can we validate that a package works as intended after build? I would hate to start producing `.deb` packages which can't reliably be used in the real world

clash-prelude/Setup.hs
martijnbastiaan: Hmm, Christiaan recently added it. @christiaanb , thoughts?

bindist/linux/debian/scripts/build.sh
martijnbastiaan: I didn't want to make another docker container due to the limits on dockerhub.com.. So the `Dockerfile` caches _some_ things to make it cheaper to experiment locally. But it should all work on base images.

bindist/linux/debian/focal/Dockerfile
martijnbastiaan: I'll remove it, but I swear it did solve an issue with GHDL before :(

bindist/linux/debian/Dockerfile
martijnbastiaan: F

martijnbastiaan:
> Should this live in .ci/ instead of at the top level? I feel like building OS packages should be a CI task by default instead of one which is manually triggered. If we manually need to build / deploy to a particular repo, that should be a special case

Agreed, although I don't look forward to fixing all the hardcoded paths :(

> Can we also try this on Debian stable instead of just Ubuntu? I'm curious whether it would work, or whether the Debian packages are too out of date

Sure thing, I'll add a `buildinfo.json`

> Do we / can we validate that a package works as intended after build? I would hate to start producing .deb packages which can't reliably be used in the real world

We don't. We should at least run the doctests and unittests. The whole testsuite is probably a bit much, but we could test whether it synthesis some "interesting" designs. I2C + something with BlackBoxHaskell?

bindist/linux/debian/focal/Dockerfile
alex-mckenna: Possibly, but I think in that case it _should_ also be possible to pull the newer version of the base image and build (provided the only difference that matters is the kernel version)

alex-mckenna:
> We don't. We should at least run the doctests and unittests. The whole testsuite is probably a bit much, but we could test whether it synthesis some "interesting" designs. I2C + something with BlackBoxHaskell?

Perhaps. Maybe it would be a good idea to have a test group in the testsuite for when we want to test, but not fully. Then you could just run that single group instead of everything. Maybe wrap the `examples/` and `Basic` groups up in some `SmokeTest` group? :thinking:

martijnbastiaan:
> Can we also try this on Debian stable instead of just Ubuntu? I'm curious whether it would work, or whether the Debian packages are too out of date

I've tried. It fails with:

```
dpkg-source: error: can't build with source format '3.0 (native)': native package version may not have a revision
```

We need revisions to differentiate the nightly builds though. We could probably try and make proper source tarballs, but honestly, I can't really be bothered.

alex-mckenna:
> but honestly, I can't really be bothered.

Fair enough, at least we know how to fix it if someone wants packaging for Debian specifically and not Ubuntu. Another thing that occurred to me: should we package source debs as well for completeness?

martijnbastiaan:
Users get a `.tar.xz` with the source :)

alex-mckenna:
I was thinking about the people who want to `apt-get source` packages, could be worth also providing for completeness given it's easier than making the binary `.deb`. Is that what you meant, or did you just mean the GitHub releases?

martijnbastiaan:
Yeah, that's what the `.tar.xz` is I believe. This is what a complete package would look like:

Selection_218

The `.buildinfo` indicates that there's a source distribution:

```
Architecture: amd64 source
```

Having said that, I still have to figure out how to properly set up an APT repository. My priority is rebasing the snap though.

alex-mckenna:
I believe you use something like `dh_make` to do it on Ubuntu. Maybe the easiest way to setup a clash-lang PPA? I also think we need a GPG key to sign packages, which we should think about how to do properly. Good thing you did Security :wink:

clash-prelude/Setup.hs
christiaanb: Oops. I forgot to add `Warning.hs`. Anyhow, you can also replace that `import` by:

```
#warning You are configuring this package without cabal-doctest installed. \
         The doctests test-suite will not work as a result. \
         To fix this, install cabal-doctest before configuring.
```

christiaanb:
With regards to the static linking warning. I've tried multiple configurations (ghc 8.6.5 vs 8.8, cabal 2.4 vs 3.2) and I've been unable to reproduce the slow-down even though the bug is supposedly not fixed: https://gitlab.haskell.org/ghc/ghc/-/issues/15524

martijnbastiaan:
I can't get tests to work: I'm unable to make a package for `quickcheck-classes-base` which we need for `clash-lib`:

```
cabal copy --builddir=dist-ghc --destdir=debian/tmp-inst-ghc
cabal: unrecognised command: copy (try --help)
make: *** [/usr/share/cdbs/1/class/hlibrary.mk:188: debian/tmp-inst-ghc] Error 1
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
```

martijnbastiaan:
Thanks @christiaanb. That confirms my experiences, so I'm going to pull out the warning.

alex-mckenna:
> I can't get tests to work: I'm unable to make a package for quickcheck-classes-base which we need for clash-lib:

`cabal copy` is a `v1` command only, I think it's replaced by some extra flags to the `v2-{build,install}` commands

martijnbastiaan:
Ok, then it's probably going to get fixed once distributions move to cabal 3.0?

alex-mckenna:
Well the command will be removed entirely I believe. I think you want something like `cabal install --installdir=DIR --install-method=copy` to do what `cabal copy` did

alex-mckenna (approved):
LGTM, I would leave a note in the `bindist` directory saying that it should be moved into `.ci/` when the CI is changed to build / test these debian packages

PR #1598, opened 2020-11-19
Rebase snap on Debian packages

martijnbastiaan:
TODO:

- [x] Move `bindist` dir to `.ci`
- [x] Change CI scripts to use logic in this PR


.ci/bindist/linux/debian/focal/build-and-publish-docker-image.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -xeo pipefail
```
alex-mckenna, line 3: No `set -u`?

alex-mckenna: Oh God, I think I wrote this script initially.

.ci/bindist/linux/snap/Dockerfile
```diff
@@ -0,0 +1,67 @@
+# This Dockerfile is:
+#
+#   https://github.com/snapcore/snapcraft/pull/3384
+#   + xenial => focal
+#
+FROM ubuntu:focal as builder
```
alex-mckenna, line 6: Multi stage dockerfile, nice

.ci/bindist/linux/snap/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+# Grab dependencies
+RUN apt-get update
+RUN apt-get dist-upgrade --yes
```
alex-mckenna, line 10: Again, do we need this `dist-upgrade`?

.ci/bindist/linux/snap/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN apt-get install --yes \
+      curl \
+      jq \
+      squashfs-tools
+
+# Grab the core snap (for backwards compatibility) from the stable channel and
+# unpack it in the proper place.
+RUN curl -L $(curl -H 'X-Ubuntu-Series: 16' 'https://api.snapcraft.io/api/v1/snaps/details/core' | jq '.download_url' -r) --output core.snap
+RUN mkdir -p /snap/core
+RUN unsquashfs -d /snap/core/current core.snap
```
alex-mckenna, line 20: I think each of these three should live in a single layer, they seem like one operation conceptually

.ci/bindist/linux/snap/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN mkdir -p /snap/snapcraft
+RUN unsquashfs -d /snap/snapcraft/current snapcraft.snap
+
+# Create a snapcraft runner (TODO: move version detection to the core of
+# snapcraft).
+RUN mkdir -p /snap/bin
+RUN echo "#!/bin/sh" > /snap/bin/snapcraft
+RUN snap_version="$(awk '/^version:/{print $2}' /snap/snapcraft/current/meta/snap.yaml)" && echo "export SNAP_VERSION=\"$snap_version\"" >> /snap/bin/snapcraft
+RUN echo 'exec "$SNAP/usr/bin/python3" "$SNAP/bin/snapcraft" "$@"' >> /snap/bin/snapcraft
+RUN chmod +x /snap/bin/snapcraft
```
alex-mckenna, line 46: Same with this, it should just be one layer IMO

.ci/bindist/linux/snap/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN apt-get update && apt-get dist-upgrade --yes && apt-get install --yes snapd sudo locales && locale-gen en_US.UTF-8
+
+# Set the proper environment.
+ENV LANG="en_US.UTF-8"
+ENV LANGUAGE="en_US:en"
+ENV LC_ALL="en_US.UTF-8"
+ENV PATH="/snap/bin:$PATH"
+ENV SNAP="/snap/snapcraft/current"
+ENV SNAP_NAME="snapcraft"
+ENV SNAP_ARCH="amd64"
```
alex-mckenna, line 67: Single layer?

alex-mckenna (approved):
LGTM, I would like some layers to be flattened if you agree it's reasonable to do so.

.ci/bindist/linux/snap/Dockerfile
martijnbastiaan: Probably not, it's copied 1:1 from Canonical's `Dockerfile`

testsuite/Main.hs
```diff
@@ -40,6 +47,79 @@ import           Test.Tasty.HUnit
...
+-- | See 'compiledWith'
+data RunWith
+  = Stack
+  | Cabal
+  | Global
+  deriving (Show)
+
+-- | Detects Clash binary the testsuite should use (in order):
+--
+--     * If STACK_EXE is present, use Stack's Clash
```
alex-mckenna, line 76: Is it a good idea to look just for `STACK_EXE`, or should we care if there is a `.stack-work` directory?

alex-mckenna: Hmm, rephrase: I find it strange on first read that we care about an envvar for stack, and the presence of a directory for cabal? What if I built with cabal but also have stack on my system - will it always choose stack?

martijnbastiaan: The logic is: 

* `STACK_EXE` should only be set when running with `stack run testsuite`. 
* If you intend to run the testsuite with a globally installed clash, the testsuite binary is (probably?) also installed globally. I didn't find a way to detect whether we're running in `cabal run`, hence the logic of looking at the presence or absence of `dist-newstyle`.

alex-mckenna: Right, that makes sense. So the goal is to make sure you use the same `clash` when spawning processes as the `clash` as the tool running the testsuite would use?

testsuite/Main.hs
```diff
@@ -93,7 +93,7 @@ setClashEnvs :: HasCallStack => RunWith -> IO ()
 setClashEnvs Global = setEnv "GHC_ENVIRONMENT" "-"
 setClashEnvs Stack = pure ()
 setClashEnvs Cabal = do
-  let cp = proc "cabal" ["--write-ghc-environment-files=always", "run", "--", "clash", "--help"]
+  let cp = proc "cabal" ["--write-ghc-environment-files=always", "new-run", "--", "clash", "--help"]
```
alex-mckenna, line 96: I believe `v2-run` is the preferred way of doing this

testsuite/Main.hs
martijnbastiaan: Yes.

PR #1607, opened 2020-12-01
Remove starter project

martijnbastiaan:
It's now available at https://github.com/clash-lang/clash-starters


basile-henry:
@martijnbastiaan I think if you split the `simple` and `simple-nix` into their own repos then you'd be able to mark each repo as a GitHub template: https://github.blog/2019-06-06-generate-new-repositories-with-repository-templates/

I also think `stack` supports fetching templates straight from GitHub.

martijnbastiaan:
Oh interesting! I'm currently using a repository analogous to https://github.com/commercialhaskell/stack-templates (that is, https://github.com/clash-lang/stack-templates). This does make `stack new clash-lang/simple` and `stack new clash-lang/simple-nix` work. GitHub templates could be even nicer though.

basile-henry:
I actually think `stack new clash-lang/simple` would look for a repo at `https://github.com/clash-lang/simple` by itself if I read the docs correctly :smile: 

From `commercialhaskell/stack-templates`'s README:
> See [Stack documentation](https://docs.haskellstack.org/en/stable/GUIDE/#templates) for instructions on providing templates to your users, for example, by publishing them on GitHub or an arbitrary URL

If this does work you would get both GitHub and stack templates fairly easily, no?

martijnbastiaan:
Where do you read that? I see:

Selection_245

So `clash-lang/simple` tries to download `clash-lang/stack-templates/simple.hsfiles`. Which is how I set it up :).

basile-henry:
You are right, I did misunderstand it! So I guess the GitHub templates aren't quite as powerful since they don't automatically fill in some details like the stack one does (like email, name). If you do decide to have both stack templates and GitHub templates that'll end up being more work to maintain both :sweat_smile:

PR #1624, opened 2021-01-04
Add support for GHC 9

christiaanb:
- [x] Upgrade `clash-prelude` to ghc9
  - [x] Have `Clash.Sized.Internal.Mod` use `ghc-bignum`
  - [x] Fix failing `clash-prelude:unittests`, suspect the source of the failure can be found in using the now `integer-gmp`-shim and its `Internal` modules incorrectly
  - [x] Fix failing `clash-prelude:doctests`
    - [x] Change expected error messages
    - [x] Change type signature of Typed TH `bLit` and `fLit` for new `Code` quoting mechanims
- [x] Upgrade `clash-lib` to ghc9
  - [x] Adapt primitives to `ghc-bignum` naming
- [x] Upgrade `clash-ghc` to ghc9
  - [x] Fix all the ghc api calls
  - [x] Add clash-specific calls to `src-bin-9.0`
  - [x] Adapt primitives to `ghc-bignum` naming
- [x] Fix bugs

If you want to explore this branch, you'll need the following in your `cabal.project.local`:

```
repository head.hackage.ghc.haskell.org
   url: https://ghc.gitlab.haskell.org/head.hackage/
   secure: True
   key-threshold: 3
   root-keys:
       f76d08be13e9a61a377a85e2fb63f4c5435d40f8feb3e12eb05905edb8cdea89
       26021a13b401500c8eb2761ca95c61f2d625bfef951b939a8124ed12ecf07329
       7541f32a4ccca4f97aea3b22f5e593ba2c0267546016b992dfadcd2fe944e55d

allow-newer: *:base, *:template-haskell, *:ghc, *:stm, *:ghc-boot, *:ghc-prim, *:ghci, *:integer-gmp, *:Cabal

source-repository-package
    type: git
    location: https://github.com/haskell-hint/hint.git
    tag: fc4651a7c327c531645c8d4d2d03d9ed7d324f9f

source-repository-package
    type: git
    location: https://github.com/utdemir/haskell-hedgehog.git
    tag: c98aa9e33bf6871098d6f4ac94eeaac10383d696
    subdir: hedgehog
```

Note that the `tests/shouldwork/Numbers/Natural.hs` test will fail on GHC-9 because the function for subtraction on `Natural` is no longer marked NOINLINE which results in two failures:
1. ANF will fail when calling `mkSelectorCase` on `Natural` because the TyCon doesn't have any registered DataCons. Even if we would hardcode projection for `Natural` the issue that pops up then is that we would project `ByteArray#` out of the Natural for the `NB` constructor, where `ByteArray#` is not representable.
2. There's an exposed FFI call which we cannot even begin to make a primitive/blackbox for until https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4832 is merged, see also https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3817#note_324128


clash-lib/prims/vhdl/GHC_Num_Integer.json
```diff
@@ -0,0 +1,304 @@
...
+, { "BlackBox" :
+    { "name"      : "GHC.Num.Integer.integerCompare"
+    , "kind"      : "Declaration"
+    , "type"      : "integerCompare :: Integer -> Integer -> Ordering"
+    , "template"  : "
+-- begin integerCompare
+~RESULT <= \"00\" when ~ARG[0] < ~ARG[1] else
+           \"01\" when ~ARG[0] = ~ARG[1] else
+           \"10\";
+-- end integerCompare
```
martijnbastiaan, line 300: It would be nice if we could encode sum types as a VHDL enums.

clash-lib/src/Clash/Core/EqSolver.hs
```diff
@@ -124,6 +136,12 @@ isAbsurdAlt
   :: TyConMap
   -> Alt
   -> Bool
+#if MIN_VERSION_base(4,15,0)
+isAbsurdAlt _tcm (DataPat dc _ _,_)
+  -- unsafeCoerce is not absurd in the way intended by /isAbsurdAlt/
+  | dcUniq dc == getKey unsafeReflDataConKey
+  = False
```
martijnbastiaan, line 143: Why is this only a thing for 9.0?

clash-prelude/src/Clash/Signal/BiSignal.hs
```diff
@@ -183,8 +183,13 @@ data BiSignalIn (ds :: BiSignalDefault) (dom :: Domain) (n :: Nat)
 --
 -- Wraps (multiple) writing signals. The semantics are such that only one of
 -- the signals may write at a single time step.
+#if MIN_VERSION_base(4,15,0)
+data BiSignalOut (ds :: BiSignalDefault) (dom :: Domain) (n :: Nat)
+  = BiSignalOut ![Signal dom (Maybe (BitVector n))]
+#else
 newtype BiSignalOut (ds :: BiSignalDefault) (dom :: Domain) (n :: Nat)
   = BiSignalOut [Signal dom (Maybe (BitVector n))]
+#endif
```
martijnbastiaan, line 192: What's up with this change? I wouldn't mind making it strict for all versions (changelog necessary though).

clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -179,7 +179,7 @@ traceSignal# traceMap period traceName signal =
           m
       , signal)
  where
-  width = snatToNum (SNat @ (BitSize a))
+  width = snatToNum (SNat @(BitSize a))
```
martijnbastiaan, line 182: No change needed for now, but this pattern can be replaced by:

```haskell
natToNum @(BitSize a)
```

clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -133,10 +138,17 @@ import Clash.XException
...
+data Index (n :: Nat) =
+    -- | The constructor, 'I', and the field, 'unsafeToInteger', are not
+    -- synthesizable.
+    I { unsafeToInteger :: !Integer }
+#else
 newtype Index (n :: Nat) =
     -- | The constructor, 'I', and the field, 'unsafeToInteger', are not
     -- synthesizable.
     I { unsafeToInteger :: Integer }
+#endif
```
martijnbastiaan, line 151: This changes the ability to do `coerce` on `Index`es. Wouldn't it make sense to have the same behavior for all ghc versions? Did this change show up in benchmarks at all?

clash-prelude/src/Clash/Sized/Internal/Signed.hs
```diff
@@ -157,10 +162,17 @@ import Clash.XException
 -- 3
 -- >>> satAdd SatSymmetric (-2) (-3) :: Signed 3
 -- -3
+#if MIN_VERSION_base(4,15,0)
+data Signed (n :: Nat) =
+    -- | The constructor, 'S', and the field, 'unsafeToInteger', are not
+    -- synthesizable.
+    S { unsafeToInteger :: !Integer}
+#else
```
martijnbastiaan, line 170: Same comment ^

clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
```diff
@@ -153,15 +171,26 @@ import Clash.XException
 -- 7
 -- >>> satSub SatSymmetric 2 3 :: Unsigned 3
 -- 0
+#if MIN_VERSION_base(4,15,0)
+data Unsigned (n :: Nat) =
+    -- | The constructor, 'U', and the field, 'unsafeToInteger', are not
+    -- synthesizable.
+    U { unsafeToNatural :: !Natural }
+#else
```
martijnbastiaan, line 179: Same comment ^

martijnbastiaan (changes requested):
Overall LGTM, I'm sorry you had to go through this :sweat_smile: . I've left some comments and questions.

The `newtype` -> `data` conversions probably need a changelog (mentioning that it only applies to 9.0 - although see my PR comments). Likewise for the strictness change in BiSignal.

clash-lib/prims/common/GHC_Num_Natural.json
```diff
@@ -0,0 +1,77 @@
...
+      "comment": "Needed to make the evaluator handle this constructor strictly"
+    }
+  },
+  {
+    "BlackBox": {
+      "name": "GHC.Num.Natural.naturalAdd",
+      "kind": "Expression",
+      "type": "naturalSub :: Natural -> Natural -> Natural",
+      "template": "~ARG[0] + ~ARG[1]",
+      "warning": "GHC.Num.Natural.naturalAdd: Naturals are dynamically sized in simulation, but fixed-length after synthesis. Use carefully."
```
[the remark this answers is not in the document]
martijnbastiaan: I meant to comment on the 'warning' part.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -3528,7 +3862,7 @@ wrapUnsigned n i = i `mod` sz
   sz = 1 `shiftL` fromInteger n
 
 wrapSigned :: Integer -> Integer -> Integer
-wrapSigned n i = if mask == 0 then 0 else res
+wrapSigned n i = if n == 0 then 0 else res
```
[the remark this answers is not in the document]
christiaanb: Yes, correct.

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -48,6 +48,44 @@ import qualified Data.Traversable            as T
...
+   isClosedSynFamilyTyConWithAxiom_maybe, expandSynTyCon_maybe, tyConArity,
+   tyConDataCons, tyConKind, tyConName, tyConUnique, isClassTyCon)
+import GHC.Core.Type (mkTvSubstPrs, substTy, coreView)
+import GHC.Core.TyCo.Rep (Coercion (..), TyLit (..), Type (..), scaledThing)
+import GHC.Types.Unique (Uniquable (..), Unique, getKey, hasKey)
+import GHC.Types.Var
+  (Id, TyVar, Var, VarBndr (..), idDetails, isTyVar, varName, varType,
+   varUnique, idInfo, isGlobalId)
+import GHC.Types.Var.Set (isEmptyVarSet)
+#else
```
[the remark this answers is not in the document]
christiaanb: Yes, I'm aware. But it's incomplete. Also, moving to the new hierarchy makes it easier to test against ghc-head or new alphas or rc's in the future, and eventually dropping the flat hierarchy once we stop supporting ghc versions older than 9.0

clash-lib/src/Clash/Core/EqSolver.hs
christiaanb: Because of the new implementation of unsafeCoerce based on pattern-matching on `UnsafeRefl` is a base 4.15 and up thing. Before that, unsafeCoerce would be translated to a cast at the GHC core level.

clash-prelude/src/Clash/Signal/BiSignal.hs
christiaanb: This is needed because for any newtype `T`. GHC will transform:

```
f :: a -> T
{-# NOINLINE f #-}
```

to:
```
f1 :: a -> RepOfT
{-# NOINLINE f1 #-}

f = f1 |> (co :: a -> RepOfT ~ a w %'Many-> T)
``` 
Which is a `LinearTypes` thing. Note how `f` no longer has the NOINLINE!!! That means we can no longer recognize our primitives. The work-around is to use `data T`. However:

```
data T = MkT a
```
is not the same as:
```
newtype T = MkT a
```
because data-constructors are lazy. Closer in spirit to the newtype is:
```
data T = MkT !a
```
It's still not the same as the newtype, as we do add another machine word to represent the data constructor `MkT`; but the field is no longer a pointer to the thunk `a` but but an actual value `a`. All in all, performance is going to be worse, so it's definitely not something we want for all versions.

clash-prelude/src/Clash/Sized/Internal/Index.hs
christiaanb: No, see my earlier comment, the ghc-9 change makes performance likely worse; but it's the only work-around that I know of.

clash-prelude/src/Clash/Sized/Internal/Signed.hs
christiaanb: Same response

clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
christiaanb: Same response

christiaanb:
@martijnbastiaan I've added the changelog entry w.r.t. the `newtype -> data` changes.

martijnbastiaan (approved):

Issue #1625, opened 2021-01-05
Primitive warning field is ignored with PrimitiveGuard annotation


JvWesterveld:
For example, use of [`clockGen`](https://github.com/clash-lang/clash-compiler/blob/b0bab9c293a075975ac5cbad71db0d7345576ce5/clash-prelude/src/Clash/Signal/Internal.hs#L816) does not result in its warning (on it not being synthesizable) being thrown during compilation because it is annotated with `HasBlackBox`:

```topEntity = delay @System clockGen enableGen True```

PR #1626, opened 2021-01-05
Retain primitive warning when `HasBlackBox` annotation is present

JvWesterveld:
A `WarnNonSynthesizable` `PrimitiveGuard` is [added by `Clash.Primitives.Util.resolvePrimitive'`](https://github.com/clash-lang/clash-compiler/blob/b0bab9c293a075975ac5cbad71db0d7345576ce5/clash-lib/src/Clash/Primitives/Util.hs#L104) whenever a primitive template has a `warning` field. In case such a primitive also has a `PrimitiveGuard` annotation, the annotation took precedence and the warning was always lost. This change makes that the warning is retained if the annotation is `HasBlackBox`; `WarnNonSynthesizable` implies `HasBlackBox` anyway.

Fixes the example mentioned in #1625. The warning is still lost when a `WarnNonSynthesizable` or `WarnAlways` annotation is present, but that's not that bad I think.

### DIFF
--- a/changelog/2021-01-05T11_10_47+01_00_keep_prim_warning
+++ b/changelog/2021-01-05T11_10_47+01_00_keep_prim_warning
@@ -1,0 +1,1 @@
+FIXED: Primitive template warning is now retained when a `HasBlackBox` annotation is present.
--- a/clash-lib/src/Clash/Primitives/Util.hs
+++ b/clash-lib/src/Clash/Primitives/Util.hs
@@ -123,30 +123,33 @@ addGuards
   -> ResolvedPrimMap
 addGuards = foldl go
  where
-  lookupPrim :: TS.Text -> ResolvedPrimMap -> Maybe ResolvedPrimitive
-  lookupPrim nm primMap = join (extractPrim <$> HashMapStrict.lookup nm primMap)
-
   go primMap (nm, guard) =
     HashMapStrict.insert
       nm
-      (case (lookupPrim nm primMap, guard) of
-        (Nothing, HasBlackBox _) ->
+      (case (prim, guardedPrim, guard) of
+        (Nothing, _, HasBlackBox _) ->
           error $ "No BlackBox definition for '" ++ TS.unpack nm ++ "' even"
                ++ " though this value was annotated with 'HasBlackBox'."
-        (Nothing, WarnNonSynthesizable _ _) ->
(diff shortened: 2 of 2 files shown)


martijnbastiaan:
This does seem to suggest we need another datatype for `PrimitiveGuard`. E.g. instead of

```haskell
data PrimitiveGuard a
  = DontTranslate
  | HasBlackBox a
  | WarnNonSynthesizable String a
  | WarnAlways String a
```

something like:

```haskell
data PrimitiveGuard a
  = DontTranslate
  | EmitWarnings [PrimitiveWarning] a

data PrimitiveWarning
  = WarnNonSynthesizable String a
  | WarnAlways String a
```

maybe even:

```haskell
data PrimitiveGuard a
  = DontTranslate
  | Warnings (NonEmpty PrimitiveWarning) a
  | NoWarnings a

data PrimitiveWarning
  = WarnNonSynthesizable String a
  | WarnAlways String a
```

what do you think?

JvWesterveld:
Yes, I think that would be a better solution, though it would break the API.
I'd go for this:
```
data PrimitiveGuard a
  = DontTranslate
  | HasBlackBox [PrimitiveWarning] a

data PrimitiveWarning
  = WarnNonSynthesizable String a
  | WarnAlways String a
```

martijnbastiaan:
> I'd go for this:

Looks good!

>  though it would break the API.

The documentation discourages the use of the constructors, instead saying users should use the functions, so I don't think this is a problem :)

martijnbastiaan:
`PrimitiveGuard` then has a sensible `Monoid` instance I think, which you might use to slap multiple together.

JvWesterveld:
Alright, I'll probably give that a go sometime soon.

> `PrimitiveGuard` then has a sensible `Monoid` instance I think, which you might use to slap multiple together.

Good idea :)

JvWesterveld:
I think a `Semigroup` instance for `PrimitiveGuard` wouldn't be helpful here because [addGuards](https://github.com/clash-lang/clash-compiler/blob/1bde4831c94e38a4359be2c9c5fbab8e871569f6/clash-lib/src/Clash/Primitives/Util.hs#L120) combines `PrimitiveGuard`s of different types.

JvWesterveld:
Superseded by #1628.

PR #1628, opened 2021-01-08
Split warnings of `PrimitiveGuard` into separate data type

### LINKED ISSUE #1625: Primitive warning field is ignored with PrimitiveGuard annotation

JvWesterveld:
Add a separate data type for representing warnings for `PrimitiveGuard`s so that multiple can be represented. Alternative approach to #1626.

Fixes #1625.

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox.hs
@@ -49,8 +49,9 @@ import           TextShow                      (showt)
 import           Util                          (OverridingBool(..))
 
 import           Clash.Annotations.Primitive
-  (PrimitiveGuard(HasBlackBox, WarnNonSynthesizable, WarnAlways, DontTranslate),
-   extractPrim)
+  ( PrimitiveGuard(HasBlackBox, DontTranslate)
+  , PrimitiveWarning(WarnNonSynthesizable, WarnAlways)
+  , extractPrim)
 import           Clash.Core.DataCon            as D (dcTag)
 import           Clash.Core.FreeVars           (freeIds)
 import           Clash.Core.Literal            as L (Literal (..))
@@ -281,10 +282,18 @@ extractPrimWarnOrFail
 extractPrimWarnOrFail nm = do
   prim <- HashMap.lookup nm <$> Lens.use primitives
   case prim of
-    Just guardedPrim ->
-      -- See if we need to warn the user, or error because we encountered
-      -- a primitive the user explicitly requested not to translate
-      go guardedPrim
+    Just (HasBlackBox warnings compiledPrim) ->
+      -- See if we need to warn the user
+      if null warnings then return compiledPrim else go warnings compiledPrim
+    Just DontTranslate -> do
+      -- We need to error because we encountered a primitive the user
+      -- explicitly requested not to translate
+      (_,sp) <- Lens.use curCompNm
+      let msg = $(curLoc) ++ "Clash was forced to translate '" ++ unpack nm
+             ++ "', but this value was marked with DontTranslate. Did you forget"
+             ++ " to include a blackbox for one of the constructs using this?"
+             ++ (if debugIsOn then "\n\n" ++ prettyCallStack callStack ++ "\n\n" else [])
+      throw (ClashException sp msg Nothing)
     Nothing -> do
       -- Blackbox requested, but no blackbox found at all!
       (_,sp) <- Lens.use curCompNm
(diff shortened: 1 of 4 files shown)


JvWesterveld:
Should I move the second commit to a separate PR?

clash-lib/src/Clash/Netlist/BlackBox.hs
```diff
@@ -281,10 +282,18 @@ extractPrimWarnOrFail
 extractPrimWarnOrFail nm = do
   prim <- HashMap.lookup nm <$> Lens.use primitives
   case prim of
-    Just guardedPrim ->
-      -- See if we need to warn the user, or error because we encountered
-      -- a primitive the user explicitly requested not to translate
-      go guardedPrim
+    Just (HasBlackBox warnings compiledPrim) ->
+      -- See if we need to warn the user
+      if null warnings then return compiledPrim else go warnings compiledPrim
```
JvWesterveld, line 287: This if statement isn't strictly necessary, but this way the same primitives should end up in the `_seenPrimitives` field of the `NetlistState` as was the case before. That might be a little faster.

clash-prelude/src/Clash/Annotations/Primitive.hs
```diff
@@ -241,37 +273,19 @@ data Primitive
...
--- @
--- {\-\# ANN f (warnNonSynthesizable "Tread carefully, user!") \#-\}
--- @
---
--- or
---
--- @
--- {\-\# ANN f (warnAlways "Tread carefully, user!") \#-\}
--- @
+-- For use, see `dontTranslate`, `hasBlackBox`, `warnNonSynthesizable` and
```
martijnbastiaan, line 280: If you use single quotes to reference functions it will get properly linked in the documentation. E.g.,

```
'dontTranslate'
```

clash-lib/src/Clash/Primitives/Util.hs
```diff
@@ -101,10 +101,10 @@ resolvePrimitive' metaPath BlackBox{template=t, includes=i, resultNames=r, resul
           <*> traverse resolveSourceM ri
           <*> resolveSourceM t
   case warning of
-    Just w  -> pure (name, WarnNonSynthesizable (TS.unpack w) bb)
-    Nothing -> pure (name, HasBlackBox bb)
+    Just w  -> pure (name, HasBlackBox [WarnNonSynthesizable $ TS.unpack w] bb)
```
martijnbastiaan, line 104: We prefer `()` over `$` :)

clash-lib/src/Clash/Primitives/Util.hs
```diff
@@ -123,29 +123,29 @@ addGuards
...
-               ++ " though this value was annotated with 'WarnAlways'"
-               ++ ", implying it has a BlackBox."
         (Just _, DontTranslate) ->
           error (TS.unpack nm ++ " was annotated with DontTranslate, but a "
-                                 ++ "BlackBox definition was found anyway.")
-        (Nothing, DontTranslate) -> DontTranslate
-        (Just p, g) -> fmap (const p) g)
+                              ++ "BlackBox definition was found anyway.")
+        (Just (ws1, p), HasBlackBox ws2 ()) ->
+          HasBlackBox (ws1 ++ ws2) p
```
martijnbastiaan, line 147: It's been a while since I've looked at this code, but I'm a little bit surprised by this. Do you know why these warnings are merged this late? Shouldn't they have been merged in all the compilation / discover steps we do?

martijnbastiaan (changes requested):
Looks good overall! I've added some suggestions and a question.

martijnbastiaan:
Next time feel free to just force push over a PR. Keeps the discussion neatly in one place :).

clash-lib/src/Clash/Primitives/Util.hs
JvWesterveld: I'm not sure what you refer to as the compilation / discover steps, or what a better place for merging the warnings could be, but the reason that they need to be merged this late is that [just a couple of lines up](https://github.com/clash-lang/clash-compiler/blob/1bde4831c94e38a4359be2c9c5fbab8e871569f6/clash-lib/src/Clash/Primitives/Util.hs#L103-L104) a warning field of a primitive template yields a new `PrimitiveGuard` with a warning, and there may already be a `PrimitiveGuard` annotation with warnings.

clash-prelude/src/Clash/Annotations/Primitive.hs
JvWesterveld: Sure, but it looks like they do the same thing: [https://haskell-haddock.readthedocs.io/en/latest/markup.html#hyperlinked-identifiers](https://haskell-haddock.readthedocs.io/en/latest/markup.html#hyperlinked-identifiers)

martijnbastiaan: Ah, I thought it didn't. In that case it's fine!

clash-lib/src/Clash/Primitives/Util.hs
martijnbastiaan: Aha, I misremember how this worked. Looks perfect!

martijnbastiaan (approved):

clash-lib/src/Clash/Primitives/Util.hs
JvWesterveld: Fixed

clash-prelude/src/Clash/Annotations/Primitive.hs
JvWesterveld: Changed it anyway while I was at it.

JvWesterveld:
Added a changelog entry.

PR #1645, opened 2021-02-03
Force accumulating argument 'o' to WHNF in 'asyncRegister#'

### LINKED ISSUE #1644: Space leak when simulating ClearOnReset RAM
In 1993, a gaming magazine reported on [Microsoft's upcoming astronautical game](https://en.wikipedia.org/wiki/Microsoft_Space_Simulator); that was a **Space Simulator leak**. This ticket, instead, is about a **simulator space leak**. 

In a situation eerily reminiscent of #1256, I have discovered that the following program keeps using more and more memory:

```
import Clash.Prelude
import Control.Arrow.Transformer.Automaton

main :: IO ()
main = do
    let go (Automaton step) = do
            let (out, next) = step ()
            printX out
            go next
    go $ signalAutomaton @System board1

board1 :: (HiddenClockResetEnable dom) => Signal dom () -> Signal dom (Unsigned 1)
board1 inp = blockRam1 @_ @_ @(Unsigned 1) ClearOnReset (SNat @1) 0 (pure 0) (pure Nothing)
```

while changing the `blockRam1`'s `ResetStrategy` to `NoClearOnReset` results in a constant memory usage. This can easily be tested by running with the flag `+RTS -M2M`, since 2M of heap is enough for the non-leaking one, while the leaking one quickly runs out of it after ~5000 simulated periods.

martijnbastiaan:
Fixes memory leak triggered on asserted resets. This went under the
radar for a long time, as reset periods are typically very short.

Fixes #1644

### DIFF
--- a/clash-prelude/src/Clash/Explicit/DDR.hs
+++ b/clash-prelude/src/Clash/Explicit/DDR.hs
@@ -180,7 +180,7 @@ ddrOut# clk rst en i0 xs ys =
     -- That is why we drop the first value of the stream.
     let (_ :- out) = zipSig xs' ys' in out
   where
-    xs' = register# clk rst en (error "ddrOut: unreachable error") i0 xs
+    xs' = register# clk rst en (errorX "ddrOut: unreachable error") i0 xs
     ys' = register# clk rst en (deepErrorX "ddrOut: initial value undefined") i0 ys
     zipSig (a :- as) (b :- bs) = a :- b :- zipSig as bs
 {-# NOINLINE ddrOut# #-}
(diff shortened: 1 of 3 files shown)


clash-prelude/src/Clash/Explicit/DDR.hs
```diff
@@ -180,7 +180,7 @@ ddrOut# clk rst en i0 xs ys =
     -- That is why we drop the first value of the stream.
     let (_ :- out) = zipSig xs' ys' in out
   where
-    xs' = register# clk rst en (error "ddrOut: unreachable error") i0 xs
+    xs' = register# clk rst en (errorX "ddrOut: unreachable error") i0 xs
```
christiaanb, line 183: @leonschoorl Do you remember why this was originally `error` and not `deepErrorX` like it is for `ys'`?

PR #1646, opened 2021-02-04
Fixed enum fix

DigitalBrains1:
This PR makes the `Enum` instance of `Fixed` conform to the laws in the [Haskell Report](https://www.haskell.org/onlinereport/haskell2010/haskellch6.html#dx13-131001) as well as the doc for the [`Enum`](https://hackage.haskell.org/package/base-4.14.1.0/docs/Prelude.html#t:Enum) class on Hackage. This PR also includes some small improvements/fixes related to the `Fixed` type.

The behaviour of `Float` and `Double` described in the Report, which also matches the behaviour of `Rational`, is implemented for `Fixed`, including the rule for the ending value of arithmetic sequences.

Because `Fixed` is an instance of both `Bounded` and `Fractional`, and both the Report and the `Enum` class documentation say `succ`/`pred` add/subtract 1, the rule in the Report:
> For any type that is an instance of class Bounded as well as Enum, the following should hold:
> * The calls succ maxBound and pred minBound should result in a runtime error.

is interpreted as:
> * If the result of a call of succ or pred cannot be represented, they should result in a runtime error.

because there are more values than just `succ maxBound` and `pred minBound` that cannot be represented. I suspect the Report was written without considering something that is both `Bounded` and `Fractional`.

It should be noted that [`Data.Fixed`](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Fixed.html) from `base` breaks the rule from the documentation (in base!) of the `Enum` class that `succ`/`pred` adds/subtracts 1, and they don't follow the extended rules for ending values of arithmetic sequences either.

A list of further comments or points of discussion:

* `Typeable` is now in `NumFixedC`, we use it to format error messages in `Enum`.
* In the unittests, Hedgehog `Size` scales the length of generated arithmetic sequences, with the largest `Size`s generating lists of up to 1000 elements. Because of how Hedgehog works, most will be shorter. Is this a good length?
* I added `Hedgehog.Extra.throwsException`, and only later discovered there was something similar at `Clash.Tests.Resize.expectExceptionNoX`. Do we harmonise this? The naming style differs: do we describe the test or the result?
* Should `throwsException` have `HasCallStack` / use `withFrozenCallStack`?
* I harmonised the tested lengths of `Fixed` over several tests. In particular, I'd like to point out that `SFixed 2 0` is the smallest type that is capable of expressing the number 1. Given that we've had issues with representing the number 1 in the past, it seems prudent to include it in our tests.
* For `Signed` and `Unsigned`, `enumFrom*` has `NOINLINE`. Do that here as well? Or is this another remnant of the never executed plans for enhanced error messages for non-synthesisable code?
* We need all of `NumFixedC` as constraint for `Enum Fixed`, this was tested. This test did reveal a redundancy: `Integral` implies both `Enum` and `Ord`, so the latter probably need not even be in `NumFixedC`.
* I based the error messages on the ones in the Haskell Prelude.
* I use `SatWrap` for speed.
* Redundant, narrower-scope tests for `Fixed` removed from `Clash.Tests.Laws.Enum`.
* `Clash.Tests.Laws.SaturatingNum` has tests for `satSucc`/`satPred`, of which only part was run for Fixed. All of them are now superfluous. The `isTotal` tests there, however, are not anywhere else yet. These tests just verify there are no runtime errors; with the infrastructure of `Clash.Tests.Fixed`, we could also verify the result.

### DIFF
--- a/clash-prelude/src/Clash/Sized/Fixed.hs
+++ b/clash-prelude/src/Clash/Sized/Fixed.hs
@@ -956,7 +956,7 @@ instance NumFixedC rep int frac => SaturatingNum (Fixed rep int frac) where
                          1 -> unpack (resize (shiftR rR sh))
                          _ -> case msb rL of
                                 0 -> maxBound
-                                _ -> succ minBound
+                                _ -> Fixed $ succ minBound
           False -> case rL of
                      0 -> unpack (resize (shiftR rR sh))
                      _ -> maxBound
--- a/clash-prelude/src/Clash/Sized/Fixed.hs
+++ b/clash-prelude/src/Clash/Sized/Fixed.hs
@@ -69,6 +69,7 @@ operator that uses truncation introduces an additional error of /0.109375/:
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE NegativeLiterals #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -382,7 +383,7 @@ Where 'NumSFixedC' refers to the @Constraints@ needed by the operators of
 the 'Num' class for the 'SFixed' datatype.
 
 Although the number of constraints for the @mac@ function defined earlier might
-be considered small, here is an \"this way lies madness\" example where you
+be considered small, here is a \"this way lies madness\" example where you
 really want to use constraint kinds:
 
 @
@@ -1094,10 +1095,10 @@ instance (NumFixedC rep int frac, Integral (rep (int + frac))) =>
      denom     = 1 `shiftL` nF
      nom       = toInteger fRep
 
-instance (FracFixedC rep int frac, NumFixedC rep int frac, Integral (rep (int + frac))) =>
-         RealFrac (Fixed rep int frac) where
+instance FracFixedC rep int frac => RealFrac (Fixed rep int frac) where
   properFraction f@(Fixed fRep) = (fromIntegral whole, fract)
     where
       whole = (fRep `shiftR` fracShift f) + offset
       fract = Fixed $ fRep - (whole `shiftL` fracShift f)
-      offset = if f < 0 then 1 else 0
+      frMask = fromInteger $ (1 `shiftL` fracShift f) - 1
+      offset = if f < 0 && fRep .&. frMask /= 0 then 1 else 0
--- a/clash-prelude/src/Clash/Sized/Fixed.hs
+++ b/clash-prelude/src/Clash/Sized/Fixed.hs
@@ -119,7 +119,7 @@ import Text.Read                  (Read(..))
 import Data.List                  (find)
 import Data.Proxy                 (Proxy (..))
 import Data.Ratio                 ((%), denominator, numerator)
-import Data.Typeable              (Typeable, TypeRep, typeRep)
+import Data.Typeable              (Typeable, TypeRep, typeRep, typeOf)
 import GHC.TypeLits               (KnownNat, Nat, type (+), natVal)
 import GHC.TypeLits.Extra         (Max)
 import Language.Haskell.TH        (Q, appT, conT, litT, mkName,
@@ -142,7 +142,7 @@ import Clash.Class.Num            (ExtendingNum (..), SaturatingNum (..),
                                    boundedMul)
 import Clash.Class.Resize         (Resize (..))
 import Clash.Promoted.Nat         (SNat, natToNum, natToInteger)
-import Clash.Prelude.BitIndex     (msb, split)
+import Clash.Prelude.BitIndex     (lsb, msb, split)
 import Clash.Prelude.BitReduction (reduceAnd, reduceOr)
 import Clash.Sized.BitVector      (BitVector, (++#))
 import Clash.Sized.Signed         (Signed)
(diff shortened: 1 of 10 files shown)


clash-prelude/tests/Clash/Tests/FixedExhaustive.hs
```diff
@@ -0,0 +1,203 @@
...
+  , limitTestWidth enumTests (Proxy @(SFixed 3 2))
+  , limitTestWidth enumTests (Proxy @(SFixed 4 1))
+  , limitTestWidth enumTests (Proxy @(SFixed 5 0))
+  , limitTestWidth enumTests (Proxy @(SFixed 0 6))
+  , limitTestWidth enumTests (Proxy @(SFixed 1 5))
+  , limitTestWidth enumTests (Proxy @(SFixed 2 4))
+  , limitTestWidth enumTests (Proxy @(SFixed 3 3))
+  , limitTestWidth enumTests (Proxy @(SFixed 4 2))
+  , limitTestWidth enumTests (Proxy @(SFixed 5 1))
+  , limitTestWidth enumTests (Proxy @(SFixed 6 0))
```
DigitalBrains1, line 174: I'm axing this.

As martijn said, at a quick glance people might conclude these datatypes are tested when in fact they are excluded through `widthLimit`. And on second thought it doesn't add useful functionality, not useful enough.

DigitalBrains1:
Changed:
* Added `{S,U}Fixed 7 7` and `121 121` to check non-power-of-two types, both within machine word length and beyond it. Discussed this regarding computation of masks that we'd like to check for non-power-of-two.
* Removed the `widthLimit` mechanism from FixedExhaustive; just test up to four bit width. No longer a strict limit on list length; just a generic and generous 1000 that'll do fine for preventing computational runaway.
* Minor fixes.

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -904,6 +905,182 @@ fLitR a = Fixed (fromInteger sat)
...
+enumFromThenToUp
+  :: forall f rep int frac
+   . ( NumFixedC rep int frac
+     , f ~ Fixed rep int frac)
+  => f
+  -> f
+  -> f
+  -> [f]
+enumFromThenToUp x1 x2 y
+  | y < x1 = let !y' = satAdd SatWrap y halfDelta  -- Never wraps
```
martijnbastiaan, line 1024:
```haskell
let !x = ....
```

basically means "when you evaluate `x` to WHNF, evaluate `x` to WHNF", which is superfluous. Bang patterns are useful in arguments, e.g.:


```haskell
foo !a !b = c
  where
    c = ....
```

where it would mean "when you evaluate the output of `foo`, first evaluate `a`, then `b`, and finally the output to WHNF. It's typically used to prevent huge chains of chunks from building up in accumulation parameters (which I think you mentioned during the call).

clash-prelude/src/Clash/Sized/Fixed.hs
```diff
@@ -904,6 +905,182 @@ fLitR a = Fixed (fromInteger sat)
...
+        rMax = toInteger (maxBound :: rep (int + frac))
+        rMin = toInteger (minBound :: rep (int + frac))
+    in if res > rMax || res < rMin then
+         error $  "Enum.toEnum{"
+               ++ show (typeRep $ Proxy @(Fixed rep int frac)) ++ "}: tag ("
+               ++ show i ++ ") is outside of bounds "
+               ++ show ( minBound :: Fixed rep int frac
+                       , maxBound :: Fixed rep int frac)
+       else
+         Fixed (fromInteger res)
```
martijnbastiaan, line 955: If you want to prevent all the indention whitespace, consider:

```haskell
  toEnum i =
    if res > rMax || res < rMin then
      error $  "Enum.toEnum{"
             ++ show (typeRep $ Proxy @(Fixed rep int frac)) ++ "}: tag ("
             ++ show i ++ ") is outside of bounds "
             ++ show ( minBound :: Fixed rep int frac
                     , maxBound :: Fixed rep int frac)
    else
       Fixed (fromInteger res)
   where
    sh   = natToNum @frac
    res  = toInteger i `shiftL` sh
    rMax = toInteger (maxBound :: rep (int + frac))
    rMin = toInteger (minBound :: rep (int + frac))
```

Especially in larger blocks, it makes it easier to see how many control structures deep you are IMO.

martijnbastiaan: We should really add something like `Clash.Util.Interpolate` to `clash-prelude`. Writing error messages like this:

https://github.com/clash-lang/clash-compiler/blob/250aea43ce6111cd55cdb83e1ad30761024c9bb0/clash-lib/src/Clash/Core/TermInfo.hs#L185-L193

is so much easier to read (and write)!

martijnbastiaan (approved):
Overall LGTM! Thanks for the explanation in the call :)

clash-prelude/src/Clash/Sized/Fixed.hs
DigitalBrains1: I've dropped all bangs since we discovered the reason for them in `Int` is unrelated.

clash-prelude/src/Clash/Sized/Fixed.hs
DigitalBrains1: I've transformed the `let` to `where` so the indentation is less.

DigitalBrains1:
I've removed some redundant constraints as general cleanup (redundant in the sense that the constraint was stated twice or was implied by a different one already).

I'd like to discuss this remaining point:
* For `Signed` and `Unsigned`, `enumFrom*` has `NOINLINE`. Do that here as well? Or is this another remnant of the never executed plans for enhanced error messages for non-synthesisable code?

Other than that, I think we're there with this PR.

PR #1647, opened 2021-02-04
Count applied transformations individually

leonschoorl:
Use debug level `DebugCount` see an overview
of how often each transformation is applied.

### DIFF
--- a/clash-lib/src/Clash/Rewrite/Types.hs
+++ b/clash-lib/src/Clash/Rewrite/Types.hs
@@ -32,7 +32,8 @@ import Control.Monad.Writer                  (MonadWriter (..))
 import Data.Binary                           (Binary)
 import Data.Hashable                         (Hashable)
 import Data.IntMap.Strict                    (IntMap)
-import Data.Monoid                           (Any)
+import Data.Map.Strict                       (Map)
+import Data.Monoid                           (Any,Sum)
 import qualified Data.Set                    as Set
 import GHC.Generics
 
@@ -74,7 +75,9 @@ data RewriteStep
 data RewriteState extra
   = RewriteState
   { _transformCounter :: {-# UNPACK #-} !Int
-  -- ^ Number of applied transformations
+  -- ^ Total number of applied transformations
+  , _transformCounters :: Map String (Sum Word)
+  -- ^ Map that tracks how many times each transformation is applied
   , _bindings         :: !BindingMap
   -- ^ Global binders
   , _uniqSupply       :: !Supply
--- a/clash-lib/src/Clash/Rewrite/Types.hs
+++ b/clash-lib/src/Clash/Rewrite/Types.hs
@@ -1,9 +1,10 @@
 {-|
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016     , Myrtle Software Ltd,
-                    2017     , Google Inc.
+                    2017     , Google Inc.,
+                    2021     , QBayLogic B.V.
   License    :  BSD2 (see the file LICENSE)
-  Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+  Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
   Type and instance definitions for Rewrite modules
 -}
@@ -31,10 +32,11 @@ import Control.Monad.State.Strict            (State)
 import Control.Monad.Writer                  (MonadWriter (..))
 import Data.Binary                           (Binary)
 import Data.Hashable                         (Hashable)
+import Data.HashMap.Strict                   (HashMap)
 import Data.IntMap.Strict                    (IntMap)
-import Data.Map.Strict                       (Map)
-import Data.Monoid                           (Any,Sum)
+import Data.Monoid                           (Any)
 import qualified Data.Set                    as Set
+import Data.Text                             (Text)
 import GHC.Generics
 
 #if EXPERIMENTAL_EVALUATOR
@@ -74,9 +76,11 @@ data RewriteStep
 -- | State of a rewriting session
 data RewriteState extra
   = RewriteState
+    -- TODO Given we now keep transformCounters, this should just be 'fold'
+    -- over that map, otherwise the two counts could fall out of sync.
   { _transformCounter :: {-# UNPACK #-} !Int
   -- ^ Total number of applied transformations
-  , _transformCounters :: Map String (Sum Word)
+  , _transformCounters :: HashMap Text Word
   -- ^ Map that tracks how many times each transformation is applied
   , _bindings         :: !BindingMap
   -- ^ Global binders
--- a/clash-lib/src/Clash/Rewrite/Util.hs
+++ b/clash-lib/src/Clash/Rewrite/Util.hs
@@ -1,9 +1,10 @@
 {-|
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016     , Myrtle Software Ltd,
-                    2017     , Google Inc.
+                    2017     , Google Inc.,
+                    2021     , QBayLogic B.V.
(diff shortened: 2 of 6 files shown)


clash-lib/src/Clash/Rewrite/Types.hs
```diff
@@ -74,7 +75,9 @@ data RewriteStep
 data RewriteState extra
   = RewriteState
   { _transformCounter :: {-# UNPACK #-} !Int
-  -- ^ Number of applied transformations
+  -- ^ Total number of applied transformations
+  , _transformCounters :: Map String (Sum Word)
+  -- ^ Map that tracks how many times each transformation is applied
```
martijnbastiaan, line 80: Although I don't expect this dictionary to get large, `Map` should generally be avoided in favor of `HashMap`:

https://github.com/haskell-perf/dictionaries

Its memory footprint is way better and it only loses sligthly for very small dicts.

martijnbastiaan: Similarly, `String` should not be used as a key for `Map`s as its comparison overhead is truly awful. Now I might say `String` is truly awful in general, but I wouldn't want to start a flamewar :relieved:

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -201,9 +202,17 @@ apply = \s rewrite ctx expr0 -> do
 
   if lvl == DebugNone
     then return expr2
-    else applyDebug lvl dbgTranss fromLimit s expr0 hasChanged expr2
+    else do
+      Monad.when (lvl >= DebugCount && hasChanged) $ updateTransformCounters s
```
martijnbastiaan, line 206: Could we move this inside `applyDebug`, just so we have it all nice and cozy in one function?

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -39,11 +39,12 @@ import           Data.Bool                   (bool)
 import           Data.Bifunctor              (bimap)
 import           Data.Coerce                 (coerce)
 import           Data.Functor.Const          (Const (..))
-import           Data.List                   (group, partition, sort)
+import           Data.List                   (group, partition, sort, sortOn)
 import qualified Data.List                   as List
 import qualified Data.List.Extra             as List
 import           Data.List.Extra             (partitionM)
 import qualified Data.Map                    as Map
```
martijnbastiaan, line 46: Lazy maps even.. Not a problem with your PR of course.

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -201,9 +202,17 @@ apply = \s rewrite ctx expr0 -> do
...
+    else do
+      Monad.when (lvl >= DebugCount && hasChanged) $ updateTransformCounters s
+      applyDebug lvl dbgTranss fromLimit s expr0 hasChanged expr2
 {-# INLINE apply #-}
 
+updateTransformCounters
+  :: String -- ^ Name of the transformation
+  -> RewriteMonad extra ()
+updateTransformCounters name =
+  transformCounters  %= (MapStrict.insertWith mappend name (Monoid.Sum 1))
```
martijnbastiaan, line 214: Any reason to not just use an `Word` and:

```haskell
transformCounters  %= MapStrict.insertWith (\_ -> succ) name 1
```

?

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -313,11 +322,14 @@ runRewriteSession :: RewriteEnv
...
-    ("Clash: Applied " ++ show (s' ^. transformCounter) ++ " transformations")
-    a
+  traceIf (_dbgLevel r >= DebugCount)
+    ("Clash: Transform counters:\n" ++ showCounters (s' ^. transformCounters)) $
+    traceIf (_dbgLevel r > DebugSilent)
+      ("Clash: Applied " ++ show (s' ^. transformCounter) ++ " transformations")
+      a
   where
     (a,s',_) = runR m r s
+    showCounters = unlines . map (\(nm,cnt) -> nm ++ ": " ++ show (Monoid.getSum cnt)) . sortOn snd . Map.toList
```
martijnbastiaan, line 332: It would be nice if we could print a MarkDown table so we can paste it in a GitHub comment. We'd need something like https://pypi.org/project/tabulate/. I wouldn't mind if this gets postponed to a later PR though.

docs/developing-hardware/flags.rst
```diff
@@ -25,7 +25,7 @@ Clash Compiler Flags
 
   - ``DebugFinal`` to show expressions after they have been completely
     normalized
-
+  - ``DebugCount`` to count how often each transformation is applied
```
martijnbastiaan, line 28: Could you document this in `docs/developing-hardware/flags.rst`?

martijnbastiaan (changes requested):
In general LGTM! but I've got some slight performance worries.

We could also store time spent in each transformation. Depending on what you're debugging that might be the more important metric.

clash-lib/src/Clash/Rewrite/Types.hs
alex-mckenna:
```haskell
'W' : 'h' : 'a' : 't' : '\'' : 's' : ' ' : 'w' : 'r' : 'o' : 'n' : 'g' : ' ' : 'w' : 'i' : 't' : 'h' : ' ' : 'S' : 't' : 'r' : 'i' : 'n' : 'g' : '?' : []
```

docs/developing-hardware/flags.rst
martijnbastiaan: Also, the "ordered debug levels" is really starting to work against us. We should rework it to a set of debug options.

leonschoorl: Didn't I just do that?

martijnbastiaan: Damnit I'm dumb, sorry.

clash-lib/src/Clash/Rewrite/Util.hs
leonschoorl: I started with lenses to go into the Map, using the [`<>=` combinator](https://ekmett.github.io/lens/Control-Lens-Setter.html#v:-60--62--61-), which needed the Monoid.
But later decided against it because the Map lenses are lazy and I think I want the Map to be strict.

clash-lib/src/Clash/Rewrite/Util.hs
gergoerdi: Maybe https://hackage.haskell.org/package/boxes ?

alex-mckenna: Leaving this for a future PR, it's just as easy to read a list of `transformation: count` then it is to read a table IMO. I can't imagine there's a nice way to hack in prettyprinting for just one type of debugging message too...

clash-lib/src/Clash/Rewrite/Util.hs
alex-mckenna: Leaving this as a `Sum Word` unless there are any strong objections. It shouldn't cost anything extra because it's a `newtype` anyway and GHC should be able to optimize things like `(<>) = coerce (+)` down to just addition anyway

martijnbastiaan: I think we should change it. Abstractions shouldn't be used if not necessary, it only adds cognitive load for the reader.

alex-mckenna: Sure, whatever. Changed it to `HashMap Text Word`

martijnbastiaan (approved):
LGTM.

Related issue: https://github.com/clash-lang/clash-compiler/issues/1800

PR #1649, opened 2021-02-05
Add GHC.Classes primitives for Int comparison

alex-mckenna:


martijnbastiaan (approved):

christiaanb:
Perhaps expand to also add rules for `leInt`, `gtInt`, `eqInt`  and `neInt` ?

PR #1651, opened 2021-02-07
Fix Stack build by adding versions of `hedgehog` and `tasty-hedgehog`

gergoerdi:
Fix Stack build by adding versions of `hedgehog` and `tasty-hedgehog` that work both with each other and with `clash-prelude`


gergoerdi:
> ```
> [ Warn  ] New GHCup version available: 0.1.12. To upgrade, run 'ghcup upgrade'
> ghcup: /root/.ghcup/ghc/8.8.4/bin: openFd: does not exist (No such file or directory)
> 
> 
> ```

martijnbastiaan:
I think the current Circle CI issues are caused by Clash using too much memory :persevere: . I guess we should limit the amount of memory Clash is allowed to use in the testsuite such that we catch these errors earlier.

PR #1655, opened 2021-02-09
Change work free checks

alex-mckenna:
When checking applications where the LHS is a variable, the function
checked if it was polymorphic / function type before checking if it
is a local id. This seems reasonable at first glance, but in the
partial evaluator this means that we can end up with core like

```haskell
\arg ->
  let f0 = \x -> x in
  let f1 = f0 in
  ...
  let fn = fm in
  fn arg
```

as the term bound to f0 is a function type, it is preserved. However,
other bindings which refer to this are not inlined as they are not
work free (due to having a function type). If we switch the order of
these checks, PE instead produces

```haskell
\arg ->
  let f0 = \x -> x in
  f0 arg
```

which significantly reduces the work needed after PE in the rewriting
pipeline. In the example program at https://github.com/clash-lang/clash-compiler/issues/1536#issuecomment-775918287 this reduces the
number of transformations from 17 to 4.

**NOTE**: This has been tested locally on GHC 9 and passes
clash-testsuite (GHC 9 is still not in CI for this PR).

### DIFF
--- a/clash-lib/src/Clash/Rewrite/WorkFree.hs
+++ b/clash-lib/src/Clash/Rewrite/WorkFree.hs
@@ -8,7 +8,6 @@ evaluation to check whether it is possible to perform changes without
 duplicating work in the result, e.g. inlining.
 -}
 
-{-# LANGUAGE MultiWayIf #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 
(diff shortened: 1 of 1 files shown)


christiaanb (commented):
This seems wrong:

```
x[localId] y[localId]
```

Seems like an expression that doesn't look to be guaranteed workfree to me.

alex-mckenna:
Hmm, yes I can see how that could end up duplicating work in some cases. Do you agree it's reasonable to move this check first when `collectArgs -> (fun, [])` though?

christiaanb:
Perhaps we could change `isWorkFree` so that it takes on `InScopeSet` of all the FVs of the expression `e` for which we determine work-freeness. And change the `Var` clause to:

```haskell
    if | isLocalId i -> pure (i `notElemInScopeSet` inScopeSet || not (isPolyFunTy (varType i)))
       | otherwise -> andM [isWorkFreeBinder cache bndrs i, allM isWorkFreeArg args]
```

christiaanb:
> 
> 
> Hmm, yes I can see how that could end up duplicating work in some cases. Do you agree it's reasonable to move this check first when `collectArgs -> (fun, [])` though?

No, some primitives are work-free, even if they have a function type.

alex-mckenna:
> No, some primitives are work-free, even if they have a function type.
This wouldn't apply though right? That would only matter if the LHS was a `Prim` instead of a `Var`?

I'm not sure I see the advantage to adding the `InScopeSet`, what is it for?

christiaanb:
> I'm not sure I see the advantage to adding the `InScopeSet`, what is it for?

If the variable is in the inscopeset (i.e. the FVs of the expression `e`), then we under-approximate work-freeness by checking whether it is a polyfunty. However, if it is not in the inscopeset (i.e. it is bound in `e`), then we check the work-freeness of the subexpression.

so for the expression:
```
(\(x :: Int -> Int) -> x) does_work 
```
it's okay to say that `x` is work_free, because the applied argument is classified as does_work, and hence the entire expression is classified as does_work. Thus the InScopeSet is mostly there to improve the accuracy of our workFreeNess assertion. Without it, we would more often under-approximate workFreeNess.

christiaanb (approved):

PR #1658, opened 2021-02-10
Raise upper bounds to allow building on ghc9

christiaanb:
Originally missed this because of allow-newer in cabal.project.local


martijnbastiaan:
Can we drop some of the `allow-newer` now too?

christiaanb:
The allow-newer also applies to our dependencies. If you e.g. drop the `*:template-haskell`, you might have to add `packageX:template-haskell`, `packageY:template-haskell`, etc.

Issue #1253, opened 2020-04-04
Term-level `hzToPeriod` has different behaviour than type-level `Div`


gergoerdi:
One would hope that the following would be a matching type-level version of `hzToPeriod`:

```
type HzToPeriod (freq :: Nat) = 1_000_000_000_000 `Div` freq
```

However, because `hzToPeriod` rounds up, and `Div` rounds down, instead we have to write

```
type HzToPeriod (freq :: Nat) = (1_000_000_000_000 + freq - 1) `Div` freq
```

Unless there is a good reason to round up, I think it would be worth changing the implementation of `hzToPeriod` (it would only change the result by 1 picosecond at most!) to be more amenable to a type-level implementation.

DigitalBrains1:
The new `hzToPeriod` should always give the same result as the simple

```
type HzToPeriod (freq :: Nat) = 1_000_000_000_000 `Div` freq
```

(the difference being that `hzToPeriod` also supports fractional frequencies)

PR #1660, opened 2021-02-10
hzToPeriod/periodToHz: use Ratio Natural

DigitalBrains1:
This will more often give the intuitively correct solution, per the
principle of least surprise. Actual functionality should not be affected
by the exact rounding, and thus this should not significantly change the
behaviour of existing designs.

Example:

Before this commit:

```
>>> hzToPeriod 100_000
10000001
```
After this commit:
```
>>> hzToPeriod 100_000
10000000
```

\<edit\>
We would like `hzToPeriod` to satisfy the following:

- Results should be intuitive
- Results should match a type-level equivalent as proposed in issue #1253

The former can be done with `Double`, by rounding to the nearest integer. This means that if there is an exact integer solution, we give that. Before this commit we had:

```
>>> hzToPeriod 100_000
10000001
```

instead of the intuitive result that a 100 kHz clock corresponds to a 10 µs period. The off-by-one has no significant consequences for the circuit, clocks are inherently not precise. It is just not intuitive.

But if we also want the result to match a simply expressed type-level equivalent, this is not enough, because the slight errors introduced by `Double` might cause a fractional part of 0.5 to be rounded either up or down seemingly randomly.

So the most straightforward fix is to use `Ratio Natural` and `floor`. If the result is an integral number of picoseconds, it will give that, and for roundings, it will match the behaviour of the straightforward type-level equivalent of:

```
type HzToPeriod (freq :: Nat) = 1_000_000_000_000 `Div` freq
```
\</edit\>

* ~~_Does this need a Changelog entry?_ Observable behaviour changes minutely, e.g., clock periods in generated waveforms might be a picosecond shorter or longer, two clocks might accumulate a slightly different phase relation than before. None of this affects functionality, so I'm inclined to say, no, a Changelog entry is too verbose. It's definitely subjective, though.~~
* Should we backport?

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Signal.hs
+++ b/clash-prelude/src/Clash/Explicit/Signal.hs
@@ -194,7 +194,6 @@ module Clash.Explicit.Signal
   , enableGen
     -- * Clock
   , Clock
-  , freqCalc  -- DEPRECATED
   , periodToHz
   , hzToPeriod
     -- ** Synchronization primitive
@@ -357,18 +356,6 @@ systemClockGen = clockGen
 -- @
 systemResetGen ::Reset System
 systemResetGen = resetGen
-
--- | Calculate the period, in __ps__, given a frequency in __Hz__
---
--- i.e. to calculate the clock period for a circuit to run at 240 MHz we get
---
--- >>> freqCalc 240e6
--- 4167
---
--- __NB__: This function is /not/ synthesizable
-freqCalc :: Double -> Integer
-freqCalc = toInteger . hzToPeriod
-{-# DEPRECATED freqCalc "Use 'hzToPeriod' instead." #-}
 
 -- ** Synchronization primitive
 -- | The 'unsafeSynchronizer' function is a primitive that must be used to
--- a/changelog/2021-02-10T17_29_11+01_00_hztoperiod_rounding
+++ b/changelog/2021-02-10T17_29_11+01_00_hztoperiod_rounding
@@ -1,0 +1,2 @@
+CHANGED: hzToPeriod now takes a `Ratio Natural` rather than a `Double`. It rounds slightly differently, leading to more intuitive results and satisfying the requested change in issue #1253. While the difference is observable in simulation, it does not affect generated circuits in a significant way.
+CHANGED: For symmetry, `periodToHz` now results in a `Ratio Natural`.
--- a/changelog/2021-02-11T14_20_26+01_00_drop_freqCalc
+++ b/changelog/2021-02-11T14_20_26+01_00_drop_freqCalc
@@ -1,0 +1,1 @@
+REMOVED: The deprecated function `freqCalc` has been removed.
--- a/clash-prelude/src/Clash/Signal/Internal.hs
+++ b/clash-prelude/src/Clash/Signal/Internal.hs
@@ -156,6 +156,7 @@ import Data.Data                  (Data)
 import Data.Default.Class         (Default (..))
 import Data.Hashable              (Hashable)
 import Data.Proxy                 (Proxy(..))
+import Data.Ratio                 (Ratio)
 import Data.Type.Equality         ((:~:))
 import GHC.Generics               (Generic)
 import GHC.Stack                  (HasCallStack)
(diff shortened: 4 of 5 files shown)


christiaanb:
We probably want to use a custom rounding function that ties half to infinity instead of even, as it's easier to implement a type-level equivalent for that.

martijnbastiaan:
I kinda missed it in the (large!) discussions, but why don't we simply switch to `Rational` instead of `Double`? It would mitigate _any_ rounding error anyway, right?

christiaanb:
No, the error wasn't due to the number format, but due to use of `ceiling`

DigitalBrains1:
I have an idea but no time to do so right away. I'll be back.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -367,7 +367,7 @@ systemResetGen = resetGen
 --
 -- __NB__: This function is /not/ synthesizable
 freqCalc :: Double -> Integer
-freqCalc = toInteger . hzToPeriod
+freqCalc = toInteger . hzToPeriod . realToFrac
```
christiaanb, line 370: You can just remove this function since we cannot backport this PR to the 1.2 branch.

DigitalBrains1: I'll do it tomorrow!

DigitalBrains1: Done

martijnbastiaan:
@DigitalBrains1 Should `periodToHz` output a `Ratio Natural`? Does this mean that this comment is obsolete:

```haskell
-- __NB__: This function is lossy. I.e., hzToPeriod . periodToHz /= id.
```

?

DigitalBrains1:
Even if `periodToHz` output a `Ratio Natural`, it would still be lossy, because we round to whole picoseconds in the period.

I don't have a strong conviction, but to me, `Ratio` suggests exact numbers. And clock frequencies are inherently inexact. Even America's prime time and frequency standard, the atomic [NIST-F1](https://en.wikipedia.org/wiki/NIST-F1) clock only has an accuracy that is pretty much the same as the accuracy of a `Double` variable (cute coincidence!).

The only reason for `Ratio` here, for me, is to get equivalency with the type-level construction.

martijnbastiaan:
Alright. Wouldn't `periodToHz` have a similar type level equivalence though? (Also, I'm a sucker for symmetry :-))

DigitalBrains1:
Yes, I just realised that that bit of documentation is now smelly anyway:
```
-- __NB__: This function is lossy. I.e., hzToPeriod . periodToHz /= id.
```
```
> :t hzToPeriod . periodToHz

<interactive>:1:14: error:
    • Couldn't match type ‘Double’
                     with ‘GHC.Real.Ratio GHC.Natural.Natural’
      Expected type: GHC.Natural.Natural
                     -> GHC.Real.Ratio GHC.Natural.Natural
        Actual type: GHC.Natural.Natural -> Double
    • In the second argument of ‘(.)’, namely ‘periodToHz’
      In the expression: hzToPeriod . periodToHz
```
So, yes, that is indeed not `id` by a long shot, but not because it's lossy :-)

Yeah, you could make a type-level `PeriodToHz` but I don't know if anybody is going to use it. I mean, you use `hzToPeriod` and `HzToPeriod` to tell Clash the frequency the circuit should run at. The reverse, have Clash tell you the frequency, isn't that useful, I think.

Anyway, I'm happy to change the other one to `Ratio` as well. I like symmetry too.

martijnbastiaan (approved):

DigitalBrains1:
I just realised the doc wasn't without issues.

Because now, unless I'm very mistaken, `hzToPeriod . periodToHz` is equal to the identity, but the juxtaposition is not. So I corrected the doc to say that `periodToHz . hzToPeriod` is the lossy one.

I feel like we're taking baby steps in this PR. But we'll get there :-).

DigitalBrains1:
I fixed the doctest failure. Note how it was the representation, not the result that changed here:
```diff
@@ -1433,7 +1434,7 @@ hzToPeriod freq = floor ((1.0 / freq) / 1.0e-12)
 -- i.e. to calculate the clock frequency of a clock with a period of 5000 ps:
 --
 -- >>> periodToHz 5000
--- 2.0e8
+-- 200000000 % 1
 --
 -- __NB__: This function is /not/ synthesizable
 periodToHz :: Natural -> Ratio Natural
```

I also introduced a line break (actualy, a paragraph break) between the two NBs in the doc.

PR #1664, opened 2021-02-11
Cleanup superfluous let-bindings in 'removeUnsedExpr'

### LINKED ISSUE #1663: Unexpected data constructor in subject of `case` in netlist stage
While investigating https://github.com/clash-lang/clash-compiler/issues/1536 I found an issue where Clash errors on valid code. The error message is:

```
    Clash.Netlist(598): Not in normal form: Not a variable reference or primitive as subject of a case-statement:
    
    DataCon (SP "GHC.Maybe.Maybe" [("GHC.Maybe.Nothing",[]),("GHC.Maybe.Just",[SP "GHC.Maybe.Maybe" [("GHC.Maybe.Nothing",[]),("GHC.Maybe.Just",[Index 32768])]])]) (DC (SP "GHC.Maybe.Maybe" [("GHC.Maybe.Nothing",[]),("GHC.Maybe.Just",[SP "GHC.Maybe.Maybe" [("GHC.Maybe.Nothing",[]),("GHC.Maybe.Just",[Index 32768])]])],1)) [DataCon (SP "GHC.Maybe.Maybe" [("GHC.Maybe.Nothing",[]),("GHC.Maybe.Just",[Index 32768])]) (DC (SP "GHC.Maybe.Maybe" [("GHC.Maybe.Nothing",[]),("GHC.Maybe.Just",[Index 32768])],0)) []]
```

The core (obtained using `-fclash-debug DebugFinal`) confirms this is the case:

```
topEntity after normalization:

[..]
  = case <prefixName>"topEntity_ds1"
      Just @(Maybe (Index 32768)) (Nothing @(Index 32768)) of
      Nothing  ->
        <prefixName>"topEntity_ds1"
        Just @(Maybe (Index 32768)) (Nothing @(Index 32768))
      Just (ipv2 :: Maybe (Index 32768)) ->
        <prefixName>"topEntity_ds1"
        Just @(Maybe (Index 32768)) (Nothing @(Index 32768))
[..]
```

i.e.

```
topEntity after normalization:

[..]
  = case Just Nothing of
      Nothing  ->
        Just Nothing
      Just ipv2 ->
        Just Nothing
[..]
```

The netlist stage can only handle `case x of` where `x` is a variable reference and therefore crashes. Usually, this would have been cleaned up by `caseCon`, as the subject of the case is known. This constructs only appears _after_ executing `rmDeadCode` however, after which `caseCon` is never run again. What happens is:

```
let
  a = let foo = someComplexThing in Nothing
in
  case a of
    Nothing -> ...
    Just _ -> ...
```

If the (superfluous) `let` is not cleaned up, `bindConstantVar` will not inline it. In turn, `caseCon` will not fire.

Reproducer:

 * https://gist.github.com/martijnbastiaan/71b561250fec4e7e2b486cd0bdfea461

Solution:

 * Cleanup unused let-bindings in `removeUnusedExpr`

martijnbastiaan:
Fixes #1663

### DIFF
--- a/clash-lib/src/Clash/Normalize/Strategy.hs
+++ b/clash-lib/src/Clash/Normalize/Strategy.hs
@@ -116,8 +116,22 @@ constantPropagation =
     -- binder might be recursive. The idea is, is that if the recursive
     -- non-representable binder is inlined once, we can get rid of the recursive
     -- aspect using the case-of-known-constructor
+    --
+    -- Note that we first do a dead code removal pass, which makes sure that
+    -- unused let-bindings get cleaned up. Only if no dead code is removed
+    -- 'inlineNonRep' is executed. We do this for two reasons:
+    --
+    --   1. 'deadCode' is an expensive operation and is therefore left out of
+    --      the hot loop 'transPropagateAndInline'.
+    --
+    --   2. In various situations 'transPropagateAndInline' can do more work
+    --      after 'deadCode' was successful. This work in turn might remove a
+    --      a construct 'inlineNonRep' would fire on - saving the compiler work.
+    --
     inlineNR :: NormRewrite
-    inlineNR = bottomupR (apply "inlineNonRep" inlineNonRep)
+    inlineNR =
+          bottomupR (apply "deadCode" deadCode)
+      >-! bottomupR (apply "inlineNonRep" inlineNonRep)
 
     specTransformations :: [(String,NormRewrite)]
     specTransformations =
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,6 +80,7 @@ aliases:
     run:
       name: Build dependencies and Clash itself with Nix
       command: .ci/nix_build.sh
+      no_output_timeout: 1h
 
   - &stack_build
     run:
@@ -99,6 +100,7 @@ aliases:
           cabal v2-run testsuite -- --hide-successes -p .VHDL
           cabal v2-run testsuite -- --hide-successes -p .Verilog
         fi
+      no_output_timeout: 1h
 
   # When testing a pull request on CircleCI,
   # fetch the result of merging it in.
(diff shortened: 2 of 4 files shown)


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -953,6 +953,11 @@ removeUnusedExpr _ e@(collectArgsTicks -> (Data dc, [_,Right aTy,Right nTy,_,Lef
            in  changed v
       _ -> return e
 
+removeUnusedExpr _ e@(Letrec binds body) =
+  case removeUnusedBinders binds body of
```
christiaanb, line 957: The issue with adding it to `removedUnusedExpr` is that the transformation is currently used in a topdown traversal, while you want to use `removeUnusedBinders` in a bottomup traversal after all the constants have been propagated and e.g. bodies of let-expressions stop refencing their binders. i.e. I would change: https://github.com/clash-lang/clash-compiler/blob/250aea43ce6111cd55cdb83e1ad30761024c9bb0/clash-lib/src/Clash/Normalize/Strategy.hs#L120

to

```haskell
inlineNR = bottomupR (apply "deadcode"  deadCode >-> apply "inlineNonRep" inlineNonRep)
```

martijnbastiaan: Done! Albeit with a small twist :-)

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -117,7 +117,9 @@ constantPropagation =
     -- non-representable binder is inlined once, we can get rid of the recursive
     -- aspect using the case-of-known-constructor
     inlineNR :: NormRewrite
-    inlineNR = bottomupR (apply "inlineNonRep" inlineNonRep)
+    inlineNR =
+          bottomupR (apply "deadCode" deadCode)
+      >-! bottomupR (apply "inlineNonRep" inlineNonRep)
```
leonschoorl, line 122: Wouldn't it be more efficient to combine the two transformations in one bottom up traversal?

martijnbastiaan: We had some discussion on this on Slack, I'll update the commit message / PR message to explain the reasoning!

martijnbastiaan: I've updated the comment right above the transformation!

PR #1668, opened 2021-02-15
Do not traverse alternatives of nonRep cases

martijnbastiaan:
More specifically: if 'inlineNonRep' successfully fires on a case's
subject, it does not traverse alternatives. The rationale is as follows:
if an inline-non-rep successfully fires in or on the subject of a case,
`caseCon` might eliminate it in the following round of
`inlineAndPropagate` or subsequent ones. If we inline in
alternatives too we risk doing too much work - as they might get
eliminated later on.

This is part of finding the root cause of #1536.

### DIFF
--- a/clash-lib/src/Clash/Normalize/Strategy.hs
+++ b/clash-lib/src/Clash/Normalize/Strategy.hs
@@ -131,7 +131,7 @@ constantPropagation =
     inlineNR :: NormRewrite
     inlineNR =
           bottomupR (apply "deadCode" deadCode)
-      >-! bottomupR (apply "inlineNonRep" inlineNonRep)
+      >-! apply "inlineNonRep" inlineNonRep
 
     specTransformations :: [(String,NormRewrite)]
     specTransformations =
--- a/clash-lib/src/Clash/Normalize/Transformations.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations.hs
@@ -424,9 +424,46 @@ caseCase _ e = return e
 {-# SCC caseCase #-}
 
 -- | Inline function with a non-representable result if it's the subject
--- of a Case-decomposition
+-- of a Case-decomposition. It's a custom topdown traversal that -for efficiency
+-- reasons- does not explore alternative of cases whose subject triggered an
+-- 'inlineNonRepWorker'.
 inlineNonRep :: HasCallStack => NormRewrite
-inlineNonRep _ e@(Case scrut altsTy alts)
+inlineNonRep ctx0 e0@(Case {}) = do
+  r <- listen (inlineNonRepWorker e0)
+  case r of
+    (e1, Monoid.getAny -> True) ->
+      return e1
+    (~(Case subj0 typ alts), _) -> do
+      -- If a term _in_ the subject triggers 'inlineNonRepWorker', inline and
+      -- propagate might eliminate this case. We therefore don't explore the
+      -- alternatives. Note that this makes it substantially different from a
+      -- 'topdownSucR' transformation.
+      let
+        TransformContext inScope ctx1 = ctx0
+        ctx2 = TransformContext inScope (CaseScrut:ctx1)
+
+      listen (inlineNonRep ctx2 subj0) >>= \case
+        (subj1, Monoid.getAny -> True) ->
+          return (Case subj1 typ alts)
+        (subj1, _) -> do
+          let (pats, rhss0) = unzip alts
+          rhss1 <- mapM (inlineNonRep ctx2) rhss0
+          pure (Case subj1 typ (zip pats rhss1))
+
+inlineNonRep ctx e =
+  -- All non-case statements are simply traversed. TODO: are there other special
+  -- cases like 'Case' that would warrant an optimization like ^ ?
+  allR inlineNonRep ctx e
+{-# SCC inlineNonRep #-}
+
+-- | Inline function with a non-representable result if it's the subject
+-- of a Case-decomposition. This worker function only tries the given term
+-- (i.e., it does not traverse it).
+--
+-- It sets the changed flag in the NormalizeSession if it successfully inlines
+-- a binder.
+inlineNonRepWorker :: HasCallStack => Term -> NormalizeSession Term
+inlineNonRepWorker e@(Case scrut altsTy alts)
   | (Var f, args,ticks) <- collectArgsTicks scrut
   , isGlobalId f
   = do
@@ -448,7 +485,7 @@ inlineNonRep _ e@(Case scrut altsTy alts)
                       , "\nRun with '-fclash-inline-limit=N' to increase"
                       , " the inlining limit to N."
                       ])
-              (return e)
+              (pure e)
       else do
         bodyMaybe   <- lookupVarEnv f <$> Lens.use bindings
         nonRepScrut <- not <$> (representableType <$> Lens.view typeTranslator
@@ -463,14 +500,14 @@ inlineNonRep _ e@(Case scrut altsTy alts)
             let scrutBody0 = mkTicks (bindingTerm b) (mkInlineTick f : ticks)
             let scrutBody1 = mkApps scrutBody0 args
 
-            changed $ Case scrutBody1 altsTy alts
+            changed (Case scrutBody1 altsTy alts)
 
-          _ -> return e
+          _ -> pure e
   where
     exception = isClassTy
 
-inlineNonRep _ e = return e
-{-# SCC inlineNonRep #-}
+inlineNonRepWorker e = pure e
+{-# SCC inlineNonRepWorker #-}
 
 
 caseCon :: HasCallStack => NormRewrite


clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,63 @@ caseCase _ e = return e
...
 -- | Inline function with a non-representable result if it's the subject
--- of a Case-decomposition
+-- of a Case-decomposition. It's a custom topdown traversal that -for efficiency
+-- reasons- does not explore alternative of cases whose subject triggered an
+-- 'inlineNonRepWorker'.
 inlineNonRep :: HasCallStack => NormRewrite
-inlineNonRep _ e@(Case scrut altsTy alts)
+inlineNonRep _ctx = fmap snd . go
+ where
+  go :: HasCallStack => Term -> NormalizeSession (Bool, Term)
```
alex-mckenna, line 433: I don't like that you return a `Bool` here which is just thrown away. `go` should just return the final `Term` IMO

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,63 @@ caseCase _ e = return e
...
+  go e0 = inlineNonRepWorker e0 >>= \case
+    (True, e1) ->
+      return (True, e1)
+    (False, e1) ->
```
alex-mckenna, line 437: Another thing: you return a `Bool` which indicates whether there was a change, but the `Writer` part of this monad tracks if there was a change (as far as I can tell anyway, it's rather unhelpfully marked as `Any` in the types module). However I have seen things like

```haskell
(e1, Monoid.getAny -> hasChanged) <- Writer.listen $ rw ctx e
```

in other parts of the codebase, which make me think you could instead write

```haskell
Writer.listen (inlineNonRepWorker e0) >>= \case
  (e1, Monoid.getAny -> True)  -> pure e1
  (e1, Monoid.getAny -> False) -> case e1 of ...
```

and use the fact that `return` and `changed` will determine what happens

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,63 @@ caseCase _ e = return e
...
+        Data _ -> return (False, e1)
+        Literal _ -> return (False, e1)
+        Prim _ -> return (False, e1)
+        Lam v t -> second (Lam v) <$> go t
+        TyLam v t -> second (TyLam v) <$> go t
+        App a0 b0 -> do
+          (change0, a1) <- go a0
+          (change1, b1) <- go b0
+          return (change0 || change1, App a1 b1)
+        TyApp t typ -> second (`TyApp` typ) <$> go t
```
alex-mckenna, line 449: I think applicative style might read nicer here, e.g.

```haskell
TyApp <$> go t <*> pure typ
```

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,63 @@ caseCase _ e = return e
...
+        Cast t0 typ0 typ1 -> do
+          (change, t1) <- go t0
+          return (change, Cast t1 typ0 typ1)
+        Tick tickInfo t -> second (Tick tickInfo) <$> go t
+        Letrec bindings_ body0 -> do
+          let (bndrs, bodies0) = unzip bindings_
+          (changes, bodies1) <- unzip <$> mapM go bodies0
+          (change, body1) <- go body0
```
alex-mckenna, line 457: Now I'm definitely more in favour of ditching `Bool` and using `Writer` to find changes. Think how much nicer it would make these cases to not have to collect / propagate whether things have changed

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,63 @@ caseCase _ e = return e
...
+          pure (or (change:changes), Letrec (zip bndrs bodies1) body1)
```
alex-mckenna, line 458: We should really specify in the style guide that one of `pure` or `return` is specified. It's especially jarring to use both in the same function though

alex-mckenna (changes requested):
Overall I like it, but I would prefer you didn't manually pass around information about whether something has changed (assuming I'm correct about it not being necessary to do so)

clash-lib/src/Clash/Normalize/Transformations.hs
alex-mckenna: Although actually you only need the `second` because you carry around the `Bool`. So it should just become an `fmap`

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: I've considered it but decided against it because you need to reset the flag if you're handling things like `App e1 e2` or `Letrec bindings_ body`. I.e., if `e1` triggers a rewrite that does _not_ mean `e2` shouldn't try to inline non rep constructs.

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: `pure` is probably preferred.

clash-lib/src/Clash/Normalize/Transformations.hs
alex-mckenna: In `inlineNonRepWorker` or `inlineNonRep.go`? If you mean `inlineNonRep.go` then I'm not sure I see what you mean, you appear to rewrite different parts of `App` and `Letrec` then just or the results at the end

clash-lib/src/Clash/Normalize/Transformations.hs
alex-mckenna: Oh definitely, it has looser constraint bounds. The point is more that I'd like to see just one, but don't mind that one being `return` in functions / modules which already heavily favour it (like transformations)

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: Yeah alright, I misunderstood what `listen` was doing. I like the advice.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,59 @@ caseCase _ e = return e
...
+          (subj1, subjHasChanged) <- listen (go subj0)
+          if Monoid.getAny subjHasChanged then
+            -- This is the crucial part of this transformation, making it
+            -- substantially different from just a bottom-up traversal with
+            -- 'inlineNonRepWorker': if any non representable binder is inlined
+            -- in the the subject of a case expression, we don't traverse the
+            -- alternatives as this inlined structure might very well give use a
+            -- chance to do 'caseCon'. Traversing the alternatives might
+            -- therefore yield a lot of work.
+            pure (Case subj1 typ alts)
```
alex-mckenna, line 465: I think this needs to be `changed`, I'm not sure the change propagates when you capture it locally with `listen`

martijnbastiaan: It does:

```haskell
listens :: (Monoid w, Monad m) => (w -> b) -> RWST r w s m a -> RWST r w s m (a, b)
listens f m = RWST $ \ r s w -> do
    (a, s', w') <- runRWST m r s
    let wt = w `mappend` w'
    wt `seq` return ((a, f w'), s', wt)
{-# INLINE listens #-}
```

The `wt` is getting propagated.

martijnbastiaan: Oh and `listen = listens id`

alex-mckenna: :+1:

alex-mckenna (approved):
Assuming CI passes, LGTM

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -131,7 +131,7 @@ constantPropagation =
     inlineNR :: NormRewrite
     inlineNR =
           bottomupR (apply "deadCode" deadCode)
-      >-! bottomupR (apply "inlineNonRep" inlineNonRep)
```
christiaanb, line 134: Why was simply
```haskell
>-! topdownSucR (apply "inlineNonRep" inlineNonRep)
```
not an option?

christiaanb: Ah, I see. Because you want to stop the traversal in the alternatives.

martijnbastiaan: I don't know, it depends. My understanding is that `topdownSucR` stops at the first success, meaning that it will either replace zero or one non-representable case subjects and never more than one. This can get the transformation stuck in an infinite loop, for example in:

```haskell
let 
  x = case nonRep of ...
  y = case nonRep of ...
in
  case y of
    ...
```

..it will keep expanding `x` (if it's recursive) even though it might need `y` to make progress instead.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,59 @@ caseCase _ e = return e
 {-# SCC caseCase #-}
 
 -- | Inline function with a non-representable result if it's the subject
--- of a Case-decomposition
+-- of a Case-decomposition. It's a custom topdown traversal that -for efficiency
+-- reasons- does not explore alternative of cases whose subject triggered an
+-- 'inlineNonRepWorker'.
 inlineNonRep :: HasCallStack => NormRewrite
-inlineNonRep _ e@(Case scrut altsTy alts)
+inlineNonRep _ctx = go
```
christiaanb, line 431: I think this could be simplified to:

```haskell
inlineNonRepTraverse ctx e = do
  (e1, hasChanged) <- listen (inlineNonRep ctx e)
  case e1 of
    Case {} -> return e1
    _ -> allR inlineNonRepTraverse ctx e1
```

where `inlineNonRep` is what you call `inlineNonRepWorker`. Or am I missing something?

clash-lib/src/Clash/Normalize/Strategy.hs
martijnbastiaan: Just to clarify, if it doesn't that'd also be subtly different from what the current transformation does. The current transformation doesn't traverse if _anything_ changes in a subject of a case, because it might be enough to make progress in `inlineAndPropagate`.

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -424,9 +424,59 @@ caseCase _ e = return e
...
+        Case subj0 typ alts -> do
+          (subj1, subjHasChanged) <- listen (go subj0)
+          if Monoid.getAny subjHasChanged then
+            -- This is the crucial part of this transformation, making it
+            -- substantially different from just a bottom-up traversal with
+            -- 'inlineNonRepWorker': if any non representable binder is inlined
+            -- in the the subject of a case expression, we don't traverse the
+            -- alternatives as this inlined structure might very well give use a
+            -- chance to do 'caseCon'. Traversing the alternatives might
+            -- therefore yield a lot of work.
```
christiaanb, line 464: The comment and the code do not match. The codes says: when the subject of an outer case-statement contains inner case-statements which `inlineNonRep` transforms, then we do not traverse the alternatives of the outer case-statements. But that is not what the comment says.

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: I don't think it's equivalent, did you mean:

```haskell
inlineNonRepTraverse ctx e = do
  r <- listen (inlineNonRep ctx e)
  case r of
    (e1, Monoid.getAny -> True) -> return e1
    (e1, _) -> allR inlineNonRepTraverse ctx e1
```

? Or more directly: did you mean to ignore `hasChanged`? Why would you match on `Case`?

christiaanb: Yes, sorry, I meant:
```haskell
inlineNonRepTraverse ctx e = do
  (e1, hasChanged) <- listen (inlineNonRep ctx e)
  case Monoid.getAny hasChanged of
    True -> return e1
    _ -> allR inlineNonRepTraverse ctx e1
```

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: Yeah I'll split this comment out over

```
(e1, hasChanged) <- listen (inlineNonRepWorker e0)
```

and this position to make it clearer.

clash-lib/src/Clash/Normalize/Transformations.hs
martijnbastiaan: Yes, that's sensible. As discussed in a video call, this has the slight downside of keeping track of the context - even though it isn't used by the worker function.

martijnbastiaan: We would still need to special case the `Case`. I've changed the transformation to use `allR`, but retain the original behavior.

christiaanb (approved):

PR #1671, opened 2021-02-18
Add CLASH_PPR_TICKS flag

alex-mckenna:
When dealing with output where a lot of inlining has occured (like PE)
ticks introduce a lot of noise. This flag means pretty printed core
need not have pesky ticks everywhere.


martijnbastiaan (approved):
Nice. I've been wanting this for a while but have been too lazy to implement it!

PR #1676, opened 2021-02-23
Various small fixes

leonschoorl:


clash-prelude/src/Clash/Prelude.hs
```diff
@@ -130,8 +130,7 @@ module Clash.Prelude
   , Lift (..)
     -- ** Type classes
     -- *** Clash
-  -- , module Clash.Class.AutoReg
-  , autoReg, deriveAutoReg
+  , AutoReg, autoReg, deriveAutoReg
```
martijnbastiaan, line 133: Did you use tooling to discover this? It would be great if we could check export lists on CI - the issue seems to be coming up every few months.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -420,7 +420,7 @@ substInExistentials is exts subst@(typeVar, _type) =
       -- it is as we would replace global variables in them.
       take (i+1) exts ++ substGlobalsInExistentials is (drop (i+1) exts) [subst]
 
--- | This checks if the substitution satisfies the invariant from 'TvSbust's
+-- | This checks if the substitution satisfies the invariant from 'TvSubst's
```
martijnbastiaan, line 423: One step closer to enabling documentation checks on clash-lib :smile:

clash-prelude/src/Clash/Prelude.hs
leonschoorl: No, I ran into this when trying to use the implicit variant of `autoReg` in a polymorphic function and trying to write down its type.

If we could get the full export list programmatically, what would you want to check it with?

martijnbastiaan: Ah I figured you made this commit to get `Clash.Prelude` in line with `Clash.Explicit.Prelude`. What I'd like to do (at some point) is to keep a file encoding the differences between the Explicit Prelude and the Implicit Prelude. Then, if you'd add something to one Prelude but forget to add it to the other, CI could complain.

martijnbastiaan: But in that case; did you check whether the explicit prelude exports this too?

leonschoorl: Good idea

And yes, the explicit prelude [exports](https://github.com/clash-lang/clash-compiler/blob/e16e9012a19b0296bf0ace7c23ce44c200060e03/clash-prelude/src/Clash/Explicit/Prelude.hs#L110) the the full AutoReg module

leonschoorl:
@martijnbastiaan I've added the "Only print inline warning when trying to inline"

clash-lib/src/Clash/Normalize/Transformations.hs
```diff
@@ -471,40 +471,45 @@ inlineNonRepWorker e@(Case scrut altsTy alts)
...
+                                              <*> pure scrutTy)
+    case (nonRepScrut, bodyMaybe) of
+      (True, Just b) -> do
+        if overLimit then
+          trace ($(curLoc) ++ [I.i|
+            InlineNonRep: #{showPpr (varName f)} already inlined
+            #{limit} times in: #{showPpr (varName cf)}. The type of the subject
+            is:
+
+              #{showPpr scrutTy}
```
martijnbastiaan, line 497: @alex-mckenna Added a comment on this commit saying we should probably use `showPpr'` to prevent it not printing anything when using `CLASH_PPR_*`.

martijnbastiaan (approved):
LGTM if the CI passes!

`showPpr` => `showPpr'` in the warning patch would be nice too :)

clash-lib/src/Clash/Normalize/Transformations.hs
alex-mckenna: I think a relatively harmless internal change might also be to make the `CLASH_PPR_*` flags only apply in certain cases, i.e. making `CLASH_PPR_TYPES` only come into effect when printing the type of a binder, or a type application. I think every other instance of a type _should_ be printed regardless of this flag - i.e. if I ask to print a type I should _never_ see an empty string

leonschoorl: That's a good idea, but for now I just used `showPpr'`

Issue #1677, opened 2021-02-23
Combinational Loops arise when updating from Clash .99


jamieChipscan:
Hi all,

I'm a bit new to Clash, but am working on bringing some legacy code from .99 up to the most recent version. The original code has lines like: 

`nsec2   = mux (count .==. 0) nsec nsec2
`
That previously compiled to 

` always @(*) begin
    if(b_1)
      result_4 = t_1;
    else
      result_4 = result_4;
  end`

However, the same lines now introduce a non-stable combinational loop when synthesized.

`assign result_2 = b ? s_0 : result_2;
`
Is there some general solution to these errors, like explicitly adding an intermediary register? I didn't notice note of this in the migration guide or blog posts.

Thanks!

christiaanb:
Interesting.

When I ask Intel Quartus to create an RTL diagram for verilog generated from:
```haskell
module Test where

import Clash.Prelude

nsec2 :: SystemClockResetEnable => Signal System (Unsigned 4)
nsec2 = mux (count .==. 0) nsec nsec2
 where
   nsec  = register @System (0 :: Unsigned 4) (nsec + 1)
   count = register (0 :: Unsigned 8) (count + 1)
{-# ANN nsec2 (defSyn "nsec2") #-}
```

I get:
![image](https://user-images.githubusercontent.com/93169/108863929-ff8e0e00-75f1-11eb-8069-3ddca346da86.png)

With the expected level-sensitive latches generated for the following verilog:
```verilog
wire [3:0] c$result_rec;
assign c$result_rec = b ? nsec : c$result_rec;
```
And, when I change that to:
```verilog
reg [3:0] c$result_rec;
always @(*) begin : result_latch
    if (b) begin
	   c$result_rec = nsec;
	 end else begin
	   c$result_rec = c$result_rec;
	 end
  end
```
I get the same level-sensitive latch:
![image](https://user-images.githubusercontent.com/93169/108864968-fcdfe880-75f2-11eb-89e7-388d4795aab4.png)

However... regardless of what style Clash generated, it doesn't correspond to what we can simulate in Clash; that is when I try:
```
clashi> sampleN_lazy 8 nsec2 
[0,0,^CInterrupted.
```
simulation gets stuck trying to output the 3rd sample, which is when `count` would be not `1` (after the `0` on "power-up", followed by the `0` out of reset); and I have to abort with ctrl+c.

[...]

PR #1678, opened 2021-02-23
Add flag to force sequential block for ifThenElse

christiaanb:
Work-around for #1677

### DIFF
--- a/Clash.hs
+++ b/Clash.hs
@@ -33,17 +33,17 @@ import Util (OverridingBool(..))
 genSystemVerilog
   :: String
   -> IO ()
-genSystemVerilog = doHDL (initBackend WORD_SIZE_IN_BITS HDLSYN True PreserveCase Nothing (AggressiveXOptBB False) :: SystemVerilogState)
+genSystemVerilog = doHDL (initBackend WORD_SIZE_IN_BITS HDLSYN True PreserveCase Nothing (AggressiveXOptBB False) (TernaryOpt True) :: SystemVerilogState)
 
 genVHDL
   :: String
   -> IO ()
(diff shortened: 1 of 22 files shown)


alex-mckenna:
`TernaryOpt` is quite specific, what about a more generic name like `PreferSequential` so the flag can be reused if there are any similar issues in the future where comb. / seq. versions have different semantics? Although documenting such a flag would be more difficult - _it's sequential if it might need to be_...

Clash.hs
```diff
@@ -33,17 +33,17 @@ import Util (OverridingBool(..))
 genSystemVerilog
   :: String
   -> IO ()
-genSystemVerilog = doHDL (initBackend WORD_SIZE_IN_BITS HDLSYN True PreserveCase Nothing (AggressiveXOptBB False) :: SystemVerilogState)
+genSystemVerilog = doHDL (initBackend WORD_SIZE_IN_BITS HDLSYN True PreserveCase Nothing (AggressiveXOptBB False) (TernaryOpt True) :: SystemVerilogState)
```
martijnbastiaan, line 36: I think this has reached a breaking point. Could we refactor this into a record?

martijnbastiaan (changes requested):

Clash.hs
alex-mckenna: Did someone say _refactor this_ :sparkles:

martijnbastiaan: :tada:

martijnbastiaan:
Given the conflicts and activity level I'm going to close this PR. Feel free to reopen if you pick up work again.

PR #1680, opened 2021-02-25
Create domain alias/helper only when they don't exist

### LINKED ISSUE #1674: `createDomain` creates definitions without documentation
This is an issue when trying to achieve 100% documentation coverage. E.g.:

```
  Missing documentation for:
    Core (src/Config.hs:11)
    vCore (src/Config.hs:11)
```

basile-henry:
fixes #1674

This makes it possible to explicitly define the type alias (or the
helper). This can be useful when writing documentation for both.

For example:

```haskell
-- | MySystem is a custom domain ...
type MySystem = ("MySystem" :: Domain)

-- | vMySystem is the domain configuration associated with `MySystem`
createDomain vSystem{vName="MySystem"}
```

### DIFF
--- a/changelog/2021-02-25T13_28_53+00_00_documentable_create_domain
+++ b/changelog/2021-02-25T13_28_53+00_00_documentable_create_domain
@@ -1,0 +1,1 @@
+FIXED: `createDomain` creates definitions without documentation [#1674] https://github.com/clash-lang/clash-compiler/issues/1674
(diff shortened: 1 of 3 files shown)


martijnbastiaan (approved):
I didn't realize you could lookup names in TH, makes perfect sense! 

Could you add a changelog (instructions [over here](https://github.com/clash-lang/clash-compiler/blob/master/changelog/README.md))?

martijnbastiaan:
Thanks @basile-henry :D

PR #1694, opened 2021-03-09
Use JSON for manifest files

martijnbastiaan:
Makese it easier to interact with external tools. It is not meant to be
a stable API yet.

----------------------------

Example:

```json
{
    "components": [
        "topEntity"
    ],
    "flags": [
        20,
        20,
        false
    ],
    "hash": 4223898113426202399,
    "version": "unstable",
    "files": [
        {
            "name": "topEntity.sdc",
            "sha256": "13d4c1aafcc09a84d81d3b60d513cb901384a10e27e5a5a16572a600f15895f3"
        },
        {
            "name": "Calculator_topEntity_types.vhdl",
            "sha256": "53299fb53c4bc5ac1d4dace45e769aa45ab5ea7ec15a9df5a299f0e1ef14f778"
        },
        {
            "name": "topEntity.vhdl",
            "sha256": "aa718fb2a4234ed45d551607f57f932c4a89068c879be463ad119b04a7053476"
        }
    ],
    "top_component": {
        "ports_flat": {
            "in": [
                {
                    "domain": "System",
                    "width": 1,
                    "name": "clk",
                    "is_clock": true,
                    "type_name": "clk.Calculator_topEntity_types.clk_System"
                },
                {
                    "domain": "System",
                    "width": 1,
                    "name": "rst",
                    "is_clock": false,
                    "type_name": "rst.Calculator_topEntity_types.rst_System"
                },
                {
                    "domain": "System",
                    "width": 1,
                    "name": "en",
                    "is_clock": false,
                    "type_name": "en.Calculator_topEntity_types.en_System"
                },
                {
                    "width": 7,
                    "name": "eta1",
                    "is_clock": false,
                    "type_name": "eta1.Calculator_topEntity_types.OPC"
                }
            ],
            "out": [
                {
                    "width": 5,
                    "name": "val",
                    "is_clock": false,
                    "type_name": "val.Calculator_topEntity_types.Maybe"
                }
            ]
        },
        "name": "topEntity"
    },
    "domains": {
        "System": {
            "period": 10000,
            "init_behavior": "Defined",
            "reset_kind": "Asynchronous",
            "active_edge": "Rising",
            "reset_polarity": "ActiveHigh"
        },
        "XilinxSystem": {
            "period": 10000,
            "init_behavior": "Defined",
            "reset_kind": "Synchronous",
            "active_edge": "Rising",
            "reset_polarity": "ActiveHigh"
        },
        "IntelSystem": {
            "period": 10000,
            "init_behavior": "Defined",
            "reset_kind": "Asynchronous",
            "active_edge": "Rising",
            "reset_polarity": "ActiveHigh"
        }
    }
}
```

-----------------------

TODO: 

 - [x] Rename blackbox primitive files


clash-lib/src/Clash/Driver/Manifest.hs
```diff
@@ -65,7 +72,58 @@ data Manifest
...
     -- ^ Names and hashes of all the generated files for the @TopEntity@. Hashes
     -- are SHA256.
-  } deriving (Show,Read)
+  } deriving (Show,Read,Eq)
+
+instance ToJSON Manifest where
+  toJSON (Manifest{..}) =
+    Aeson.object
+      [ "version" .= ("unstable" :: Text)
+      , "hash" .= manifestHash
```
alex-mckenna, line 81: I wonder if this should also be a sha256 field / hash, if only for uniformity?

clash-lib/src/Clash/Driver/Manifest.hs
```diff
@@ -65,7 +72,58 @@ data Manifest
   , fileNames :: [(FilePath, ByteString)]
     -- ^ Names and hashes of all the generated files for the @TopEntity@. Hashes
     -- are SHA256.
-  } deriving (Show,Read)
+  } deriving (Show,Read,Eq)
+
+instance ToJSON Manifest where
+  toJSON (Manifest{..}) =
+    Aeson.object
+      [ "version" .= ("unstable" :: Text)
```
alex-mckenna, line 80: I assume this is manifest format version. Do we also care about clash version? I would expect if a user updates Clash then components should be rebuilt as the compiler may output something different

clash-lib/src/Clash/Driver/Manifest.hs
```diff
@@ -65,7 +72,58 @@ data Manifest
...
+      , "hash" .= manifestHash
+      , "flags" .= successFlags
+        -- TODO: add nested ports (i.e., how Clash split/filtered arguments)
+      , "components" .= componentNames
```
alex-mckenna, line 84: If there is a manifest for each entity directory, when would components ever be more than one thing? I'm slightly unsure what this field is anyway

clash-lib/src/Clash/Driver/Manifest.hs
```diff
@@ -65,7 +72,58 @@ data Manifest
...
+        <*> (topComponent >>= (.: "name"))
+        <*> do
+              files <- v .: "files"
+              forM files $ \obj -> do
+                fName <- obj .: "name"
+                sha256 <- obj .: "sha256"
+                -- Note that we don't particularly care about hash decode
+                -- failures. Realistically it shouldn't happen, and if it does
+                -- there's almost no chance it would result in accidental caching.
+#if MIN_VERSION_base16_bytestring(1,0,0)
```
alex-mckenna, line 122: Can we not set this bound in the `.cabal` file?

clash-lib/tests/Clash/Tests/Driver/Manifest.hs
```diff
@@ -0,0 +1,51 @@
...
+
+genFilePath :: Q.Gen FilePath
+genFilePath = Text.unpack . coerce @ArbitraryText <$> Q.arbitrary
+
+genManifest :: Q.Gen Manifest
+genManifest =
+  Manifest
+    <$> Q.arbitrary -- hash
+    <*> Q.arbitrary -- flags
+    <*> fmap (map (coerce @ArbitraryText @Text)) Q.arbitrary -- in ports
```
alex-mckenna, line 37: Do you need to `map` the `coerce` here? I thought for many types `fmap coerce == coerce` meaning

```haskell
fmap coerce (Q.arbitrary @ArbitraryText)
```

would suffice

clash-lib/src/Clash/Driver/Manifest.hs
martijnbastiaan: This change just got out, so it's not on any Stack LTS yet </3

clash-lib/tests/Clash/Tests/Driver/Manifest.hs
martijnbastiaan: You're absolutely right!

clash-lib/src/Clash/Driver/Manifest.hs
alex-mckenna: :crying_cat_face:

clash-lib/tests/Clash/Tests/Driver/Manifest.hs
christiaanb: do you even need to `fmap` then? won't the following work?:
```haskell
coerce @(Q ArbitraryText) @(Q Text)
```

christiaanb: assuming the role of the type parameter of whatever applicative you're in is `representational`

alex-mckenna: Oh good point, I didn't even think about type roles. Yes you should be able to write

```haskell
coerce (Q.arbitrary @ArbitraryText)
```

clash-lib/src/Clash/Driver/Manifest.hs
martijnbastiaan: Yeah, that's all included in the `hash` field. This is just there to indicate to users that they really shouldn't rely on any API defined by the file. (We might later stabilize it though.)

martijnbastiaan: Having said that, having the Clash version+git hash that generated the file would be cool even if it's just for diagnostics.

clash-lib/src/Clash/Driver/Manifest.hs
alex-mckenna: Right makes sense, I didn't realise we called these components. Is there any benefit in keeping this field then, it seems to just be the information from `files` but without the hashes?

martijnbastiaan: Hmmyeah. In _theory_ the filenames can be wildly different from component names. Clash itself doesn't really generate such files, but it might happen once we do some magic with vendor IP?

clash-lib/src/Clash/Driver/Manifest.hs
martijnbastiaan: Sure thing

clash-lib/src/Clash/Driver/Manifest.hs
alex-mckenna: I've used [this](https://hackage.haskell.org/package/githash) before to bake it into a constant. We could probably use some `CPP` magic to have some `clashVersion` constant set to either the git tree + hash or the actual version number

martijnbastiaan: Sure, I'm going to leave this for a separate PR though as it's not really something we include right now. (And I think it might affect some other things we need to discuss too.)

clash-lib/tests/Clash/Tests/Driver/Manifest.hs
martijnbastiaan: We wouldn't get anything done without type checkers ;-)

```haskell
coerce @(Q.Gen [ArbitraryText]) @(Q.Gen [Text])
```

PR #1696, opened 2021-03-10
 Restore port types in manifest files 

martijnbastiaan:
See: https://github.com/gergoerdi/clash-issue-1686/pull/1#issuecomment-795198464. Sorry @gergoerdi!

Also adds some additional info:

* For clock, reset, and enable ports it now includes the synthesis domain (future work: add for all signal types too)
* Whether a port is clock
* Port width
* All domains and their properties


gergoerdi:
While you're at it, could you make it something other than just `Text`? It's quite crazy that I need to [parse it](https://github.com/gergoerdi/clashilator/blob/master/src/Clash/Clashilator.hs#L66) when I'm pretty sure it's not originally generated as text:)

martijnbastiaan:
I also want to add "is_clock" and domain information, which I believe would come in quite handy in (future iterations of) `clashilator`. Do you need anything else?

Issue #1629, opened 2021-01-11
Unify `Bundle` for `Signal` and `DSignal` (and other `Traversable` `Applicative` functors...)


gergoerdi:
Currently, there are two typeclasses (specific to `Signal` and `DSignal` unbundling) that implement the same logic: turning, e.g., `f (a, b)` into `(f a, f b)`. Couldn't these be unified into a single typeclass that doesn't care about the choice of `f`, as long as it is a traversable applicative functor?

To illustrate, here's what it would look like:

```
class Bundle a where
    type Unbundled (f :: Type -> Type) a = res | res -> f a
    type Unbundled f a = f a

    bundle :: (Applicative f) => Unbundled f a -> f a

    default bundle :: (Applicative f, Unbundled f a ~ f a) => Unbundled f a -> f a
    bundle = id

    unbundle :: (Traversable f) => f a -> Unbundled f a

    default unbundle :: (Traversable f, Unbundled f a ~ f a) => f a -> Unbundled f a
    unbundle = id
```

And as a proof-of-concept, here are some instances: for an "atomic" type, for a tuple, and for `Vec`:

```
instance Bundle Bit where
    bundle = id
    unbundle = id

instance Bundle (a, b) where
    type Unbundled f (a, b) = (f a, f b)
    bundle (x, y) = (,) <$> x <*> y
    unbundle xy = (fst <$> xy, snd <$> xy)

instance (KnownNat n) => Bundle (Vec n a) where
    type Unbundled f (Vec n a) = Vec n (f a)
    bundle = traverse# id
    unbundle = sequenceA . fmap lazyV
```

Is there any hidden gotcha I am missing? If not, I would prefer moving over to this typeclass, to avoid having to manually dispatch between the `Signal` and the `DSignal` one.

christiaanb:
Some code archeology doesn't explain whether the above approach was tried but rendered unsatisfactory type inference results. Perhaps @blaxill can tell us whether he tried the above approach before submitting the original PR for the `Signal.Delayed.Bundle` module?

martijnbastiaan:
I don't see any immediate reason not to do it. 

Just a heads up, the various `bundle` and `unbundle` primitives are currently hardcoded in the compiler:

https://github.com/clash-lang/clash-compiler/blob/1bde4831c94e38a4359be2c9c5fbab8e871569f6/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs#L371-L387

[...]

PR #1699, opened 2021-03-10
Bundle class based on MPTC+FD instead of TF

christiaanb:
TODO:
- [x] Generate Bundle tuple instances for `DSignal dom d` and for any `Applicative f`
- [x] Add `vecBundleD#` primitive
- [x] Fix `ghc-typelits-natnormalize` so it solves inside class constraints (reason for `clash-cores` failure): https://github.com/clash-lang/ghc-typelits-natnormalise/pull/53

NB: This is mostly an experiment for now: it generalizes Bundle to work over any `Applicative f`


gergoerdi:
What's the status of this?

christiaanb:
It's existence slipped my mind. I'll see whether I can get it in a working state soon.

christiaanb:
@gergoerdi if it passes CI, the implementation is pretty much ready.

christiaanb:
So CI failed on Nix and Stack because they aren’t configured with the latest `ghc-typelits-natnormalise` plugin.

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -535,7 +546,8 @@ coreToTerm primMap unlocs = term
               | f == "Clash.Signal.Internal.appSignal#" -> return (appSignalTerm xType)
               | f == "Clash.Signal.Internal.traverse#"  -> return (traverseTerm xType)
               | f == "Clash.Signal.Internal.joinSignal#" -> return (joinTerm xType)
-              | f == "Clash.Signal.Bundle.vecBundle#"   -> return (vecUnwrapTerm xType)
+              | f == "Clash.Prelude.Bundle.vecBundle#"   -> return (vecUnwrapTerm xType)
+              | f == "Clash.Prelude.Bundle.vecBundleD#"  -> return (vecUnwrapDTerm xType)
```
martijnbastiaan, line 550: Perhaps `'show vecBundle#`?

nix/sources.json
```diff
@@ -5,10 +5,10 @@
         "homepage": null,
         "owner": "clash-lang",
         "repo": "ghc-tcplugins-extra",
-        "rev": "380e4510ae94e6b6cfb1bcacf633437e2c7226a4",
-        "sha256": "0fwfjvqw4k9lk23xrysiw54khfcif2s2cg52zwkgd0wbhjg6rrmd",
+        "rev": "684f3aa53c193220974c542ed337ec94e5265291",
+        "sha256": "19sczl14hcaai20ckrd8gwqbrzyaby9pyv488ljxizq4ynqdx380",
         "type": "tarball",
-        "url": "https://github.com/clash-lang/ghc-tcplugins-extra/archive/380e4510ae94e6b6cfb1bcacf633437e2c7226a4.tar.gz",
+        "url": "https://github.com/clash-lang/ghc-tcplugins-extra/archive/684f3aa53c193220974c542ed337ec94e5265291.tar.gz",
```
martijnbastiaan, line 11: If you didn't already; niv can help you do this automatically: https://hackage.haskell.org/package/niv

martijnbastiaan (approved):
We should add a chapter on `Bundle` in clash-by-example, it's a pretty complex beast.

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
christiaanb: Done

nix/sources.json
christiaanb: Thanks! I didn't, so I had to do this manually.

clash-prelude/src/Clash/Prelude/Bundle.hs
```diff
@@ -118,53 +122,62 @@ class Bundle a where
...
-  type Unbundled t (Vec n a) = Vec n (Signal t a)
-  -- The 'Traversable' instance of 'Vec' is not synthesizable, so we must
-  -- define 'bundle' as a primitive.
+instance {-# OVERLAPPABLE #-} (KnownNat n, Applicative f, Traversable f) => Bundle f (Vec n a) (Vec n (f a)) where
+  type Unbundled f (Vec n a) = Vec n (f a)
+  bundle   = traverse# id
+  unbundle = sequenceA . fmap lazyV
+
+instance KnownNat n => Bundle (Signal dom) (Vec n a) (Vec n (Signal dom a)) where
+  type Unbundled (Signal dom) (Vec n a) = Vec n (Signal dom a)
```
leonschoorl, line 159: What's the point of the more specific `OVERLAPPING` instances?
Does it lead to better performance?

leonschoorl (changes requested):
I wonder if this is net gain.
The upside is having only one `Bundle` class, instead of two.
The downside is types for the common case (Signal) become more complicated.

clash-prelude/src/Clash/Prelude/Bundle.hs
christiaanb: Better run-time performance, no space leakage, and with the ability filter out the applications of the methods also better Haskell-to-HDL compile-time performance.

christiaanb:
@leonschoorl I created it in response to https://github.com/clash-lang/clash-compiler/issues/1629; perhaps @gergoerdi can better articulate why he prefers the solution in this PR over the status quo.

clash-prelude/src/Clash/Prelude/Bundle.hs
christiaanb: Ah, sorry, I thought the question was about all of the `Bundle` instances. So in the `Vec` case, the overlapping instances are there so can filter out the applications of `vecBundle#` (resp `vecBundleD#`) in GHC2Core. We want this because `traverse# :: Applicative f => (a -> f b) -> Vec n a -> f (Vec n b)` is always unrolled by: https://github.com/clash-lang/clash-compiler/blob/5f5739a775b9713f971ba33557771659fc1eb2aa/clash-lib/src/Clash/Normalize/Transformations/Reduce.hs#L199-L205

leonschoorl:
It also hurts type inference :(
This used to compile:
```haskell
x :: Signal System Bool
x = pure False

xs = bundle (x,x)
```

But now fails with:
```
    • Ambiguous type variables ‘f0’,
                               ‘a0’ arising from a use of ‘bundle’
      prevents the constraint ‘(Bundle
                                  f0
                                  a0
                                  (Signal System Bool, Signal System Bool))’ from being solved.
      Relevant bindings include
        xs :: f0 a0 (bound at BundleCompat.hs:9:1)
      Probable fix: use a type annotation to specify what ‘f0’,
                                                          ‘a0’ should be.
      These potential instances exist:
        instance [overlappable] Applicative f =>
                                Bundle f (a1, a2) (f a1, f a2)
          -- Defined in ‘Clash.Prelude.Bundle’
        instance Bundle
                   (Signal dom) (a1, a2) (Signal dom a1, Signal dom a2)
          -- Defined in ‘Clash.Prelude.Bundle’
    • In the expression: bundle (x, x)
      In an equation for ‘xs’: xs = bundle (x, x)
  |
9 | xs = bundle (x,x)
```

gergoerdi:
> @leonschoorl I created it in response to #1629; perhaps @gergoerdi can better articulate why he prefers the solution in this PR over the status quo.

Well, fundamentally it is an operation that is well-defined for any (traversable) applicative functor, so it'd be nice to present it as such instead of requiring separate imports (and qualified occurrences) for the two special cases. 

In https://github.com/clash-lang/clash-compiler/issues/1629 I raised this asking if there's some reason it wasn't done this way initially. The answer seemed to be that it wasn't an avenue investigated at all; then some technical problems came up, and then it seemed Christiaan and I addressed those. So if there are more technical problems, this juice might not be worth the squeeze and we could just shelf it again.

martijnbastiaan:
Given that the current type families are already hard for beginners to parse, and this PR would worsen that and type inference.. I'm inclined to reject it. @christiaanb Do you want to go through with this? If not, could we close the PR?

PR #1703, opened 2021-03-11
Preserve build order in Manifest files

martijnbastiaan:
Manifest files now list files in reverse topological order. This makes
sure files can be fed to EDA tools in that order, without triggering
compilation errors.

### DIFF
--- a/clash-lib/src/Clash/Driver/Manifest.hs
+++ b/clash-lib/src/Clash/Driver/Manifest.hs
@@ -119,6 +119,9 @@ data Manifest
     -- ^ Names of all the generated components for the @TopEntity@ (does not
     -- include the names of the components of the @TestBench@ accompanying
     -- the @TopEntity@).
+    --
+    -- This list is reverse topologically sorted. I.e., a component might depend
+    -- on any component listed before it, but not after it.
   , topComponent :: Text
     -- ^ Design entry point. This is usually the component annotated with a
     -- @TopEntity@ annotation.
--- a/clash-lib/src/Clash/Driver/Manifest.hs
+++ b/clash-lib/src/Clash/Driver/Manifest.hs
@@ -43,9 +43,10 @@ import           Text.Read (readMaybe)
 
 import           Clash.Annotations.TopEntity.Extra ()
 import           Clash.Backend (Backend (hdlType), Usage (External))
+import           Clash.Core.Name (nameOcc)
 import           Clash.Driver.Types
 import           Clash.Primitives.Types
-import           Clash.Core.Var (Id)
+import           Clash.Core.Var (Id, varName)
 import           Clash.Netlist.Types
   (TopEntityT, Component(..), HWType (Clock), hwTypeDomain)
 import qualified Clash.Netlist.Types as Netlist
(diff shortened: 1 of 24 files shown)


clash-lib/src/Clash/Driver/Manifest.hs
```diff
@@ -161,6 +168,8 @@ instance ToJSON Manifest where
             ]
           )
         | (domNm, VDomainConfiguration{..}) <- HashMap.toList domains ]
+      , "dependencies" .= Aeson.object
+        [ "transitive" .= transitiveDependencies ]
```
alex-mckenna, line 172: Does this need to be an object with a single key? Are you planning on adding other types of dependencies later?

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -281,13 +283,16 @@ data NetlistEnv
   -- ^ (Maybe) user given instance/register name
   }
 
+type ComponentMap = OMap Unique ([Bool], SrcSpan, IdentifierSet, Component)
```
alex-mckenna, line 286: If you're giving it a name, you _could_ also make it a record :eyes:

clash-lib/src/Clash/Util.hs
```diff
@@ -184,6 +186,25 @@ makeCachedU key l create = do
...
+  -> m v
+  -- ^ Action to cache
+  -> m v
+makeCachedO key l create = do
+  cache <- use l
+  case OMap.lookup (getUnique key) cache of
+    Just value -> return value
+    Nothing -> do
+      value <- create
+      l %= (flip (OMap.|>)) (getUnique key, value)
```
alex-mckenna, line 205: I'm not really a fan of qualifying operators, I'd rather see it in the import list and used "normally"

alex-mckenna:
Looks nice overall though :+1:

clash-lib/src/Clash/Driver/Manifest.hs
martijnbastiaan: I intended to add only direct dependencies (which I think is neater), but only the transitive closure was easily obtainable. At some point I do want to add the former though. Given that the JSON is probably going to be parsed by Python's `json.load`, it'd be nice if it didn't break on adding it.

clash-lib/src/Clash/Netlist/Types.hs
martijnbastiaan: Yeah good point

clash-lib/src/Clash/Util.hs
martijnbastiaan: Hmm yeah, I really don't like obscure operators like these and I generally just avoid them by using their named versions, but `OMap` doesn't export those. But I'll apply your suggestion anyway.

christiaanb: Please keep the qualification, I’d confuse it with the `Data.Sequence` operator otherwise.

martijnbastiaan: 2:1, sorry Alex

PR #1705, opened 2021-03-12
Move Stack and Nix builds to GitLab CI

martijnbastiaan:


christiaanb:
Will this still run the circle ci builds for external contributors?

martijnbastiaan:
"Normal" tests: yes, stack/nix: no. There's already a number of tests GitLab CI only (haddock, bindist tests), so I figured we should just treat Circle CI as the first line of defense and use its minutes sparingly. If we see any dependency changes in a PR, we should just clone and push to a local directory to run everything on GitLab. 

I think we should spend a day on building a bot that does the clone+push for us on a, say, `@clash-lang-builder run ci` comment. I've talked to @leonschoorl and @alex-mckenna (independently, just during "coffee moments" so-to-speak) about this before. It should be pretty easy with GitHub Actions and/or probot:

 * https://github.com/probot/example-github-action
 * https://github.com/probot/adapter-github-actions
 * GitHub Actions's issue_comment: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#issue_comment

martijnbastiaan:
Going to go ahead and merge this so we can use it for https://github.com/clash-lang/clash-compiler/pull/1708.

PR #1722, opened 2021-03-26
Eagerly reduce nested type family applications

### LINKED ISSUE #1721: Fails Generating VHDL When Typeclass Has Many Instances
The compiler will fail to generate an output port on noinlined functions when a typeclass has many specialized instances.
It's suspected that the compiler skips generating an output port if the output of a function is one of many instances. Without the output port, the Netlist generator thinks the function is being applied to too many I/O components and throws an UnderApplied error.

An example is given in the attached file (this will fail to generate, and dump a long Core message).
[clash-fix.zip](https://github.com/clash-lang/clash-compiler/files/6207888/clash-fix.zip)

christiaanb:
Fixes #1721

### DIFF
--- a/clash-lib/src/Clash/Core/Type.hs
+++ b/clash-lib/src/Clash/Core/Type.hs
@@ -490,7 +490,10 @@ funSubst tcm (Just s) = uncurry go
     -- to different types, so this is OK for our purposes.
     go (AppTy a1 r1) (AppTy a2 r2) = do
       s1 <- funSubst tcm (Just s) (a1, a2)
-      funSubst tcm (Just s1) (r1, r2)
+      funSubst tcm (Just s1)
+                   ( r1
+                   , argView tcm r2 -- See [Note: Eager type families]
+                   )
 
     go ty1@(ConstTy _) ty2 =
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Core/Type.hs
```diff
@@ -490,7 +490,10 @@ funSubst tcm (Just s) = uncurry go
     -- to different types, so this is OK for our purposes.
     go (AppTy a1 r1) (AppTy a2 r2) = do
       s1 <- funSubst tcm (Just s) (a1, a2)
-      funSubst tcm (Just s1) (r1, r2)
+      funSubst tcm (Just s1)
+                   ( r1
+                   , argView tcm r2 -- See [Note: Eager type families]
```
martijnbastiaan, line 495: We could add to the note that:

* There's no behavior (as far as we can tell) that depends on lazy evaluation
* GHC already resolved the type family, so we should never get stuck (and it is, therefore, safe to evaluate strictly)

martijnbastiaan (approved):
LGTM

PR #1725, opened 2021-03-26
Partial Evaluator II

alex-mckenna:
Preliminary work on adding primitives to the partial evaluator. Adding primitives naturally also exposed many holes in the semantics, which have been chunked up into commits according to which construct is evaluated. Still TODO:

  - [x] Add a flag for selective enabling / disabling of partial evaluation (`-fclash-partial-evaluation`). This flag should probably be opt-in for now while any defects are ironed out
  - [ ] Figure out an approach for sensibly testing with or without the partial evaluator. I suspect this means having a means to run all the tests in `clash-testsuite` with some clash options that apply to all tests (so all tests can have partial evaluation turned on). While in the short term this is okay (branches starting `partial-evaluator-` would turn it on) this isn't so good long term until the partial evaluator is on by default. I expect people to have opinions about this
  - [ ] Handle `Args` more consistently. At some point I changed arguments to be delayed in a way which would mean impredicate data would delay correctly - a similar change should really apply to `forceArgs`. The change does not affect `Args` which are not impredicative, so is harmless for normal types
  - [ ] `RTree` primitives. Turns out I never implemented any, but it doesn't seem to matter for now. I assume the old transformation pipeline is picking up the slack
  - [ ] Iron out issues caused by the use of work free / expandable checks which prevent a more complete constant folding. While this doesn't lead to many test failures, there is no reason for partial evaluation to _not_ fully fold constants where they appear
  - [ ] Double check that all small changes to `clash-lib` are actually needed. Some may be redundant in the current form of the partial evaluator, and should be removed if possible


martijnbastiaan:
Closing due to inactivity - please reopen if there is an intention to move through with it.

PR #1726, opened 2021-03-26
Add true, dual-port blockram

martijnbastiaan:
TODO:

- [ ] VHDL primitive
- [ ] SystemVerilog primitive (same as Verilog?)
- [ ] Test with clock `A` + `C`
- [ ] Test conflict behavior


christiaanb:
@martijnbastiaan @leonschoorl the distinction between slow and fast is wrong, at least, `veryUnsafeSynchronizer` doesn't have a notion of slow and fast:
```haskell
veryUnsafeSynchronizer
  :: Int
  -- ^ Period of clock belonging to @dom1@
  -> Int
  -- ^ Period of clock belonging to @dom2@
  -> Signal dom1 a
  -> Signal dom2 a
veryUnsafeSynchronizer t1 t2
  -- this case is just an optimization for when the periods are the same
  | t1 == t2 = same

  | otherwise = go 0

  where
  same :: Signal dom1 a -> Signal dom2 a
  same (s :- ss) = s :- same ss

  go :: Int -> Signal dom1 a -> Signal dom2 a
  go relativeTime (a :- s)
    | relativeTime <= 0 = a :- go (relativeTime + t2) (a :- s)
    | otherwise = go (relativeTime - t1) s
```
so when `t1` is larger than `t2` (i.e. `t1` is _slower_ than `t2`)  then you need to add `t2` multiple times to `relativeTime` to catch up to the subtraction of one `t1` from `relativeTime`. Whereas when `t2` is larger than `t1` (i.e. `t2` is _slower_ than `t1`) then you need to subtract `t1` multiple times from `relativeTime` to catch up to the addition of one `t2` to `relativeTime`.

I guess what I'm trying to say is:
1. Is this concept of slow and fast clocks something we really need? or
2. Can we just make the portA and portB behavior symmetric? i.e. it behaves correctly regardless of whether portA is faster than portB or partA is faster than portB

martijnbastiaan:
The purpose of the slow/fast clocks is simply because we know the _slow_ clock will consume exactly one cycle from one of its streams. I.e., the next iteration of `go` will be spent in the section handling the fast clock. This allows us to set the "current conflict" back to `Nothing`:

https://github.com/clash-lang/clash-compiler/blob/efe12084e1213fd1c8161f8cd14506c946af857c/clash-prelude/src/Clash/Explicit/BlockRam.hs#L1258

Without switching the clocks at the very start you'd need to check explicitly whether the "next" tick is going to be spent in the "other" part of go (or precalculate how many cycles you spend, but that's "ugly" IMO as you need to do it again and again).

christiaanb:
Alright, that makes sense.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1038,3 +1046,241 @@ readNew clk rst en ram rdAddr wrM = mux wasSame wasWritten $ ram rdAddr wrM
...
+  -- | Outputs data on /next/ cycle. If write enable is @True@, the data written
+  -- will be echoed. If write enable is @False@, the read data is returned.
+  (Signal domA a, Signal domB a)
+trueDualPortBlockRam clkA writeA addrA clkB writeB addrB = case compareSNat (SNat @(DomainPeriod domB)) (SNat @(DomainPeriod domA)) of
+  SNatLE ->
+    trueDualPortBlockRam#
+      clkA (isJust <$> writeA) addrA (fromJustX <$> writeA)
+      clkB (isJust <$> writeB) addrB (fromJustX <$> writeB)
+  SNatGT ->
+    swap (trueDualPortBlockRam clkB writeB addrB clkA writeA addrA)
```
christiaanb, line 1094:
```suggestion
    swap (trueDualPortBlockRam# clkB (isJust <$> writeB) addrB (fromJustX <$> writeA) 
                                clkA (isJust <$> writeA) addrA (fromJustX <$> writeB))
```
This reduces the burdon on the rewrite mechanism, and also ensures `trueDualPortBlockRam` is not needlessly recursive, making it a candiate for inlining.

martijnbastiaan: Alternatively, we should simply do this in the primitive itself. This would prevent surprising HDL (swapped ports) too.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -386,7 +386,7 @@ data SDomainConfiguration (dom :: Domain) (conf :: DomainConfiguration) where
 
 deriving instance Show (SDomainConfiguration dom conf)
 
-type KnownConfiguration dom conf = (KnownDomain dom, KnownConf dom ~ conf)
+type KnownConfiguration dom conf = (KnownDomain dom, KnownConf dom ~ conf, KnownNat (DomainConfigurationPeriod conf))
```
martijnbastiaan, line 389: Could we add this is a superclass of `KnownDomain`?

christiaanb: No, this is the wrong approach to begin with. We should extract the SNat out of the domain configuration in the `trueDualPortRam` function instead of having the `KnownNat` constraint. i.e. we already have `KnownDomain` constraint, get the `SDomainConfiguration` out of that, and use the `SNat` inside of it.

martijnbastiaan: Agreed. @leonschoorl This works too:

```haskell
case knownDomain @someDom of
  SDomainConfiguration _ sPeriod _ _ _ _ _ _ _ _
```

Ideally this would be a record type, but I didn't know you could do that for GADTs when I implemented synthesis domains.

martijnbastiaan: Or is there another reason you opted for `KnowNat` instead?

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1038,3 +1046,241 @@ readNew clk rst en ram rdAddr wrM = mux wasSame wasWritten $ ram rdAddr wrM
...
+
+toMaybeX :: a -> MaybeX a
+toMaybeX a =
+  case isX a of
+    Left _ -> IsX
+    Right _ -> IsDefined a
+
+data MaybeX a = IsX | IsDefined !a
```
martijnbastiaan, line 1102: I added this type, but wasn't sure whether it was really necessary. Perhaps a `Maybe` + documentation in the record would do. I like the explicitness of it though.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1038,3 +1045,280 @@ readNew clk rst en ram rdAddr wrM = mux wasSame wasWritten $ ram rdAddr wrM
...
+--
+--   * Writes over reads
+--   * Undefineds over reads/writes
+--   * @Just@s over @Nothing@s
+--
+mergeMaybeConflicts :: Maybe Conflict -> Maybe Conflict -> Maybe Conflict
+mergeMaybeConflicts Nothing Nothing = Nothing
+mergeMaybeConflicts (Just c) Nothing = Just c
+mergeMaybeConflicts Nothing (Just c) = Just c
+mergeMaybeConflicts (Just c1) (Just c2) = Just (mergeConflicts c1 c2)
```
alex-mckenna, line 1130: `mergeConflicts` looks to be a `Semigroup` instance for `Conflict`, and `mergeMaybeConflicts` is then just `<>` on `(Semigroup a) => Maybe a`

clash-prelude/src/Clash/Explicit/BlockRam.hs
leonschoorl: I've moved this swapping inside the (haskell) simulation model, clash never sees it now.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1064,3 +1071,258 @@ readNew clk rst en ram rdAddr wrM = mux wasSame wasWritten $ ram rdAddr wrM
...
+  => Clock domA
+  -- ^ Clock for port A
+  -> Signal domA (Maybe a)
+  -- ^ Write data port A
+  -> Signal domA (Index nAddrs)
+  -- ^ Address to read from or write to on port A
+
+  -> Clock domB
+  -- ^ Clock for port B
+  -> Signal domB (Maybe a)
```
rowanG077, line 1102: I would rather have a new type for this. I assume if you pass in a `Just` here then it means you want to write and if you pass `Nothing` then a value is read. Why not just make a special `Either` here. `data DualPortRamOp n a = DualPortWrite (Index n) a | DualPortRead (Index n)` or something.

PR #1754, opened 2021-04-07
Improve and fix documentation

DigitalBrains1:
- Several functions defined and documented in
`Clash.Signal.Delayed.Internal` also show up in the documentation for
`Clash.Signal.Delayed`. The former module uses explicit clocks, resets
and enables, but the latter uses hidden parameters. This can make the
documentation confusing. The confusion specifically dealt with in this
commit is that the `sampleN` from `Clash.Prelude` does not work with
`DSignal` whereas `sampleN` from `Clash.Explicit.Prelude` does. To make
the code valid for either one, an application of `toSignal` is included
in the examples. However, none of the examples for code defined in
`Clash.Explicit.Signal.Delayed` end up in confusing locations, so all
`toSignal`s there have been removed for consistenty (it was half with,
half without).

- Some examples did not typecheck, or the executed code differed from the
code in the documentation.

- The example of `InlinePrimitive` is improved by guiding the reader to
use `QuasiQuotes`.

### DIFF
--- a/clash-prelude/src/Clash/Annotations/Primitive.hs
+++ b/clash-prelude/src/Clash/Annotations/Primitive.hs
@@ -241,6 +241,7 @@ data HDL
 -- nicer multiline strings.
 --
 -- @
+-- {\-\# LANGUAGE QuasiQuotes \#-\}
 -- module InlinePrimitive where
 --
 -- import           Clash.Annotations.Primitive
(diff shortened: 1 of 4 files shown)


martijnbastiaan:
> The example of InlinePrimitive is improved by guiding the reader to use QuasiQuotes.

I think you forgot to commit / push this

martijnbastiaan (approved):

PR #1763, opened 2021-04-13
Add 'Clash.Class.Counter'

martijnbastiaan:
I keep reinventing this class each time I do hardware development with Clash. I propose we add it to the prelude.

### DIFF
--- a/clash-prelude/src/Clash/Class/Counter.hs
+++ b/clash-prelude/src/Clash/Class/Counter.hs
@@ -1,0 +1,62 @@
+{-|
+  Copyright   :  (C) 2021     , QBayLogic B.V.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+  Utilities for wrapping counters consisting of multiple individual counters
+-}
+
+module Clash.Class.Counter
+  ( Counter
+  , countSucc
+  , countPred
+  ) where
+
+import Clash.Class.Counter.Internal
+
+-- $setup
+-- >>> import Clash.Class.Counter
+-- >>> import Clash.Sized.BitVector (BitVector)
+-- >>> import Clash.Sized.Index (Index)
+-- >>> import Clash.Sized.Signed (Signed)
+-- >>> import Clash.Sized.Unsigned (Unsigned)
+
+-- | Successor of a counter.
+--
+-- Examples:
+--
+-- >>> type T = (Unsigned 2, Unsigned 2)
+-- >>> countSucc @T (1, 1)
+-- (1,2)
+-- >>> countSucc @T (1, 2)
+-- (1,3)
+-- >>> countSucc @T (1, 3)
+-- (2,0)
+-- >>> countSucc @T (3, 3)
+-- (0,0)
+-- >>> countSucc @(Index 9, Index 2) (0, 1)
+-- (1,0)
+-- >>> countSucc @(Either (Index 9) (Index 9)) (Left 8)
+-- Right 0
+countSucc :: Counter a => a -> a
+countSucc = snd . countSuccOverflow
+
+-- | Predecessor of a counter
+--
+-- Examples:
+--
+-- >>> type T = (Unsigned 2, Unsigned 2)
+-- >>> countPred @T (1, 2)
+-- (1,1)
+-- >>> countPred @T (1, 3)
+-- (1,2)
+-- >>> countPred @T (2, 0)
+-- (1,3)
+-- >>> countPred @T (0, 0)
+-- (3,3)
+-- >>> countPred @(Index 9, Index 2) (1, 0)
+-- (0,1)
+-- >>> countPred @(Either (Index 9) (Index 9)) (Right 0)
+-- Left 8
+countPred :: Counter a => a -> a
+countPred = snd . countPredOverflow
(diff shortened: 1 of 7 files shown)


clash-prelude/src/Clash/Class/Counter.hs
```diff
@@ -0,0 +1,151 @@
...
+-- number of drawn frames. To do so, the three counters are setup with different
+-- types. On each _round_ of the horizontal counter the vertical counter should
+-- be increased. On each _round_ of the vertical counter the frame counter should
+-- be increased. With this class you could simply use the type:
+--
+--   (FrameCount, VerticalCount, HorizontalCount)
+--
+-- and have 'countSucc' work as described.
+--
+class (Eq a, Bounded a) => Counter a where
```
christiaanb, line 39: Could you drop the `Bounded a` super-class constraint so we can also have up-down counters with a non-50% duty cycle using e.g. `Counter (Either (Index 4) (Index 10))`: `Either` doesn't have a `Bounded` instance.

martijnbastiaan: Sure, following the current logic that'd mean that:

```haskell
>>> satSucc (Either (0 :: Index 4) (9 :: Index 10)) 
Either 1 0
```

I'd have to think a bit about what to change to the class. I'll get to it tomorrow.

clash-prelude/src/Clash/Class/Counter.hs
```diff
@@ -0,0 +1,151 @@
...
+  countPred :: a -> a
+  default countPred :: SaturatingNum a => a -> a
+  countPred = satPred SatWrap
+
+instance (1 <= n, KnownNat n) => Counter (Index n)
+instance KnownNat n => Counter (Unsigned n)
+instance KnownNat n => Counter (Signed n)
+instance KnownNat n => Counter (BitVector n)
+
+instance (Counter a, Counter b) => Counter (a, b) where
```
alex-mckenna, line 73: All these tuple instances seem pretty mechanical, do you think it would be worth using template Haskell?

martijnbastiaan: Yes! It's even a TODO in that file :)

christiaanb: If only Haskell were like BlueSpec, and any n-tuple larger than a 2-tuple would be desugared to a nested 2-tuple, meaning you only ever have to define a 2-tuple instances.

clash-prelude/src/Clash/Class/Counter/Internal.hs
```diff
@@ -0,0 +1,223 @@
...
+-- >>> countSucc @T (Right 0)
+-- Right 1
+instance (Counter a, Counter b) => Counter (Either a b) where
+  countMin = Left countMin
+  countMax = Right countMax
+
+  countSuccOverflow e =
+    case bimap countSuccOverflow countSuccOverflow e of
+      Left (overflow, a)  -> (False, if overflow then Right countMin else Left a)
+      Right (overflow, b) -> (overflow, if overflow then Left countMin else Right b)
```
christiaanb, line 94: Sadly, Clash's case-of-case transformation only fires when the subject of the outer case (the alternatives of the inner case) are non-representable: https://github.com/clash-lang/clash-compiler/blob/12fa89070d2cf55c133ab93af2faab2b611659f3/clash-lib/src/Clash/Normalize/Transformations.hs#L392-L397

I don't know whether GHC does a case-of-case transformation, but if it doesn't then this code introduces two multiplexers where one would be sufficient. You should check the output. If a case-of-case doesn't fire (either in GHC or Clash), then you should rewrite to:

```haskell
countSuccOverflow e = case e of
  Left a -> let (overflow,a1) = countSuccOverflow a in (False, if overflow then Right countMin else Left a1)
  Right b -> let (overflow,b1) = countSuccOverflow b in (overflow, if overflow then Left countMin else Right b1)
```

alex-mckenna: > I don't know whether GHC does a case-of-case transformation
It does in the simplifier IIRC

martijnbastiaan: It only generates one :)

clash-prelude/src/Clash/Class/Counter.hs
```diff
@@ -0,0 +1,64 @@
+{-|
+  Copyright   :  (C) 2021     , QBayLogic B.V.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+  Utilities for wrapping counters consisting of multiple individual counters
+-}
+
+{-# LANGUAGE TypeFamilies #-}
```
leonschoorl, line 9: I don't see any type families here

clash-prelude/src/Clash/Class/Counter/Internal.hs
```diff
@@ -0,0 +1,223 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TypeFamilies #-}
```
leonschoorl, line 3: No type families here either

alex-mckenna: :family:

martijnbastiaan: Without it, you can't write `1 <= n` as a class constraint. The other file doesn't need it anymore though!

clash-prelude/src/Clash/Class/Counter.hs
martijnbastiaan: tnx

PR #1764, opened 2021-04-14
Only generate up to 3-tuple instances for Haddock

martijnbastiaan:
We currently show up to 12, which hides instance documention on all but
the widest of screens.

### DIFF
--- a/clash-prelude/src/Clash/Class/AutoReg/Instances.hs
+++ b/clash-prelude/src/Clash/Class/AutoReg/Instances.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
+-- {-# OPTIONS_GHC -ddump-splices #-}
 
 module Clash.Class.AutoReg.Instances where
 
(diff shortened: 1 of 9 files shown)


clash-prelude/src/Clash/Class/Counter/Internal.hs
```diff
@@ -115,7 +115,13 @@ instance (Counter a, Counter b) => Counter (Either a b) where
 -- (0,1,1)
 -- >>> countSucc @T (0, 1, 1)
 -- (1,0,0)
-instance (Counter a, Counter b) => Counter (a, b) where
+--
+-- __N.B.__: The documentation only shows the instances up to _3_ tuples. By
+-- default, instances up to and including /12/-tuples will exist. If the flag
+-- @clash-prelude@ is set instances up to the GHC imposed limit will exist. The
```
DigitalBrains1, line 121:
```suggestion
-- @large-tuples@ is set instances up to the GHC imposed limit will exist. The
```

martijnbastiaan: woops, thanks

leonschoorl:
I've added the same message to NFDataX and ShowX.

And it turns out you can just put a haddock comment before a TH call and it will apply to the (last?) thing generated by that TH call.
So undid you manual AutoReg instance.

martijnbastiaan:
> and it will apply to the (last?) thing generated by that TH call.

That's horrible. But also, nice job :P.

clash-prelude/src/Clash/Class/AutoReg/Instances.hs
```diff
@@ -6,6 +6,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 
 {-# OPTIONS_GHC -Wno-orphans #-}
+-- {-# OPTIONS_GHC -ddump-splices #-}
```
DigitalBrains1, line 9: Should this be included in the commit?

martijnbastiaan: Oh yes, I like having that in files with template haskell splices. It makes it a bit easier to debug!

PR #1765, opened 2021-04-14
Use effective number of CPUs on CI, not available cores

martijnbastiaan:
Before this patch, CI would use `nproc` to determine the number of CPUs
in a Docker instance. This only works when the docker container is
pinned to certain cpus, not if it's just rate limited. This script
detects that rate limiting, allowing our GitLab runners to unpin the
runners. This would allow us to "oversell" GitLab runners, hopefully
reducing the idle time imposed by single-threaded workloads.

### DIFF
--- a/.ci/effective_cpus.sh
+++ b/.ci/effective_cpus.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# `nproc` doesn't account for limits set by cgroups/docker. This script tries
+# to determine the effective number of cpus we can use by inspecting the shares
+# it has been given.
+set -euo pipefail
+IFS=$'\n\t'
+
+cfs_quota_us=$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+cfs_period_us=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us)
+
+if [[ ${cfs_quota_us} == -1 ]]; then
+  # No limits set
+  nproc
+else
+  expr "${cfs_quota_us}" / "${cfs_period_us}"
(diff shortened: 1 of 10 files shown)


.ci/effective_cpus.sh
```diff
@@ -0,0 +1,16 @@
...
+IFS=$'\n\t'
+
+cfs_quota_us=$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+cfs_period_us=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us)
+
+if [[ ${cfs_quota_us} == -1 ]]; then
+  # No limits set
+  nproc
+else
+  ghci -e "quot ${cfs_quota_us} ${cfs_period_us}"
```
alex-mckenna, line 15: I would use `expr` here instead

```bash
expr ${cfs_quota_us} / ${cfs_period_us}
```

Note the lack of quote marks: add them and it doesn't work as you expect (even ")

alex-mckenna (approved):
Other than the comment, LGTM

.ci/effective_cpus.sh
alex-mckenna: Although actually, you can do

```bash
expr "${cfs_quota_us}" / "${cfs_period_us}"
```

it's just putting quotes around the entire expr that makes it not work as expected

martijnbastiaan: Ah, I didn't know about `expr`! I was tempted to use `bc`, but I didn't feel like adding a new dependency to the docker image.

PR #1774, opened 2021-04-19
Add isLike to BitPack module

alex-mckenna:


clash-prelude/src/Clash/Sized/Internal/BitVector.hs
```diff
@@ -1237,14 +1237,14 @@ undefined# =
 --
 -- >>> let expected = $$(bLit "1.") :: BitVector 2
 -- >>> let checked  = $$(bLit "11") :: BitVector 2
--- >>> checked  `isLike` expected
+-- >>> checked  `isLike#` expected
 -- True
--- >>> expected `isLike` checked
+-- >>> expected `isLike#` checked
```
martijnbastiaan, line 1242: As you linked to it in `isLike` I read the documentation of this function. I think it could use some work:

* I think we should introduce it as "Checks for equality but ignores undefined bits in its second argument". I think "Check if one BitVector is like another." is too vague/imprecise. 
* Explicitly mention what happens when bits are set to undefined in the _first_ argument.

clash-prelude/src/Clash/Class/BitPack.hs
```diff
@@ -134,6 +135,19 @@ packXWith f x =
...
 
+-- | Check if the bit patterns of two values are similar. For most types this
+-- is equivalent to '(==)', but for types which contain 'BitVector' or 'Bit'
+-- it can behave more leniantly when certain bits are _don't care_. See
+-- 'Clash.Sized.Internal.BitVector.isLike#' for more information.
+--
+isLike
+  :: (BitPack a)
+  => a
+  -> a
```
martijnbastiaan, line 146: Have you thought about writing:

```haskell
isLike :: (BitPack a, BitPack b, BitSize a ~ BitSize b) => a -> b -> Bool
```

I'm on the fence on whether this is a good idea.

martijnbastiaan (changes requested):

clash-prelude/src/Clash/Sized/Internal/BitVector.hs
alex-mckenna: > "Check if one BitVector is like another." is too vague/imprecise

At least it's correct :wink: 

But no, point taken on the documentation here and in the other module. I'll make it better tomorrow

martijnbastiaan (approved):

alex-mckenna:
@martijnbastiaan Backport?

martijnbastiaan:
My heart says yes, but my head says no. Strictly seen it breaks the API as it introduces a new function. Let's not backport now, and reevaluate when we need it in 1.4.

Issue #1573, opened 2020-11-08
Logical right shift missing


rongcuid:
I know Haskell does not expose the logical right shift operator, but this is an operator used in hardware. Should this be provided by Clash, or do I need to do `uncheckedIShiftRL#`?

rowanG077:
`shiftR` is exposed under the [`Bits`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Bits.html#v:shiftR) type class. Is this not what you need?

martijnbastiaan:
I partly blame this on our documentation. Given that bit manipulation is such an integral part of hardware design, perhaps we should just re-export `Data.Bits` in clash-prelude such that it [shows up under "Models"](https://hackage.haskell.org/package/clash-prelude-1.2.4)?

rowanG077:
What would be most logical to me would be to roll `BitIndex` and `BitReduction` into a single page and there re-expose the `Data.Bits` functions on that page as well. That would keep all bit level stuff together. But i'm not sure if that's possible with haddock.

martijnbastiaan:
`Clash.Bits`!

rongcuid:
I asked on Slack channel, and found out I misunderstood the docs. Haskell `Data.Bits` says `shiftR` performs arithmetic right shift, but it turns out Unsigned and BitVector types actually perform logical right shift. I would say this is a documentation problem.

Speaking of which, I find the docs of Clash quite hard to navigate, because the Prelude docs don't show many methods, and I have to guess their names and look inside Index or other modules.

martijnbastiaan:
Thanks for the feedback @rongcuid. If you've got any specific suggestions on how to improve our documentation (e.g., where to show certain functionality), we'd be happy to hear it!

rongcuid:
I would like a page describing the most important data types of ClaSH, and a list of operations. If that sounds too vague, maybe it is good to provide a quick reference and comparison between, say, Verilog operators. Currently the docs are very much scattered around.

alex-mckenna:
We have [this page](https://clash-lang.readthedocs.io/en/latest/developing-hardware/prelude.html) in our readthedocs, but it only covers the main types in `Clash.Prelude`, not their full APIs. It may help give you an idea which types / modules to then look at in the haddock documentation though.

rongcuid:
Ah, I didn't even see this site. The website Documentation page directs straight to the Hackage documents. Is there a contributing guide or something? I might consider doing so later.

[...]

PR #1775, opened 2021-04-20
Move `BitIndex` and `BitReduction` under `BitPack`

martijnbastiaan:
And export functions from 'Clash.Class.BitPack' to make them more
discoverable.

-----------

Part of: https://github.com/clash-lang/clash-compiler/issues/1573


clash-prelude/src/Clash/Class/BitPack.hs
```diff
@@ -1,24 +1,13 @@
+
 {-|
-Copyright  :  (C) 2013-2016, University of Twente,
-                  2016-2017, Myrtle Software Ltd
```
alex-mckenna, line 3: _Removing_ a copyright notice?

clash-prelude/src/Clash/Class/BitPack.hs
```diff
@@ -30,420 +19,22 @@ module Clash.Class.BitPack
   , bitToBool
   , packXWith
 
-  -- * Internals
-  , GBitPack(..)
+  -- * Bit indexing
```
alex-mckenna, line 22: These are section titles, it should be "Bit Indexing"

clash-prelude/src/Clash/Class/BitPack/BitIndex.hs
```diff
@@ -0,0 +1,148 @@
+{-|
+Copyright  :  (C) 2013-2016, University of Twente
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
```
alex-mckenna, line 4: * Update the copyright notice
  * Save @christiaanb the burden of being the sole maintainer :wink:

clash-prelude/src/Clash/Class/BitPack/BitIndex.hs
```diff
@@ -0,0 +1,148 @@
...
+-- >>> slice d6 d4 (7 :: Unsigned 6)
+-- <BLANKLINE>
+-- <interactive>:...
+--     • Couldn't match type ‘7 + i0’ with ‘6’
+--         arising from a use of ‘slice’
+--       The type variable ‘i0’ is ambiguous
+--     • In the expression: slice d6 d4 (7 :: Unsigned 6)
+--       In an equation for ‘it’: it = slice d6 d4 (7 :: Unsigned 6)
+slice :: (BitPack a, BitSize a ~ ((m + 1) + i)) => SNat m -> SNat n -> a
+      -> BitVector (m + 1 - n)
```
alex-mckenna, line 63: I know this is probably lifted as-is from somewhere else, but I would take the time to fix the formatting of that type signature while the diff is all green anyway

clash-prelude/src/Clash/Class/BitPack/BitIndex.hs
```diff
@@ -0,0 +1,148 @@
...
+{-# INLINE split #-}
+-- | Split a value of a bit size @m + n@ into a tuple of values with size @m@
+-- and size @n@.
+--
+-- >>> pack (7 :: Unsigned 6)
+-- 00_0111
+-- >>> split (7 :: Unsigned 6) :: (BitVector 2, BitVector 4)
+-- (00,0111)
+split :: (BitPack a, BitSize a ~ (m + n), KnownNat n) => a
+      -> (BitVector m, BitVector n)
```
alex-mckenna, line 75: Same here

clash-prelude/src/Clash/Class/BitPack/BitIndex.hs
```diff
@@ -0,0 +1,148 @@
...
+-- >>> setSlice d6 d5 0 (-5 :: Signed 6)
+-- <BLANKLINE>
+-- <interactive>:...
+--     • Couldn't match type ‘7 + i0’ with ‘6’
+--         arising from a use of ‘setSlice’
+--       The type variable ‘i0’ is ambiguous
+--     • In the expression: setSlice d6 d5 0 (- 5 :: Signed 6)
+--       In an equation for ‘it’: it = setSlice d6 d5 0 (- 5 :: Signed 6)
+setSlice :: (BitPack a, BitSize a ~ ((m + 1) + i)) => SNat m -> SNat n
+         -> BitVector (m + 1 - n) -> a -> a
```
alex-mckenna, line 119: Another one!

clash-prelude/src/Clash/Class/BitPack/BitReduction.hs
```diff
@@ -0,0 +1,82 @@
+{-|
+Copyright  :  (C) 2013-2016, University of Twente
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
```
alex-mckenna, line 4: Copyright notice + maintainer here as well

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -1,102 +1,441 @@
 {-|
-Copyright  :  (C) 2019, QBayLogic B.V.
+Copyright  :  (C) 2013-2016, University of Twente,
+                  2016-2017, Myrtle Software Ltd
```
alex-mckenna, line 3: Does this copyright notice match up with the one removed in the other file?

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -1,102 +1,441 @@
...
+import Prelude                        hiding (map)
+import System.IO.Unsafe               (unsafeDupablePerformIO)
+
+import Clash.Promoted.Nat             (SNat(..), snatToNum)
+import Clash.Annotations.Primitive    (hasBlackBox)
+import Clash.Class.BitPack.Internal.TH (deriveBitPackTuples)
+import Clash.Class.Resize             (zeroExtend, resize)
+import Clash.Sized.Internal.BitVector
+  (BitVector, Bit, pack#, split#, checkUnpackUndef, undefined#, unpack#,
+   unsafeToNatural, (++#), isLike#)
```
alex-mckenna, line 47: another nitpick: I feel if the diff is going to be all green anyway we may as well drop the padding before the import lists. Saves them from being partially out of line / all changed in a big diff later

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -1,102 +1,441 @@
...
+import Clash.XException
+
+{- $setup
+>>> :set -XDataKinds
+>>> import Clash.Prelude
+-}
+
+-- | Convert to and from a 'BitVector'
```
alex-mckenna, line 55: I wonder if it would be worth expanding on the docs? Give some rationale for users about what types can be converted to / from BV (i.e. why is there no list instance), or why you might even want to do it at all from a Haskell POV?

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -1,102 +1,441 @@
...
+  type BitSize () = 0
+  pack   _ = minBound
+  unpack _ = ()
+
+-- | __N.B.__: The documentation only shows instances up to /3/-tuples. By
+-- default, instances up to and including /12/-tuples will exist. If the flag
+-- @large-tuples@ is set instances up to the GHC imposed limit will exist. The
+-- GHC imposed limit is either 62 or 64 depending on the GHC version.
+instance (BitPack a, BitPack b) =>
+    BitPack (a,b) where
```
alex-mckenna, line 309: Another nitpick: why is there a line break here after the `=>`?

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -1,102 +1,441 @@
...
+
+  gUnpack c cc b =
+    let cLeft = snatToNum (SNat @(GConstructorCount f)) in
+    if c < cc + cLeft then
+      L1 (gUnpack c cc f)
+    else
+      R1 (gUnpack c (cc + cLeft) g)
+
+   where
+    -- It's a thing of beauty, if I may say so myself!
```
alex-mckenna, line 377: :sparkles:

alex-mckenna (changes requested):
Overall I like it. I've been a bit picky here because there's no way to do this without a large diff so it seems like a good time to also fix style issues

clash-prelude/src/Clash/Class/BitPack/Internal.hs
alex-mckenna: Also alphabetize it!

leonschoorl: I feel the padding makes it more readable, even if it isn't all perfectly aligned.

alex-mckenna: I do too, but there are some files where it just gets out of sync. We kind of have a weird mixup of styles for imports in different modules and they all end up getting messy, so I would prefer picking the least-easily messed up option

alex-mckenna: And things like adding a qualified module and having to indent all the other import names, or aligning the import lists then having to realign just makes diffs noisier. We avoid this in other places by not aligning things like alternatives in case expressions (which also makes it less readable IMO) so it seems weird to take that stance in some places and not others

We could also leave it for now and bikeshed it later. I do think we should be more consistent with style though

martijnbastiaan: https://gist.github.com/martijnbastiaan/26be40480c2edcc216ee28d0191bcda7 <= best style IMO (until `qualified` can be used in the post position on all supported ghcs)

martijnbastiaan: which would make it look like https://gist.github.com/martijnbastiaan/8245012dfe0b6a11dca36454ad741c0a

martijnbastiaan: What I dislike most about alignment is that it's not forced, so you end up with garbage-looking imports over time. I wouldn't mind tools enforcing it, but not as a separate compile step. If we're going to do that it needs to be something [like this](https://github.com/ocharles/hlint-source-plugin).

clash-prelude/src/Clash/Class/BitPack/Internal.hs
martijnbastiaan: Yes. This is apart of a rename:

```
BitVector -> Internal
Internal -> TH
```

clash-prelude/src/Clash/Class/BitPack/Internal.hs
alex-mckenna: For the time being I'm leaving this alone. I would like to change it but holding the PR back while we bikeshed style seems counterproductive. I have alphabetised the imports though, and moved the `import Prelude` to the start because it's somewhat "special" as an import

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -52,7 +54,28 @@ import Clash.XException
 >>> import Clash.Prelude
 -}
 
--- | Convert to and from a 'BitVector'
+-- | Convert data to/from a 'BitVector'. This allows functions to be defined
+-- on the underlying representation of data, while exposing a nicer API using
+-- pack / unpack at the boundaries. For example:
```
martijnbastiaan, line 59:
```suggestion
-- 'pack' / 'unpack' at the boundaries. For example:
```

martijnbastiaan (commented):
LGTM

(Can't approve because I opened this PR :-|)

PR #1779, opened 2021-04-20
Account for undefined Vec spines in 'showX'

### LINKED ISSUE #1777: `showX` misbehaves if spine of Vector is `XException`
For example:

```haskell
>>> showX (1 :> errorX "abc" :> 3 :> Nil)
"<1,X,3>"
>>> showX (1 :> 2 :> errorX "def")
"<1,*** Exception: X: def
  errorX, called at <interactive>:3:29 in interactive:Ghci3
```

martijnbastiaan:
Fixes #1777

### DIFF
--- a/clash-prelude/tests/Clash/Tests/Vector.hs
+++ b/clash-prelude/tests/Clash/Tests/Vector.hs
@@ -16,7 +16,7 @@ case_showXVector :: Assertion
 case_showXVector = "<1,2,X" @=? showX (1 :> i 2 :> errorX "def")
 
 case_showX2DVector :: Assertion
-case_showX2DVector = "<<1,X,<3,5" @=? showX ((1 :> errorX "def") :> (3 :> i 5 :> Nil) :> Nil)
+case_showX2DVector = "<<1,X,<3,5>>" @=? showX ((1 :> errorX "def") :> (3 :> i 5 :> Nil) :> Nil)
 
 tests :: TestTree
 tests = testGroup "All"
--- a/clash-prelude/tests/Clash/Tests/Vector.hs
+++ b/clash-prelude/tests/Clash/Tests/Vector.hs
@@ -1,0 +1,33 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Clash.Tests.Vector where
+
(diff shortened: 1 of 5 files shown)


clash-prelude/tests/Clash/Tests/Vector.hs
```diff
@@ -0,0 +1,33 @@
...
+import           Test.Tasty.TH
+
+i :: Int -> Int
+i = id
+
+case_showXVector :: Assertion
+case_showXVector = "<1,2,X" @=? showX (1 :> i 2 :> errorX "def")
+
+case_showX2DVector :: Assertion
+case_showX2DVector = "<<1,X,<3,5" @=? showX ((1 :> errorX "def") :> (3 :> i 5 :> Nil) :> Nil)
```
christiaanb, line 19: Why doesn't this print as:
```
<<1,X,<3,5>>
```

alex-mckenna: I think https://github.com/clash-lang/clash-compiler/pull/1779/files#diff-0983c302bba1b7e92851789687ff742546e7453d6b5b9cf032076e6131fa0599R321, but I was just pulling locally to check

christiaanb: Ah, right, that should probably become:
```haskell
punc (Cons x (isX -> Right Nil)) = showsX x . ('>':)
```

alex-mckenna: I don't think the tests are even being run! When I change the test to:

```haskell
case_showX2DVector = "penguins" @=? showX ((1 :> errorX "def") :> (3 :> i 5 :> Nil) :> Nil) 
```

it still passes. Which seems _slightly_ wrong

martijnbastiaan: Good point Alex, it hasn't been added to the file collecting all tests.

alex-mckenna: It has now :P

alex-mckenna (approved):

PR #1788, opened 2021-04-23
Fix #1786.

### LINKED ISSUE #1786: `~ISACTIVENABLE` doesn't handle `Enable`s after #1368
In #1368 we changed Clash to handle `Enable` lines separately, allow us to prettyprint it in the HDL. We forgot to account for this change in:

https://github.com/clash-lang/clash-compiler/blob/eb48e7e1d2e1ad0a6bc386a6f1d6a0396b5308c8/clash-lib/src/Clash/Netlist/BlackBox/Util.hs#L457-L468

This makes this:

```haskell
module Enable where

import Clash.Explicit.Prelude

top :: Clock System -> Signal System Bool -> Signal System Bool
top clk = delay clk enableGen undefined
```

Generate this:

```verilog
    if (1'b1) begin
      result_1 <= c$arg;
    end
```

while it could have simply omitted the if-statement.

DigitalBrains1:
Some functionality is added: when `~ISACTIVEENABLE` is used on a signal
of type `Signal dom Bool` and it is always true, `~ISACTIVEENABLE` now
evaluates to true as well. Before this commit, it was not an error to
use `~ISACTIVEENABLE` on a `Signal dom Bool`, but it would not be
checked whether it was always true or not.

### DIFF
--- a/changelog/2021-04-23T11_48_42+02_00_fix_1786
+++ b/changelog/2021-04-23T11_48_42+02_00_fix_1786
@@ -1,0 +1,1 @@
+FIXED: ~ISACTIVEENABLE in blackboxes works again, and now acts on `Signal dom Bool` in addition to `Enable dom`. Since [#1368](https://github.com/clash-lang/clash-compiler/pull/1368), enable lines were always generated even if they were known to be always enabled. Fixes [#1786](https://github.com/clash-lang/clash-compiler/issues/1786).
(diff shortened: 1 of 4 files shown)


martijnbastiaan (approved):
Very nice! Clever use of the inline primitives. My first instinct would have been to try and detect the change when synthesizing `register` but this is much better.

PR #1798, opened 2021-05-02
Add SymbiYosys compatible SVA

expipiplus1:
Also add Assume directive for verification

Write a description of your PR here. Briefly describe the solution, taking more care with larger changes to outline the intention of the PR, or any compromises made in the final design. If there is no associated issue, it is helpful to also include the motivation for this PR.

## Still TODO:


clash-lib/src/Clash/Backend/Verilog.hs
```diff
@@ -567,7 +568,8 @@ seq_ (SeqDecl sd) = case sd of
 
 seqs :: [Seq] -> VerilogM Doc
 seqs [] = emptyDoc
-seqs (SeqDecl (TickDecl id_):ds) = "//" <+> stringS id_ <> line <> seqs ds
+seqs (SeqDecl (TickDecl (Comment c)):ds) = "//" <+> stringS c <> line <> seqs ds
```
christiaanb, line 571: Might as well fix this and change it to:
```
seqs (SeqDecl (TickDecl (Comment c)):ds) = comment "//" c <> line <> seqs ds
```

expipiplus1:
Still to do:

- Wrapping all this in `ifdef FORMAL`
- Adding blackboxes for `$prev` `$stable` and all those.

expipiplus1:
I've added guards with `FORMAL`.

`$prev`, `$stable`, `$rose`, `$fell` can all be implemented in user code, so I've not included them.

I think this is ready for a review now.

expipiplus1:
An example:

Here we will check that our blink program only generated primary colors, and
that each of red, green and blue are represented.

- Set up `Blink.hs` and `blink.sby` as below.
- Compile with `clash --systemverilog`
- Run `sby -f blink.sby`
- Observe that it fails to prove that we always output a primary color!

    `SBY 16:12:37 [blink] engine_0.induction: ##   0:00:00  Assert failed in blink: isPrimary`

- Look at the trace using `gtkwave blink/engine_0/trace_induct.vcd`
- Observe that `isPrimary` is not upheld because we turn off the green LED
  before turning on the blue LED!
- Fix the example by changing b to be `b = (>= 2000000) <$> c` instead of using `>`


`Blink.hs`

```haskell
{-# LANGUAGE TypeApplications #-}
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TemplateHaskellQuotes #-}

module Blink where

import           Clash.Prelude
import           Clash.Verification             ( RenderAs(YosysSVA)
                                                , assert
                                                , checkI
                                                , cover
                                                )

{-# ANN topEntity
  (Synthesize
    { t_name   = "blink"
    , t_inputs = [PortName "clk_i"]
    , t_output = PortProduct "" [PortName "led_r", PortName "led_g", PortName "led_b"]
    }) #-}
topEntity :: Clock System -> Signal System (Bool, Bool, Bool)
topEntity clk =
  let rst = resetGen @System
      en  = enableGen
  in  withClockResetEnable clk rst en go

go :: HiddenClockResetEnable dom => Signal dom (Bool, Bool, Bool)
go =
  let c   = counter @2500000
      r   = (< 1000000) <$> c
      g   = ((>= 1000000) <&&> (< 2000000)) <$> c
      b   = (> 2000000) <$> c
      out = bundle (r, g, b)
      p   = isPrimary <$> out
  in  checkI "isPrimary" YosysSVA (assert p)
        . checkI "R" YosysSVA (cover r)
        . checkI "G" YosysSVA (cover g)
        . checkI "B" YosysSVA (cover b)
        $ out

isPrimary :: (Bool, Bool, Bool) -> Bool
isPrimary (r, g, b) =
  (r && not (g || b)) || (g && not (b || r)) || (b && not (r || g))

counter :: (KnownNat n, HiddenClockResetEnable dom) => Signal dom (Index n)
counter = x where x = register 0 (x + 1)

(<&&>) :: Applicative f => f Bool -> f Bool -> f Bool
(<&&>) = liftA2 (&&)
```

`blink.sby`

```sby
[options]
mode prove

[engines]
smtbmc z3

[script]
read -formal -sv blink_types.sv blink.sv
prep -top blink

[files]
systemverilog/Blink.topEntity/blink.sv
systemverilog/Blink.topEntity/blink_types.sv
```

christiaanb:
@kloonbot run_ci a41d61c

martijnbastiaan:
@christiaanb You need to set your team membership to public, @kloonbot can't see you're a privileged user at the moment.

martijnbastiaan:
That'd be great @expipiplus1. The PR looks good in general; no real comments from me except for the changelog. It'd be pretty nice if we could get this working on CI, to prevent regressions. The weather report promises lots of rain tomorrow, so I might look into it then.

martijnbastiaan:
In preparation for adding SymbiYosys to the testsuite I've split up the various testsers (GHDL, Iverilog, ModelSim) into separate modules. This should make it a bit easier to add a new one (SymiYosys). For the next couple of months, I'll be fully dedicated to another project, so I doubt I'll have the time to pick it up. Branch: https://github.com/clash-lang/clash-compiler/tree/symbiyosys.

expipiplus1:
OK, I added a SymbiYosys backend to the test suite (under verification rather than simulation). I tried to make it minimally invasive...

expipiplus1:
x`

> In preparation for adding SymbiYosys to the testsuite I've split up the various testsers (GHDL, Iverilog, ModelSim) into separate modules. This should make it a bit easier to add a new one (SymiYosys). For the next couple of months, I'll be fully dedicated to another project, so I doubt I'll have the time to pick it up. Branch: https://github.com/clash-lang/clash-compiler/tree/symbiyosys.

Thanks! this was very helpful

expipiplus1:
> I might also see about adding the verilog yosys extension too, it's basically the same as this one with slightly different syntax.

In fact it's the same syntax!

expipiplus1:
I don't have a circle CI account so can't see the failures there, but no doubt it's at least partly due to not having `sby` or `z3` available.

Otherwise from my end I think this is ready

alex-mckenna:
> Cannot locate program `sby` in the PATH

Yep, I guess we need yosys + symbiyosys + z3 added to the CI image. I'll patch it today / tomorrow

expipiplus1:
Thanks!

alex-mckenna:
@expipiplus1 This should be testable when #1836 is merged / rebased on

alex-mckenna:
@expipiplus1 still failing from CI unfortuantely. It seems the latest stable yosys release is missing the dffunmap pass. I'll update the CI image to use a newer commit.

alex-mckenna:
CI patched (again) in #1838, sorry for the inconvenience

expipiplus1:
ok, rebased and pushed again.

martijnbastiaan:
@kloonbot run_ci ede6365

alex-mckenna:
Okay, this seems to be passing in the testsuite but failing in `debian-bindist` (looks like the debian package is still being built off an older version of the CI image). Assuming this is ready for review, I think it's more productive to review then I can fix `debian-bindist` after review (so you're not waiting around too long to know if there are any desired changes)

clash-lib/clash-lib.cabal
```diff
@@ -279,6 +279,8 @@ Library
                       Clash.Util.Interpolate
                       Clash.Pretty
 
+                      Clash.Verification.PrettyPrinters
```
alex-mckenna, line 282: `clash-lib` has some other modules just for pretty-printing which are called `Pretty` instead of `PrettyPrinters`. I think it would be good to rename this to be consistent

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -1419,6 +1420,9 @@ inst_ (InstDecl entOrComp libM _ nm lbl gens pms0) = do
 inst_ (BlackBoxD _ libs imps inc bs bbCtx) =
   fmap Just (Ap (column (renderBlackBox libs imps inc bs bbCtx)))
 
+inst_ (ConditionalDecl cond ds) = Just <$>
+  "#ifdef" <+> pretty cond <> line <> insts ds <> line <> "#endif"
+
```
alex-mckenna, line 1425: For my own understanding: does VHDL support conditional compilation like this? Or is this SymbiYosys only and supported there?

alex-mckenna (commented):
Thanks for your work @expipiplus1, this is a really nice looking PR. Just a few small comments, and I'll get the CI fixed for `debian-bindist` tomorrow

clash-lib/src/Clash/Backend/VHDL.hs
expipiplus1: Hmm, not sure actually. For symbiyosys it's up to `ghdl`. With this style though any C preprocessor could be used 

Maybe it's best to use 

```vhdl
if FORMAL generate
   some logic using the optional port
end generate;
```

Although I'm not sure how recent a feature this is.

alex-mckenna: `if ... generate` is supported in VHDL '93 which is what Clash aims to produce, so maybe this is a better option? `FORMAL` is defined as a macro in verilog though, AFAIK VHDL doesn't have macros, so I'm wondering if `ghdl` would be passed anything to set `FORMAL` :thinking: 

I'll see if any of my colleagues with more VHDL experience have any suggestions

martijnbastiaan: Not necessarily more experience, but a TheInternetSays `if ... generate` is a good option for VHDL 93.

Silly question, but, wouldn't we have seen a GHDL error if it's not valid?

alex-mckenna: The tests added in this PR don't test against VHDL, so possibly not. Having had a brief skim of `ghdl`'s command line options I'm not sure there is a way to pass options like this anyway. If there's no way to pass `FORMAL` without the C preprocessor I wonder if it's a good idea, it means you can no longer necessarily pass VHDL generated by Clash directly to a vendor tool (without running it through `cpp` or manually removing the relevant sections of code first).

expipiplus1: TBH symbiyosys with ghdl is pretty clunky. The only example I could find of it doesn't do any conditional compilation: https://github.com/tmeissner/psl_with_ghdl/blob/master/src/psl_next.vhd

expipiplus1: > The tests added in this PR don't test against VHDL, so possibly not. Having had a brief skim of `ghdl`'s command line options I'm not sure there is a way to pass options like this anyway. If there's no way to pass `FORMAL` without the C preprocessor I wonder if it's a good idea, it means you can no longer necessarily pass VHDL generated by Clash directly to a vendor tool (without running it through `cpp` or manually removing the relevant sections of code first).

yeah, that's a good point, perhaps better to just fail compilation with clash, or not emit any vhdl and emit a warning?

alex-mckenna: I think emitting a warning seems more reasonable than throwing out everything. Other parts of the VHDL backend use tracing functions for warnings, so maybe:

```haskell
inst_ ConditionalDecl{} = do
  traceM "WARNING: Conditional compilation is not supported in VHDL, any conditionally compiled code will be discarded."
  return Nothing
```

alex-mckenna: You could also then add another `runTest` for the tests on VHDL only which expects Clash to "fail" with an exit code of 0, and the warning printed to stderr

expipiplus1: Done, I mentioned the condition in the message to for greater clarity.

expipiplus1: > You could also then add another `runTest` for the tests on VHDL only which expects Clash to "fail" with an exit code of 0, and the warning printed to stderr

hmm, this is tricky actually because currently `YosysSVA` already selects PSL in comments for VHDL, and never actually generates any ConditionalDecls.

expipiplus1: (and there's no other way of generating ConditionalDecls for vhdl)

alex-mckenna: Ah, that's fine then. I assumed something in the new tests would already hit the warning

alex-mckenna:
@kloonbot run_ci 4fa2b5d

alex-mckenna:
Sorry for the delay fixing the `debian-bindist` part of the CI @expipiplus1, if you rebase on `master` one final time I can get this merged for you (I'd like to make sure the bindist part of CI is actually fixed and this PR is the only thing that will use `sby` in CI)

expipiplus1:
done, thanks!

alex-mckenna:
@kloonbot run_ci 4e05cb6

alex-mckenna:
Turns out in the spirit of confusion, the job to test the `debian-bindist` wasn't using the same image used to build the bindist. I've pushed (yet another) fix to master, which I really hope passes CI when rebased on. :sweat_smile:

expipiplus1:
No problem.

I think you should have permission to push to this branch too BTW.

martijnbastiaan:
Also, if it's not too much effort @alex-mckenna, I think we should try and move over our Circle runners to GitHub Actions. The closed logs are just meh.

alex-mckenna:
@expipiplus1 
> I think you should have permission to push to this branch too BTW.

You know, it never occurred to me I would have write access to the fork. In that case don't worry about this, I'll fight with CI and get it merged for you :+1: 

@martijnbastiaan 
> I think we should try and move over our Circle runners to GitHub Actions

Well if I'm already experiencing this much DevOps pain, doing DevOps can't be any worse :wink: I'll open a ticket

alex-mckenna:
@kloonbot run_ci b493ff8

alex-mckenna:
@kloonbot run_ci ac8cee8

expipiplus1:
Thanks!

PR #1802, opened 2021-05-04
Spring cleaning transformations

alex-mckenna:
The normalization part of `clash-lib` has grown a lot over the years, and before this PR the transformations module is the largest in the library (at ~2000 lines). This PR starts the process of cleaning up the normalization modules, by grouping related transformations in more manageable submodules.

Anecdotally this has a slight improvement in compile time, as there is less recompilation. However, real gains won't be achieved until the `Util` modules for `Rewrite` and `Normalize` can be broken down further.

## Still TODO:


martijnbastiaan:
LGTM @ general idea. There's a few things a want to do in a proper review, I'll do that later. Also, sorry for the broken runner, I'll fix it.

---------------

I see you format the imports in the new files in a particular way. Given that we've had a number of (dead-ended) discussions on this I'd like to have @christiaanb make a decision so we can put it to rest. @christiaanb I think there are roughly three choices:

1. Align all the things:

```haskell
import qualified Data.HashMap               as HashMap
import           Foo                        (foo) 
import           Some.Very.Long.Module.Name (bar)
```

2. Align modules:

```haskell
import qualified Data.HashMap as HashMap
import           Foo (foo) 
import           Some.Very.Long.Module.Name (bar)
```

3. Align nothing:

```haskell
import qualified Data.HashMap as HashMap
import Foo (foo)
import Some.Very.Long.Module.Name (bar)
```

I think (1) is most pleasing to the eye, but I also think it doesn't scale. Extra / removed imports can cause massive realignments and very long module imports force other imports to insert a lot of whitespace (making it harder to read).

Two separate but related decisions:

* Whether to put package names on top in a comment (e.g., `-- unordered-containers`).
* Whether to force a certain import format using (for example) stylish-haskell - note that it has plugins for all kinds of editors so it shouldn't need to be a separate compile step.

christiaanb:
The way I've grown to feel about import style aesthetics is that it is somewhat of a "lipstick on a pig" situation: it's basically horrible that the first half, and often more, of any file we open is import statements and we have to scroll down before we get to the meat of the module. So we make it look aesthetically pleasing so as not to be reminded of this fact.

So what do I use the import list for as a _reader_ of code: I want to know the origin of function being called in the module. For that, I:
* Use search, either in my editor or web-browser
* Depend on my IDE/Editor to give me the answer using mouse-over/hover

None of this requires aesthetics or alphabetical ordering. So instead I've grown to prefer to optimize for minimal diffs, for which style 3 is the best style: we don't lose 10 characters by having a forced offset for non-qualified imports. This in turn means we don't have to introduce a new line as quickly as we import more functions from a module. Minimal diffs reduces cognitive load during reviewing and makes code archeology more precise using `git blame`.

Also, if you really want aesthetically pleasing alignment, style 3 future-proofs us for [`ImportQualifiedPost`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/import_qualified_post.html) for when we drop pre GHC 8.10 support.

christiaanb:
With regards to package origins of modules, comments can only go out of sync. At best we can use [`PackageImports`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/package_qualified_imports.html).

With regard to forcing a certain style for import statements: See my earlier response; I don't think there's an automated tool that optimizes for minimal diffs.

martijnbastiaan:
Alright, let's go with (3) then. I'll investigate what's possible in terms of tooling.

> So what do I use the import list for as a reader of code: I want to know the origin of function being called within the module. 

Personally, I regularly skim the imports to get a feel of the "scope" of the module. I haven't heard of anyone else doing this, so I'm guessing this is a minor point.

> None of this requires aesthetics or alphabetical ordering. 

To me style of imports is purely psychological: if it's inconsistent or otherwise a mess it ticks me off a bit - as I expect the rest of the module to not be in a better shape. So I'd still advocate for _some_ consistency (alphabetization being one of them).

christiaanb:
> 
> 
> Alright, let's go with (3) then. I'll investigate what's possible in terms of tooling.
> 
> > So what do I use the import list for as a reader of code: I want to know the origin of function being called within the module.
> 
> Personally, I regularly skim the imports to get a feel of the "scope" of the module. I haven't heard of anyone else doing this, so I'm guessing this is a minor point.
>

That's a valid point. But I think grouping and adding comments to each group (e.g. `-- json pretty-print and parsing`, `-- fast lookup containers`, `-- fast mutate containers`, etc.) would help more.
 
> > None of this requires aesthetics or alphabetical ordering.
> 
> To me style of imports is purely psychological: if it's inconsistent or otherwise a mess it ticks me off a bit - as I expect the rest of the module to not be in a better shape. So I'd still advocate for _some_ consistency (alphabetization being one of them).

Yeah, alphabetical order of the modules, plus perhaps some grouping by functionality/origin, makes sense. However, under the minimal diff guideline, the import list for every imported module should be treated as a snoc list for additions made in subsequent commits; i.e. no alphabetic **re**-ordering. 

The only exception to that no **re**-ordering should be when you eliminate swaths of  the import list during a refactor. In that sense, a alphabetical+chronological ordering gives a good indication that a module might need some refactoring: many discontinuous alphabetical orderings => time to refactor. Reordering on every commit brings us into "lipstick on a pig" territory again.

christiaanb:
Lastly, I want to say that I don't want to be dogmatic about minimal diffs either. I just prefer minimal diffs because I tend to review more these days than write new code, or go over complete modules.

martijnbastiaan:
For my understanding, what do you mean with re-ordering? If you start out with an ordered list and enforce that rule, when would you have to re-order?

christiaanb:
So let's say the currently the import list is:

```haskell
import GHC.Core.TyCon
  (AlgTyConRhs (..), TyCon, algTyConRhs, isAlgTyCon, isFamilyTyCon,
   isFunTyCon, isNewTyCon, isPrimTyCon, isTupleTyCon)
```
then when we want to add `expandSynTyCon_maybe` to this list, we add it at the end of the list, i.e.:
```haskell
import GHC.Core.TyCon
  (AlgTyConRhs (..), TyCon, algTyConRhs, isAlgTyCon, isFamilyTyCon,
   isFunTyCon, isNewTyCon, isPrimTyCon, isTupleTyCon, expandSynTyCon_maybe)
```
Why? Because an alphabetic insertion would cause a shuffle because of the 80 character line limit, i.e. it would turn into this two-line change:
```haskell
import GHC.Core.TyCon
  (AlgTyConRhs (..), TyCon, algTyConRhs, expandSynTyCon_maybe, isAlgTyCon, 
   isFamilyTyCon, isFunTyCon, isNewTyCon, isPrimTyCon, isTupleTyCon)
```

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -0,0 +1,675 @@
...
+caseFlat _ e = return e
+{-# SCC caseFlat #-}
+
+collectFlat :: Term -> Term -> Maybe [(Pat,Term)]
+collectFlat scrut (Case (collectEqArgs -> Just (scrut', val)) _ty [lAlt,rAlt])
+  | scrut' == scrut
+  = case collectArgs val of
+      (Prim p,args') | isFromInt (primName p) ->
+        go (last args')
+      (Data dc,args')    | nameOcc (dcName dc) == "GHC.Types.I#" ->
```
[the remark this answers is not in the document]
alex-mckenna: I should probably go through _all_ the new modules and fix the various "broken windows" like these. The diffs are all green anyway so no worse for archeology.

This feels like karma for suggesting the same to @martijnbastiaan in #1775 :thinking:

alex-mckenna:
> I think grouping and adding comments to each group (e.g. -- json pretty-print and parsing, -- fast lookup containers, -- fast mutate containers, etc.) would help more.

FWIW, I don't like this style of commenting over groups of imports. My preference for grouping is simply

```haskell
-- Explicitly importing Prelude, as this overrides an implicit "normal" behaviour

-- modules from external packages, e.g. base, text, containers etc.

-- modules from clash-prelude, if used

-- modules from clash-lib, if used

-- modules from clash-ghc, if used

-- modules from the package itself (if it is something auxiliary that uses clash, e.g. clash-protocols)
```

with modules where reasonable (new modules added in order where the import list is already tidy enough, otherwise at the end if the imports are too disorderly). I hadn't considered adding extra things to an explicit using list for an import, but I agree that makes sense to be snoc only).

> Personally, I regularly skim the imports to get a feel of the "scope" of the module.

I do this too, which is why I like them alphabetised. Makes it easier to skim-read without having to double-take, or miss that a module is imported.

Also, if we reach any conclusion here, we should remember to add it to the style guide with rationale so we don't end up back here in the future.

martijnbastiaan:
I've [played around a bit](https://gitlab.com/martijnbastiaan/styleless-haskell) and all of the mentioned checks would be perfectly possible to enforce with a source plugin (I've only implemented an enforced explicit package import - if we were to go for that). This would mean that the enforcement would be part of normal development loop, instead of either a review or CI roundtrip.

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -74,25 +74,25 @@ constantPropagation =
   dec >->
   conSpec
   where
-    etaTL              = apply "etaTL" etaExpansionTL !-> topdownR (apply "applicationPropagation" appPropFast)
+    etaTL              = apply "etaTL" etaExpansionTL !-> topdownR (apply "applicationPropagation" appProp)
```
martijnbastiaan, line 77: Can we get an `apply$transformation` for every transformation? I.e., the module exporting `appProp` would also export `applyAppProp`. Names go out of sync easily it seems, and its a lot of repetition in this file.

clash-lib/src/Clash/Normalize/Transformations/ANF.hs
```diff
@@ -0,0 +1,334 @@
...
+-- error #24 (https://github.com/christiaanb/clash2/issues/24), where we
+-- try to get the first field out of the 'Vec's 'Nil' constructor.
+--
+-- Ultimately we should stop treating Signal as a "transparent" type and deal
+-- handling of the Signal type, and the involved co-recursive functions,
+-- properly. At the moment, Clash cannot deal with this recursive type and the
+-- recursive functions involved, hence the need for special-casing code. After
+-- everything is done properly, we should remove the two lines below.
+collectANF _ e@(Case _ _ [(DataPat dc _ _,_)])
+  | nameOcc (dcName dc) == "Clash.Signal.Internal.:-" = return e
```
martijnbastiaan, line 233:
```suggestion
  | nameOcc (dcName dc) == show '(:-) = return e
```

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -0,0 +1,673 @@
...
+   where
+    isNum0 (tyView -> TyConApp (nameOcc -> tcNm) [arg])
+      | tcNm `elem`
+        ["Clash.Sized.Internal.BitVector.BitVector"
+        ,"Clash.Sized.Internal.Unsigned.Unsigned"
+        ,"Clash.Sized.Internal.Signed.Signed"
+        ]
+      = isLitX 0 arg
+      | tcNm ==
+        "Clash.Sized.Internal.Index.Index"
```
martijnbastiaan, line 311: ```show ''Type```

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -0,0 +1,673 @@
...
+
+  | primName pr == "Clash.Transformations.eqInt"
+  , [Left scrut,Left val] <- args
+  = Just (mkTicks scrut ticks,val)
+ where
+  clashNumEqs =
+    [ "Clash.Sized.Internal.Index.eq#"
+    , "Clash.Sized.Internal.Signed.eq#"
+    , "Clash.Sized.Internal.Unsigned.eq#"
+    ]
```
martijnbastiaan, line 557: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -0,0 +1,673 @@
...
+collectEqArgs (collectArgsTicks -> (Prim pr, args, ticks))
+  | primName pr == "Clash.Sized.Internal.BitVector.eq#"
+  , [_, _, Left scrut, Left val] <- args
+  = Just (mkTicks scrut ticks, val)
+
+  | primName pr `elem` clashNumEqs
+  , [_, Left scrut, Left val] <- args
+  = Just (mkTicks scrut ticks, val)
+
+  | primName pr == "Clash.Transformations.eqInt"
```
martijnbastiaan, line 549: ```show 'term```

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -0,0 +1,673 @@
...
+    isTrueDcPat (DataPat p _ _)
+      = ((== "GHC.Types.True") . nameOcc . dcName) p
+    isTrueDcPat _ = False
+
+collectFlat _ _ = Nothing
+{-# SCC collectFlat #-}
+
+collectEqArgs :: Term -> Maybe (Term,Term)
+collectEqArgs (collectArgsTicks -> (Prim pr, args, ticks))
+  | primName pr == "Clash.Sized.Internal.BitVector.eq#"
```
martijnbastiaan, line 541: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -0,0 +1,673 @@
...
+          where
+            -- Substitute solution in existentials and applied types
+            is3   = extendInScopeSetList is2 exts0
+            xs1   = map (substTyInVar (extendTvSubstList (mkSubst is3) sols)) xs0
+            exts1 = substInExistentialsList is2 exts0 sols
+
+            -- Substitute solution in term.
+            is4       = extendInScopeSetList is3 xs1
+            subst     = extendTvSubstList (mkSubst is4) sols
+            term1     = substTm "Replacing tyVar due to solved eq" subst term0
```
martijnbastiaan, line 668: I'm not going to comment on style too much, but this is just weird :-)

clash-lib/src/Clash/Normalize/Transformations/DEC.hs
```diff
@@ -27,79 +25,217 @@
...
+import Data.Monoid (All (..))
+import qualified Data.Text as Text
+import GHC.Stack (HasCallStack)
 
 #if MIN_VERSION_ghc(8,10,0)
-import           GHC.Hs.Utils                     (chunkify,mkChunkified)
+import GHC.Hs.Utils (chunkify,mkChunkified)
 #else
-import           HsUtils                          (chunkify,mkChunkified)
+import HsUtils (chunkify,mkChunkified)
```
martijnbastiaan, line 61: Commenting because you're changing it for style anyway: could you insert spaces between the commas?

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -0,0 +1,638 @@
...
+            | TyConApp (nameOcc -> "Clash.Explicit.SimIO.SimIO") _ <- tyView aTy
+            -> True
+          _ -> False
+      | id_ `notElemVarSet` bodyFVs
+      = case tm of
+          Prim pInfo
+            | primName pInfo `elem`
+                [ "Clash.Explicit.SimIO.openFile"
+                , "Clash.Explicit.SimIO.fgetc"
+                , "Clash.Explicit.SimIO.feof"
```
martijnbastiaan, line 355: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -0,0 +1,638 @@
...
+            , TExpr <- kind p
+            , Just occ <- lookupVarEnv id_ allOccs
+            , occ < 2
+            -> True
+            | otherwise
+            -> primName pInfo `elem` ["Clash.Explicit.SimIO.bindSimIO#"]
+          Case _ _ [_] -> True
+          Data _ -> True
+          Case _ aTy (_:_:_)
+            | TyConApp (nameOcc -> "Clash.Explicit.SimIO.SimIO") _ <- tyView aTy
```
martijnbastiaan, line 346: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -0,0 +1,638 @@
...
+      = case tm of
+          Prim pInfo
+            | let nm = primName pInfo
+            , Just (extractPrim -> Just p@(BlackBox {})) <- HashMap.lookup nm prims
+            , TExpr <- kind p
+            , Just occ <- lookupVarEnv id_ allOccs
+            , occ < 2
+            -> True
+            | otherwise
+            -> primName pInfo `elem` ["Clash.Explicit.SimIO.bindSimIO#"]
```
martijnbastiaan, line 342: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -0,0 +1,638 @@
...
+            | primName pInfo `elem`
+                [ "Clash.Explicit.SimIO.openFile"
+                , "Clash.Explicit.SimIO.fgetc"
+                , "Clash.Explicit.SimIO.feof"
+                ]
+            , Just occ <- lookupVarEnv id_ allOccs
+            , occ < 2
+            -> True
+            | otherwise
+            -> primName pInfo `elem` ["Clash.Explicit.SimIO.bindSimIO#"]
```
martijnbastiaan, line 361: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -0,0 +1,638 @@
...
+          Case _ _ [(DataPat dcE _ _,_)]
+            -> let nm = (nameOcc (dcName dcE))
+               in -- Inlines WW projection that exposes internals of the BitVector types
+                  nm == "Clash.Sized.Internal.BitVector.BV"  ||
+                  nm == "Clash.Sized.Internal.BitVector.Bit" ||
+                  -- Inlines projections out of constraint-tuples (e.g. HiddenClockReset)
+                  "GHC.Classes" `Text.isPrefixOf` nm
+          Case _ aTy (_:_:_)
+            | TyConApp (nameOcc -> "Clash.Explicit.SimIO.SimIO") _ <- tyView aTy
```
martijnbastiaan, line 370: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -0,0 +1,425 @@
...
+    usedArgs0 =
+      case Monad.join (extractPrim <$> bbM) of
+        Just (BlackBoxHaskell{usedArguments}) ->
+          case usedArguments of
+            UsedArguments used -> Just used
+            IgnoredArguments ignored -> Just ([0..length args - 1] \\ ignored)
+        Just (BlackBox pNm _ _ _ _ _ _ _ _ _ inc r ri templ) -> Just $
+          if | isFromInt pNm -> [0,1,2]
+             | primName pInfo `elem` [ "Clash.Annotations.BitRepresentation.Deriving.dontApplyInHDL"
+                                     , "Clash.Sized.Vector.splitAt"
```
martijnbastiaan, line 99: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -0,0 +1,425 @@
...
+
+    go _ _ _ [] = return []
+    go tcm !n used (Right ty:args') = do
+      args'' <- go tcm n used args'
+      return (Right ty : args'')
+    go tcm !n used (Left tm : args') = do
+      args'' <- go tcm (n+1) used args'
+      case tm of
+        TyApp (Prim p0) _
+          | primName p0 == "Clash.Transformations.removedArg"
```
martijnbastiaan, line 130: Add this as a real thing and `show 'term`?

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -0,0 +1,425 @@
...
+             else return  (Left tm : args'')
+
+removeUnusedExpr _ e@(Case _ _ [(DataPat _ [] xs,altExpr)])
+  | xs `localIdsDoNotOccurIn` altExpr = changed altExpr
+  | otherwise = return e
+
+-- Replace any expression that creates a Vector of size 0 within the application
+-- of the Cons constructor, by the Nil constructor.
+removeUnusedExpr _ e@(collectArgsTicks -> (Data dc, [_,Right aTy,Right nTy,_,Left a,Left nil],ticks))
+  | nameOcc (dcName dc) == "Clash.Sized.Vector.Cons"
```
martijnbastiaan, line 146: `show 'term`

clash-lib/src/Clash/Normalize/Transformations/XOptimize.hs
```diff
@@ -0,0 +1,157 @@
...
+xOptimize :: HasCallStack => NormRewrite
+xOptimize (TransformContext is0 _) e@(Case subj ty alts) = do
+  runXOpt <- Lens.view aggressiveXOpt
+
+  if runXOpt then do
+    defPart <- List.partitionM (isPrimError . snd) alts
+
+    case defPart of
+      ([], _)    -> return e
+      (_, [])    -> changed (Prim (PrimInfo "Clash.XException.errorX" ty WorkConstant SingleResult))
```
martijnbastiaan, line 74: `show 'term`

martijnbastiaan (changes requested):

clash-lib/src/Clash/Normalize/Transformations/ANF.hs
alex-mckenna: > `show '(:-)`

Sorcery!

alex-mckenna: I didn't know you could do that though, _neat_

clash-lib/src/Clash/Normalize/Transformations/Case.hs
alex-mckenna: Yes, I missed those aligned `=`. Spend enough time reading these files and your eyes slowly glaze over

clash-lib/src/Clash/Normalize/Transformations/Case.hs
alex-mckenna: Not this one. The function doesn't really exist

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
alex-mckenna: This one I can't do: `bindSimIO#` is not imported, and there isn't a `Clash.Explicit.SimIO.Internal` module or anything

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
alex-mckenna: Same as above, can't do this one

alex-mckenna:
w.r.t. the things I haven't addressed:

  * There's some cleanup I want to do to `Strategy` but I haven't fully got it figured out yet, so I've left out the suggestion about `apply` for now
  * The "special" primitives namespaced under "Clash.Transformations" don't really exist. I think it would better to in another PR make a module that defines these primitives, and then change the rest of the codebase to import and use those `PrimInfo` instead. The `PrimInfo` can be given the same name as the primitive and their names can be changed to fit with the new module (i.e. change `Clash.Transformations.eqInt` everywhere it appears to `Clash.Normalize.Primitives.eqInt` or something similar). Having a new module for this also makes it a lot easier to document because AFAIK the existence of these primitives isn't documented _anywhere_
  * Making a `SimIO.Internal` module also feels outside the scope of this PR, so I would like to do it in another PR as well
  * I couldn't figure out how to get the names inside patterns to work with this quoting trick. If you have any more wisdom for me @martijnbastiaan, I'd appreciate it

martijnbastiaan:
> I couldn't figure out how to get the names inside patterns to work with this quoting trick. If you have any more wisdom for me @martijnbastiaan, I'd appreciate it

Perhaps:

```haskell
          Case _ aTy (_:_:_)
            | TyConApp nm _ <- tyView aTy
            , nameOcc nm == show ''SimIO
```

Agreed on all the other points!

alex-mckenna:
Alright, changed those last two in the view patterns, you can set auto-merge if you're happy @martijnbastiaan

martijnbastiaan:
> Edit: Maybe not, appears I've broken something somewhere

I'm betting it's the brittleness of hardcoded function names :smile:

PR #1804, opened 2021-05-06
Identify recursive groups of global binders

### LINKED ISSUE #1796: Certain test bench inlining causes Clash to enter infinite loop
I could not reproduce the following issue using "plain" Clash invocation. Rather, they occur when using a Clash starter project with cabal. I've created a starter project with this code at https://github.com/DigitalBrains1/testbenchinline.git for your convenience. Note that it expects current Clash master git repository at `../clash-compiler`. My ghc version is 8.8.4.

This code:
```haskell
module TestBenchInline where

import Clash.Prelude

topEntity
  :: Signal System Bool
  -> Signal System Bool
topEntity = id
{-# NOINLINE topEntity #-}

tb
  :: Signal System Bool
tb = done
  where
    testInput = stimuliGenerator clk rst (singleton True)
    expectOutput = outputVerifier' clk rst (singleton True)
    done = expectOutput $ topEntity testInput
    clk = tbClockGen @System (not <$> done)
    rst = resetGen @System
{-# INLINE tb #-}

testBench = tb
-- {-# NOINLINE testBench #-}
```
works. Reproduction:
```sh
$ git clone https://github.com/DigitalBrains1/testbenchinline.git
$ cd testbenchinline
$ cabal update
$ cabal run clash -- --vhdl TestBenchInline
```

However, either removing the `INLINE` on `tb` or adding the `NOINLINE` to `testBench` will cause Clash to go into an endless loop:

```
$ cabal run -- --vhdl TestBenchInline
Up to date
Loaded package environment from /home/peter/src/clash/testbenchinline/.ghc.environment.x86_64-linux-8.8.4
GHC: Parsing and optimising modules took: 1.001s
GHC: Loading external modules from interface files took: 0.000s
GHC: Parsing annotations took: 0.002s
Clash: Parsing and compiling primitives took 0.163s
GHC+Clash: Loading modules cumulatively took 1.438s
Clash: Compiling TestBenchInline.topEntity
Clash: Normalization took 0.002s
Clash: Compiling TestBenchInline.testBench
```
and there it stays looping and eating gradually more and more memory.

This also occurs with Clash 1.4.1.

alex-mckenna:
When loading modules, makeRecursiveGroups now also makes edges for
mutually recursive global binders. Fixes #1796.

## Still TODO:


martijnbastiaan (changes requested):
LGTM. Could you copy the test case @DigitalBrains1 provided in https://github.com/clash-lang/clash-compiler/issues/1796?

christiaanb:
Test case won't work in our current testsuite. I needs to be compiled in another package and loaded from the interface file.

martijnbastiaan:
We could add it to testsuite-the-lib and load it from the interface file directly. I think this should work without any changes, but we'll have to see. I do propose adding all the tests loaded from modules to a separate test-directory to prevent any confusion.

martijnbastiaan:
@alex-mckenna Modules are currently compiled by listing their file on the command line:

https://github.com/clash-lang/clash-compiler/blob/0bc08c17cebc4a459877a565b33cff714636b116/testsuite/src/Test/Tasty/Clash.hs#L795

There's two ways of fixing this up to support precompiled libraries:

* The Proper Way: rename _all_ existing test benches to have a fully qualified module name. I.e., [`T1187.hs`](https://github.com/clash-lang/clash-compiler/blob/dd80f7313fa30f71348be10fe995790c8d891276/tests/shouldwork/Issues/T1187.hs) is now called `T1187` but would then be called `Issues.T1187`. You could then change the call to Clash to include `-itests/shouldwork`, make it not  call a file but a module, and it should all work out of the box.
* The Quick Way: Add an option to `TestOptions`: https://github.com/clash-lang/clash-compiler/blob/0bc08c17cebc4a459877a565b33cff714636b116/testsuite/src/Test/Tasty/Clash.hs#L107-L130. Case on it in https://github.com/clash-lang/clash-compiler/blob/0bc08c17cebc4a459877a565b33cff714636b116/testsuite/src/Test/Tasty/Clash.hs#L795 to decide whether to call a file or a module.

DigitalBrains1:
Please test in isolation first if this actually works for #1796, because I might have been stupid but I could not get it to work (i.e., produce the problem).

alex-mckenna:
I know anecdotally it works from recreating the big (with your repo) locally. I have the original example as the test (at least locally) so I'll double check the test fails without the fix

testsuite/src/Test/Tasty/Clash.hs
```diff
@@ -269,15 +273,17 @@ instance IsTest ClashTest where
...
-      [ target
-      , ctSourceDirectory </> ctModName <.> "hs"
-      , "-i" ++ ctSourceDirectory
-      , "-fclash-hdldir", oDir
+      target : modFlags <>
+      [ "-fclash-hdldir", oDir
       , "-odir", oDir
       , "-hidir", oDir
       , "-fclash-debug", "DebugSilent"
-      , "-package", "clash-testsuite"
```
martijnbastiaan, line 280: Could you see if Stack needs this line by any chance? So something like `stack run testsuite -- -p T1796`.

testsuite/src/Test/Tasty/Clash.hs
```diff
@@ -269,15 +273,17 @@ instance IsTest ClashTest where
         (testExitCode testExit) "clash" (args oDir) NoGlob PrintNeither False
         (specificExitCode testExit) (ExpectStdErr expectedErr) Nothing
 
+    modFlags = if ctLoadFromInterface
+                 then [ctModName]
+                 else ["-i" <> ctSourceDirectory
+                      , ctSourceDirectory </> ctModName <.> "hs"]
```
martijnbastiaan, line 279: I don't think you even need the switch. Could you test if the following works:

1. Always use the `-i` stuff
2. Always call a module by its name, not its file

Loading from interface files should take precedence over loading from local files. If that works we could simply remove the flag from the options record.

alex-mckenna: Actually scratch that

martijnbastiaan (approved):
Could you move the `tests/shouldwork/LoadModules/T1796.hs` to `tests/shouldwork/LoadModules/Tests/Issues/T1796.hs`? It's a bit of a mouthful now, but this should probably become the default way to write tests:

* We can share common test functionality now (yay)
* Our IDEs can give hints thanks to a proper project structure / HLS
* GHC only has to run once for Verilog/VHDL/SystemVerilog
* API changes will be way more comfortable if we don't have to run the testsuite - instead relying on HLS/`cabal build`/`ghcid`

PR #1823, opened 2021-05-19
Fix floating point literal alpha equivalence

DigitalBrains1:
When we switched to using `Float` and `Double` in our
`Clash.Core.Literal` data type in #1811, our `Eq` and `Hashable`
instances for `Literal` were made incorrect. We need floating point
literals `x` and `y` to be equal iff `pack x == pack y`.

We already assume that `Float` is single-precision IEEE-754 and
`Double` is double-precision IEEE-754, so we might as well represent
them as 32/64-bits words to fix equality and hashing.

## Still TODO:


clash-lib/src/Clash/Core/Literal.hs
```diff
@@ -42,8 +43,8 @@ data Literal
   | Int64Literal    !Integer
   | Word64Literal   !Integer
   | StringLiteral   !String
-  | FloatLiteral    !Float
-  | DoubleLiteral   !Double
+  | FloatLiteral    !Word32
+  | DoubleLiteral   !Word64
```
alex-mckenna, line 47: I think you should add a note about why we use `Word{32,64}` here instead of (1) `Rational` and (2) `Float`/`Double`.
Notes typically look something like this in source:

```haskell
{-
Note [name of note]

This is a note. It notes something.
-}
```

alex-mckenna (approved):
LGTM, needs an implementation note in `Clash.Core.Literal` for traceability though

clash-lib/src/Clash/Core/Literal.hs
DigitalBrains1: Ah, right, that line of source code only triggers on old versions of GHC, that's why it didn't show up for me.

I'll add the code comment tomorrow-ish.

PR #1825, opened 2021-05-20
Don't render negative shift/rotate in VHDL

### LINKED ISSUE #1810: VHDL shift_right shift amount must be a natural:
As can be seen on https://groups.google.com/g/clash-language/c/CQTJhkDrb1Q/m/6YcY0Ur-AAAJ GHDL raises an error when the shift amount to the `shift_right` call is a negative number. As we can see in:

https://github.com/ghdl/ghdl/blob/c2b3f93dfa40ee92de53b54e38e1819cfc33cb04/libraries/ieee/numeric_std.vhdl#L556-L560
```vhdl
  -- Id: S.2
  function SHIFT_RIGHT (ARG: UNSIGNED; COUNT: NATURAL) return UNSIGNED;
  -- Result subtype: UNSIGNED(ARG'LENGTH-1 downto 0)
  -- Result: Performs a shift-right on an UNSIGNED vector COUNT times.
  --         The vacated positions are filled with '0'.
  --         The COUNT rightmost elements are lost.
```

and

https://github.com/ghdl/ghdl/blob/c2b3f93dfa40ee92de53b54e38e1819cfc33cb04/libraries/ieee/numeric_std.vhdl#L570-L574
```vhdl
  function SHIFT_RIGHT (ARG: SIGNED; COUNT: NATURAL) return SIGNED;
  -- Result subtype: SIGNED(ARG'LENGTH-1 downto 0)
  -- Result: Performs a shift-right on a SIGNED vector COUNT times.
  --         The vacated positions are filled with the leftmost
  --         element, ARG'LEFT. The COUNT rightmost elements are lost.
```
the shift count must be a natural number.

However, the default implementation for the `shift` method of the `Bits` class is: http://hackage.haskell.org/package/base-4.14.1.0/docs/src/Data.Bits.html#shift
```haskell
    x `shift`   i | i<0       = x `shiftR` (-i)
                  | i>0       = x `shiftL` i
                  | otherwise = x
```
which gets translated to concurrent assignments by clash, i.e. that `shift` method will get translated to something like the following VHDL:
```vhdl
alt1 <= shift_right(x,-i);
alt2 <= shift_left(x,i);
res <= alt1 when (i < 0) else
       alt2 when (i > 0) else
       x;
```

The solution is to create primitives for the `shift` instance methods which we can control: `Signed`, `Unsigned`, `Bit` and `BitVector`; so that we can generate the following VHDL instead:

```vhdl
process (x,i) begin
  if (i < 0) then
    res <= shift_right(x,-i);
  elsif (i > 0) then
    res <= shift_left(x,i)
  else
    res <= x;
  end if;
end;
```

ensuring that the `shift_right` is not evaluated concurrently to the `shift_left` when `i > 0`.

alex-mckenna:
This PR stops clash rendering calls to the shift and rotate functions in numeric_std with negative amounts to shift/rotate by. This should prevent errors in simulation from unexpected negative shift amounts.

Fixes #1810 

## Still TODO:

### DIFF
--- a/clash-lib/prims/vhdl/GHC_Num_Integer.primitives
+++ b/clash-lib/prims/vhdl/GHC_Num_Integer.primitives
@@ -189,16 +189,26 @@ end block;
   }
 , { "BlackBox" :
     { "name"      : "GHC.Num.Integer.integerShiftR#"
-    , "kind"      : "Expression"
-    , "type"      : "integerShiftR :: Integer -> Word# -> Integer"
-    , "template"  : "shift_right(~ARG[0],to_integer(~ARG[1]))"
+    , "kind"      : "Declaration"
+    , "type"      : "integerShiftR# :: Integer -> Word# -> Integer"
+    , "template"  :
+"~RESULT <= shift_right(~ARG[0], to_integer(~VAR[count][1](30 downto 0)))
+    -- pragma translate_off
+    when (~ARG[1] >= 0) else (others => 'X')
+    -- pragma translate_on
+    ;"
     }
   }
 , { "BlackBox" :
     { "name"      : "GHC.Num.Integer.integerShiftL#"
-    , "kind"      : "Expression"
-    , "type"      : "integerShiftR :: Integer -> Word# -> Integer"
-    , "template"  : "shift_left(~ARG[0],to_integer(~ARG[1]))"
+    , "kind"      : "Declaration"
+    , "type"      : "integerShiftL# :: Integer -> Word# -> Integer"
+    , "template"  :
+"~RESULT <= shift_left(~ARG[0], to_integer(~VAR[count][1](30 downto 0)))
+    -- pragma translate_off
+    when (~ARG[1] >= 0) else (others => 'X')
+    -- pragma translate_on
+    ;"
     }
   }
 , { "BlackBox" :
--- a/changelog/2021-05-20T14_43_21+02_00_fix_negative_shift_rotate.md
+++ b/changelog/2021-05-20T14_43_21+02_00_fix_negative_shift_rotate.md
@@ -1,0 +1,1 @@
+FIXED: Clash no longer generates calls to {shift,rotate}_{left,right} in VHDL where the count is a negative number. [#1810](https://github.com/clash-lang/clash-compiler/issues/1810).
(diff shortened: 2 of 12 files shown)


alex-mckenna:
Somewhat disappointingly, the test doesn't normalize when using `integer-gmp`:

```
Unmatchable constant as case subject: GHC.Integer.Type.zeroBigNat
WHNF is: GHC.Integer.Type.zeroBigNat
Unmatchable constant as case subject: GHC.Integer.Type.oneBigNat
WHNF is: GHC.Integer.Type.oneBigNat
Unmatchable constant as case subject: GHC.Integer.Type.czeroBigNat
WHNF is: GHC.Integer.Type.czeroBigNat
Unmatchable constant as case subject: GHC.Integer.Type.nullBigNat
WHNF is: GHC.Integer.Type.nullBigNat
```

I guess these need evaluator rules before this can pass CI...

christiaanb:
I once added rules for those. It leads to severe performance degradation because it lead Clash to evaluating branches which will later be pruned. I do not recommend adding evaluator rules for those constants at this point.

christiaanb:
My suggestion: only do the shift/rotets tests on `Natural` on GHC 9+

clash-lib/prims/vhdl/GHC_Num_Integer.primitives
```diff
@@ -189,16 +189,28 @@ end block;
...
-    , "kind"      : "Expression"
-    , "type"      : "integerShiftR :: Integer -> Word# -> Integer"
-    , "template"  : "shift_right(~ARG[0],to_integer(~ARG[1]))"
+    , "kind"      : "Declaration"
+    , "type"      : "integerShiftR# :: Integer -> Word# -> Integer"
+    , "template"  :
+"-- INTEGER_SHIFT_R
+~RESULT <= shift_right(~ARG[0], to_integer(~VAR[count][1](30 downto 0)))
+    -- pragma translate_off
+    when (~ARG[1] >= to_unsigned(0, ~SIZE[~TYP[1]]-1)) else (others => 'X')
```
christiaanb, line 198:
```suggestion
    when (~ARG[1] >= 0) else (others => 'X')
```
When we were debugging this, I didn't really know what was going on. But given that just comparing against the `integer` 0 seems to work for the other primitives, let's simplify the ones that shift by a `Word#` amount.

christiaanb (approved):
LGTM, just simplify the shift primitives that shift by a `Word#` amount a little more.

PR #1833, opened 2021-05-28
Remove reducers from clash-lib

alex-mckenna:
The only part of reducers that is used in clash-lib is the `Mon` type, which is simply the `Ap` type from Data.Monoid (but from a time before `Ap` existed / had all the instances it has now).

Given that what will become the 1.6 branch already has some spring cleaning PRs merged, this seems a good point to do this.

## Still TODO:


christiaanb (approved):

martijnbastiaan:
Is this API compatible? I know one of our clients is going to notice this change, so a changelog entry under "internal changes" would be nice :)

christiaanb:
I guess since we had to change some code in `clash-cores` it will affect other people writing primitives as well.

alex-mckenna:
> Is this API compatible?

Functionally yes, but it may upset people writing their own primitives. I'll add an internal changelog entry for traceability in a new PR.

Issue #1766, opened 2021-04-14
Clash generates duplicate singal names


martijnbastiaan:
I've got a fairly small example that triggers Clash to generate duplicate signal names:

```haskell
module Knightrider where

import Clash.Class.Counter
import Clash.Explicit.Prelude

modulate :: Unsigned 4 -> Unsigned 4 -> (Unsigned 4, Bool)
modulate counter threshold = (counter + 1, counter < threshold)

modulateVec :: Unsigned 4 -> Vec 8 (Unsigned 4) -> (Unsigned 4, Vec 8 Bool)
modulateVec counter thresholds = (counter + 1, map (counter <) thresholds)

distance :: (Ord a, Num a) => a -> a -> a
distance a b = if a > b then a - b else b - a

brightness :: Unsigned 3 -> Unsigned 3 -> Unsigned 4
brightness epoch led =
  maxBound - satMul SatBound distanceToEpoch stepSize
 where
  distanceToEpoch = distance (extend epoch) (extend led)
  stepSize = 4 -- 2^4 / 4

slope :: Unsigned 3 -> Vec 8 (Unsigned 4)
slope epoch = map (brightness epoch) (iterateI (+1) 0)

knightrider ::
  Clock System ->
  Reset System ->
  Enable System ->
  Signal System (BitVector 8)
knightrider clk rst ena = pack <$> leds
 where
  leds = mealy clk rst ena modulateVec 0 (slope <$> ledPos)

  initial :: (Unsigned 3, Index 10)
  initial = (0, 0)

  (ledPos, delayCount) =
      unbundle
    $ register clk rst ena initial
    $ fmap countSucc
    $ bundle (ledPos, delayCount)

knightriderTop ::
  Clock System ->
  Signal System Bool ->
  Signal System (BitVector 8)
knightriderTop clk rstBool = knightrider clk rst enableGen
 where
  rst = unsafeFromLowPolarity rstBool
```

Compile as:

```
clash -- Test.hs --verilog -main-is knightriderTop
```

Note that

```
clash -- Test.hs --verilog -main-is knightrider
```

doesn't exhibit the behavior. Signal declarations in the Verilog:

```verilog
  // Test.hs:32:1-11
  wire  rst;
  wire [7:0] result;
  wire [7:0] c$app_arg;
  wire [7:0] c$ds_app_arg;
  // Test.hs:12:1-11
  reg [3:0] counter = 4'd0;
  wire [31:0] result;
[...]

PR #1837, opened 2021-06-02
flattenLet nested letrec; error on unexpected change

alex-mckenna:
The flattenLet transformation now also flattens nested letrec
expressions, by deshadowing and merging the bindings into a single
expression.

The error for when setChanged is not called was never being called,
this is now fixed so that transformations that change the core
without declaring throw an error at runtime.

Fixes: #1766 

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
@@ -70,7 +70,7 @@ import Clash.Primitives.Types (Primitive(..), UsedArguments(..))
 import Clash.Rewrite.Types
   (TransformContext(..), bindings, curFun, extra, tcCache, workFreeBinders)
 import Clash.Rewrite.Util
-  (changed, isFromInt, isUntranslatable, mkTmBinderFor, removeUnusedBinders)
+  (changed, isFromInt, isUntranslatable, mkTmBinderFor, removeUnusedBinders, setChanged)
 import Clash.Rewrite.WorkFree
 import Clash.Unique (lookupUniqMap)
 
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -162,9 +162,8 @@ apply = \s rewrite ctx expr0 -> do
       isRelevantTrans = s `Set.member` dbgTranss || Set.null dbgTranss
   traceIf (isTryLvl && isRelevantTrans) ("Trying: " ++ s) (pure ())
 
-  (expr1,anyChanged) <- Writer.listen (rewrite ctx expr0)
+  !(expr1,anyChanged) <- Writer.listen (rewrite ctx expr0)
```
christiaanb, line 165:
```suggestion
  (!expr1,anyChanged) <- Writer.listen (rewrite ctx expr0)
```

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -176,6 +176,14 @@ removeUnusedExpr _ e = return e
 --
 -- NB: must only be called in the cleaning up phase.
 flattenLet :: HasCallStack => NormRewrite
+flattenLet ctx@(TransformContext is0 _) (Letrec binds0 body0@Letrec{}) = do
+  -- deshadow binds1, so nothing bound in binds0 is bound
```
christiaanb, line 180:
```suggestion
  -- deshadow binds1, so binds0 and binds1 don't conflict when merged
```

christiaanb (changes requested):
Could we make this two commits, one to fix `apply` and the other to fix `flattenLet`. It can still stay in one PR though.

christiaanb:
Could you also add the `fixes #1766` to the `flattenLet` commit message?

christiaanb (approved):

PR #1840, opened 2021-06-03
New function: memFile

DigitalBrains1:
A helper function for producing files with memory contents for such
modules as Clash.Explicit.BlockRam.File, etcetera.

I know you agreed to my initial proposal of

```haskell
showsBitPackFile
  :: BitPack a
  => a
  -> ShowS

-- used as
bla file es =
  writeFile file (foldr showsBitPackFile "" es)
```

but I changed my mind. The point of `ShowS` is that you can combine it with other datatypes that have a `Show` instance. But the memory file only contains words of a single length `n`. If people really want to put multiple datatypes into one memory, they can just `pack` it into `BitVector n`. So the function doesn't need to be combined in ways as `ShowS`, and you would in practice always invoke it as `foldr showsBitPackFile ""`, so why not incorporate that directly so people don't have to write it.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/BlockRam/File.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam/File.hs
@@ -1,9 +1,10 @@
 {-|
 Copyright  :  (C) 2015-2016, University of Twente,
-                  2017     , Google Inc.
-                  2019     , Myrtle Software Ltd
+                  2017     , Google Inc.,
+                  2019     , Myrtle Software Ltd,
+                  2021     , QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
-Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
 = Initializing a BlockRAM with a data file #usingramfiles#
 
@@ -29,6 +30,12 @@ For example, a data file @memory.bin@ containing the 9-bit unsigned number
 000001011
 000001100
 000001101
+@
+
+Such a file can be produced with 'memFile':
+
+@
+writeFile "memory.bin" (memFile Nothing [7 :: Unsigned 9 .. 13])
 @
 
 We can instantiate a BlockRAM using the content of the above file like so:
@@ -87,12 +94,15 @@ module Clash.Explicit.BlockRam.File
   ( -- * BlockRAM synchronized to an arbitrary clock
     blockRamFile
   , blockRamFilePow2
+    -- * Producing files
+  , memFile
     -- * Internal
   , blockRamFile#
   , initMem
   )
 where
 
+import Data.Bits             ((.&.), (.|.), shiftL, xor)
 import Data.Char             (digitToInt)
 import Data.Maybe            (isJust, listToMaybe)
 import qualified Data.Sequence as Seq
@@ -101,13 +111,17 @@ import GHC.TypeLits          (KnownNat)
 import Numeric               (readInt)
 import System.IO.Unsafe      (unsafePerformIO)
 
-import Clash.Promoted.Nat    (SNat (..), pow2SNat)
-import Clash.Sized.BitVector (BitVector)
+import Clash.Class.BitPack   (BitPack, BitSize, pack)
+import Clash.Promoted.Nat    (SNat (..), pow2SNat, natToNum)
+import Clash.Sized.Internal.BitVector (Bit(..), BitVector(..))
 import Clash.Signal.Internal
   (Clock(..), Signal (..), Enable, KnownDomain, fromEnable, (.&&.))
 import Clash.Signal.Bundle   (unbundle)
 import Clash.Sized.Unsigned  (Unsigned)
 import Clash.XException      (errorX, maybeIsX, seqX, fromJustX)
+
+-- $setup
+-- >>> import Clash.Prelude.BlockRam.File
 
 
 -- | Create a blockRAM with space for 2^@n@ elements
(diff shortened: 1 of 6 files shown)


clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -205,6 +221,79 @@ blockRamFile = \clk gen sz file rd wrM ->
...
+-- 100001000
+--
+memFile
+  :: forall a
+   . ( BitPack a
+     , HasCallStack)
+  => Maybe Bit
+  -- ^ Value to map don't care bits to. Nothing means throwing an error on
+  -- don't care bits.
+  -> [a]
```
alex-mckenna, line 265: I wonder if people may want to convert a `Vec` of values? Maybe `[a]` should be `(Foldable f) => f a`?

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -205,6 +221,79 @@ blockRamFile = \clk gen sz file rd wrM ->
...
+memFile care = foldr (\e -> showsBV $ pack e) ""
+ where
+  showsBV :: BitVector (BitSize a) -> String -> String
+  showsBV (BV mask val) s =
+    if n == 0 then
+      '0' : '\n' : s
+    else
+      case care of
+        Just (Bit 0 0) -> go n (val .&. (mask `xor` fullMask)) ('\n' : s)
+        Just (Bit 0 1)  -> go n (val .|. mask) ('\n' : s)
```
alex-mckenna, line 278: It's unlikely that `low` and `high` will change, but maybe it's better to do

```haskell
case care of
  Just b
    | b == low -> ...
    | b == high -> ...
```

alex-mckenna:
Also, you've ticked the changelog box but haven't pushed a changelog. Have you forgotten a `git add`?

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -205,6 +221,79 @@ blockRamFile = \clk gen sz file rd wrM ->
...
+        _ -> if mask /= 0 then
+               err
+             else
+               go n val ('\n' : s)
+   where
+    n = natToNum @(BitSize a) @Int
+    fullMask = (1 `shiftL` n) - 1
+    err = withFrozenCallStack . error $
```
christiaanb, line 286:
```suggestion
    err = withFrozenCallStack $ error $
```
should fix the ghc9 error

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: That'll error on don't care bits!

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: I wrote this function because `Vec`s were impossibly slow and memory hungry for me :grin:. But I could do `Foldable` if you prefer. Shall I change it to that?

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: > Also, you've ticked the changelog box but haven't pushed a changelog. Have you forgotten a `git add`?

Again? Man....

DigitalBrains1:
> Also, you've ticked the changelog box but haven't pushed a changelog. Have you forgotten a `git add`?

Again? Man....

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: Thanks for the hint!

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
alex-mckenna: I don't care about those :wink:

But more seriously, I thought they had a non-zero mask anyway? Or is that only undefined bits?

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: Done

clash-prelude/src/Clash/Prelude/BlockRam/File.hs
```diff
@@ -31,6 +32,12 @@ For example, a data file @memory.bin@ containing the 9-bit unsigned number
 000001101
 @
 
+Such a file can be produced with 'memFile':
```
christiaanb, line 35:
```suggestion
Such a file can be produced with 'E.memFile':
```
should fix the haddock error

DigitalBrains1: Check. Were that all the errors in CI? I can't see the old result anymore since I force pushed.

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: s/bits/bit/ :-)

clash-prelude/src/Clash/Prelude/BlockRam/File.hs
christiaanb: All the haddock errors at least.

DigitalBrains1: Okay, tnx. Then let's see how the latest one fares.

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
DigitalBrains1: I never saw your comment before, odd. I was s/-ing my own "undefined bits" comment. The problem is that `Eq Bit` correctly stumbles over undefined bits. After all, you want `lsb $(bLit ".") == 0` to throw an error instead of taking a wild stab at a boolean result.

alex-mckenna: Ah yes, I see now. I guess you would need to use something like `Clash.Class.BitPack.isLike` to compare in a way that's friendly to don't care values

DigitalBrains1: The semantics of that here would mean that you want don't care bits to be mapped to don't care bits. And that is precisely the semantics of the current code. `Just $ lsb $(bLit ".")` is handled the same as `Nothing` by the `_` pattern match.

DigitalBrains1: User code should probably not peek under the hood of `Bit`, but we can do that, it's the same body of code as where it is defined.

PR #1841, opened 2021-06-03
Look through type synonyms in SynthesisAttributes

### LINKED ISSUE #1771: Synthesis attributes not found when using type aliases
This:

```haskell
module Test where

import Clash.Explicit.Prelude
import Clash.Annotations.SynthesisAttributes

top :: BitVector 1 `Annotate` [ 'StringAttr "chip_pin" "XX" ] -> BitVector 1 
top a = a
```

produces:

```verilog
(* chip_pin = "XX" *) input [0:0] a
```

When using an intermediate type synonym, this fails. E.g., this:

```haskell
module Test where

import Clash.Explicit.Prelude
import Clash.Annotations.SynthesisAttributes

type X = '[ 'StringAttr "chip_pin" "XX" ]

top :: BitVector 1 `Annotate` X -> BitVector 1 
top a = a
```

produces:

```verilog
input [0:0] a
```

while I would expect them to behave identically.  Weirdly, when using:

```haskell
module Test where

import Clash.Explicit.Prelude
import Clash.Annotations.SynthesisAttributes

-- No []
type X = 'StringAttr "chip_pin" "XX"

top :: BitVector 1 `Annotate` X -> BitVector 1 
top a = a
```

Clash fails explicitly:

```
<no location info>: error:
    Clash error call:
    Expected StringAttr, IntegerAttr, BoolAttr or Attr constructor, got:Test.X
      error, called at src-ghc/Clash/GHC/GHC2Core.hs:824:9 in clash-ghc-1.4.1-Fkb2bPkpKD58uHIycIlSqz:Clash.GHC.GHC2Core
```

----------------

This is an issue in `GHC2Core`. `coreToAttrs` & friends should account for type synonyms.

alex-mckenna:
When extracting SynthesisAttributes from types in GHC2Core, there
were some cases where the attributes were not visible to coreToAttr
or coreToAttrs, meaning they would not be included in the
generated HDL. These cases involved type synonyms which were not
being looked though.

These functions now look through type synonyms when they fail to
find an attribute, but do find a synyonm which can be expanded.

Fixes #1771 

## Still TODO:

### DIFF
--- a/changelog/2021-06-03T14_58_09+02_00_synthesis_annotation_synyonms.md
+++ b/changelog/2021-06-03T14_58_09+02_00_synthesis_annotation_synyonms.md
@@ -1,0 +1,1 @@
+FIXED: SynthesisAnnotations can now be defined in type synoynms without being excluded from the generated HDL [#1771](https://github.com/clash-lang/clash-compiler/issues/1771).
--- a/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
@@ -746,8 +746,12 @@ _ATTR_NAME = "Clash.Annotations.SynthesisAttributes.Attr"
 
 -- | Flatten a list type structure to a list of types.
 listTypeToListOfTypes :: Type -> [Type]
+-- TyConApp ': [kind, head, tail]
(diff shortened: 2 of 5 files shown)


christiaanb (approved):

alex-mckenna:
This works, but with a non-obvious caveat that arguments with annotations have to appear eta-expanded or they may be dropped. So this is very much a best effort fix, and I should update the haddock to explain this before it gets merged.

PR #1845, opened 2021-06-04
Fix bitrot in profiling code

leonschoorl:
The clash-profile-normalization-run broke at run time because the deserialized data had a different type signature then the serialiazed data in clash-profile-normalization-prepare.
This was broken by the refactor in: https://github.com/clash-lang/clash-compiler/pull/1057/commits/df551ecdd65c5c19474fc71108a0ed41d95bd0ab

There is now one type alias shared by both, resulting in compile errors
in both files in future refactors like this.

Also I've remove the unused tuple element leading to this mess.

## Still TODO:


benchmark/profiling/prepare/instances/SerialiseInstances.hs
```diff
@@ -11,11 +11,17 @@ import           Clash.Annotations.Primitive                  (PrimitiveGuard)
...
+import qualified Clash.Netlist.Types                          as CL (BlackBox, TopEntityT)
 import qualified Clash.Netlist.BlackBox.Types                 as CL
 
 import qualified Data.HashMap.Strict                          as HM
 
+import           Clash.Annotations.BitRepresentation.Internal (CustomReprs)
+import           Clash.Core.Var (Id)
+import           Clash.Core.TyCon (TyConMap,TyConName)
+import           Clash.Driver.Types (BindingMap)
+import           Data.IntMap.Strict           (IntMap)
```
alex-mckenna, line 23: Weird bit of indentation here

alex-mckenna (approved):

martijnbastiaan:
So much nicer <3. Could you reword the "Same for netlist" commit message to something that's readable without context?

leonschoorl:
@martijnbastiaan I've kept the commits separate for easier reviewing, but I was planning to do a squash-merge.
So it would get its context that way.

benchmark/profiling/prepare/instances/SerialiseInstances.hs
leonschoorl: copy-pasta from different places
Fixed now

PR #1847, opened 2021-06-08
Improved `-fclash-debug-*` flags

### LINKED ISSUE #1800: Refactor `DebugLevel` logic to not use `Ord`
Over time we've added more and more debug levels:

https://github.com/clash-lang/clash-compiler/blob/5eee599f24cbb933fce1ec17d99139fea09f7063/clash-lib/src/Clash/Driver/Types.hs#L81-L97

PR https://github.com/clash-lang/clash-compiler/pull/1647 will add yet another one. We're currently using `Ord` to determine whether an option is enabled (well, most of the time). The number of debug options has outgrown this strategy. For example; enabling a count summary should probably not affect whether the core of a normalized binder is printed.

alex-mckenna:
The old `DebugLevel` style debug options are replaced with a more granular style, where different options can be set independently. This allows things not previously possible, such as counting  transformations without having to print any normalized core.
    
The old `-fclash-debug` flag still works in the same way for backwards compatibility. Additional command line flags are now available to directly set options from the new type:
    
  * `-fclash-debug-invariants` checks for invariants and displays warnings / errors. Setting this alone is equivalent to `DebugSilent`.
  * `-fclash-debug-info` sets the amount of information to display when applying transformations. This overlaps a lot with the old `DebugLevel` type, but is named more consistently.
  * `-fclash-debug-count-transformations` counts the number of transformations applied (without showing any normalized core like `DebugCount` did)

Additionally, the debugging options are now in their own type (`DebugOpts`) which is accessed directly through the `RewriteEnv` in `apply` / `applyDebug` instead of having the individual debugging options passed as inputs to the function. This simplifies the logic slightly in these functions and avoids needing to call `applyDebug` recursively with modified values.

Fixes #1800 

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver/Types.hs
+++ b/clash-lib/src/Clash/Driver/Types.hs
@@ -133,11 +133,11 @@ data ClashOpts = ClashOpts
   -- ^ List the transformations that are to be debugged.
   --
   -- Command line flag: -fclash-debug-transformations
-  , opt_dbgTransformationsFrom :: Int
+  , opt_dbgTransformationsFrom :: Word
   -- ^ Only output debug information from (applied) transformation n
   --
   -- Command line flag: -fclash-debug-transformations-from
-  , opt_dbgTransformationsLimit :: Int
+  , opt_dbgTransformationsLimit :: Word
   -- ^ Only output debug information for n (applied) transformations. If this
   -- limit is exceeded, Clash will stop normalizing.
   --
--- a/clash-lib/src/Clash/Driver/Types.hs
+++ b/clash-lib/src/Clash/Driver/Types.hs
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2013-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017     , QBayLogic, Google Inc.
-                    2020     , QBayLogic
+                    2020-2021, QBayLogic
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -24,6 +24,8 @@ import           Control.DeepSeq                (NFData)
 import           Data.Binary                    (Binary)
 import           Data.Fixed
 import           Data.Hashable
+import           Data.Maybe                     (isJust)
+import           Data.Set                       (Set)
 import qualified Data.Set                       as Set
 import           Data.Text                      (Text)
 import           Data.Text.Prettyprint.Doc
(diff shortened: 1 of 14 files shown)


clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -78,25 +80,172 @@ data Binding a = Binding
...
+    opt_invariants `hashWithSalt`
+    opt_transformationInfo `hashSet`
+    opt_transformations `hashWithSalt`
+    opt_countTransformations `hashWithSalt`
+    opt_transformationsFrom `hashWithSalt`
+    opt_transformationsLimit `hashWithSalt`
+    opt_historyFile
+   where
+    hashSet = Set.foldl' hashWithSalt
+    infixl 0 `hashSet`
```
martijnbastiaan, line 156: Any reason to not just derive `Hashable`?

alex-mckenna: I did wonder about that. `Hashable` isn't derived for `ClashOpts`, so when I defined this one I just moved the relevant lines from that instance + added the ones for the new fields to be safe

martijnbastiaan: Hmmm, my guess is that that's a leftover from a feature/hack we've since moved to the right place?:

https://github.com/clash-lang/clash-compiler/blob/aaf15590d74b9477cfca96c0ab40e27e1b678ab3/clash-lib/src/Clash/Driver/Manifest.hs#L370-L408

I think we can just remove it and derive `Hashable` - in a separate commit of course so we can rollback if necessary.

alex-mckenna: Ah, `Set String` has no `Hashable` instance, so deriving would fail

alex-mckenna: And same for `OverridingBool` in `ClashOpts`. So I guess without making any more `*.Extra` modules we wouldn't be able to derive these instances

martijnbastiaan: Ah. Do we rely on the ordering? `HashSet` is generally preferable anyway. Which would leave `OverridingBool` - I guess an orphan instance wouldn't hurt.

alex-mckenna: > Do we rely on the ordering?

No, it's just a set of transformations which are whitelisted for showing in the debug logs. I had thought that there should probably be a later commit to change it from `String` to `Text`, maybe that should be a commit to change it from `Set String` to `HashSet Text`?

martijnbastiaan: Yesss :)

alex-mckenna: I think I'm going to save this kind of type shake-up for a future PR, there are a few places where the types in rewrite / normalize could be improved and fiddling with types seems a bit tangential to this PR

alex-mckenna:
@leonschoorl I've changed `applyDebug` so it reports unexpected changes when `dbg_invariants` is on, as opposed to when the new analogue of `DebugApplied` is on. So if something like #1848 happens again the testsuite will pick it up

leonschoorl:
Wow, the term in IntegralTB with 8.10 is so big is doesn't even get to print the actually error.

alex-mckenna:
That's how you know it's ~working~ doing something. I guess I need to actually deal with the type change error instead of just downgrading it to a trace, it feels wrong to not include it in `opt_invariants`, or make it depend on an unrelated debugging option

christiaanb:
Don't change **when** type-change-errors are reported compared to how it is now on HEAD of `master`. The reporting is very fragile (asserts often) because we throw away all the casts. Until we properly support casts I wouldn't worry too much about these type-change-errors.

alex-mckenna:
> Don't change **when** type-change-errors are reported compared to how it is now on HEAD of `master`.

Yeah, seems reasonable. It'll be a little non-obvious to read with the new check (what does the amount of info printed about transformed terms have to do with type errors?) but the behaviour will be identical. I'll leave a comment for archaeology / to reference to #1064 so it's clearer to anyone who happens across it in the future

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -212,82 +194,87 @@ applyDebug
...
+                       , "substitution."
+                       ])
+
+      -- TODO This check should not have the `hasDebugInfo` call in it, as
+      -- setting dbg_invariants should be all that is necessary to check this.
+      -- However, currently this error is very fragile, as Clash currently does
+      -- not keep casts, so "illegally" changing between `Signal dom a` and `a`
+      -- will trigger this error for many designs.
+      --
+      -- This should be changed when #1064 (PR to keep casts in core) is merged.
```
leonschoorl, line 254: Apart from casts, wasn't this also triggered by [`separateArguments`](https://github.com/clash-lang/clash-compiler/blob/be7558062e2832c446b9dcf61be3852872b646bb/clash-lib/src/Clash/Normalize/Transformations/SeparateArgs.hs#L46)?

leonschoorl (approved):

clash-lib/src/Clash/Rewrite/Util.hs
alex-mckenna: The implicit params thing? Yeah, I didn't include that because by the time the signal cast PR is in it might not be an issue / there might be other small issues so I didn't want to get too into detail past the most immediate obstacle

PR #1848, opened 2021-06-09
Report changes in simpleCSE

### LINKED ISSUE #1846: Unexpected changes in CSE
In f68204ff44bafdade476aa006dce0229aa5a2742, a mistake was fixed that was stopping Clash from identifying when an expression changed unexpectedly. Currently on `master`, we can see this in action by running

```bash
clash --vhdl -fclash-debug DebugApplied examples/Blinker.hs
```

which reports

```
<no location info>: error:                       
    Clash error call:                                                                                                                                                                                              
    Clash.Rewrite.Util(279): Expression changed without notice(CSE): beforeletrec
    ...
```

There is likely a path in CSE where a change is not reported, i.e. `return` is used instead of `changed`.

alex-mckenna:
The simpleCSE transformation previously never reported when it had
changed the resulting AST. With the fix from f68204f this meant
that CSE would often cause failures when using DebugApplied or
higher.

Fixes #1846 

## Still TODO:

  - [ ] Make sure the testsuite can detect such errors
 
### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
@@ -391,9 +391,10 @@ isClassConstraint _ = False
 -- On the two examples that were tested, Reducer and PipelinesViaFolds, this new
 -- version of CSE removed the same amount of let-binders.
 simpleCSE :: HasCallStack => NormRewrite
-simpleCSE (TransformContext is0 _) (inverseTopSortLetBindings -> Letrec bndrs body) = do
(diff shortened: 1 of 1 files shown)


leonschoorl (approved):

leonschoorl:
The testsuite should have caught this

alex-mckenna:
> The testsuite should have caught this

Discussed offline, the testsuite runs with `DebugSilent` which is too low to catch this failure (which needs `DebugApplied` or higher). The best course of action is to change when checks like this are performed in #1800 which touches that logic anyway

PR #1849, opened 2021-06-10
New: mergeEnable

DigitalBrains1:
The "hidden" counterpart to `Clash.Explicit.Signal.enable`.

Also corrected some outdated documentation from before we added enables,
and some copy-paste mistakes in the documentation.

* The current implementation of `mergeEnable` was written for components with single clock domains. Is this enough?

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Signal.hs
+++ b/clash-prelude/src/Clash/Signal.hs
@@ -1,9 +1,10 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente,
                   2016-2019, Myrtle Software Ltd,
-                  2017     , Google Inc.
+                  2017     , Google Inc.,
+                  2021     , QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
-Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
 Clash has synchronous 'Signal's in the form of:
 
@@ -149,7 +150,7 @@ module Clash.Signal
   , toEnable
   , fromEnable
   , S.enableGen
-    -- * Hidden clocks and resets
+    -- * Hidden clocks, resets and enables
     -- $hiddenclockandreset
 
     -- ** Hidden clock
@@ -193,6 +194,7 @@ module Clash.Signal
 #endif
   , SystemClockResetEnable
     -- * Basic circuit functions
+  , mergeEnable
   , dflipflop
   , delay
   , delayMaybe
@@ -313,9 +315,9 @@ let countSometimes = s where
 -- * Hidden clock, reset, and enable arguments
 
 {- $hiddenclockandreset #hiddenclockandreset#
-Clocks and resets are by default implicitly routed to their components. You can
-see from the type of a component whether it has hidden clock or reset
-arguments:
+Clocks, resets and enables are by default implicitly routed to their components.
+You can see from the type of a component whether it has hidden clock, reset or
+enable arguments:
 
 It has a hidden clock when it has a:
 
@@ -333,8 +335,16 @@ g :: 'HiddenReset' dom => ...
 
 Constraint.
 
-Or it has both a hidden clock argument and a hidden reset argument when it
-has a:
+Or it has a hidden enable when it has a:
+
+@
+g :: 'HiddenEnable' dom => ...
+@
+
+Constraint.
+
+Or it has a hidden clock argument, a hidden reset argument and a hidden enable
+argument when it has a:
 
 @
 h :: 'HiddenClockResetEnable' dom  => ..
@@ -342,8 +352,9 @@ h :: 'HiddenClockResetEnable' dom  => ..
 
 Constraint.
 
-Given a component with an explicit clock and reset arguments, you can turn them
-into hidden arguments using 'hideClock', 'hideReset', and 'hideEnable'. So given a:
+Given a component with explicit clock, reset and enable arguments, you can turn
+them into hidden arguments using 'hideClock', 'hideReset', and 'hideEnable'. So
+given a:
 
 @
 f :: Clock dom -> Reset dom -> Enable dom -> Signal dom a -> ...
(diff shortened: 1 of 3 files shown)


tests/shouldwork/Signal/MergeEnable.hs
```diff
@@ -0,0 +1,36 @@
...
+        -> Signal dom (Unsigned 8)
+      f enM0 = mergeEnable enM0 $ register 0
+  in withClockResetEnable clk rst en (f enM)
+
+testBench
+  :: Signal System Bool
+testBench = done
+ where
+  enM = True :> True :> False :> True :> Nil
+  en  = True :> True :> True  :> True :> False :> True :> Nil
```
leonschoorl, line 28: For completeness maybe test all 4 possible combinations of the two enable signals.
(You're missing the case where both are False at the same time.)

clash-prelude/src/Clash/Signal.hs
```diff
@@ -373,37 +384,38 @@ f :: 'HiddenClockResetEnable' dom
   -> Signal dom Int
 @
 
-which has hidden clock and routed reset arguments, we expose those hidden
-arguments so that we can explicitly apply them:
```
leonschoorl, line 377: Does `routed` have some special meaning here?
My feeling is to just say "... hidden clock, reset and enable arguments"

DigitalBrains1: I have no idea, I thought maybe it had a special meaning, so I kept it.

tests/shouldwork/Signal/MergeEnable.hs
DigitalBrains1: Sure, I'll add it. It might catch some weird mistake we might make in the future and it costs us virtually nothing to cover the case here.

leonschoorl:
For multiple clock you could copy what `exposeClock` and friends do: https://github.com/clash-lang/clash-compiler/blob/aaf15590d74b9477cfca96c0ab40e27e1b678ab3/clash-prelude/src/Clash/Signal.hs#L521-L565
with a `mergeEnable` as a wrapper around `mergeSpecificEnable`

tests/shouldwork/Signal/MergeEnable.hs
leonschoorl: I don't really expected you to accidentally write an XNOR (or is it NXOR?).
But it's such a small change to test all cases, you might as well test them all.

clash-prelude/src/Clash/Signal.hs
leonschoorl: It got shortened from "implicitly routed clock/reset" in [this big patch](https://github.com/clash-lang/clash-compiler/commit/26973e75d65e9de4942d298e4b55f2cec3e7b7f3#diff-3997ddd6d1b20d5deb35e5fd5056f8fc31ca6a8826afa19ac9c1982b5795a832L217-L218):
```diff
-which has an implicitly routed clock and implicitly routed clock, we can use
-it in a function with explicit clock and reset arguments like so:
+which has hidden clock and routed reset arguments, we expose those hidden
+arguments so that we can explicitly apply them:
```
So you can remove the "routed".
It's meaningless without the "implicitly".

DigitalBrains1:
Should we perhaps just call this thing `enable`? WIth other functions that have an explicit and an implicit one they are named the same, and though this has a quite different signature than the explicit one, it provides the same functionality.

The downside is that if someone imported both `Clash.Prelude` as well as `Clash.Explicit.Signal.enable`[1] unqualified, their code "breaks" because suddenly the name is ambiguous.

[1] Also available in `Clash.Explicit.Prelude.Safe`, `Clash.Explicit.Prelude`

clash-prelude/src/Clash/Signal.hs
DigitalBrains1: Oh thanks for the archeology! It hadn't occurred to me to do that!

leonschoorl:
Yeah I was wondering the same thing.

But name clashes when doing unqualified imports from both `Clash.Prelude` and `Clash.Explicit....` isn't a very strong argument, there are lots of names that clash there.

martijnbastiaan:
@DigitalBrains1 Did you reach a conclusion on what you want to name the function? :) What should happen for this PR to move forward?

DigitalBrains1:
It will be named `andEnable`, and `Clash.Explicit.Signal.enable` will be deprecated in favour of this new name.

I have not pushed the latest work-in-progress to this PR, it's on a separate branch [mergeenable-work](https://github.com/clash-lang/clash-compiler/tree/mergeenable-work), I thought it made more sense for any pushes here to be at least a suggestion of how to go forward, not merely "that's it for today, push a fixup commit and continue some other day".

It turned out we had hidden a large part of our documentation on the whole hidden business 😅 due to a CPP mistake. The only thing left to do is extend the documentation of our hidden parameters, and explain a bit about avoiding that the monomorphism restriction ignores your use of `andEnable`, and add a note to the developers on the relation between `WithSingleDomain dom r => expose...` and `exposeSpecific...`, because it is not very self-explanatory, and we do ourselves a service by spelling it out.

DigitalBrains1:
This PR is 90% documentation and 10% code. Extensive changes to the documentation have been made.

Note that there is cruft in the generated documentation when compiling without `multiple-hidden` due to Haddock issue https://github.com/haskell/haddock/issues/1382. The proposed workaround in https://github.com/haskell/haddock/issues/1382#issuecomment-889922782 has its drawbacks for us as devs. I'm not sure where to go with this. Perhaps we should try to get the issue fixed and the fix implemented on Hackage before this new documentation graduates to a Clash release. Note that the current documentation doesn't include this cruft because that whole piece of the documentation is just not rendered! It's there, but it's just as invisible as the actual documentation that follows the cruft.

DigitalBrains1:
Hmmm it seems `tests/shouldwork/Signal/AndSpecificEnable.hs` was actually unfinished. I'll finish it tomorrow. Please disregard it when reviewing for now :-).

DigitalBrains1:
Hopefully that's it, I think it's ready for inclusion now.

alex-mckenna (approved):
LGTM, especially the documentation improvements :tada:

PR #1852, opened 2021-06-14
Report changes in constantSpec / reduceNonRepPrim

### LINKED ISSUE #1851: Unexpected changes in transformations
By changing #1800 so that the issue that caused #1848 could not happen again, more issues were exposed in the testsuite. These are:

  * unexpected changes in `reduceRepNonPrim` and `constantSpec`. These transformations need to ensure they use `setChanged` / `changed` where necessary
  * unexpected type changes from `Signal dom a` to `a` in some transformations. This is known about but is now an error. Perhaps the check should compare on the coreView of types, or be downgraded to a warning
  * unexpected type changes from implicit params being split up in types. I'm not entirely sure what the right thing to do here is

alex-mckenna:
These transformations were previously not reporting all times when
they were applied / changed the input term. However, the check for
this was only enabled with DebugApplied or higher, which meant CI
would not identify these mistakes (as CI uses DebugSilent).

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
@@ -22,7 +22,9 @@ module Clash.Normalize.Transformations.Specialize
 
 import qualified Control.Lens as Lens
 import Control.Monad.Extra (orM)
+import qualified Control.Monad.Writer as Writer (listen)
 import qualified Data.Either as Either
+import qualified Data.Monoid as Monoid (getAny)
 import GHC.Stack (HasCallStack)
 
 import Clash.Core.FreeVars (termFreeTyVars, typeFreeVars)
@@ -44,7 +46,7 @@ import Clash.Rewrite.Combinators (topdownR)
 import Clash.Rewrite.Types
   ( TransformContext(..), bindings, censor, customReprs, tcCache
   , typeTranslator, workFreeBinders)
-import Clash.Rewrite.Util (mkDerivedName, mkTmBinderFor, setChanged)
+import Clash.Rewrite.Util (mkDerivedName, mkTmBinderFor, setChanged, changed)
 import Clash.Rewrite.WorkFree (isWorkFree)
 import Clash.Normalize.Types (NormRewrite, NormalizeSession)
 import Clash.Normalize.Util
@@ -234,11 +236,13 @@ constantSpec ctx@(TransformContext is0 tfCtx) e@(App e1 e2)
          else do
            -- Parts of e2 are constant
            let is1 = extendInScopeSetList is0 (fst <$> csrNewBindings specInfo)
-           Letrec newBindings
-            <$> specializeNorm
-                  (TransformContext is1 tfCtx)
-                  (App e1 (csrNewTerm specInfo))
+           (body, isSpec) <- Writer.listen $ specializeNorm
+             (TransformContext is1 tfCtx)
+             (App e1 (csrNewTerm specInfo))
 
+           if Monoid.getAny isSpec
+             then changed (Letrec newBindings body)
+             else return e
        else
         -- e2 has no constant parts
         return e
--- a/clash-lib/src/Clash/Rewrite/Util.hs
+++ b/clash-lib/src/Clash/Rewrite/Util.hs
@@ -275,7 +275,7 @@ applyDebug lvl _transformations _fromLimit name exprOld hasChanged exprNew =
                      ]
             ) (return ())
 
-  Monad.when (lvl >= DebugApplied && not hasChanged && not (exprOld `aeqTerm` exprNew)) $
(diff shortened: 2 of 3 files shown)


clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -234,11 +234,11 @@ constantSpec ctx@(TransformContext is0 tfCtx) e@(App e1 e2)
...
            let is1 = extendInScopeSetList is0 (fst <$> csrNewBindings specInfo)
-           Letrec newBindings
-            <$> specializeNorm
-                  (TransformContext is1 tfCtx)
-                  (App e1 (csrNewTerm specInfo))
+           body <- specializeNorm
+             (TransformContext is1 tfCtx)
+             (App e1 (csrNewTerm specInfo))
 
+           changed (Letrec newBindings body)
```
christiaanb, line 241: This is surprising, `specializeNorm` already does a `setChanged` in all cases except when it doesn't find the binding for the global variable it wants to specialize on.

I would say that when `specializeNorm` doesn't do a `setChanged`, then we don't want do `constantSpec` at all. i.e. this code should probably do a `Writer.listen` and use that as a predicate on whether it does a `changed (Letrec newBindings body)` or a `return e`.

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -829,9 +829,9 @@ specialise' specMapLbl specHistLbl specLimitLbl (TransformContext is0 _) e (Var
                       -- are inlined, meaning the state-transition-function
                       -- and the memory element will be in a single function.
                       gTmM <- fmap (lookupUniqMap (varName g)) $ Lens.use bindings
-                      return (g,maybe inl bindingSpec gTmM, maybe specArg (Left . (`mkApps` gArgs) . bindingTerm) gTmM)
-                    else return (f,inl,specArg)
-                _ -> return (f,inl,specArg)
+                      changed (g,maybe inl bindingSpec gTmM, maybe specArg (Left . (`mkApps` gArgs) . bindingTerm) gTmM)
+                    else changed (f,inl,specArg)
+                _ -> changed (f,inl,specArg)
```
christiaanb, line 834: We already do a `changed` on https://github.com/clash-lang/clash-compiler/pull/1852/files#diff-ef03bca01fe87dd3c262b523212da81f7626bdc2a0f9af328fe0376bcc3f65b1L843 why would we need these?

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
alex-mckenna: I think `constantSpecInfo` says that there are bindings it can make, but if the body of that resulting `letrec` is an application on a top entity it won't be specialised by `specialise` so `changed` is never set:

https://github.com/clash-lang/clash-compiler/blob/master/clash-lib/src/Clash/Rewrite/Util.hs#L756

I was still seeing the failure in some tests without adding the `changed` here because

```
Not specializing TopEntity: topEntity

                <no location info>: error:
                    Clash error call:
                    Clash.Rewrite.Util(279): Expression changed without notice(constantSpec): beforetopEntity[GlobalId]
                      (C:(%,,%) testBench_clk[LocalId]
                         (resetGenN removedArg removedArg (SNat 1))
                         (Enable True))
                    after:
                    letrec
                      testBench_done_bindCsr
                      = testBench_clk[LocalId]
                      testBench_done_bindCsr
                      = resetGenN removedArg removedArg (SNat 1)
                    in topEntity[GlobalId]
                         (C:(%,,%) testBench_done_bindCsr[LocalId]
                            testBench_done_bindCsr[LocalId]
                            (Enable True))
                      error, called at src/Clash/Rewrite/Util.hs:279:5 in clash-lib-1.5.0-inplace:Clash.Rewrite.Util
```

clash-lib/src/Clash/Rewrite/Util.hs
alex-mckenna: We might not need these, I should double check. The one on line 834 was just the first thing I found in common on all the tests where the "changed without notice" error was being thrown

alex-mckenna: Yep, we don't need any of these

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
christiaanb: Reading up on why we created the let-bindngs: it was to allow constant-specialisation on expression that are in WHNF but not NF due to some FVs. But if we're not specializing to begin with, we don't need to create those let-bindings either. So I still think we should do a `Writer.listen` on `specializeNorm` and only create the let-bindings when we actually specialize.

PR #1855, opened 2021-06-15
Clarify what we expect from participants in the CoC

christiaanb:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,7 +1,8 @@
 # Guidelines for Respectful Communication
 
 As members of a Haskell Foundation affiliated project, we commit ourselves to
-a high standard of public behaviour. We have one over-arching rule:
+a high standard of public behaviour and we expect the same of participants on
+the project-related public forums. We have one over-arching rule:
 
     We strive to treat every person with respect.
 


martijnbastiaan (approved):

CODE_OF_CONDUCT.md
```diff
@@ -1,7 +1,8 @@
 # Guidelines for Respectful Communication
 
 As members of a Haskell Foundation affiliated project, we commit ourselves to
-a high standard of public behaviour. We have one over-arching rule:
+a high standard of public behaviour and we expect the same of participants on
```
DigitalBrains1, line 4: Our public-facing language of choice is US English, right, so that should be
```suggestion
a high standard of public behavior and we expect the same of participants on
```

I see a few more down, but I can't edit them in the review process. Shall I do a force-push?

DigitalBrains1 (changes requested):

CODE_OF_CONDUCT.md
DigitalBrains1: Or can't we even change the spelling of the text further down :-)

christiaanb: No, I'd rather keep the original text as much as possible.

PR #1863, opened 2021-06-21
Move from CircleCI to GitHub Actions

alex-mckenna:


.ci/setup.sh
```diff
@@ -46,7 +46,7 @@ fi
 cabal --version
 ghc --version
 
-# This might happen when running on Circle CI or during tags
+# This might happen during tags
```
martijnbastiaan, line 49: Could you clarify this is a GitLab CI thingy?

PR #1868, opened 2021-06-24
Fix subst warnings in substWithTyEq and tvSubstWithTyEq

alex-mckenna:
Also make sure we don't reorder TyLam's and Lam's. The old implementation kept 2 lists one with `Id`s and one with `TyVar`s. Now we use one `[Either Id TyVar]`.

Fixed by @leonschoorl already, fix verified to work on rebased branch

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize/Util.hs
+++ b/clash-lib/src/Clash/Normalize/Util.hs
@@ -37,7 +37,7 @@ module Clash.Normalize.Util
 import           Control.Lens            ((&),(+~),(%=),(.=))
 import qualified Control.Lens            as Lens
 import           Data.Bifunctor          (bimap)
-import           Data.Either             (lefts)
+import           Data.Either             (lefts,rights)
 import qualified Data.List               as List
 import qualified Data.List.Extra         as List
 import qualified Data.Map                as Map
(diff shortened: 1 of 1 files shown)


clash-lib/src/Clash/Normalize/Util.hs
```diff
@@ -480,25 +477,22 @@ tvFirst _ = Nothing
...
-  go []  [] = ty0 -- no eq constraints, returning original type
-  go eqs [] = substTy subst ty2
-   where
-     subst = extendTvSubstList (mkSubst emptyInScopeSet) eqs
-     args2 = args0 List.\\ (map (Left . fst) eqs)
-     ty2 = mkPolyFunTy tyRes args2
+    , Just (tv,ty) <- tvFirst tcArgs
+    = let
+        subst = extendTvSubst (mkSubst $ mkInScopeSet $ mkVarSet $ lefts argsOut2) tv ty
+        argsOut2 = Right arg : (argsOut List.\\ [Left tv])
```
christiaanb, line 491: Perhaps reverse these two statements? I know the order doesn't matter, but I still read let-expressions from top to bottom.

christiaanb (approved):

PR #1869, opened 2021-06-25
clash-cores New: Xilinx float addition

DigitalBrains1:
Single-precision floating point addition, realized through instantiation of the Xilinx Floating-Point LogiCORE IP v7.1.

A supporting repository integrating Clash with Vivado to run the testsuite can be found at
https://github.com/DigitalBrains1/xilinxfloat

## Still TODO:


DigitalBrains1:
Note that plans for the future are to make this work on single precision, half precision and extendable precision (that's IEEE754-speak for variable size mantissa and exponent).

That makes the current name `addFloat` a point of debate: perhaps in anticipation we should already call it `addFloating` even though it only takes the type `Float` at this time. Or pick a different naming scheme altogether, I'm not really fond of the name.

clash-cores/src/Clash/Cores/Floating/Xilinx/BlackBoxes.hs
```diff
@@ -0,0 +1,120 @@
...
+  DataCon _ (DC (Sum _ cfgDspUsageConstrs, cfgDspUsageTag)) _ = cfgDspUsageExpr
+  cfgDspUsage = cfgDspUsageConstrs !! cfgDspUsageTag
+  tclDspUsage :: String
+  tclDspUsage =
+    case cfgDspUsage of
+      "Clash.Cores.Floating.Xilinx.Internal.NoDspUsage" -> "No_Usage"
+      "Clash.Cores.Floating.Xilinx.Internal.MediumDspUsage" -> "Medium_Usage"
+      "Clash.Cores.Floating.Xilinx.Internal.FullDspUsage" -> "Full_Usage"
+      "Clash.Cores.Floating.Xilinx.Internal.MaxDspUsage" -> "Max_Usage"
+      _ -> error "Unknown FloatingDspUsage constructor"
```
martijnbastiaan, line 84: To make refactors easier I'd suggest:

```haskell

  tclDspUsage
    | cfgDspUsage == show 'NoDspUsage -> "No_Usage"
    | cfgDspUsage == show 'MediumDspUsage -> "Medium_Usage"
    ...
    | otherwise -> error ...
```

What do you think?

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
```diff
@@ -0,0 +1,104 @@
...
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Support for the [Xilinx Floating-Point LogiCORE IP v7.1](https://www.xilinx.com/support/documentation/ip_documentation/floating_point/v7_1/pg060-floating-point.pdf).
+
+The functions in this module make it possible to use the Xilinx IP in Clash.
+Compilation will instantiate the Xilinx IP. Clash will output a TCL script
+named @floating_point@/.../@.tcl@ which needs to be executed in the Vivado
+project to create the proper entity. Simulation in Clash produces bit-identical
+results to synthesis.
```
martijnbastiaan, line 12: Is this comment up-to-date? I remember some discussion on Slack, but didn't actively follow it.

martijnbastiaan (commented):
Quite a bit of work, looks very nice!

Two overall comments:

* I don't think we should add more blackboxes based on the template language. They're hard to read, debug, and update. 
* It would be nice to understand the differences between Haskell simulation and Xilinx primitives a bit more. I think it'd be a good idea to add a test that tests the carthesian product of various "suspicious" floats over some binary operators. E.g. the ones mentioned in [this class](https://hackage.haskell.org/package/ieee754-0.8.0/docs/Numeric-IEEE.html). To be clear, I don't think we need 100% bit compatibility with the Xilinx primitives, but documenting it would be good.

clash-cores/src/Clash/Cores/Floating/Xilinx/BlackBoxes.hs
DigitalBrains1: Looks good! Thanks for the hint!

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
DigitalBrains1: It's correct for addition. A few unary operators have a 1 ULP (unit in last place) error in Xilinx, which I don't think we can simulate without knowing how the Xilinx IP works (it's "encrypted"). The plan is to add this to this documentation when those operators are added.

DigitalBrains1:
> * It would be nice to understand the differences between Haskell simulation and Xilinx primitives a bit more. I think it'd be a good idea to add a test that tests the carthesian product of various "suspicious" floats over some binary operators. E.g. the ones mentioned in [this class](https://hackage.haskell.org/package/ieee754-0.8.0/docs/Numeric-IEEE.html). To be clear, I don't think we need 100% bit compatibility with the Xilinx primitives, but documenting it would be good.

What would you like to see tested in addition to the 95(!) tests in [addFloatBasicSamples](https://github.com/clash-lang/clash-compiler/pull/1869/files#diff-034970d638ce1c4321f05ddd8b77c860900d11e4c8ba6ea4bc06bc9aa4850155R39)? I thought I already picked a nice few interesting samples, but I'm happy to see more!

(Okay, 66 of those 95 tests are just combinations of NaN and one :-))

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
DigitalBrains1: Oh right, we **can** simulate it bit-accurate, but it requires a foreign function interface and then using Xilinx's C simulation shared lib. Or perhaps the portion of that library that contains the necessary information is even open source, which would enable us to take a close look at it. I haven't looked at the C sim lib from Xilinx yet.

DigitalBrains1:
I suppose some subnormals as input should be added to the test. I think I test most of the other interesting things in `Numeric.IEEE`, although not as a Cartesian product, just interesting combinations I could come up with. So product of my brain, not product of Descartes ;-).

martijnbastiaan:
> (Okay, 66 of those 84 tests are just combinations of NaN and one :-))

Those are great on their own and they should be kept :). I think the carthesian product apporach has a chance of being more thorough / more obviously complete. Ideally (but I don't know how much effort this is) the expected values would be computed with TH:

```haskell
interesting = [infinity, minDenormal, minNormal, maxFinite, epsilon, nan, 1, 2, 42]
inputs = carthesianProduct interesting interesting
expected = $(model inputs)
```

I think it's a nice-to-have though, so not really necessary if it's hard.

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
martijnbastiaan: Oh yeah, don't bother with that IMO.

clash-cores/clash-cores.cabal
```diff
@@ -74,6 +74,10 @@ library
   hs-source-dirs: src
 
   exposed-modules:
+    Clash.Cores.Floating.Xilinx
+    Clash.Cores.Floating.Xilinx.Annotations
+    Clash.Cores.Floating.Xilinx.BlackBoxes
+    Clash.Cores.Floating.Xilinx.Internal
```
alex-mckenna, line 80: I think I'd rather see these under `Clash.Core.Xilinx.Floating` on the grounds that

  * existing cores for Lattice IP are under `Clash.Cores.LatticeSemi`
  * if we add more Xilinx IP that isn't FP-related then the Xilinx IP won't be all together, which I personally think is less intuitive for someone browsing the source

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
```diff
@@ -0,0 +1,104 @@
...
+-- For those operations that support it, the default options are:
+--
+-- * Speed-optimized architecture ('SpeedArch').
+-- * Full DSP slice usage ('FullDspUsage').
+-- * Exponential operator does not use block memory.
+defFloatingC :: FloatingConfig
+defFloatingC = FloatingConfig
+  { floatingArchOpt = SpeedArch
+  , floatingDspUsage = FullDspUsage
+  , floatingBMemUsage = False
```
alex-mckenna, line 74: I know there are only two options for block memory usage (I want to say none and full?), but I still think it might be worth making a type for this. Then you get the bonus of it still being clear if you don't use record syntax:

```haskell
mySpecificConfig :: FloatingConfig
mySpecificConfig =
  FloatingConfig LatencyArch FullDspUsage FullBmemUsage
```

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
```diff
@@ -0,0 +1,104 @@
...
+  -> DSignal dom n Float
+  -> DSignal dom (n + d) Float
+addFloat cfg = withFrozenCallStack $ hideEnable . hideClock $ addFloat# cfg
+{-# INLINE addFloat #-}
+
+-- | The default delay for floating point addition with default customization.
+type AddFloatDefDelay = 11
+
+-- | Floating point addition with default settings.
+addFloat'
```
alex-mckenna, line 95: I'm not sure about the use of `'` here, mostly because it makes me think of strictness. In my own code I would tend to do something more like

```haskell
addFloatWith :: FloatingConfig -> ...
addFloatWith = ...

addFloat :: ...
addFloat =
  withFrozenCallStack $ addFloatWith defFloatingC
```

so it's clearer as someone reading just the haddock that one is configurable and one isn't

clash-cores/src/Clash/Cores/Floating/Xilinx/BlackBoxes.hs
```diff
@@ -0,0 +1,120 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RecordWildCards #-}
+module Clash.Cores.Floating.Xilinx.BlackBoxes
+  ( addFloatTclTF
+  ) where
+
+import Prelude
+
+import Control.Monad.State (State())
```
alex-mckenna, line 10: You don't need the extra parens after `State`, right?

clash-cores/src/Clash/Cores/Floating/Xilinx/BlackBoxes.hs
```diff
@@ -0,0 +1,120 @@
...
+import Data.Maybe (isJust, fromJust)
+import Data.String (fromString)
+import Data.String.Interpolate (i)
+
+import Clash.Backend (Backend())
```
alex-mckenna, line 15: Same here with `Backend`

clash-cores/clash-cores.cabal
DigitalBrains1: Right! I had reasoned the other way round, we'll add Floating support to more targets, and they are put together. But looking at the prelude, we have `Clash.Intel.ClockGen` and `Clash.Xilinx.ClockGen`, so your solution has precedent!

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
DigitalBrains1: Will do!

DigitalBrains1:
> > (Okay, 66 of those 84 tests are just combinations of NaN and one :-))
> 
> Those are great on their own and they should be kept :). I think the carthesian product apporach has a chance of being more thorough / more obviously complete. Ideally (but I don't know how much effort this is) the expected values would be computed with TH:
> 
> ```haskell
> interesting = [infinity, minDenormal, minNormal, maxFinite, epsilon, nan, 1, 2, 42]
> inputs = carthesianProduct interesting interesting
> expected = $(model inputs)
> ```
> 
> I think it's a nice-to-have though, so not really necessary if it's hard.

I think this is trivial to do, a good addition, and a negligible impact on test run time, I'll do it!

DigitalBrains1:
Can we bikeshed the name some more? I think maybe the shed looks the best in the colour `addF`. So we'd have `addF`, `mulF`, `divF`, expF`, etcetera. Right now these all take `Float`, but eventually it'll be a class which takes single precision, half precision, and extendable precision.

clash-cores/src/Clash/Cores/Floating/Xilinx/BlackBoxes.hs
DigitalBrains1: Right

clash-cores/src/Clash/Cores/Floating/Xilinx.hs
DigitalBrains1: I like the "with". The prime to me just suggests relation between prime and non-prime, but it's also a really small glyph, that's annoying.

martijnbastiaan:
Why not `add` and rely on users to do a `qualified` import? Especially if we plan on adding even more functions in the future.

DigitalBrains1:
I'd also be completely fine with that.

DigitalBrains1:
Do we also want a version with explicit clock and enable?

martijnbastiaan:
> Do we also want a version with explicit clock and enable?

Yes. Implicit clocks are hard to use for multi-clock designs unfortunately. Typically we write the primitives with an explicit clock/reset/enable and then write a wrapper with  `Hidden`* constraints.

DigitalBrains1:
Here's another idea: We could add infix components. Either in a separate `Infix` module, or just with the rest. It would allow you to do:
```haskell
import qualified Clash.Cores.Xilinx.Floating as F
import qualified Clash.Cores.Xilinx.Floating.Infix as F

f :: ( HiddenClock dom
     , HiddenEnable dom
     )
  => DSignal dom n Float
  -> DSignal dom n Float
  -> DSignal dom (n + F.AddDefDelay) Float
f x y = x F.+ y
```

(Not that this `f` is particularly useful, it's about the infix use and the type of that infix operator)

DigitalBrains1:
I've implemented all suggestions, except @martijnbastiaan's suggestion of rewriting the primitives to avoid the template language, and my own suggestion of adding an infix operator. @christiaanb noted that we can always rewrite the primitives in a later commit, and he'd rather see the PR merged without doing that extra work.

I implemented testing the Cartesian product of the following interesting values: infinity, minimum subnormal value, maximum subnormal value, minimum normal value, maximum normal value, epsilon, NaN and the basic numbers 1, 2 and 42.

DigitalBrains1:
Some of the manual tests (including two I added since the previous version) overlap with the Cartesian product, but this is for documentation purposes: it describes the result and why that is the result, unlike the Cartesian product part. Given the length of the test, these few manual ones come pretty much for free.

clash-cores/src/Clash/Cores/Xilinx/Floating/Explicit.hs
```diff
@@ -0,0 +1,129 @@
...
+addWith !_ clk en (conditionFloatF -> x) (conditionFloatF -> y) =
+  delayI und en clk . conditionFloatF $ x + y
+ where
+  und = withFrozenCallStack $ deepErrorX "Initial values of addFloat undefined"
+{-# NOINLINE addWith #-}
+{-# ANN addWith
+      (vhdlBinaryPrim
+         "Clash.Cores.Xilinx.Floating.Explicit.addWith"
+         "add"
+         "Clash.Cores.Xilinx.Floating.BlackBoxes.addTclTF"
```
martijnbastiaan, line 91: Is it possible to use `show 'name` here, or do staging restrictions get in the way?

clash-cores/src/Clash/Cores/Xilinx/Floating/Explicit.hs
```diff
@@ -0,0 +1,129 @@
...
+      ) #-}
+{-# ANN addWith
+      (veriBinaryPrim
+         "Clash.Cores.Xilinx.Floating.Explicit.addWith"
+         "add"
+         "Clash.Cores.Xilinx.Floating.BlackBoxes.addTclTF"
```
martijnbastiaan, line 97: Same here

clash-cores/src/Clash/Cores/Xilinx/Floating/Explicit.hs
DigitalBrains1: Yes, that turns out to work fine. Thanks for your vigilance, this is much better.

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
```diff
@@ -0,0 +1,339 @@
...
+    -- Minimum subnormal
+  , encodeFloat 1 (minExp - digits)
+    -- Maximum subnormal
+  , encodeFloat (2 ^ (digits - 1) - 1) (minExp - digits)
+    -- Minimum normal
+  , encodeFloat 1 (minExp - 1)
+    -- Maximum normal
+  , encodeFloat (2 ^ digits - 1) (maxExp - digits)
+    -- Epsilon (smallest positive x such that 1 + x is representable)
+  , encodeFloat 1 (1 - digits)
```
martijnbastiaan, line 207: Perhaps a dumb question, but why not use the names from https://hackage.haskell.org/package/ieee754-0.8.0/docs/Numeric-IEEE.html?

DigitalBrains1: Because they don't define anywhere near everything I want to use and I thought consistency was more desirable than being able to use a few short function names when they were available.

DigitalBrains1: I've gone back on my earlier decision; your comment made me realise I actually prefer using the `Numeric.IEEE` names where possible. This is more readable.

DigitalBrains1:
I think this is ready for merging now, based on the received and resolved feedback so far.

PR #1873, opened 2021-07-05
VHDL: no div/rem/mod by zero in gen'd code

christiaanb:
Because Clash translates to concurrent assignments, things like

```
if y /= 0 then quot x y else 0
```

would result in a VHDL sim error (at least on GHDL) because all subterms of the if-then-else expression would be evaluated
concurrently, and `x / y` would throw a sim error whenever `y=0` even though that result could never be observed.


martijnbastiaan:
Nice work on all the fixes, could you add changelogs? :)

martijnbastiaan (approved):

PR #1874, opened 2021-07-05
VHDL: don't overflow natural range in shift/rotate

christiaanb:
Haskell `shift{R;L}` and `rotate{R;L}` take an `Int` as the shift amount. In VHDL, `shift_{left;right}` and `rotate_{left;right}` take a `natural` as a shift amount. On 64 bit systems Haskell's `maxBound :: Int` is larger than VHDL's `natural'high` (in (nearly) all VHDL simulators).

So for `shift_{left;right}` we saturate the shift amount to `natural'high`, and for `rotate_{left;right}` we mask the shift amount with `natural'high`, all in order for the VHDL simulator not to error out with an out-of-bounds exception when the shift/rotate amount is larger than `natural'high`.

### DIFF
--- a/clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives
+++ b/clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives
@@ -541,11 +541,24 @@ end process;
     , "kind"      : "Declaration"
     , "type"      : "shiftL# :: KnownNat n => BitVector n -> Int -> BitVector n"
     , "template"  :
-"~RESULT <= std_logic_vector(shift_left(unsigned(~ARG[1]), to_integer(~ARG[2])))
-    -- pragma translate_off
-    when (~ARG[2] >= 0) else (others => 'X')
-    -- pragma translate_on
-    ;"
+"~GENSYM[~RESULT_shiftL][0] : block
+  signal ~GENSYM[sh][1] : natural;
+begin
+  ~SYM[1] <=
+      -- pragma translate_off
+      natural'high when (~VAR[shI][2](~SIZE[~TYP[2]]-1 downto 31) /= 0) else
+      -- pragma translate_on
+      to_integer(~VAR[shI][2]
+      -- pragma translate_off
+      (30 downto 0)
+      -- pragma translate_on
+      );
+  ~RESULT <= std_logic_vector(shift_left(unsigned(~ARG[1]),~SYM[1]))
+      -- pragma translate_off
+      when (~ARG[2] >= 0) else (others => 'X')
+      -- pragma translate_on
+      ;
+end block;"
     }
   }
 , { "BlackBox" :
@@ -553,11 +566,24 @@ end process;
     , "kind"      : "Declaration"
     , "type"      : "shiftR# :: KnownNat n => BitVector n -> Int -> BitVector n"
     , "template"  :
-"~RESULT <= std_logic_vector(shift_right(unsigned(~ARG[1]),to_integer(~ARG[2])))
-    -- pragma translate_off
-    when (~ARG[2] >= 0) else (others => 'X')
-    -- pragma translate_on
-    ;"
+"~GENSYM[~RESULT_shiftR][0] : block
+  signal ~GENSYM[sh][1] : natural;
+begin
+  ~SYM[1] <=
+      -- pragma translate_off
+      natural'high when (~VAR[shI][2](~SIZE[~TYP[2]]-1 downto 31) /= 0) else
+      -- pragma translate_on
+      to_integer(~VAR[shI][2]
+      -- pragma translate_off
+      (30 downto 0)
+      -- pragma translate_on
+      );
+  ~RESULT <= std_logic_vector(shift_right(unsigned(~ARG[1]),~SYM[1]))
+      -- pragma translate_off
+      when (~ARG[2] >= 0) else (others => 'X')
+      -- pragma translate_on
+      ;
+end block;"
     }
   }
 , { "BlackBox" :
@@ -565,7 +591,11 @@ end process;
     , "kind"      : "Declaration"
     , "type"      : "rotateL# :: KnownNat n => BitVector n -> Int -> BitVector n"
     , "template"  :
-"~RESULT <= std_logic_vector(rotate_left(unsigned(~ARG[1]),to_integer(~ARG[2])))
+"~RESULT <= std_logic_vector(rotate_left(unsigned(~ARG[1]),to_integer(~ARG[2]
+    -- pragma translate_off
+    mod ~SIZE[~TYP[1]]
+    -- pragma translate_on
+    )))
     -- pragma translate_off
     when (~ARG[2] >= 0) else (others => 'X')
     -- pragma translate_on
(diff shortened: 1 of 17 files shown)


martijnbastiaan (commented):
Similar to https://github.com/clash-lang/clash-compiler/pull/1875 I'm not sure what it would achieve in practice. Perhaps naively I'd think this would occlude a valid failure. Could you elaborate on why this isn't the case?

christiaanb:
So when you do:
```
f :: Signed 64 -> Signed 64 -> Signed 64
f x y = if y <= 0 then 0 else shiftL x (fromIntegral y)
```
VHDL simulation will throw an exception whenever `y` is negative, because the `shiftL` is executed concurrently. So in VHDL we would see an exception that's unobservable in Haskell.

christiaanb:
Also, you will only observe differences in simulation for values that actually can be shifted by more than 2^31-1 positions. That's a lot of bits! Clash/Haskell sim would probably already go OOM for such data types.

DigitalBrains1:
I'm worried that the masking for `rotate` only works properly for power-of-two bitlengths, when I define "works properly" as "simulation in Clash and in HDL give the same result for really large rotate amounts":

```haskell
>>> import qualified Prelude as P
>>> (1 :: BitVector 5) `rotateR` (2 P.^ 31 + 5)
0b0_0100
>>> (1 :: BitVector 5) `rotateR` ((2 P.^ 31 + 5) `mod` (2 P.^ 31))
0b0_0001
```

clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives
```diff
@@ -531,31 +531,57 @@ end process;
...
+end block;"
     }
   }
 , { "BlackBox" :
     { "name"      : "Clash.Sized.Internal.BitVector.rotateL#"
     , "kind"      : "Declaration"
     , "type"      : "rotateL# :: KnownNat n => BitVector n -> Int -> BitVector n"
     , "template"  :
-"~RESULT <= std_logic_vector(rotate_left(unsigned(~ARG[1]),to_integer(~ARG[2])))
+"~RESULT <= std_logic_vector(rotate_left(unsigned(~ARG[1]),to_integer(~VAR[sh][2](30 downto 0))))
```
DigitalBrains1, line 584: I'm worried that the masking for `rotate` only works properly for power-of-two bitlengths, when I define "works properly" as "simulation in Clash and in HDL give the same result for really large rotate amounts":

```haskell
>>> import qualified Prelude as P
>>> (1 :: BitVector 5) `rotateR` (2 P.^ 31 + 5)
0b0_0100
>>> (1 :: BitVector 5) `rotateR` ((2 P.^ 31 + 5) `mod` (2 P.^ 31))
0b0_0001
```

We should probably put "modulo bitlength" inside a `--pragma translate off` instead.

DigitalBrains1 (changes requested):

christiaanb:
@DigitalBrains1 could you add tests that you think will fail? Thanks!

clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives
DigitalBrains1: Indeed, if I change `tests/shouldwork/Numbers/ShiftRotateBase.hs` to contain a really large amount, the testbench fails in GHDL:

```diff
@@ -5,12 +5,13 @@ import Clash.Prelude
 import Clash.Explicit.Testbench
 import Data.Word
 import Data.Int
+import qualified Prelude as P
 
 testpattern = 0b1010011000
 
 -- test with shift/rotate amounts: 0..16,200
 amounts :: Vec _ Int
-amounts = $(listToVecTH [0::Int ..16]) :< 200
+amounts = $(listToVecTH [0::Int ..16]) :< 200 :< (2 P.^ 31 + 5)
 
 testall v i
   = ( testAs @(Unsigned 8)  v i
```

DigitalBrains1:
Oh wow, I thought I had a test that works on both 32- and 64-bit operating systems (although on 32-bit operating systems this patch to `rotate` is a no-op anyway as you take the lowest 32 bits of a 32 bits number), but I get an interesting error instead:
```diff
--- a/tests/shouldwork/Numbers/ShiftRotateBase.hs
+++ b/tests/shouldwork/Numbers/ShiftRotateBase.hs
@@ -10,7 +10,7 @@ testpattern = 0b1010011000
 
 -- test with shift/rotate amounts: 0..16,200
 amounts :: Vec _ Int
-amounts = $(listToVecTH [0::Int ..16]) :< 200
+amounts = $(listToVecTH [0::Int ..16]) :< 200 :< (maxBound - 5)
 
 testall v i
   = ( testAs @(Unsigned 8)  v i
```
leads to
```haskell
$ cabal run clashi
Up to date
Clashi, version 1.5.0 (using clash-lib, version 1.5.0):
https://clash-lang.org/  :? for help
Loaded package environment from /home/peter/src/clash/clash-compiler/.ghc.environment.x86_64-linux-8.10.4
Loaded Clashi configuration from /home/peter/.clashi
Clash.Prelude P> :cd tests/shouldwork/Numbers/
Clash.Prelude P> :l ShiftRotate.hs 
[1 of 2] Compiling ShiftRotateBase  ( ShiftRotateBase.hs, interpreted )
[2 of 2] Compiling ShiftRotate      ( ShiftRotate.hs, interpreted )

ShiftRotate.hs:26:12: error:
    • Exception when trying to run compile-time code:
        heap overflow
      Code: lift $ map (testall testpattern) amounts
    • In the untyped splice:
        $(lift $ map (testall testpattern) amounts)
   |
26 | expected = $(lift $ map (testall testpattern) amounts)
   |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, one module loaded.
*ShiftRotateBase P> 
```

clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives
christiaanb: We can do:
```
#include "MachDeps.h"
```
Which gives us access to
```
WORD_SIZE_IN_BITS
```
So we can differentiate between 32 and 64 bit wrt the inputs we feed to the DUT and the outputs we expect from the DUT using CPP if.

DigitalBrains1: What code do you have in mind where knowing `WORD_SIZE_IN_BITS` would help?

On a 32-bit OS, the current code is a no-op and there was no issue to be fixed, the original code already worked fine.

On a 64-bit OS, the current code is broken, it only works for power-of-two data sizes.

If however we add a "modulo size" inside a `--pragma translate off`, that fixes the issue, it doesn't need `WORD_SIZE_IN_BITS`.

christiaanb: I was referring to the Haskell side of things.

DigitalBrains1:
Adding `{-# NOINLINE topEntity #-}` in the test bench breaks the test bench. Investigation is needed why this is. I've marked it as draft for now as there clearly is something going on.

DigitalBrains1 (changes requested):
Good job finding the issue!

I think the commits need some cleanup and ordering. Commit https://github.com/clash-lang/clash-compiler/pull/1874/commits/1569c1b7fcbcf079658292d6b5cfa44b2b8163a9 wasn't meant to be part of the PR, it was meant as an easy means of issue isolation/comparison. I think commit https://github.com/clash-lang/clash-compiler/pull/1874/commits/b5427a1a3e8277943bd37ed286bb67db0352a7ae (VHDL: Truncate in Signed.fromInteger when result is small) should come before the commit that changed `tests/shouldwork/Numbers/ShiftRotateBase.hs` (https://github.com/clash-lang/clash-compiler/pull/1874/commits/d422b5352c3956439182768c11f92d44b8cb3c91) and that commit should add the `{-# NOINLINE #-}` as well. That way, every commit passes both CI and manually running the test bench within Clash.

I can do this myself next week, but I wanted to make this change request to prevent it getting merged before I have the time to do that.

christiaanb:
I hope I fixed all the errors now.

DigitalBrains1:
I forgot to update a copyright message on my commit, and the commit message for the last commit had not been updated to reflect the changes in the commit.

With those two fixed, I think we're finally there now :-).

DigitalBrains1 (approved):

PR #1875, opened 2021-07-06
VHDL: don't overflow integer range in RAM/ROM

christiaanb:
Most RAM and ROM functions take an `Int` as either a read or write index. On 64-bit systems `maxBound :: Int` and `minBound :: Int` exceed VHDL's `integer'high` and `integer'low` (on (nearly) all VHDL simulators). This poses an issue, because the blackbox
implementations use `to_integer` to convert the `signed (63 downto 0)` (then translation for `Int`) to VHDL's `integer`. So any value falling outside of the range would trigger an overflow exception in the VHDL simulator.

So we now masks all indexes to fit the `integer'low to integer'high` range.

### DIFF
--- a/tests/shouldwork/Signal/RomNegative.hs
+++ b/tests/shouldwork/Signal/RomNegative.hs
@@ -1,0 +1,27 @@
+module RomNegative where
+
+import Clash.Prelude
+import qualified Clash.Explicit.Prelude as Explicit
+import Clash.Explicit.Testbench
+
+topEntity
+  :: Clock System
+  -> Reset System
+  -> Enable System
+  -> Signal System (Signed 64)
+  -> Signal System (Unsigned 8)
+topEntity = exposeClockResetEnable go
+ where
+  go rd = mux ((< 0) <$> rd) 0 (unpack <$> romFile d256 "memory.list" rd)
+{-# NOINLINE topEntity #-}
+
+testBench :: Signal System Bool
+testBench = done
+  where
+    testInput      = Explicit.register clk rst enableGen minBound
+                       (testInput + 1)
+    expectedOutput = outputVerifier' clk rst $ replicate d10 0
+    done           = expectedOutput . ignoreFor clk rst enableGen d1 0 $
+                       topEntity clk rst enableGen testInput
+    clk            = tbSystemClockGen (not <$> done)
+    rst            = systemResetGen
--- a/changelog/2021-07-14T16_36_27+02_00_vhdl_to_integer_no_overflow
+++ b/changelog/2021-07-14T16_36_27+02_00_vhdl_to_integer_no_overflow
@@ -1,0 +1,1 @@
+FIXED: Don't overflow VHDL's integer type when addressing RAM/ROM in simulation.Addresses are masked to 32 bits to be sure to keep it within the simulator's range.
--- a/changelog/2021-07-16T16_35_52+02_00_mem_xexception
+++ b/changelog/2021-07-16T16_35_52+02_00_mem_xexception
@@ -1,0 +1,1 @@
+CHANGED: RAM/ROM functions: They now throw `XExeception` for out-of-bounds address inputs, so this condition no longer aborts simulation.
--- a/clash-lib/prims/vhdl/Clash_Explicit_BlockRam.primitives
+++ b/clash-lib/prims/vhdl/Clash_Explicit_BlockRam.primitives
@@ -21,13 +21,13 @@
   signal ~GENSYM[rd][4]  : integer range 0 to ~LENGTH[~TYP[5]] - 1;
   signal ~GENSYM[wr][5]  : integer range 0 to ~LENGTH[~TYP[5]] - 1;
 begin
-  ~SYM[4] <= to_integer(~ARG[6])
+  ~SYM[4] <= to_integer(~VAR[rdI][6](31 downto 0))
   -- pragma translate_off
                 mod ~LENGTH[~TYP[5]]
   -- pragma translate_on
                 ;
 
-  ~SYM[5] <= to_integer(~ARG[8])
+  ~SYM[5] <= to_integer(~VAR[wrI][8](31 downto 0))
   -- pragma translate_off
                 mod ~LENGTH[~TYP[5]]
   -- pragma translate_on
@@ -80,13 +80,13 @@ end block;
   signal ~GENSYM[rd][4]  : integer range 0 to ~LIT[5] - 1;
   signal ~GENSYM[wr][5]  : integer range 0 to ~LIT[5] - 1;
 begin
-  ~SYM[4] <= to_integer(~ARG[6])
+  ~SYM[4] <= to_integer(~VAR[rdI][6](31 downto 0))
   -- pragma translate_off
                 mod ~LIT[5]
   -- pragma translate_on
                 ;
 
-  ~SYM[5] <= to_integer(~ARG[8])
+  ~SYM[5] <= to_integer(~VAR[wrI][8](31 downto 0))
   -- pragma translate_off
                 mod ~LIT[5]
   -- pragma translate_on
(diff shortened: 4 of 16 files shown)


tests/shouldwork/Signal/RomNegative.hs
```diff
@@ -0,0 +1,33 @@
...
+import qualified Clash.Explicit.Prelude as Explicit
+import Clash.Explicit.Testbench
+
+zeroAt0
+  :: HiddenClockResetEnable dom
+  => Signal dom (Unsigned 8)
+  -> Signal dom (Unsigned 8)
+zeroAt0 a = mux en a 0
+  where
+    en = register False (pure True)
```
DigitalBrains1, line 13: We have something for this already! :-) I think this is just `ignoreFor d1` : [ignoreFor](https://hackage.haskell.org/package/clash-prelude-1.4.2/docs/Clash-Prelude-Testbench.html#v:ignoreFor). I feel inclined to make it part of `testBench` rather than `topEntity`.

DigitalBrains1: This `zeroAt0` construction is already used extensively in other tests, so it fits perfectly with its peers. Only one of the ten other instances puts it in `testBench`, despite my personal preferences ;-).

martijnbastiaan (commented):
@DigitalBrains1 Feel free to create a PR removing all instances of `zeroAt` while replacing it with `ignoreFor`. Seems a lot neater!

I don't understand this patch: wouldn't failing loudly be preferable over silently failing and introducing a Clash simulation vs RTL simulation difference? Wouldn't you only see this issue if you're surpassing your simulator's capabilities anyway?

christiaanb:
@martijnbastiaan isn't the point of the patch clear from the testsuite addition? In Haskell, we wouldn't see the exception (X value) from indexing in the rom with a negative address, becausee we won't observe the output of the rom when the address is negative. However, when we naively translate to concurrent VHDL statements, the VHDL simulator would choke on those exceptions, in this case two even:
1. Indexing with an out-of-bound index (already handled before this pathc)
2. to_integer fails when the argument is outside of the integer range (handled in this patch)
This will unlikely pose a problem in practice, until we get blockRAMs with more than 2^31-1 addressable words. It's also less of a problem because Haskell simulations will likely run out of memory when you have 2^31-1 addressable words.

DigitalBrains1:
> In Haskell, we wouldn't see the exception (X value) from indexing in the rom with a negative address, becausee we won't observe the output of the rom when the address is negative.

Are you sure about this? Maybe I'm misunderstanding what you are saying, but I don't see where that happens...

```
$ cabal run clashi
Up to date
Clashi, version 1.5.0 (using clash-lib, version 1.5.0):
https://clash-lang.org/  :? for help
Loaded package environment from /home/peter/src/clash/xilinxfloat/.ghc.environment.x86_64-linux-8.10.4
Loaded Clashi configuration from /home/peter/dotfiles-git/home/.clashi
Clash.Prelude P> :l RomNegative.hs 
[1 of 1] Compiling RomNegative      ( RomNegative.hs, interpreted )
Ok, one module loaded.
*RomNegative P> sampleN 10 testBench 
[False,False*** Exception: Ix{Int}.index: Index (-9223372036854775808) out of range ((0,255))
*RomNegative P> 
```

DigitalBrains1:
Thinking about it, it could very well be that the issue shows up if the testbench is changed a bit such that the output isn't observed when the address is negative, leading to a normally running testbench in Haskell, but failing in VHDL simulation because of the concurrent nature.

So perhaps the issue is real and the fix is fine as well, but the testbench needs tweaking.

christiaanb:
Try running the test added in this PR on head of master, and you’ll see.

DigitalBrains1:
The test fails in `master` when run through GHDL.

The test fails in both `master` and this branch when run in Clash.

The latter is the problem here (well, alternatively, there is no problem, because the test fails in Clash, so GHDL can do that as well).

christiaanb:
Ah, right, the output needs to be ignored for 2 cycles, not just on the first.

DigitalBrains1:
All addresses presented by the `testBench` are negative, so all of them will give an exception. As I mentioned on our Slack, I hit a wall while trying to change the test bench. I don't know why, but it insists on evaluating `romFile` even though the `mux` correctly behaves in the following.

```diff
--- a/tests/shouldwork/Signal/RomNegative.hs
+++ b/tests/shouldwork/Signal/RomNegative.hs
@@ -20,14 +20,14 @@ topEntity
   -> Signal System (Unsigned 8)
 topEntity = exposeClockResetEnable go where
   go rd = zeroAt0 (unpack <$> dout) where
-    dout = romFile d256 "memory.list" rd
+    dout = mux ((< 0) <$> rd) 0 (romFile d256 "memory.list" rd)
 {-# NOINLINE topEntity #-}
 
 testBench :: Signal System Bool
 testBench = done
   where
     testInput      = Explicit.register clk rst enableGen minBound (testInput + 1)
-    expectedOutput = outputVerifier' clk rst $(listToVecTH [0::Unsigned 8,0,1,2,3,4,5,6,7,8])
+    expectedOutput = outputVerifier' clk rst (replicate d10 0)
     done           = expectedOutput (topEntity clk rst enableGen testInput)
     clk            = tbSystemClockGen (not <$> done)
     rst            = systemResetGen
```
gives
```haskell
>>> P.takeWhile not $ sample testBench
[False,False*** Exception: Ix{Int}.index: Index (-9223372036854775808) out of range ((0,255))
```

the following change makes it work but defeats the purpose:
```diff
--- a/tests/shouldwork/Signal/RomNegative.hs
+++ b/tests/shouldwork/Signal/RomNegative.hs
@@ -20,7 +20,7 @@ topEntity
   -> Signal System (Unsigned 8)
 topEntity = exposeClockResetEnable go where
   go rd = zeroAt0 (unpack <$> dout) where
-    dout = mux ((< 0) <$> rd) 0 (romFile d256 "memory.list" rd)
+    dout = mux ((< 0) <$> rd) 0 (romFile d256 "memory.list" ((.&. 255) <$> rd))
 {-# NOINLINE topEntity #-}
 
 testBench :: Signal System Bool
```

I have yet to discover where the strictness is coming from.

DigitalBrains1:
And just to make it clear, no, using `ignoreFor` does not work for any value:
```diff
--- a/tests/shouldwork/Signal/RomNegative.hs
+++ b/tests/shouldwork/Signal/RomNegative.hs
@@ -18,9 +18,7 @@ topEntity
   -> Enable System
   -> Signal System (Signed 64)
   -> Signal System (Unsigned 8)
-topEntity = exposeClockResetEnable go where
-  go rd = zeroAt0 (unpack <$> dout) where
-    dout = romFile d256 "memory.list" rd
+topEntity = exposeClockResetEnable (romFile d256 "memory.list")
 {-# NOINLINE topEntity #-}
 
 testBench :: Signal System Bool
@@ -28,6 +26,7 @@ testBench = done
   where
     testInput      = Explicit.register clk rst enableGen minBound (testInput + 1)
     expectedOutput = outputVerifier' clk rst $(listToVecTH [0::Unsigned 8,0,1,2,3,4,5,6,7,8])
-    done           = expectedOutput (topEntity clk rst enableGen testInput)
+    done           = expectedOutput . ignoreFor d2 0
+                       $ topEntity clk rst enableGen testInput
     clk            = tbSystemClockGen (not <$> done)
     rst            = systemResetGen
```
gives
```haskell
>>> P.takeWhile not $ sample testBench
[False,False*** Exception: Ix{Int}.index: Index (-9223372036854775808) out of range ((0,255))
```
and for `ignoreFor d20` (which should just give mismatched values because the output should be all zeroes for the length we run the test bench) it's the same:
```haskell
>>> P.takeWhile not $ sample testBench 
[False,False*** Exception: Ix{Int}.index: Index (-9223372036854775808) out of range ((0,255))
```

The `romFile` is evaluated too eagerly, but the test bench's logic is wrong for any value of `ignoreFor`, it should be the `mux` from my previous comment.

DigitalBrains1:
There were two remaining issues which I just fixed:

* The new test bench in https://github.com/clash-lang/clash-compiler/commit/4e139c48314aff98c93860cc549e4346dca1f899 did not test the "overflow VHDL integer range" problem at all.
* The order of commits, first https://github.com/clash-lang/clash-compiler/commit/fca9ac28c7bfda5a45e0b1aaff72ba163033da14 then https://github.com/clash-lang/clash-compiler/commit/4e139c48314aff98c93860cc549e4346dca1f899,  meant that if someone were to run the added test bench in Clash in https://github.com/clash-lang/clash-compiler/commit/fca9ac28c7bfda5a45e0b1aaff72ba163033da14, that would fail. While we don't run test benches in Clash in current CI, it seems bad form to have a failing test bench committed to master. So I reversed the order.

DigitalBrains1:
Ah, and I added a missing changelog and updated copyright messages :-)

PR #1876, opened 2021-07-14
clash-cores New: Xilinx float sub, mul and div

DigitalBrains1:
Single-precision floating point operations, realized through instantiation of the Xilinx Floating-Point LogiCORE IP v7.1.

Further work on `Clash.Cores.Xilinx.Floating`, with more to come.

## Still TODO:


martijnbastiaan (approved):
Should we provide a non-`DSignal` variant too? I'm asking because I feel like `DSignal` (while cool) isn't nearly as ubiquitous as `Signal` in Clash designs. Doesn't have to happen for this PR though.

PR #1880, opened 2021-07-23
Add fromInteger @Signed template to known

DigitalBrains1:
Per PR #1120, we have a list of known `BlackBoxFunction`s and
`TemplateFunction`s, achieving quicker startup. Commit 2e4096a9f added a
new one, and this should be added to the list. VHDL generation now no
longer emits the line

```
Hint: Interpreting Clash.Primitives.Sized.Signed.fromIntegerTF
```

I don't think this needs a Changelog. It's a fix on `master` for an issue from a change that was merged 5 days ago and isn't in a release.

I noticed that the name `fromIntegerTF` is more ambiguous than the other names used under the qualifier `P`. It could be that in the future a slight change is needed to disambiguate this `fromIntegerTF` from a future TF with the same name.

## Still TODO:


alex-mckenna (approved):
> I don't suppose your :thumbsup: on PR #1880 is a formal approving review?

No, but this is

PR #1887, opened 2021-08-02
clash-cores New: UART

rowanG077:
Adds a simple 8N1 UART with type level configurable baud rate. Has been verified in hardware on a Lattice ECP5 upto a baudrate of 921600. Mostly because my USB uart cable doesn't go higher then this. Originally I had wanted to use the design in RetroClash but this pulled in many internal dependencies that I didn't really want so I implemented this instead.

## Still TODO:


martijnbastiaan:
Thanks @rowanG077! The CI issues are being looked into.

DigitalBrains1:
I saw last week that @rowanG077 had asked for a UART a while back (somehow Slack flat out refuses to send me any notifications for those two public Slack channels, and I haven't figured out why). It came to me that perhaps the UART I wrote in 2015 and ported last year could be a nice thing to include in `clash-cores`. It can be found [here](https://github.com/DigitalBrains1/clash-toolbox/blob/28aebaf98692e46deaf48d615fe82679336e61cc/src/Toolbox/Serial.hs). In hindsight, I should have said something last week, but I thought @rowanG077 was going to use the RetroClash one.

I'm also perfectly fine with using this one. And mine does some non-idiomatic things with calculation of the baud rate divider due to the fact it was written in a version of Clash that was already obsolete even then (and perhaps my level of experience back then). One thing my UART does add, though, is supersampling. Every bit is divided into 16 periods, and the levels in the three periods around the midpoint of the bitperiod are sampled and a majority-vote is taken to filter glitches.

Perhaps it would be good to equip this UART with a glitch filter as well. I think this can be added later in an API-preserving way; it need not be part of this PR. Although with the current implementation of baud rate generator, it would make the imprecision of the baud rate slightly worse. With the current design here, the fractional part of the division of clock rate and baud rate is the error in the generated baud rate. If you increase the baud rate 16-fold, the error grows as well. If this were to become an issue, we could add, again in an API-preserving manner, a fractional baud rate generator that is more precise. The [clock scaler](https://github.com/DigitalBrains1/clash-toolbox/blob/28aebaf98692e46deaf48d615fe82679336e61cc/src/Toolbox/ClockScaler.hs) in the repository that also holds my UART implementation actually does not accumulate any error at all, but it can cost quite a lot of resources in the FPGA depending on the rates.

The baud rate of a UART needn't be super precise; it needs to be precise enough to have both parties agree where to sample 10 bits (8N1). After that, they re-synchronise on the next start bit. Given a high enough frequency of the circuit, the baud rate will probably be good enough in practice.

alex-mckenna:
CI hopefully shouldn't cause you any more problems, but if it does just @ me

Another question: when you test this on your lattice board, does it infer some lattice-specific UART IP (or is there one available)? If so, maybe we also want some function that produces this via a Clash black box (but still uses the implementation you've given here for simulation). [We already do something similar for SPI](https://github.com/clash-lang/clash-compiler/blob/master/clash-cores/src/Clash/Cores/SPI.hs#L330).

rowanG077:
@alex-mckenna  Afaik SPI uses a special IP because it requires a tri-state buffer. Uart is much simpler. I have never heard of a hard UART in an FPGA. So it doesn't  infer anything special.

rowanG077:
@DigitalBrains1 It's true that that are two possible improvement points of my UART. I didn't implement an especially accurate period because as you say UART can be off 10% and still function fine. At the clock rates an FPGA runs at I don't expect you will ever run into this limitation.

But  a glitch filter should indeed be better. I will implement this since it's only a couple of lines.

DigitalBrains1:
The code is pleasant to read, by the way! I like it.

Actually I think 10% is a bit much. I like the description of UARTs in the documentation of AVR microcontrollers; if I look at [Asynchronous Operational Range](https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2486-8-bit-AVR-microcontroller-ATmega8_L_datasheet.pdf#page=144) in the ATMega8 datasheet, it's more like 4.6% to be shared between sender and receiver, so 2.3% each if both have comparable error. But like you say, since an FPGA will run at a rate much higher than is usual as a bitrate for asynchronous serial, the error will stay low.

That 4.6% from that datasheet is with taking the middle three samples of a 16× higher samplerate, by the way. Noise can also chip away at the percentage, because the receiver might not have synchronised perfectly on the start edge. Sometimes, serial lines are relatively long and unshielded; noise is a real thing there.

On the bottom line, I don't think the baudrate error will be a problem with an FPGA design. And if it does occur, we can always make a version which takes its clock as an input (I don't mean the Clash `Clock` datatype, that seems overkill), and people can make the clock as precise and as runtime configurable as they want.

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,207 @@
...
+uartRxT state@UartRxDataBits{..} b = (nextState, Nothing)
+  where
+    periodDone = uartRxPeriodCounter == maxBound
+    lastBit = uartRxWordCounter == maxBound
+    finishedReceive = periodDone && lastBit
+
+    nextState | finishedReceive = UartRxStopBit { uartRxWord = uartRxWord }
+              | periodDone      = UartRxDataBits
+                                     { uartRxWord = replaceBit (7 :: Index 8)
+                                                      b (shiftR uartRxWord 1)
```
martijnbastiaan, line 158: We should have a [`+>>`](https://hackage.haskell.org/package/clash-prelude-1.4.2/docs/Clash-Explicit-Prelude.html#v:-43--62--62-) for bitvector.

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,207 @@
...
+       }
+   | UartRxDataBits
+   -- ^ Receive data bits
+       { uartRxWord :: BitVector 8 -- The data received
+       , uartRxPeriodCounter :: Index period -- Counter that overflows after a
+                                             -- full uart period
+       , uartRxWordCounter :: Index 9 -- Which bit of uart we are currently
+                                      -- waiting for. This is 9 because we also
+                                      -- wait until we check the stop bit.
+       }
```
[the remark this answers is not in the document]
martijnbastiaan: Same comment for other records

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,207 @@
...
+   . KnownNat period
+  => UartTxState period
+  -- ^ The current state
+  -> Maybe (BitVector 8)
+  -- ^ The word to transmit
+  -> ( UartTxState period -- The next state
+     , ( Bit -- The output bit to the tx line
+       , Bool -- An acknowledgement that the word to transmit has been received
+       )
+     )
```
martijnbastiaan, line 55: Could you move the documentation to a single line and add `-- ^`? That way it'll show up in the Haddock.

martijnbastiaan: This shows up a couple of times in this PR, so same for the others :)

martijnbastiaan (changes requested):
Looks good overall. (And I really can't wait for `NoFieldSelectors` and `RecordDotSyntax` to be usable for us.)

@DigitalBrains1 I'm trusting you verified the actual functionality is OK, so I can just do the easy part of criticizing minor things :-)

@rowanG077 If you need some lightweight method of annotating return types, check out [NamedTypes](https://hackage.haskell.org/package/clash-prelude-1.4.2/docs/Clash-NamedTypes.html)

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,207 @@
...
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE NumericUnderscores #-}
+
+module Clash.Cores.UART(HzToPeriod, uartTx, uartRx, uart) where
+
+import Clash.Prelude
+
+-- | Convert frequency in hertz to period in picoseconds
+type HzToPeriod (rate :: Nat) = Div 1_000_000_000_000 rate
```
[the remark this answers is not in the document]
rowanG077: I agree there are multiple cases where I have used this. A PWM module, An SDRAM module and more...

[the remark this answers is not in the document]
DigitalBrains1: The first I saw of it was in a [message from Érdi Gergő](https://groups.google.com/g/clash-language/c/vmPHLqdy4rQ/m/feoTrfr7CAAJ?utm_medium=email&utm_source=footer). We ended up [making hzToPeriod round exactly the same](https://groups.google.com/g/clash-language/c/DYKLt_v_Xjg/m/PYgXRl31AQAJ), for least‍-‍surprising behaviour. I think adding it to Prelude makes a lot of sense.

clash-cores/src/Clash/Cores/UART.hs
rowanG077: Fixed

clash-cores/src/Clash/Cores/UART.hs
rowanG077: I changed the internal representation to a Vec and convert at the interface boundary.

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,207 @@
...
+       , uartRxPeriodCounter :: Index period -- Counter that overflows after a
+                                             -- full uart period
+       , uartRxWordCounter :: Index 9 -- Which bit of uart we are currently
+                                      -- waiting for. This is 9 because we also
+                                      -- wait until we check the stop bit.
+       }
+   | UartRxStopBit
+   -- ^ Check the stop bit is correct
+       { uartRxWord :: BitVector 8 -- The completely received word
+       }
```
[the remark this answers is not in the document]
rowanG077: Fixed

clash-cores/src/Clash/Cores/UART.hs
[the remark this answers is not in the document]
rowanG077: I agree but I think that should be a seperate PR.

clash-cores/src/Clash/Cores/UART.hs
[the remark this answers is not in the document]
rowanG077: Fixed

rowanG077:
@DigitalBrains1 Thank you! 

I'm fine either way. I can add glitch filtering and a better clock generator to this but it's also fine if we add your UART. I'm not particularly invested in this implementation or anything. So it's your call.

clash-cores/src/Clash/Cores/UART.hs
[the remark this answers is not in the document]
alex-mckenna: Somewhat unrelated, but I also have

```haskell
data TimeUnit = Ps | Ns | Us | Ms | Sec | Min | Hr

type Ps  = 'Ps
type Ns  = 'Ns
type Us  = 'Us
type Ms  = 'Ms
type Sec = 'Sec
type Min = 'Min
type Hr  = 'Hr

-- Turn it into picoseconds
type FromUnit :: TimeUnit -> Nat -> Nat
type family FromUnit u n where
  FromUnit 'Ps  n = n
  FromUnit 'Ns  n = n * 1000
  FromUnit 'Us  n = n * 1000000
  FromUnit 'Ms  n = n * 1000000000
  FromUnit 'Sec n = n * 1000000000000
  FromUnit 'Min n = n * 60000000000000
  FromUnit 'Hr  n = n * 600000000000000
```

So I can write things like

```haskell
FromUnit Sec 10
```

instead of actually bothering to figure out how long some amount of time is. Maybe this is also useful?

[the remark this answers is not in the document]
gergoerdi: I have [a simpler version of this](https://github.com/gergoerdi/retroclash-lib/blob/668e723bda32ec689600dd17574b1ac5cc008fa5/src/RetroClash/Clock.hs#L19-L25), is there any benefit to a separate `FromUnit` type family?

```
type Seconds      (s  :: Nat) = Milliseconds (1_000 * s)
type Milliseconds (ms :: Nat) = Microseconds (1_000 * ms)
type Microseconds (us :: Nat) = Nanoseconds  (1_000 * us)
type Nanoseconds  (ns :: Nat) = Picoseconds  (1_000 * ns)
type Picoseconds  (ps :: Nat) = ps
```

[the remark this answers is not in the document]
alex-mckenna: > is there any benefit to a separate FromUnit type family

No, this is just the first thing I happened to write. Then I threw it into a _bag-of-miscellany_ module and didn't look at it again

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,204 @@
...
+  deriving (Show, Generic, NFDataX)
+
+-- | Create a word that includes start a stop bits from a raw word
+createTxWord
+  :: Vec 8 Bit
+  -- ^ The word to transmit
+  -> Vec 10 Bit
+  -- ^ The word including start and stop bit
+createTxWord d = (1 :> Nil) ++ d ++ (0 :> Nil)
```
DigitalBrains1, line 43:
```suggestion
createTxWord d = (1 :> d)  :< 0
```

rowanG077: Fixed

clash-cores/src/Clash/Cores/UART.hs
martijnbastiaan: Ah sorry for the confusion, I wasn't suggesting you'd change anything. The version on bitvectors is probably better in terms of simulation performance (both memory and cpu) - so for now that's preferable. I was just saying (worded poorly) that I think we should add `+>>`/`<<+`/..  equivalents for bitvectors at some point.

rowanG077: I have reverted it. In a perfect world `Vec n Bit` and `BitVector n` would really be the same.

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,205 @@
...
+    nextState = UartTxTransmitting
+                  { uartTxWord = createTxWord d
+                  , uartTxPeriodCounter = 0
+                  , uartTxWordCounter = 0
+                  }
+uartTxT UartTxIdle Nothing = (UartTxIdle, (1, False))
+uartTxT state@UartTxTransmitting{..} _ = (nextState, (lsb uartTxWord, False))
+  where
+    periodDone = uartTxPeriodCounter == maxBound
+    lastBit = uartTxWordCounter == maxBound
```
DigitalBrains1, line 67: `uartTxWordCounter` is superfluous: since the last bit in `uartTxWord` is always a one (stop bit), this suffices:
```suggestion
    lastBit = uartTxWord' == repeat 0
```
when
```
  uartTxWord' = 0 +>> uartTxWord
```
(this is equal to the record update a few lines down)

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,205 @@
...
+                                   , uartRxPeriodCounter = 0
+                                   , uartRxWordCounter = 0
+                                   }
+              | halfPeriodDone = UartRxIdle
+              | otherwise      = UartRxStartBit
+                                   { uartRxHalfPeriodCounter = uartRxHalfPeriodCounter + 1 }
+uartRxT state@UartRxDataBits{..} b = (nextState, Nothing)
+  where
+    periodDone = uartRxPeriodCounter == maxBound
+    lastBit = uartRxWordCounter == maxBound
```
DigitalBrains1, line 152: Getting rid of `uartRxWordCounter` is a bit more involved than getting rid of `uartTxWordCounter`, but still possible. If we initialize `uartRxWord` to `1 :> repeat 0`, we can tell that we've gotten all databits as soon as we shift out a 1; while we are still receiving a word, we shift out zeroes.

DigitalBrains1:
> @DigitalBrains1 Thank you!
> 
> I'm fine either way. I can add glitch filtering and a better clock generator to this but it's also fine if we add your UART. I'm not particularly invested in this implementation or anything, and your implementation already has those! So it's your call.

I prefer your version, and I think the clock generation is something we can let lie until someone makes a good case that it is annoying for them. Because I think the current API is nicely concise, and I'd rather add an alternative rather than replace the current API which will suffice for many applications.

I do think supersampling with majority voting and reducing the size of the state by getting rid of `uartTxWordCounter` and `uartRxWordCounter` are nice to have, but they can be implemented in a later PR as well, they need not hold up this PR if you'd rather have it merged as is.

clash-cores/src/Clash/Cores/UART.hs
DigitalBrains1: So this comment is slightly outdated due to the update from `uartTxWord :: Vec 10 Bit` to `uartTxWord :: BitVector 10`.

```suggestion
    lastBit = uartTxWord' == 0
```

```
  uartTxWord' = shiftR uartTxWord 1
```

DigitalBrains1:
```suggestion
    lastBit = uartTxWord == 1
```
if you'd rather match on current state. Sometimes you overlook the obvious :-).

rowanG077:
@DigitalBrains1 I have added your suggestion. I didn't notice I could get rid of those counters so easily, thanks! I also added a simple super sampler over half the uart period. I'm not sure what exactly you had in mind so if this is not what you meant please let me know.

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,211 @@
+{-|
+  Copyright   :  (C) 2021, LUMI GUIDE FIETSDETECTIE B.V.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  Rowan Goemans <goemansrowan@gmail.com>
+
+  UART transmitter and receiver module
+-}
+
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
martijnbastiaan, line 10: I'm guessing this is for `baudRate`? Could we make that an `SNat`? Ambiguous types can make debugging type errors incredibly hard/unexpected.

rowanG077: Yes! It works without it if `SNat` is used. So fixed.

DigitalBrains1:
>  I'm not sure what exactly you had in mind so if this is not what you meant please let me know.

Ah, no, I meant like the AVR UART (as I quoted before) does: [Asynchronous Data Recovery](https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2486-8-bit-AVR-microcontroller-ATmega8_L_datasheet.pdf#page=143) (how they do the start bit is one page earlier). You split each bit period into 16 equal sized segments, and you sample the three middle ones. You can just add the bits in an `Unsigned 2`, and then the MSB of that `Unsigned 2` is your resultant data bit: it is high if there were two or three high bits sampled.

DigitalBrains1:
> I didn't notice I could get rid of those counters so easily, thanks!

Ah, you needed 9 bits because you also time the stop bit in the `UartRxDataBits` state! If you move that, you only need 8 bits. I've done what I meant in this commit: https://github.com/clash-lang/clash-compiler/commit/61e4911ce11b094891d30c69d58a5fcb98f1dd7e on a temporary branch [peter/uart](https://github.com/clash-lang/clash-compiler/tree/peter/uart)

rowanG077:
@DigitalBrains1  I changed the super sampling method which is now more inline with what is described in the AVR UART document. I didn't use a counter because I like to keep this super sampling separate from the main state machine as this would introduce a lot of extra clutter.

DigitalBrains1:
Thanks for all the work! I have to leave now unfortunately, I'll have a look later.

rowanG077:
Confirmed to still work in hardware after the changes with a clock of 50Mhz and a 921600 baudrate by streaming an image: 

Uploading vid.mp4…

A single lost word would desync this animation. It's a little laggy because uart is not that fast.

martijnbastiaan:
@kloonbot run_ci  2e359bc

DigitalBrains1:
Haha, that's a cool way to demonstrate it works!

But I think there might be some issues with the sampling yet.

I couldn't readily tell from the code where the samples lie in the bitperiod, so I drew [a picture](https://wavedrom.com/editor.html?%7Bsignal%3A%20%5B%0A%20%20%7Bname%3A%20%27b%27%2C%20wave%3A%271.0...............%3D....%27%2C%20data%3A%270%27%7D%2C%0A%20%20%7Bname%3A%20%27b%20samples%27%2C%20wave%3A%27z%3D%3D%3Dz....%3D%3D%3Dz..........%27%2C%20data%3A%27a%20b%20c%20d%20e%20f%27%7D%2C%0A%20%20%7Bname%3A%20%27supersampled%27%2C%20wave%3A%271...0.................%3D%27%2C%20phase%3A%20-0.25%2C%20data%3A%270%27%2C%20node%3A%27....a.......b%27%7D%2C%0A%20%20%7Bname%3A%20%27supersampled%20samples%27%2C%20wave%3A%27z%3D%3D%3Dz....%3D%3D%3Dz..........%27%2C%20phase%3A%20-0.25%2C%20data%3A%27a%20b%20c%20d%20e%20f%27%7D%2C%0A%5D%2C%0A%20%20edge%3A%20%5B%27a%3C-%3Eb%20half%20bit%20period%27%5D%2C%0A%20foot%3A%20%7Btick%3A%20-1%7D%7D%0A%0A).

At the top the synchronised UART RxD. At the bottom the 16 sampling periods for the start bit are indicated, same numbering as in the ATMega8 datasheet. The falling edge of the start bit is detected in the shift register as samples a, b and c. `supersampled` lags two clock cycles behind, shown here as a slight shift to the right. `uartRxT` waits half a bitperiod, so the start bit is sampled in the shift register as samples d, e and f. This seems to be right: `b` seems to be sampled at supersample numbers 8, 9 and 10, perfectly in the middle as in the ATMega datasheet. It should be perfectly in the middle between edges for the best immunity to noise and clock rate error.

However, I constructed a test input for the UART receiver that will only produce the correct output when the supersampler indeed samples numbers 8, 9 and 10, and that test indicated that samples 7, 8 and 9 were sampled instead. By moving the start bit test in `uartRxT` a measly one clock cycle later, we get the right samples, so it looked at the state of the supersampling register one single clock cycle before it was updated.

For this test, the bit rate was chosen such that the clock scalers are exact: the clock frequency is exactly 64 times the bit rate.

This test and the one clock cycle fix is in commit https://github.com/clash-lang/clash-compiler/commit/bc44edf09108bd2e1391e9ae0ea25a70b5ada50c. I made the clock source in the test explicitly 100 MHz: if we were to ever change `System`, the `sampleN` in your tests might not be enough samples, and my new test would break altogether.

Furthermore, in this new supersampling design, there are two clock scalers running, one for the supersampling, and one in `uartRxT`. And because both might very well not be exact, that leads to a problem. Since the supersampling period (the period of the shift register) is either `shiftR uartPeriod 4` or `shiftR uartPeriod 4 + 1`, the bitperiod might very well either gain or lose compared to the supersampling period. Losing even a single clock cycle means we shift to sampling supersamples 7, 8 and 9. We might lose enough samples to end up sampling 6, 7 and 8, or gain enough to sample 9, 10 and 11.

Additionally, it is also a waste of the size of the state to have two clock scalers that are supposed to be in lock step.

I had a go at changing the implementation such that the FSM only progresses when the right supersamples are in the shift register (mind the two clock cycle delay). So there is only a single clock divider clocking the supersampling, and there is a counter `superSampleCnt` that counts off the 16 supersamples in each bit period. Additionally, the FSM also runs in the idle state; either picking up the start edge or moving past the "output result" stage so there is only a single clock cycle where we produce a result. It is in commit https://github.com/clash-lang/clash-compiler/commit/d6ea4c6dfe81c6408f6aba0188eae0583a8e42a4, but it doesn't work yet. Thing is, stuff has come up which means I don't have the time to work on this anymore. I've already spent more time than I should, and I really need to stop know. Sorry!

A difference with the AVR UART is that in your solution, the starting edge is also subjected to a low-pass filter. I'm not sure whether or not that's a good thing for clock recovery, but I'm fine with it anyway, it should work well enough.

By the way, I noticed you put yourself as maintainer of the module. We can also shoulder that responsibility! We are fine with assuming maintainership of this code if you want.

rowanG077:
Thanks for the pointers! I will finish this then this weekend. I will try to backup my implementation with some calculations to ensure the error is within limits. I will implement a proper clock scaler as well. If we want to truly minimize the error this is necessary anyway.

DigitalBrains1:
> If we want to truly minimize the error this is necessary anyway.

I was looking at these clock scalers wrongly. I thought their relative drifting was more of an issue than the imprecision of the rate in the first place. I think the right way to look at it is: of the two, the bitperiod counter is the most precise, so it most accurately times the true center of the bitperiod. That _is_ the place you want to look at your supersampling output. The point where relative drift between the counters causes the sampling to jump from one supersample window to another is a bit jumpy, but it is still close to the center of the incoming data bits.

So we don't need to truly minimise the error.

_However_. The code I dumped here as not working here yesterday was getting rather ugly, IMO, I wrote something similar that looks better now. And it turns out that if you get rid of the bitperiod counter and just express all receive timing in the supersampling counter, the error in the clock scaling does become significant. The test no longer works for 921,600 baud due to the drift. See commit https://github.com/clash-lang/clash-compiler/commit/6c2dd857ecef51d692eee0bd2e7df77acd6d6125.

So if we want to get rid of a large part of the state by using a single counter, we need a more accurate baudrate generator.
 I have an idea for a fractional baudrate generator. Let's see if I can write it quickly.

rowanG077:
@DigitalBrains1  I had already written a baud rate generator but didnt integrate it yet. Feel free to use it:

```
data UartClockScaler (dom :: Domain)
   = UartClockScaler
       { uartEnx16 :: Signal dom Bool
       , uartEnx2  :: Signal dom Bool
       , uartEnx1  :: Signal dom Bool
       }

uartClockScaler
  :: forall (baudRate :: Nat)
            (dom :: Domain)
            (tick :: Nat)
   . ( HiddenClockResetEnable dom
     , KnownNat (DomainPeriod dom)
     , KnownNat tick
     , tick ~ Div (16 * baudRate * 2 ^ 16) (PeriodToHz (DomainPeriod dom))
     , 1 <= baudRate
     , 1 <= (DomainPeriod dom)
     )
  => SNat baudRate
  -> UartClockScaler dom
uartClockScaler SNat = UartClockScaler x16 x2 x1
  where
    tick :: Unsigned 21
    tick = snatToNum $ SNat @tick
    superSampleCnt = register tick (fmap (tick +) superSampleCnt)
    bitsChanged = register 0 $ liftA2 xor (register 0 superSampleCnt) superSampleCnt
    x16 = fmap (flip testBit 16) bitsChanged
    x2 = fmap (flip testBit 19) bitsChanged
    x1 = fmap (flip testBit 20) bitsChanged
```

rowanG077:
@DigitalBrains1 
Actually do you have time to meet about this via the slack channel or something? I think we can resolve this much faster that way then keep going back and forth in this issue tracker 😄.

DigitalBrains1:
Oh, that looks like a nice clock scaler, I didn't know that particular method. I'll just believe you when you say it works properly :-). Mine has a subtraction that this method avoids by inverting the division, by the looks of it.

Actually, I don't really have the time but I felt bad about leaving it like this, so I'm jamming some coding into my schedule. Since you already have a clock scaler with better precision, I''d like to just call it a day now. In fact, I've taken a few days off and won't come back to work until Wednesday. But I'll keep looking here a bit.

If you feel fine with the version I put in commit https://github.com/clash-lang/clash-compiler/commit/6c2dd857ecef51d692eee0bd2e7df77acd6d6125 combined with your fractional baudrate generator, I suggest we go with that solution.

rowanG077:
@DigitalBrains1 I think it's good now implementation wise

@martijnbastiaan The haddock is ugly in some places because of long constraints :( I'm not sure how to fix that. Is it possible to make haddock break that up into multiple lines? For example:

2021-08-07-211209_2494x1771_scrot

And also shouldn't this package be released on hackage? Or how do you intend for users to find and use it?

alex-mckenna:
> I'm not sure how to fix that.

Maybe there's a sensible synonym you can give for things like the type equality constraint between `ValidBaudrate` and `UartBaudGenTick`? I'm not sure there's a way to make haddock nicely layout things like large constraint tuples / types

> And also shouldn't clash-cores as a package be released on hackage? Or how do you intend for users to find and use it?

We have talked internally about this a few times, I believe the current feeling is that we shouldn't make an actual release until we can guarantee some degree of API stability (esp. for things like the xilinx floating point core). The finer points of API stability for that would be for @DigitalBrains1 to discuss.

But yes, it should absolutely be published on hackage _at some point_

rowanG077:
I just found out that if write multiple constraints not as a tuple but as separate constraints then haddock will break the second constraint after a `forall` to the next line. I have rewritten some constraints like this and also removed all unnecessary quantifications. This solves the problem.

martijnbastiaan:
> I just found out that if write multiple constraints not as a tuple but as separate constraints then haddock will break every constraint after the constraint in the forall to the next line. 

That's good to know. We settled on using constraint tuples in our style guide a few years back, but this seems a great argument for using separate `=>`.

> But yes, it should absolutely be published on hackage at some point

Couldn't we release it as 0.1 / experimental?

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+  -> Signal dom Bit
+  -- ^ The UART receive line
+  -> Signal dom (Maybe (BitVector 8))
+  -- ^ The received words
+uartRxNoBaudrateGen (BaudrateGenerator enable) b = register Nothing out
+  where
+    out = mux (isRising False $ isJust <$> fsmOut) fsmOut (pure Nothing)
+    (uartS, fsmOut) = unbundle
+                        $ regEn (UartRxIdle, Nothing) enable
+                        $ uartRxT <$> uartS <*> supersampled
```
DigitalBrains1, line 265:
```suggestion
    (uartS, fsmOut) =
      unbundle $ uartRxT <$> regEn UartRxIdle enable uartS <*> supersampled
```

`fsmOut` is getting a lot of registers, let's drop one since it doesn't seem very useful.

I tried the same in `uartTxNoBaudrateGen`, but it was being obstinate. The extra register on `ackOut`  there would have been much more difficult to eliminate, and I wonder if it is at all possible to reduce the size of the state, so it can stay.

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+    nextCounter = uartRxSampleCounter + 1
+    out | validStopBit = Just uartRxWord
+        | otherwise    = Nothing
+
+    nextState | middleSample = UartRxIdle
+              | otherwise    = UartRxStopBit uartRxWord nextCounter
+uartRxT _ _ = (UartRxIdle, Nothing)
+
+-- | The UART receiver.
+-- This version requires the baud rate generator to be passed in. Create one
```
DigitalBrains1, line 250:
```suggestion
-- | The UART receiver
--
-- This version requires the baud rate generator to be passed in. Create one
```

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+                        $ regEn (UartRxIdle, Nothing) enable
+                        $ uartRxT <$> uartS <*> supersampled
+    syncB = register 1 $ register 1 b
+    -- Supersample according to doc:
+    -- https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2486-8-bit-AVR-microcontroller-ATmega8_L_datasheet.pdf#page=143
+    shiftReg = regEn (repeat 1 :: Vec 3 Bit) enable (liftA2 (+>>) syncB shiftReg)
+    majority s = boolToBit $ testBit (popCount $ v2bv s) 1
+    supersampled = register 1 $ fmap majority shiftReg
+
+-- | The UART receiver.
```
DigitalBrains1, line 273:
```suggestion
-- | The UART receiver
```

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+  => SNat baudrate
+  -- ^ The UART baudrate
+  -> Signal dom Bit
+  -- ^ The UART receive line
+  -> Signal dom (Maybe (BitVector 8))
+  -- ^ The received words
+uartRx baud rx = uartRxNoBaudrateGen (baudrateGenerator baud) rx
+
+-- | The UART module. consists of both a UART transmitter and a UART receiver.
+-- This version requires the baud rate generator to be passed in. Create one
```
DigitalBrains1, line 286:
```suggestion
-- | The UART module; consists of both a UART transmitter and a UART receiver
--
-- This version requires the baud rate generator to be passed in. Create one
```

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+  :: HiddenClockResetEnable dom
+  => BaudrateGenerator dom
+  -- ^ The baudrate generator
+  -> Signal dom Bit
+  -- ^ UART receive line
+  -> Signal dom (Maybe (BitVector 8))
+  -- ^ transmit request
+  -> (Signal dom (Maybe (BitVector 8)), Signal dom Bit, Signal dom Bool)
+  -- ^ Words that have been received, the UART transmit line and the
+  -- acknowledgement that the word to transmit has been received.
```
DigitalBrains1, line 298:
```suggestion
  -- ^ Words that have been received, the UART transmit line and the
  -- acknowledgement that the word to transmit has been received
```

clash-cores/test/Test/Cores/UART.hs
```diff
@@ -0,0 +1,211 @@
...
+
+import           Clash.Prelude (exposeClockResetEnable)
+import           Clash.Explicit.Prelude hiding (interleave)
+
+import           Clash.Cores.Internal.UART
+
+import qualified GHC.Float as F
+
+import Debug.Trace (trace)
+
```
DigitalBrains1, line 27: The tests in the testbench should not write stuff to stdout or stderr (outside of the mechanisms provided by Tasty). This remnant of debugging should be removed in the commit.

clash-cores/test/Test/Cores/UART.hs
```diff
@@ -0,0 +1,211 @@
...
+-- | Ensures the uart clock scaler has <-0.625% error over a single uart "session"
+-- We use a baudrate of 606060 because a single supersampled
+-- uart period(transmitting/receiving a single bit) takes 165.000165 clock cycles.
+-- This is the worst cases for errors since a mismatch of a single
+-- clock cycles has the highest impact.
+baudrateGeneratorErr
+  :: forall (baudrate :: Nat)
+   . baudrate ~ 606060
+  => Bool
+baudrateGeneratorErr = trace (show errPercent) (errPercent <= 0.625)
```
DigitalBrains1, line 44: See previous comment.

```suggestion
baudrateGeneratorErr = errPercent <= 0.625
```

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+import Data.Maybe (isJust)
+
+-- | Division that rounds the result
+type DivRound (a :: Nat) (b :: Nat) = Div (a + (Div b 2)) b
+
+-- | Utility type that represents a second
+type Second = 1_000_000_000_000 -- picoseconds
+
+-- | Calculate frequency from period at type level
+type PeriodToHz (period :: Nat) = DivRound Second period
```
DigitalBrains1, line 27: As noted [in this mailing list message](https://groups.google.com/g/clash-language/c/DYKLt_v_Xjg/m/PYgXRl31AQAJ), we have defined the term-level `hzToPeriod` to give the same result as `Div`, not `DivRound`. This was chosen to keep the type-level thing simple, arguing that if you need the picosecond precision that is the difference between them, you should be specifying a period rather than a rate. This is also in the 1.4.0 [changelog](https://hackage.haskell.org/package/clash-prelude-1.4.0/changelog), if you search for 1253, you'll find:
> * `hzToPeriod` now takes a `Ratio Natural` rather than a `Double`. It rounds slightly differently, leading to more intuitive results and satisfying the requested change in #1253. Clash expresses clock rate as the clock period in picoseconds. If picosecond precision is required for your design, please use the exact method of specifying a clock period rather than a clock frequency.

So if we add `HzToPeriod` to our Prelude, it will most likely be just `Div`, not `DivRound`. `PeriodToHz` is actually the same function, though they obviously do not round-trip because of rounding.

I suggest you use just `Div` here. I don't expect `UartBaudGenTick` to create a worse tick for it, right?

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
+{-|
+  Copyright   :  (C) 2021, LUMI GUIDE FIETSDETECTIE B.V.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  Rowan Goemans <goemansrowan@gmail.com>
```
DigitalBrains1, line 4: Do you want to keep the responsibility of maintainership? We can do that as well!

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+  -- ^ transmit request
+  -> (Signal dom (Maybe (BitVector 8)), Signal dom Bit, Signal dom Bool)
+  -- ^ Words that have been received, the UART transmit line and the
+  -- acknowledgement that the word to transmit has been received.
+uartNoBaudrateGen baudGen rxB txM = (rxM, txB, ackTx)
+  where
+    rxM = uartRxNoBaudrateGen baudGen rxB
+    (txB, ackTx) = uartTxNoBaudrateGen baudGen txM
+
+-- | The UART module. consists of both a UART transmitter and a UART receiver.
```
DigitalBrains1, line 304:
```suggestion
-- | The UART module; consists of both a UART transmitter and a UART receiver
```

clash-cores/src/Clash/Cores/Internal/UART.hs
```diff
@@ -0,0 +1,317 @@
...
+  => ValidBaudrate dom baudrate
+  => SNat baudrate
+  -- ^ The baudrate
+  -> Signal dom Bit
+  -- ^ UART receive line
+  -> Signal dom (Maybe (BitVector 8))
+  -- ^ transmit request
+  -> (Signal dom (Maybe (BitVector 8)), Signal dom Bit, Signal dom Bool)
+  -- ^ Words that have been received, the UART transmit line and the
+  -- acknowledgement that the word to transmit has been received.
```
DigitalBrains1, line 316:
```suggestion
  -- acknowledgement that the word to transmit has been received
```

DigitalBrains1 (changes requested):
* I've changed the punctuation of the documentation such that only full sentences end in a period.
* Could you please change all type and variable names of the form `Baudrate` to `BaudRate` (and `baudrate` to `baudRate`)? If we look at it from a linguistic point of view, the word "baudrate" does not exist. And neither does the phrase "baud rate", but at least that's a common perversion in this field of work. "Baud" is a unit with the same dimension as "hertz". So "baud rate" makes no more sense than "hertz rate" does, but it's so common I'm fine with it. However, it is still two words, so it should have a capital R. We went wrong here ourselves with the phrase "test bench" and it is always annoying having to remember whether a specific type or module or function is "Testbench" or "TestBench". I'd rather not commit more of these mistakes.

DigitalBrains1:
> > But yes, it should absolutely be published on hackage at some point
> 
> Couldn't we release it as 0.1 / experimental?

As long as we are clear that at this point in time, we don't give any API guarantees for `Xilinx.Floating`, that's fine with me!

rowanG077:
@DigitalBrains1 To be sure we are withing error limits I evaluated the error function for all baud rates larger then 4800. Not rounding made a small difference but not really significant. But it did show that 16 bit counter is not really enough for very low baud rates.

This is the error percentage of the baud rate generator plotted vs the requested baud rate at 100Mhz main clock:

With a 16 bit counter:
withoutRound16bit

With a 20 bit counter:
withoutRound20bit

It's possible to calculate an appropriate width of the counter at the type level. But when I did that I had to increase `fconstraint-solver-iterations` to 6 which I don't think we want to do since it will leak into user code. Another option would be to just use a large enough counter that should work for even extreme cases (500Mhz clock with 50 baud or something).

alex-mckenna:
> As long as we are clear that at this point in time, we don't give any API guarantees for Xilinx.Floating, that's fine with me!

I have no experience using hackage as a maintainer, but it might be possible to have just a release candidate and no actual releases for the package. Maybe that's what we want?

christiaanb:
> 
> 
> > As long as we are clear that at this point in time, we don't give any API guarantees for Xilinx.Floating, that's fine with me!
> 
> I have no experience using hackage as a maintainer, but it might be possible to have just a release candidate and no actual releases for the package. Maybe that's what we want?

I think so.. but what's the point? just hosting documentation?

alex-mckenna:
> I think so.. but what's the point? just hosting documentation?

More thinking out loud. It's a point best left to slack though instead of cluttering up this PR so I'll stop :upside_down_face:

DigitalBrains1:
> @DigitalBrains1 To be sure we are withing error limits I evaluated the error function for all baud rates larger then 4800. Not rounding made a small difference but not really significant. But it did show that 16 bit counter is not really enough for very low baud rates.

Great work on the baud rate generator, with the fancy custom error message, the unit tests and the analysis!

It seems that rounding (on 16 bit) doesn't help the worst case: it seems we're trading about -1.35% error for +1.5% error! Even the worst cases on an interval don't seem better all along the analysis range. Maybe it helps in other cases than the worst case?

> It's possible to calculate an appropriate width of the counter at the type level. But when I did that I had to increase `fconstraint-solver-iterations` to 6 which I don't think we want to do since it will leak into user code. Another option would be to just use a large enough counter that should work for even extreme cases (500Mhz clock with 50 baud or something).

I think it's fine as it is now. And since you now also have the option of passing in a baud rate generator, that's another way out for people with "extreme" demands. If we document that a `BaudRateGenerator` runs at 16× the actual baud rate and is `True` for a single cycle with that 16× rate, people can just provide their own tailored funtions. A comment added to the `baudRateGenerator` function indicating the maximal ratio between clock rate and baud rate would be helpful, so people can determine from that when to resort to a different solution.

PS: I thought it would help precision by changing this integral computation:

```
(16 * baudRate * 2^16) / (10^12 / domainPeriod)
```

to

```
(16 * baudRate * 2^16 * domainPeriod) / 10^12
```

because you do all multiplications before the division. But thinking about it, the relative error in `10^12 / domainPeriod` is only large if `domainPeriod` is really large, so it's not helpful in practice.

rowanG077:
I did a different method. `baudGenerator` will work in every case now. Close to the threshold where it will become inaccurate it will instead use `riseEvery`. Since it only becomes inaccurate when the difference between the clockrate and baud is very large this method is then accurate enough. 

Unit test are passing. I can test this in hardware on thursday.

clash-cores/clash-cores.cabal
```diff
@@ -81,6 +81,8 @@ library
     Clash.Cores.Xilinx.Floating.Explicit
     Clash.Cores.Xilinx.Floating.Internal
     Clash.Cores.SPI
+    Clash.Cores.UART
+    Clash.Cores.Internal.UART
```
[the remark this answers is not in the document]
rowanG077: Fixed. I removed the Internal module.

clash-cores/src/Clash/Cores/Internal/UART.hs
rowanG077: I have set `QBayLogic B.V. <devops@qbaylogic.com>` to the maintainer.

rowanG077:
I have now implemented the `baudGenerator` that it will automatically choose whether to use the overflowing counter based approach or using Clash `riseEvery`. This works well since at the point that the counter based is approach is inaccurate the `riseEvery` approach is accurate enough so it's always safe to switch:

So here for some graphs at the threshold where it switches between the counter based approach and `riseEvery`:

At 1Mhz:

1Mhz

at 20Mhz:

20Mhz

at 100Mhz:

100Mhz

at 500Mhz:

500Mhz

rowanG077:
@DigitalBrains1 I have also implemented all your review comments. I removed the usage of baud rate entirely and just use baud. I think this is the most accurate.

DigitalBrains1:
@kloonbot run_ci 5e374a4

rowanG077:
Yeah quite confident this works correctly now:

At 100 Mhz:
100Mhz

At 50 Mhz:
50Mhz

At 25 Mhz:
25Mhz

At 10 Mhz:
10Mhz

At 5 Mhz:
5Mhz

At 1 Mhz:
1Mhz

at 0.1Mhz:
0 1Mhz

martijnbastiaan:
@rowanG077 Kloonbot only listens to maintainers, sorry. It's a precautionary measure as the GitLab tests run on our local machines / networks.

@kloonbot run_ci 79693e4

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,357 @@
...
+     , 1 <= DomainPeriod dom
+     , 1 <= 16 * (DomainPeriod dom)
+     , 1 <= baud
+     )
+
+-- | Generates an accurate baudrate for usage with the UART modules
+--
+-- If multiple UARTs are instantiated with the same baudrate it's more
+-- efficient to have a single `baudGenerator` and manually pass them to
+-- the @*NoBaudrateGen@ functions.
```
DigitalBrains1, line 101:
```suggestion
-- | Generates an accurate baud for usage with the UART modules
--
-- If multiple UARTs are instantiated with the same baud it's more efficient to
-- have a single `baudGenerator` and manually pass them to the @*NoBaudGen@
-- functions.
```

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,357 @@
...
+      | doneTx     = case mD of
+                       Just d  -> (UartTxTransmitting (createTxWord d) 0, True)
+                       Nothing -> (UartTxIdle, False)
+      | periodDone = (UartTxTransmitting nextWord 0, False)
+      | otherwise  = (UartTxTransmitting uartTxWord nextCounter, False)
+
+-- | The UART transmitter
+--
+-- This version requires the baud rate generator to be passed in. Create one
+-- using `Clash.Cores.UART.baudGenerator`.
```
DigitalBrains1, line 187:
```suggestion
-- This version requires the baud generator to be passed in. Create one using
-- `Clash.Cores.UART.baudGenerator`.
```

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,357 @@
...
+        | otherwise    = Nothing
+
+    nextState | middleSample = UartRxIdle
+              | otherwise    = UartRxStopBit nextCounter uartRxWord
+uartRxT _ _ = (UartRxIdle, Nothing)
+
+-- | The UART receiver
+--
+-- This version requires the baud rate generator to be passed in. Create one
+-- using `Clash.Cores.UART.baudGenerator`.
```
DigitalBrains1, line 291:
```suggestion
-- This version requires the baud generator to be passed in. Create one using
-- `Clash.Cores.UART.baudGenerator`.
```

clash-cores/src/Clash/Cores/UART.hs
```diff
@@ -0,0 +1,357 @@
...
+  -> Signal dom Bit
+  -- ^ The UART receive line
+  -> Signal dom (Maybe (BitVector 8))
+  -- ^ The received words
+uartRx baud rx = uartRxNoBaudGen (baudGenerator baud) rx
+
+-- | The UART module; consists of both a UART transmitter and a UART receiver
+--
+-- This version requires the baud rate generator to be passed in. Create one
+-- using `Clash.Cores.UART.baudGenerator`.
```
DigitalBrains1, line 327:
```suggestion
-- This version requires the baud generator to be passed in. Create one using
-- `Clash.Cores.UART.baudGenerator`.
```

DigitalBrains1 (commented):
> I removed the usage of baud rate entirely and just use baud. I think this is the most accurate.

Haha, I hope this wasn't influenced by me calling _baud rate_ a perversion, that was tongue-in-cheek 😁. I noticed you forgot to change the reference `*NoBaudrateGen` in the Haddock for `baudGenerator` (intended to refer to `uartTxNoBaudGen`, `uartRxNoBaudGen` and `uartNoBaudGen`). So I grepped the source for the word _rate_ and got rid of a few remnant stragglers, for completeness' sake. Apart from that one reference I don't think it's actually important, but you're making such an effort in this PR I'd like to put a bow on it :-D.

DigitalBrains1 (changes requested):

clash-cores/src/Clash/Cores/UART.hs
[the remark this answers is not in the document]
DigitalBrains1: I'm resolving this change requesst for now, if and when we add `HzToPeriod` to Prelude, we can change the UART code to use it then.

martijnbastiaan (approved):

rowanG077:
@DigitalBrains1 Fixed the comments. I will ping here tomorrow when I verified with actual hardware. It took a lot more effort then I expected to get it in but at least it now should for all baud rates and if not you get a type error 😊

martijnbastiaan:
@kloonbot run_ci  efb6def

DigitalBrains1:
> It took a lot more effort then I expected to get it in but at least it now should for all baud rates and if not you get a type error

I know that feeling. "Well this shouldn't be hard to implement"... queue montage indicating a lot of time passes while the programmer sits coding at their computer non-stop... food and beverage on screen for mere fractions of a second...

But the end result is really awesome! Automatic switching between implementations, precisely at the point where they begin to outperform one another, so cool. And all the worst-case analysis! \o/

I'll merge this!

DigitalBrains1:
Furthermore, we'll look into this:
```haskell
-- | Any function that instantiates a baud generator
-- with polymorphic arguments requires this constraint.
type ValidBaud (dom :: Domain) (baud :: Nat)
   = ( KnownNat (UartBaudGenTick dom baud)
     , KnownNat (DomainPeriod dom)
     , 1 <= DomainPeriod dom
     , 1 <= 16 * (DomainPeriod dom)
     , 1 <= baud
     )
```

Usually, Clash would be able to infer `1 <= 16 * a` from `1 <= a` by way of the monotonicity of multiplication, but here it fails to do that. We'll investigate why.

rowanG077:
I tested it in hardware under these configs with no issues:

15Mhz and 921600 baud (Close to the edge of 16 baud > clock freq)
125Mhz and 921600 baud
125Mhz and 300 baud
125Mhz and 4800 baud

DigitalBrains1:
I created an issue for the solver stumbling where it was expected to work: https://github.com/clash-lang/clash-compiler/issues/1901

Thanks for doing all those tests! I was already very confident, but the tests are stronger evidence than my feelings.

PR #1891, opened 2021-08-05
Added zeroWidthSpec transformation

### LINKED ISSUE #508: Clash isn't smart enough when constant folding zero-width constructs
Consider:

```haskell
data AB = A | B

ab :: KnownNat n => Index n -> AB -> AB
ab n A = if n >  0 then A else B
ab n B = if n == 0 then B else A
{-# NOINLINE ab #-}

topEntity = ab @1
{-# NOINLINE topEntity #-}
```

After normalization, the type specialized ab becomes:

```
λ(eta4755801206503243778 :: Clash.Sized.Internal.Index.Index
                              1) ->
λ(eta14755801206503243777 :: Imap.AB) ->
letrec
  #case_alt2817 :: Imap.AB
  = case Clash.Sized.Internal.Index.gt# @1
           eta4755801206503243778
           (Clash.Sized.Internal.Index.fromInteger# @1
              1
              0) of
      GHC.Types.False3891110078048108556   ->
        Imap.B8214565720323842355
      GHC.Types.True3891110078048108589   ->
        Imap.A8214565720323842354
  result2819 :: Imap.AB
  = case eta14755801206503243777 of
      Imap.A8214565720323842354   ->
        #case_alt2817
      Imap.B8214565720323842355   ->
        Imap.B8214565720323842355
in result2819
```

Netlist generation correctly deduces that - because the index is a zero-width construct - it can throw away the first argument:

```
entity imap_ab is
  port(eta1   : in imap_types.ab;
       result : out imap_types.ab);
end;

architecture structural of imap_ab is
  signal \#case_alt\               : imap_types.ab;
  signal \#case_alt_selection_res\ : boolean;

begin
  \#case_alt_selection_res\ <= false;

  \#case_alt\ <= "0" when \#case_alt_selection_res\ else
                 "1";

  with (eta1) select
    result <= \#case_alt\ when "0",
              "1" when others;
end;
```

This whole entity boils down to "1", which Clash could have recognized if only it had realized that:

```
eta4755801206503243778 :: Clash.Sized.Internal.Index.Index 1 

~ 

0 :: Clash.Sized.Internal.Index.Index 1
```

alex-mckenna:
A new transformation has been defined, which replaces lambda-bound variables for a zero-width type with the only possible value of that type in the body of the lambda. e.g.

```haskell
data AB = A | B

ab :: KnownNat n => Index n -> AB -> AB
ab n A = if n >  0 then A else B
ab n B = if n == 0 then B else A
```

would see the values of n in the function body be replaced with 0 (the only possible value of the type `Index 1`)

```haskell
ab _ A = if 0 > 0 then A else B
ab _ B = if 0 == 0 then B else A
```

the normalizer is then free to further normalize this to

```haskell
ab _ _ = B
```

where previously it would not be able to improve this during normalization.

Fixes #508

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize/Strategy.hs
+++ b/clash-lib/src/Clash/Normalize/Strategy.hs
@@ -1,7 +1,8 @@
 {-|
-  Copyright  :  (C) 2012-2016, University of Twente
+  Copyright  :  (C) 2012-2016, University of Twente,
+                (C) 2021,      QBayLogic B.V.
   License    :  BSD2 (see the file LICENSE)
-  Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+  Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
   Transformation process for normalization
 -}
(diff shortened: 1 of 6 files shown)


alex-mckenna:
I've changed the hard-coded strings for type constructors / `fromInteger#` to be TH quotes for @martijnbastiaan

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -137,6 +138,7 @@ constantPropagation =
     specTransformations =
       [ ("typeSpec"    , typeSpec)
       , ("nonRepSpec"  , nonRepSpec)
+      , ("zeroWidthSpec", zeroWidthSpec)
```
christiaanb, line 141: The CL says:

> the normalizer is then free to further normalize this to

But that only happens accidentally, as a result of e.g. `caseCon` being part of `anf`. Or as a result of the late-stage function hierarchy flattening/inline that happens after regular normalization.

It's worth adding a note wrt the above.

christiaanb (approved):

clash-lib/src/Clash/Normalize/Strategy.hs
alex-mckenna: I've left a source note in the `Strategy` module for this

christiaanb: Thanks, LGTM

PR #1896, opened 2021-08-08
introduce `WorkIdentity Int [Int]` and corresponding `inlineIdentity`…

### LINKED ISSUE #779: Remove redundant generate mappings from verilog output
The verilog output of https://github.com/ncihnegn/clash-with-stack/blob/master/src/PlayClash/Adder.hs
```
refAdder :: KnownNat n => Bit -> Vec n Bit -> Vec n Bit -> (Bit, Vec n Bit)
refAdder cin a b = (cout, res)
    where cout :> res = bv2v $ (v2bv a `add` v2bv b) + boolToBV (bitToBool cin)
```
 contains two redundant generate mappings:

```
  // map begin
  genvar i_0;
  generate
  for (i_0=0; i_0 < 32; i_0 = i_0 + 1) begin : map_0
    wire  map_in_0;
    assign map_in_0 = b[i_0*1+:1];
    wire [0:0] map_out_0;
      assign map_out_0 = map_in_0;


    assign c$scrut_app_arg_0[i_0*1+:1] = map_out_0;
  end
  endgenerate
  // map end

  // map begin
  genvar i_1;
  generate
  for (i_1=0; i_1 < 32; i_1 = i_1 + 1) begin : map_1
    wire  map_in_1;
    assign map_in_1 = a[i_1*1+:1];
    wire [0:0] map_out_1;
      assign map_out_1 = map_in_1;


    assign c$scrut_app_arg_1[i_1*1+:1] = map_out_1;
  end
  endgenerate
  // map end
```

I'd like to take this task if anyone can point me to the right direction.

ZabelTech:
This patch introduces an noop analysis on Prims to prevent redundant HDL generation.
Closes #779

I deviated from the plan discussed in #779 by separating the identity analysis from `inlineCleanup`.

This is because inline cleanup as is would return the unchanged `Letrec`, if there was nothing found via `isInteresting`.
However there might have been 'uninteresting' terms, that were identified as noops and as such coerced to one of their arguments. In that case, those would have been not evaluated. 
To change that, the check that determines what to return (changed vs. original) needed to be adjusted, which would have made the logic to complex for my taste.
That's why I created a separate `inlineIdentity` inlining stage, which does only one thing and is run directly after `inlineCleanup`

### DIFF
--- a/clash-lib/src/Clash/Core/Type.hs
+++ b/clash-lib/src/Clash/Core/Type.hs
@@ -50,6 +50,7 @@ module Clash.Core.Type
   , findFunSubst
   , reduceTypeFamily
   , undefinedTy
+  , unsafeCoerceTy
   , isIntegerTy
   , normalizeType
   , varAttrs
@@ -662,6 +663,14 @@ undefinedTy =
   let aNm = mkUnsafeSystemName "a" 0
       aTv = (TyVar aNm 0 liftedTypeKind)
   in  ForAllTy aTv (VarTy aTv)
+
+unsafeCoerceTy :: Type
+unsafeCoerceTy =
+  let aNm = mkUnsafeSystemName "a" 0
+      aTv = TyVar aNm 0 liftedTypeKind
+      bNm = mkUnsafeSystemName "b" 1
+      bTv = TyVar bNm 1 liftedTypeKind
+  in ForAllTy aTv (ForAllTy bTv (mkFunTy (VarTy aTv) (VarTy bTv)))
 
 isIntegerTy :: Type -> Bool
 isIntegerTy (ConstTy (TyCon nm)) = nameUniq nm == getKey integerTyConKey
--- a/clash-lib/src/Clash/Normalize/Strategy.hs
+++ b/clash-lib/src/Clash/Normalize/Strategy.hs
@@ -55,6 +55,8 @@ normalization =
     cse        = topdownR (apply "CSE" simpleCSE)
     xOptim     = bottomupR (apply "xOptimize" xOptimize)
     cleanup    = topdownR (apply "etaExpandSyn" etaExpandSyn) >->
+                 -- See [Note] relation `collapseRHSNoops` and `inlineCleanup`
+                 topdownSucR (apply "collapseRHSNoops" collapseRHSNoops) >->
                  topdownSucR (apply "inlineCleanup" inlineCleanup) !->
                  innerMost (applyMany [("caseCon"        , caseCon)
                                       ,("bindConstantVar", bindConstantVar)
--- a/clash-lib/src/Clash/Normalize/Transformations/Inline.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Inline.hs
@@ -24,6 +24,7 @@ module Clash.Normalize.Transformations.Inline
   , inlineCast
   , inlineCleanup
   , inlineHO
+  , collapseRHSNoops
   , inlineNonRep
   , inlineOrLiftNonRep
   , inlineSimIO
@@ -33,8 +34,10 @@ module Clash.Normalize.Transformations.Inline
 
 import qualified Control.Lens as Lens
 import qualified Control.Monad as Monad
-import Control.Monad.Writer (listen)
+import Control.Monad.Trans.Maybe (MaybeT(..))
+import Control.Monad.Writer ((>=>),lift,listen)
 import Data.Default (Default(..))
+import Data.Either  (lefts)
 import qualified Data.HashMap.Lazy as HashMap
 import qualified Data.List as List
 import qualified Data.Maybe as Maybe
@@ -60,12 +63,12 @@ import Clash.Core.Name (Name(..), NameSort(..))
 import Clash.Core.Pretty (PrettyOptions(..), showPpr, showPpr')
 import Clash.Core.Subst
 import Clash.Core.Term
-  ( CoreContext(..), Pat(..), PrimInfo(..), Term(..), collectArgs
+  ( CoreContext(..), Pat(..), PrimInfo(..), Term(..), WorkInfo(..), collectArgs
   , collectArgsTicks, mkApps , mkTicks, stripTicks)
 import Clash.Core.TermInfo (isLocalVar, isPolyFun, termSize, termType)
 import Clash.Core.Type
   (TypeView(..), isClassTy, isPolyFunCoreTy, tyView)
-import Clash.Core.Util (isSignalType)
+import Clash.Core.Util (isSignalType, primUCo)
 import Clash.Core.Var (Id, Var(..), isGlobalId, isLocalId)
 import Clash.Core.VarEnv
   ( InScopeSet, VarEnv, VarSet, elemUniqInScopeSet, elemVarEnv, elemVarSet
@@ -378,6 +381,54 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
 
 inlineCleanup _ e = return e
 {-# SCC inlineCleanup #-}
+
+{- [Note] relation `collapseRHSNoops` and `inlineCleanup`
+The `collapseRHSNoops` transformation replaces functions/primitives that are the identity
+in HDL, but not in Haskell, by `unsafeCoerce`.
+`inlineCleanup` subsequently inlines these `unsafeCoerce` calls.
+The end result of all of this is that we get no/fewer assignments in HDL where the RHS is
+simply a variable reference. See issue #779 -}
+
+-- | Takes a binding and collapses its term if it is a noop
+collapseRHSNoops :: HasCallStack => NormRewrite
+collapseRHSNoops _ (Letrec binds body) = do
+  binds1 <- mapM runCollapseNoop binds
+  return $ Letrec binds1 body
+  where
+    runCollapseNoop orig =
+      runMaybeT (collapseNoop orig) >>= Maybe.maybe (return orig) changed
+
+    collapseNoop (iD,term) = do
+      (Prim info,args) <- return $ collectArgs term
+      identity         <- getIdentity info $ lefts args
+      collapsed        <- collapseToIdentity iD identity
+      return (iD,collapsed)
+
+    collapseToIdentity iD identity = do
+      tcm <- Lens.view tcCache
+      let aTy = termType tcm identity
+          bTy = varType iD
+      return $ primUCo `TyApp` aTy `TyApp` bTy `App` identity
+
+    getIdentity primInfo termArgs = do
+      WorkIdentity idIdx noopIdxs <- return $ primWorkInfo primInfo
+      mapM_ (getTermArg termArgs >=> isNoop >=> Monad.guard) noopIdxs
+      getTermArg termArgs idIdx
+
+    getTermArg args i = do
+      Monad.guard $ i <= length args - 1
+      return $ args !! i
+
+    isNoop (Var i) = do
+      binding     <- MaybeT $ lookupVarEnv i <$> Lens.use bindings
+      isRecursive <- lift $ isRecursiveBndr $ bindingId binding
+      Monad.guard $ not isRecursive
+      isNoop $ bindingTerm binding
+    isNoop (Prim PrimInfo{primWorkInfo=WorkIdentity _ []}) = return True
+    isNoop _ = return False
+
+collapseRHSNoops _ e = return e
+{-# SCC collapseRHSNoops #-}
 
 -- | Inline a function with functional arguments
 inlineHO :: HasCallStack => NormRewrite
--- a/changelog/2021-08-12T12_30_00+02_00_collapse_noops.md
+++ b/changelog/2021-08-12T12_30_00+02_00_collapse_noops.md
@@ -1,0 +1,10 @@
+ADDED: `collapseRHSNoops` inlining stage and `WorkIdentity` constructor.
+
+It is now possible to define primitives to be identical to one of their arguments via the newly introduced `WorkIdentity` constructor.
+This constructor effectivly marks a primitve to be a noop, which further can be conditioned upon multiple of its arguments being noops themselves.
+For an example see `Clash.Sized.Vector.map`.
+
+There is a new inlining stage `collapseRHSNoops` which runs just before `inlineCleanup`.
+It will find noop-primitives defined in such way and `unsafeCoerce#` them to their identity argument.
+
+The goal of all of this is to prevent redundant HDL output. (See Issue #779)
--- a/clash-lib/prims/commonverilog/Clash_Sized_Internal_BitVector.primitives
+++ b/clash-lib/prims/commonverilog/Clash_Sized_Internal_BitVector.primitives
@@ -34,7 +34,7 @@
   }
 , { "BlackBox" :
     { "name"      : "Clash.Sized.Internal.BitVector.pack#"
-    , "workInfo"  : "Never"
+    , "workInfo"  : "Identity 0 []"
     , "kind"      : "Expression"
     , "type"      : "pack# :: Bit -> BitVector 1"
     , "template"  : "~ARG[0]"
@@ -42,7 +42,7 @@
   }
 , { "BlackBox" :
     { "name"      : "Clash.Sized.Internal.BitVector.unpack#"
-    , "workInfo"  : "Never"
+    , "workInfo"  : "Identity 0 []"
     , "kind"      : "Expression"
     , "type"      : "unpack# :: BitVector 1 -> Bit"
     , "template"  : "~ARG[0]"
--- a/clash-lib/prims/verilog/Clash_Sized_Vector.primitives
+++ b/clash-lib/prims/verilog/Clash_Sized_Vector.primitives
@@ -98,7 +98,7 @@ end
   }
 , { "BlackBox" :
     { "name"      : "Clash.Sized.Vector.map"
-    , "workInfo"  : "Never"
+    , "workInfo"  : "Identity 1 [0]"
     , "kind"      : "Declaration"
     , "type"      : "map :: (a -> b) -> Vec n a -> Vec n b"
     , "template"  :
--- a/clash-lib/src/Clash/Core/Term.hs
+++ b/clash-lib/src/Clash/Core/Term.hs
@@ -155,6 +155,10 @@ data WorkInfo
   | WorkAlways
   -- ^ Performs work regardless of whether the variables are constant or
   -- variable; these are things like clock or reset generators
+  | WorkIdentity Int [Int]
+  -- ^ A more restrictive version of 'WorkNever', where the value is the
+  -- argument at the given position if all arguments for the given list of
+  -- positions are also 'WorkIdentity'
   deriving (Eq,Show,Generic,NFData,Hashable,Binary)
 
 -- | Term reference
--- a/clash-lib/src/Clash/Core/Util.hs
+++ b/clash-lib/src/Clash/Core/Util.hs
@@ -428,6 +428,14 @@ primCo
   -> Term
 primCo ty = Prim (PrimInfo "_CO_" ty WorkNever SingleResult)
 
+-- | Make an unsafe coercion
+primUCo :: Term
+primUCo =
+  Prim PrimInfo { primName        = "GHC.Prim.unsafeCoerce#"
+                , primType        = unsafeCoerceTy
+                , primWorkInfo    = WorkNever
+                , primMultiResult = SingleResult }
+
 substArgTys
   :: DataCon
   -> [Type]
--- a/clash-lib/src/Clash/Primitives/Types.hs
+++ b/clash-lib/src/Clash/Primitives/Types.hs
@@ -49,6 +49,7 @@ import qualified Data.Text                    as S
 import           Data.Text.Lazy               (Text)
 import           GHC.Generics                 (Generic)
 import           GHC.Stack                    (HasCallStack)
+import           Text.Read                    (readMaybe)
 
 -- | An unresolved primitive still contains pointers to files.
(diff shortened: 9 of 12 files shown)


ZabelTech:
I deviated from the plan discussed in #779 by separating the identity analysis from `inlineCleanup`.

This is because inline cleanup as is would return the unchanged `Letrec`, if there was nothing found via `isInteresting`.
However there might have been 'uninteresting' terms, that were identified as noops and as such coerced to one of their arguments. In that case, those would have been not evaluated. 
To change that, the check that determines what to return (changed vs. original) needed to be adjusted, which would have made the logic to complex for my taste.
That's why I created a separate `inlineIdentity` inlining stage, which does only one thing and is run directly after `inlineCleanup`

ZabelTech:
currently, the output of 
```
topEntity :: Vec 4 Bit -> Vec 4 Bit
topEntity = bv2v . v2bv
```
is
```
module topEntity
    ( // Inputs
      input  x_0
    , input  x_1
    , input  x_2
    , input  x_3

      // Outputs
    , output wire  result_0
    , output wire  result_1
    , output wire  result_2
    , output wire  result_3
    );
  wire [3:0] x;
  wire [3:0] result;

  assign x = {x_0,   x_1,   x_2,   x_3};

  assign result = (((x)));

  assign result_0 = result[3:3];

  assign result_1 = result[2:2];

  assign result_2 = result[1:1];

  assign result_3 = result[0:0];


endmodule
```

which is think is not ideal ?

alex-mckenna:
Hi @ZabelTech 

Thanks for the good work, that output is what we wanted to see from the issue. While it could be written more nicely by hand, Clash's codegen is fairly systematic so while it should never generate incorrect HDL it may generate some redundant code. What's happening here is basically

  1. `Vec` is split into multiple ports when it appears as input / output of an entity
  2. The original `Vec` is reconstructed and used in the rest of the generated circuit
  3. The `Vec` for output is obtained from the rest of the logic
  4. The output `Vec` is split into multiple ports

So ignoring the redundant extra signals from the vectors being split up (and the redundant parentheses it rendered), what we're left with is effectively `assign out = in;` which is much better than the `generate for` it would produce before.

I'll give it a proper review tomorrow, but one thing that immediately jumps out is the whitespace-only changes from spaces to tabs in some files. If you can revert those to make the diff smaller that would be appreciated

ZabelTech:
kk, fine with me then. also the tab-removal is reverted.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+  (untouched,collapsed) <- partitionEithers <$> mapM identityAnalysis binds
+  if null collapsed
+    then return (Letrec binds body)
+    else changed (Letrec (untouched ++ collapsed) body)
+inlineIdentity _ e = return e
+
+identityAnalysis :: HasCallStack => (Id,Term) -> NormalizeSession (Either (Id,Term) (Id,Term))
+identityAnalysis old@(iD,collectArgs -> (term,partitionEithers -> (termArgs,typeArgs))) =
+  runIdentityAnalysis $ do
+    Prim primInfo <- return term
```
alex-mckenna, line 397: As this is limited to things where `collectArgs` has a `Prim primInfo` in the function position, we could match for this earlier. The line is already quite long so maybe [pattern guards](https://wiki.haskell.org/Pattern_guard) would be more readable, e.g.

```haskell
identityAnalysis old@(collectArgs -> (term, args))
  | Prim primInfo <- term
  , (termArgs, typeArgs) <- partitionEithers args
  = do ...
```

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+    identity      <- getIdentity primInfo
+    let resTy = polyResTyView $ primType primInfo
+    return (iD,primUCo `App` identity `mkTyApps` (resTy:typeArgs))
```
christiaanb, line 400: And then with the change to `primUCo` you can change this to:
```haskell
identityAnalysis old@(iD,collectArgs -> (term,partitionEithers -> (termArgs,_))) =
  runIdentityAnalysis $ do
    Prim primInfo <- return term
    identity      <- getIdentity primInfo
    tcm           <- Lens.view tcCache
    let aTy = termType tcm identity
        bTy = varType iD
    return (iD,primUCo `TyApp` aTy `TyApp` bTy `App` identity)
```

clash-lib/src/Clash/Core/Type.hs
```diff
@@ -348,6 +350,11 @@ splitFunForallTy = go []
     go args (tyView -> FunTy arg res) = go (Right arg:args) res
     go args ty                        = (reverse args,ty)
 
+-- | A view on the polymorphic result type of a polymorphic function
+polyResTyView :: Type -> Type
+polyResTyView (splitFunForallTy -> (lefts -> tvs,resTy)) =
+  foldr ForAllTy resTy tvs
+
```
christiaanb, line 357: This will no longer be needed with the change to `primUCo`

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
alex-mckenna, line 400: With the corrected type for `unsafeCoerce#` this should be the type applications followed by the term application. We have a function `mkApps :: Term -> [Either Term Type] -> Term` which could be used here, i.e.

```haskell
return (iD, mkApps primUnsafeCoerce [Right a, Right b, Left x])
```

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+  where
+    runIdentityAnalysis =
+      fmap (Maybe.maybe (Left old) Right) . runMaybeT
+
+    getTermArg i = do
+      Monad.guard $ i <= length termArgs - 1
+      return $ termArgs List.!! i
```
alex-mckenna, line 407: Stylistically I'm not a fan of qualified operators. I think it would be better to

```haskell
import Data.List ((!!))
```

at the top of the file

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+
+    -- guards that a given term is a noop
+    noopGuard :: Term -> MaybeT NormalizeSession ()
+    noopGuard (Var i) = do
+        binding <- MaybeT $ lookupVarEnv i <$> Lens.use bindings
+        noopGuard $ bindingTerm binding
+    noopGuard (Prim PrimInfo{primWorkInfo=WorkIdentity _ []}) = return ()
+    noopGuard _ = fail "is not a noop"
```
alex-mckenna, line 415: This is the same as `MaybeT (return Nothing)`, but given how infrequently we use `fail` someone reading might not immediately realise that and expect it to behave the same as `error`. I think it would be clearer to either use `MaybeT (return Nothing)` or something like `empty` (from `Alternative`)

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -56,6 +56,7 @@ normalization =
     xOptim     = bottomupR (apply "xOptimize" xOptimize)
     cleanup    = topdownR (apply "etaExpandSyn" etaExpandSyn) >->
                  topdownSucR (apply "inlineCleanup" inlineCleanup) !->
+                 topdownSucR (apply "inlineIdentities" inlineIdentity) >->
```
christiaanb, line 59: I think you want to change these around. Since your implementation of `inlineCleanup` doesn't actually inline the let-binders: it replaces the RHS of a let-binding by a proper call to `unsafeCoerce#`. `inlineCleanup` will do the actual inlining. It's probably worth adding a comment w.r.t. to the above.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -60,12 +63,12 @@ import Clash.Core.Name (Name(..), NameSort(..))
 import Clash.Core.Pretty (PrettyOptions(..), showPpr, showPpr')
 import Clash.Core.Subst
 import Clash.Core.Term
-  ( CoreContext(..), Pat(..), PrimInfo(..), Term(..), collectArgs
-  , collectArgsTicks, mkApps , mkTicks, stripTicks)
+  ( CoreContext(..), Pat(..), PrimInfo(..), Term(..), WorkInfo(..), collectArgs
+  , collectArgsTicks, mkApps , mkTicks, mkTyApps, stripTicks)
```
christiaanb, line 67: The import of `mkTyApp` will no longer be needed with my suggestion.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -60,12 +63,12 @@ import Clash.Core.Name (Name(..), NameSort(..))
...
 import Clash.Core.TermInfo (isLocalVar, isPolyFun, termSize, termType)
 import Clash.Core.Type
-  (TypeView(..), isClassTy, isPolyFunCoreTy, tyView)
-import Clash.Core.Util (isSignalType)
+  (TypeView(..), isClassTy, isPolyFunCoreTy, polyResTyView, tyView)
```
christiaanb, line 70: `polyResTyView` will no longer be needed with my suggestion.

christiaanb (changes requested):
Thanks for the PR! LGTM overall modulo some minor changes.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
 inlineCleanup _ e = return e
 {-# SCC inlineCleanup #-}
 
+-- | Takes a binding and collapses its term if it is a noop
+inlineIdentity :: HasCallStack => NormRewrite
+inlineIdentity _ (Letrec binds body) = do
+  (untouched,collapsed) <- partitionEithers <$> mapM identityAnalysis binds
+  if null collapsed
+    then return (Letrec binds body)
+    else changed (Letrec (untouched ++ collapsed) body)
```
christiaanb, line 391: Perhaps we can leave the order of the let-bindings untouched by changing `identityAnalysis` slightly:
```haskell
inlineIdentity _ (Letrec binds body) = do
  binds1 <- mapM identityAnalysis binds
  return (Letrec binds1 body)
```

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+inlineIdentity _ e = return e
+
+identityAnalysis :: HasCallStack => (Id,Term) -> NormalizeSession (Either (Id,Term) (Id,Term))
```
christiaanb, line 394: By changing the return type:
```haskell
identityAnalysis :: HasCallStack => (Id,Term) -> NormalizeSession (Id,Term)
```

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+identityAnalysis old@(iD,collectArgs -> (term,partitionEithers -> (termArgs,typeArgs))) =
+  runIdentityAnalysis $ do
+    Prim primInfo <- return term
+    identity      <- getIdentity primInfo
+    let resTy = polyResTyView $ primType primInfo
+    return (iD,primUCo `App` identity `mkTyApps` (resTy:typeArgs))
+  where
+    runIdentityAnalysis =
+      fmap (Maybe.maybe (Left old) Right) . runMaybeT
```
christiaanb, line 403: And changing `runIdentityAnalysis` to:
```haskell
runIdentityAnalysis m =
  runMaybeT m >>= Maybe.maybe (return old) changed
```

christiaanb (commented):
Leaving the order of the let-bindings untouched will probably improve the run-time of the compiler.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -379,6 +382,45 @@ inlineCleanup (TransformContext is0 _) (Letrec binds body) = do
...
+    then return (Letrec binds body)
+    else changed (Letrec (untouched ++ collapsed) body)
+inlineIdentity _ e = return e
+
+identityAnalysis :: HasCallStack => (Id,Term) -> NormalizeSession (Either (Id,Term) (Id,Term))
+identityAnalysis old@(iD,collectArgs -> (term,partitionEithers -> (termArgs,typeArgs))) =
+  runIdentityAnalysis $ do
+    Prim primInfo <- return term
+    identity      <- getIdentity primInfo
+    let resTy = polyResTyView $ primType primInfo
```
alex-mckenna, line 399: I don't think this new `polyResTyView` function is what you want. What I think makes more sense is:

  1. Use [`piResultTys`](https://hackage.haskell.org/package/clash-lib-1.4.2/docs/Clash-Core-TermInfo.html#v:piResultTys) to apply the type arguments given to the `Prim` to it's `primType`
  2. Split this type using [`splitFunTys`](https://hackage.haskell.org/package/clash-lib-1.4.2/docs/Clash-Core-Type.html#v:splitFunTys) to get the (now monomorphic) result type and arguments
  3. Index the list of argument types, and use this argument and the result type as type arguments to `unsafeCoerce#`

alex-mckenna (changes requested):

christiaanb:
@ZabelTech seems some of the review suggestions of @alex-mckenna and me overlap. Pick whichever style that has your preference w.r.t. suggested changes.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
ZabelTech: Intuitively I assumed that I needed to call `changed` in case identities have been found and collapsed.
So yeah, if that's not necessary, this can be simplified.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
ZabelTech: ahh, I see that changed has been put in here now. :+1:

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
ZabelTech: mhm, I did realize that, however I think that `empty` does not clearly communicate that the computation gets short-circuited.
`MaybeT (return Nothing)` on the other hand does that clearly, but I think that is 'clunky'.
I think it's best to use the actual `Monad.guard` here.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
ZabelTech: Point taken, we can abort here earlier. 
I'm not a big fan of those pattern-matching + pattern-guards + fallback clauses, so I am going to propose an alternative.

ZabelTech:
@alex-mckenna @christiaanb since this came up with both of you.
Out of curiosity, is there an actual difference between the two? 
`forall a. a -> forall a. a` and `forall a b. a -> b`
I get that it makes sense to give different names to different independent type variables, i.e 'a' and 'b'.
But other than that, are those two definitions not equivalent?

christiaanb:
No, they are not. `forall a b . a -> b` you could type-instantiate with two concrete types first:
```
unsafeCoerce :: forall a b . a -> b
unsafeCoerce @Int @Char :: Int -> Char
unsafeCoerce @Int @Char (0 :: Int) :: Char
```
while you cannot do that with `(forall a . a) -> (forall a . a)`:
```
notUnsafeCoerce :: (forall a . a) -> (forall a.a)
notUnsafeCoerce (0 :: Int) -- type error: expected `forall a . a`, actual: `Int`
```

ZabelTech:
uhm, I think I see now. 
My version specifies a function where the argument- and result-type are polymorphic (i.e. undetermined).
Yours specifies a polymorphic function that is polymorphic in both it's argument and it's result.

So my function can only be called with undefined-like types, not with concrete types. And the result type can only be 'concretised' after the function has been called. It's not ok to do that before.

mkay, but why did it actually work for me then? 
because I have seen quite some type errors while working on this patch, but not with the above definition.

clash-lib/src/Clash/Normalize/Strategy.hs
ZabelTech: right, I'm aware that I do stretch the semantics of 'inline' here a lot.
what's a good name? `collapseRHSNoops` ?

christiaanb:
> mkay, but why did it actually work for me then?
> because I have seen quite some type errors while working on this patch, but not with the above definition.

The types of (sub-)expressions aren't "calculated" all the time, so you can create type-incorrect expressions without necessarily breaking the compiler. So yeah, it's just a "happy accident" that the compiler doesn't error out on the "incorrect" expression.

clash-lib/src/Clash/Normalize/Strategy.hs
christiaanb: Yeah, that sounds good. I'd still like to see a comment along the lines of:
```haskell
{- [Note] relation `collapseRHSNoops` and `inlineCleanup`

The `collapseRHSNoops` transformation replaces functions/primitives that are the identity in HDL, 
but not in Haskell, by `unsafeCoerce`. `inlineCleanup` subsequently inlines these `unsafeCoerce` calls.
The end result of all of this is that we get no/fewer assignments in HDL where the RHS is simply a 
variable reference.

See issue #779
-}
```
in the haddock comments for `collapseRHSNoops` function and then do:
```haskell
-- See [Note] relation `collapseRHSNoops` and `inlineCleanup`
topdownSucR (apply "collapseRHSNoops" collapseRHSNoops) >->
topdownSucR (apply "inlineCleanup" inlineIdentity) !->
```

ZabelTech:
@alex-mckenna @christiaanb your feedback was very helpful and I think we are good now!
please have a second look whenever time permits.

christiaanb (approved):

ZabelTech:
This concern comes to my mind: 
```
    isNoop (Var i) = do
      binding <- MaybeT $ lookupVarEnv i <$> Lens.use bindings
      isNoop $ bindingTerm binding
    isNoop (Prim PrimInfo{primWorkInfo=WorkIdentity _ []}) = return True
    isNoop _ = return False
```

that predicate can potentially result in an infinite loop, in case `Var a` resolves to `Var b` that resolves again to `Var a`, for example.
Can I assume this does not happen, or can I assume that there are no chained vars, i.e. a `Var` never resolves to another `Var`?

christiaanb:
> that predicate can potentially result in an infinite loop, in case `Var a` resolves to `Var b` that resolves again to `Var a`, for example.
> Can I assume this does not happen, or can I assume that there are no chained vars, i.e. a `Var` never resolves to another `Var`?

Good catch, in principle this could happen, but only for self-recursive top-level bindings, e.g.:
```
f :: Int -> Int
f = f
```
but not for mutually recursive top-level bindings, because mutually recursive top-level bindings are turned into self-recursive ones at a very early stage of the compiler, e.g. it transforms:
```
f :: Int -> Int
f = g

g :: Int -> Int
g = f
```
into:
```
f :: Int -> Int
f = let g :: Int -> Int 
        g = f 
     in g

g :: Int -> Int
g = let f :: Int -> Int 
        f = g 
     in f
```
You can check whether you're dealing with a reference to a recursive binder by using: https://github.com/clash-lang/clash-compiler/blob/e6b9e516534131c26397e55886ca0878ea981e67/clash-lib/src/Clash/Normalize/Util.hs#L178-L180
and then use the result as a guard as to whether you'll continue checking with `isNoop` recursively, or return False.

alex-mckenna (approved):
On the whole I'm happy with this. Two things that would tie it all together nicely:

  * an output test, to show that the generated verilog / systemverilog no longer contains redundant `generate` blocks for the topentity in the original issue. You can see [this test](https://github.com/clash-lang/clash-compiler/tree/master/changelog) for an example. You'll want to add it to the testsuite like so

    ```haskell
    clashLibTest "T779" def{hdlTargets=[Verilog,SystemVerilog]}
    ```

    in the `shouldwork.Issues` test group in `tests/Main.hs`.

  * a changelog entry, see [this directory](https://github.com/clash-lang/clash-compiler/tree/master/changelog) for examples

ZabelTech:
@alex-mckenna 
I take it that we deal with a noop for system-verilog as well.
The link to your test did not work btw., so I looked around a bit.
I think it's ok like this, just saying in case you wanted me to do it in a specific way.

Other than that, ready when you are.

alex-mckenna:
> I take it that we deal with a noop for system-verilog as well.

I assumed this, but it turns out I was mistaken. These aren't identity for SystemVerilog for the same reason they aren't in VHDL (you need to do a type conversion in HDL). Sorry about that, can you remove the `mainSystemVerilog` and `SystemVerilog` from the `hdlTargets`.

Other than that, the test you've written looks good :+1:

ZabelTech:
no problem! 
:rocket:

alex-mckenna:
@kloonbot run_ci 0aff6c7

christiaanb:
@ZabelTech So since this was an issue with a bounty on it over at https://www.works-hub.com/issue/524cc2e9-f1d8-4fdf-808e-38e4b9e4032b, we want to transfer the money. Could you contact me by email christiaan@qbaylogic.com so we can transfer via paypal (or some other method).

PR #1900, opened 2021-08-12
Preserve boolean combinatorial logic in HDL

### LINKED ISSUE #1881: GHC inlining Boolean combinational logic.
It looks like GHC automatically inlines combinational logic. The inlining leads to Clash compiling what should be logic gates as muxes.

What follows is an example of code with both the current and desired behaviours:
```
import Data.Bool

import qualified GHC.Magic



-- Custom Boolean operator suggested by Christiaan as a workaround.
(&!) = GHC.Magic.noinline (&&)


-- Sample circuit will return the normally generated result, alongside the fixed/noinlined version

topEntity :: Bool -> Bool -> (Bool, Bool)
topEntity a b = (a && b, a &! b)
```
and the resulting circuit in Verilog:
```
** GENERATED BY CLASH 1.5.0. DO NOT MODIFY.
*/
`timescale 100fs/100fs
module topEntity
    ( // Inputs
      input  a
    , input  b

      // Outputs
    , output wire  result_0
    , output wire  result_1
    );
  wire  c$app_arg;
  wire [1:0] result;

  assign c$app_arg = a ? b : 1'b0;

  assign result = {c$app_arg,   a & b};

  assign result_0 = result[1:1];

  assign result_1 = result[0:0];


endmodule
```

alex-mckenna:
When rendering HDL, clash now better preserves boolean functions from GHC which may be targeted by the simplifier. This helps ensure they are rendered in HDL as the expected operator instead of being rendered as a multiplexer.

Fixes #1881.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/HaskellPrelude.hs
+++ b/clash-prelude/src/Clash/HaskellPrelude.hs
@@ -1,5 +1,6 @@
 {-|
   Copyright   :  (C) 2019, QBayLogic B.V.
+                 (C) 2021, QBayLogic B.V.
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V <devops@qbaylogic.com>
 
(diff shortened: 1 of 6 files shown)


clash-prelude/src/Clash/HaskellPrelude.hs
```diff
@@ -14,11 +15,27 @@ splitAt, tail, take, unzip, unzip3, zip, zip3, zipWith, zipWith3.
...
+(&&) :: Bool -> Bool -> Bool
+(&&) = noinline (Prelude.&&)
+
+infixr 2 ||
+
+(||) :: Bool -> Bool -> Bool
+(||) = noinline (Prelude.||)
+
+not :: Bool -> Bool
+not = noinline Prelude.not
```
martijnbastiaan, line 41: Perhaps a comment on why we put the `noinline` there?

alex-mckenna: That's a helpful note. I'll add one of my own in the source

alex-mckenna: @martijnbastiaan It's done, but it appears above this span of code so the comment doesn't appear as outdated

martijnbastiaan (approved):

PR #1902, opened 2021-08-13
Add missing warnings to Integer/Natural prims

alex-mckenna:
The dynamic-size warning on primitives using Integer / Natural is not consistently included in the .primitives files. This is down to
a mix of legacy (the warning existed after integer prims) and new primitives being added and not having the warning (either by
forgetting or the primitives around them not having the warning).

To be conservative, the warning applies to all primitives in `GHC.Integer.Type`, `GHC.Integer.Logarithms`, `GHC.Num.Integer`, `GHC.Natural` and `GHC.Num.Natural`. As I understand it, the warning only triggers when the primitive is rendered in HDL so this should be okay.

## Still TODO:


martijnbastiaan (approved):
> As I understand it, the warning only triggers when the primitive is rendered in HDL so this should be okay.

This understanding is correct. LGTM!

PR #1908, opened 2021-08-24
Safe Numeric Wrapper Types

alex-mckenna:
This PR introduces a range of newtypes which provide different overflow behaviour. The following modes are provided

  * Erroring, where out of bounds throws an `XException`
  * Saturating, where out of bounds saturates on the maximum / minimum value
  * Wrapping, where out of bounds wraps to the maximum / minimum value
  * Zeroing, where out of bounds always results in 0

Additionally, there is an `Overflowing` type, which behaves like `Wrapping` but also keeps a flag for whether an overflow has occurred at any point during a calculation (this flag can be reset).

Points to consider:

  * In some places, the assumption is made that the default behaviour is to wrap when going out of bounds. While this holds true for types provided by GHC / most types in `clash-prelude`, it does not necessarily hold true.
  * The `fromInteger` / `fromRational` functions do not behave differently from the underlying type, as it is difficult to check bounds here while keeping operations synthesizable.
  * The behaviour of subtraction on `Index` is changed, the old behaviour did not make sense, but this change may break existing code and should be scrutinized appropriately 

See related:

  * #688
  * #1903

## Still TODO:


martijnbastiaan:
Can `Overflowing` be used to simplify [Clash.Class.Counter](https://github.com/clash-lang/clash-compiler/blob/1afe0084b16a0674fb58379d3d44585ff2b47720/clash-prelude/src/Clash/Class/Counter.hs)?

alex-mckenna:
I suppose, you could probably get away with an instance `(Num a) => Counter a` where the body used the `Overflowing` wrapper to implement `countSuccOverflow` / `countPredOverflow`

alex-mckenna:
Although maybe you don't want to change it anyway. You would be limiting counters to `SaturatingNum` in the default case, which is slightly less general than what you have now. Although that requirement could trivially be removed from the `Enum` instance for `Overflowing` and `Saturating`: `satSucc SatWrap` isn't exactly a difficult thing to implement...

martijnbastiaan:
Yeah, agreed. It's probably good enough as it is.

PR #1912, opened 2021-08-26
Store argument strictness in DataCon

alex-mckenna:
The clash `DataCon` type now keeps track of the strictness of data constructor arguments, as determined by the `HsImplBang` in
the GHC core. This can be used in the evaluator / transformations to help ensure that Clash's evaluation matches GHC's more closely.

## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
@@ -60,7 +60,8 @@ import GHC.Core
    collectArgs, rhssOfAlts, unfoldingTemplate)
 import GHC.Core.DataCon
   (DataCon, dataConExTyCoVars, dataConName, dataConRepArgTys, dataConTag,
-   dataConTyCon, dataConUnivTyVars, dataConWorkId, dataConFieldLabels, flLabel)
+   dataConTyCon, dataConUnivTyVars, dataConWorkId, dataConFieldLabels, flLabel,
+   HsImplBang(..), dataConImplBangs)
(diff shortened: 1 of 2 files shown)


clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -730,11 +732,16 @@ coreToDataCon dc = do
...
              , C.dcArgTys      = repTys
+             , C.dcArgStrict   = repBangs
              , C.dcUnivTyVars  = uTvs
              , C.dcExtTyVars   = eTvs
              , C.dcFieldLabels = fLabels
              }
 
+hsImplBangToBool :: HsImplBang -> C.DcStrictness
+hsImplBangToBool HsLazy = C.Lazy
+hsImplBangToBool _ = C.Strict
```
martijnbastiaan, line 743: What about `HsUnpack`? Its documentation says it's strict too (due it's "unpackedness", I guess)

martijnbastiaan (approved):
LGTM once the `hsImplBangToBool` question is cleared up :)

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
alex-mckenna: Unpacking requires strictness, and I couldn't think of any benefit to Clash knowing if something is unpacked or not

martijnbastiaan: If you don't mind, I'd prefer this to be a complete pattern match so we get warnings if there are ever any added constructors. Other than that LGTM!

PR #1913, opened 2021-08-27
Small fixes

alex-mckenna:
## Still TODO:

### DIFF
--- a/.ci/cabal.project.local
+++ b/.ci/cabal.project.local
@@ -10,7 +10,7 @@ package *
 
 package clash-prelude
   ghc-options: -Werror
-  flags: +doctests +multiple-hidden -experimental-evaluator
+  flags: +doctests +multiple-hidden
   tests: True
   benchmarks: True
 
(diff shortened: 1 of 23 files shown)


martijnbastiaan (approved):
It seems like this PR "activates" both evaluators. Could you add a comment (either in the PR description or commit messages, for archaeology purposes) on why this is?

alex-mckenna:
I just noticed that the PR to remove the experimental evaluator flag has this at the start anyway:

> The partial evaluator should not replace the old evaluator
everywhere it is currently used, rather it should _only_ be used
for partial evaluation. This is because the partial evaluator may
not result in something that still repsects invariants (e.g. it
may not preserve invariants expected to hold after ANF).

Which I think covers what you wanted anyway @martijnbastiaan?

martijnbastiaan:
Okay, so my confusion was that I thought PE would be enabled now on `master`. But as it stands, nothing calls it, so it's effectively disabled. Only the PE branch uses it.

alex-mckenna:
> I thought PE would be enabled now on `master`

If I sneak that in on a PR titled "Small fixes" then please somebody just fire me :eyes:

martijnbastiaan:
@christiaanb ^

PR #1915, opened 2021-08-31
Add HasType / InferType classes

alex-mckenna:
This PR introduces some new classes for inspecting things in Clash core:

```haskell
class HasType a where
  coreTypeOf :: a -> Type

class InferType a where
  inferCoreTypeOf :: TyConMap -> a -> Type
```

These functions can then be used to extract type information from any typed _thing_ that exists in core.

## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/PartialEval/Eval.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/PartialEval/Eval.hs
@@ -38,13 +38,13 @@ import           BasicTypes (InlineSpec(..))
 #endif
 
 import           Clash.Core.DataCon (DataCon(..))
+import           Clash.Core.HasType
 import           Clash.Core.Literal (Literal(..))
 import           Clash.Core.PartialEval.AsTerm
 import           Clash.Core.PartialEval.Monad
 import           Clash.Core.PartialEval.NormalForm
 import           Clash.Core.Subst (substTy)
 import           Clash.Core.Term
-import           Clash.Core.TermInfo
 import           Clash.Core.TyCon (tyConDataCons)
 import           Clash.Core.Type
 import           Clash.Core.TysPrim (integerPrimTy)
@@ -602,8 +602,8 @@ apply val arg = do
     f ->
       error ("apply: Cannot apply " <> show arg <> " to " <> show f)
  where
-  -- Somewhat of a cheat, but very quick to implement.
-  valueType tcm = termType tcm . asTerm
+  -- TODO Write an instance for InferType Value and use that instead
+  valueType tcm = inferCoreTypeOf tcm . asTerm
 
 applyTy :: Value -> Type -> Eval Value
 applyTy val ty = do
--- a/clash-lib/src/Clash/Core/HasType.hs
+++ b/clash-lib/src/Clash/Core/HasType.hs
@@ -1,0 +1,220 @@
+{-|
+Copyright   : (C) 2021, QBayLogic B.V.
+License     : BSD2 (see the file LICENSE)
+Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
+
+Utility class to extract type information from data which has a type.
+-}
+
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+
+module Clash.Core.HasType
+  ( HasType(..)
+  , InferType(..)
+  , applyTypeToArgs
+  , piResultTy
+  , piResultTys
+  ) where
+
+import qualified Data.Text as Text (isInfixOf)
+import Data.Text.Prettyprint.Doc (line)
+import GHC.Stack (HasCallStack)
+
+import Clash.Core.DataCon (DataCon(dcType))
+import Clash.Core.FreeVars
+import Clash.Core.Literal (Literal(..))
+import Clash.Core.Name (Name(nameOcc))
+import Clash.Core.Pretty
+import Clash.Core.Subst
+import Clash.Core.Term (Term(..), IsMultiPrim(..), PrimInfo(..), collectArgs)
+import Clash.Core.TyCon (TyConMap)
+import Clash.Core.Type
+import Clash.Core.TysPrim
+import Clash.Core.Var (Var(varType))
+import Clash.Core.VarEnv
+import Clash.Debug (debugIsOn)
+import Clash.Util (pprPanic)
+import qualified Clash.Util.Interpolate as I
+
+class HasType a where
+  coreTypeOf :: a -> Type
+
+instance HasType DataCon where
+  coreTypeOf = dcType
+
+instance HasType Literal where
+  coreTypeOf = \case
+    IntegerLiteral _ -> integerPrimTy
+    IntLiteral _ -> intPrimTy
+    WordLiteral _ -> wordPrimTy
+    StringLiteral _ -> stringPrimTy
+    FloatLiteral _ -> floatPrimTy
+    DoubleLiteral _ -> doublePrimTy
+    CharLiteral _ -> charPrimTy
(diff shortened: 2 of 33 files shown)


clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -0,0 +1,245 @@
...
+    case primMultiResult pr of
+      SingleResult -> primType pr
+      MultiResult
+        | let (primArgs, primResTy) = splitFunForallTy (primType pr)
+        , TyConApp tupTcNm tupArgs <- tyView primResTy
+        , Text.isInfixOf "GHC.Tuple.(" (nameOcc tupTcNm)
+        -> mkPolyFunTy primResTy (primArgs <> fmap Right tupArgs)
+
+        | otherwise
+        -> error "PrimInfo.coreTypeOf: MultiResult primitive without tuple type"
```
alex-mckenna, line 70: This really needs a note here to explain why it behaves as it does, e.g. the rationale behind the old `multiPrimType` that this replaces

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -0,0 +1,245 @@
...
+ where
+  go opTy' []               = opTy'
+  go opTy' (Right ty:args') = goTyArgs opTy' [ty] args'
+  go opTy' (Left _:args')   = case splitFunTy m opTy' of
+    Just (_,resTy) -> go resTy args'
+    _ -> error $ unlines ["applyTypeToArgs:"
+                         ,"Expression: " ++ showPpr e
+                         ,"Type: " ++ showPpr opTy
+                         ,"Args: " ++ unlines (map (either showPpr showPpr) args)
+                         ]
```
alex-mckenna, line 123: While this is being moved it might be nice to neaten up the error message

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -0,0 +1,245 @@
...
+-- variable at a time; instead use 'piResultTys'
+piResultTy
+  :: HasCallStack
+  => TyConMap
+  -> Type
+  -> Type
+  -> Type
+piResultTy m ty arg = case piResultTyMaybe m ty arg of
+  Just res -> res
+  Nothing  -> pprPanic "piResultTy" (ppr ty <> line <> ppr arg)
```
alex-mckenna, line 141: I wonder if this can just be replaced with `piResultTy tcm ty arg = piResultTys tcm ty [arg]` and the `piResultTyMaybe` function then removed

clash-lib/src/Clash/Core/HasType.hs
alex-mckenna: Updated: the doc in `Clash.Normalize.Transformation.MultiPrim` was extended and this now points to that.

clash-lib/src/Clash/Core/HasType.hs
alex-mckenna: Updated: needs to be checked by someone else

clash-lib/src/Clash/Core/HasType.hs
alex-mckenna: Updated: needs to be checked by someone else

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -0,0 +1,222 @@
...
+
+        to something with the non-function type
+
+          #{showPpr opTy'}
+      |]
+
+  goTyArgs opTy' revTys (Right ty:args') = goTyArgs opTy' (ty:revTys) args'
+  goTyArgs opTy' revTys args'            = go (piResultTys m opTy' (reverse revTys)) args'
+
+-- | Like 'piResultTyMaybe', but errors out when a type application is not
```
leonschoorl, line 140: `piResultTyMaybe` was deleted?

alex-mckenna: It was only used to implement `piResultTy`, which I removed (because I didn't see any noticeable change in perf by changing `piResultTy` to be defined in terms of `piResultTys`)

alex-mckenna: Oh, but the comment!

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -0,0 +1,222 @@
...
+
+      Letrec _ x -> go x
+      Case _ _ [] -> undefinedTy
+      Case _ _ (x:_) -> go (snd x)
+      Cast _ _ a -> a
+      Tick _ x -> go x
+
+-- | Get the result type of a polymorphic function given a list of arguments
+applyTypeToArgs
+  :: Term
```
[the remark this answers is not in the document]
alex-mckenna: The comment is on the line below, so this doesn't show as outdated

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -0,0 +1,221 @@
...
+        case collectArgs x of
+          (fun, args) -> applyTypeToArgs x tcm (go fun) args
+
+      x@TyApp{} ->
+        case collectArgs x of
+          (fun, args) -> applyTypeToArgs x tcm (go fun) args
+
+      Letrec _ x -> go x
+      Case _ _ [] -> undefinedTy
+      Case _ _ (x:_) -> go (snd x)
```
christiaanb, line 106: why not just:
```
Case _ ty _ -> ty
```
like we did in `termType`?

Since we seem to have replaced all occurrences of `termType` by `inferCoreType`, traversing the alternative here probably makes the compiler slower.

christiaanb: Also, since we eliminate casts in our GHC2Core pass, when patterns contains existentials, the RHS of the alternative might not even have the correct type. Like:
```
case (x :: Vec n a) of
  Cons m (n ~ m + 1) (y :: a) (ys :: m) -> Cons m CO y ys |> (Vec (m+1) a ~ Vec n a)
```
with the cast removed would return:
```
Vec (m + 1) a
```
instead of the correct type:
```
Vec n a
```

clash-lib/src/Clash/Core/HasType.hs
christiaanb: I mean, if we don't use `piResultTyMaybe` anywhere any longer, we might as well remove it.

alex-mckenna: We don't, it was more of a _maybe this is some perf trick that I'm just not seeing_

clash-lib/src/Clash/Core/HasType.hs
alex-mckenna: > why not just:
>
> Case _ ty _ -> ty
> like we did in termType?

Isn't this wrong though? I thought this was the type of the scrutinee, not the result

> when patterns contains existentials, the RHS of the alternative might not even have the correct type

Yeah... that would be a problem. I guess if it shouldn't be changed back to `Case _ ty _ -> ty` then it needs a note

leonschoorl (approved):
Nice use of typeclasses

clash-ghc/src-ghc/Clash/GHC/PartialEval/Eval.hs
```diff
@@ -603,7 +603,7 @@ apply val arg = do
       error ("apply: Cannot apply " <> show arg <> " to " <> show f)
  where
   -- Somewhat of a cheat, but very quick to implement.
-  valueType tcm = termType tcm . asTerm
```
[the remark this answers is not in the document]
alex-mckenna: This is actually why I wanted this and the `HasFreeVars` class in the first place, but I intended to make those instances in a PE branch instead of here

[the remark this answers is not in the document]
leonschoorl: TODO: add TODO

PR #1917, opened 2021-09-02
Add HasFreeVars class

alex-mckenna:
This PR introduces a new class for inspecting free variables in Clash core:

```haskell
class HasFreeVars a where
  {-# MINIMAL freeVarsOf #-}

  freeVarsOf :: a -> VarSet

  isClosed :: a - > Bool
  isClosed = nullVarSet . freeVarsOf
```

Instances are available for extracting information from a single "thing", e.g. `Term`, as well as a collection of things (any `Foldable f`).

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Core/FreeVars.hs
+++ b/clash-lib/src/Clash/Core/FreeVars.hs
@@ -17,16 +17,8 @@ module Clash.Core.FreeVars
   , freeLocalIds
   , globalIds
   , termFreeTyVars
-  , tyFVsOfTypes
-  , localFVsOfTerms
-  , hasLocalFreeVars
-  -- * Fast
-  , noFreeVarsOfType
   -- * occurrence check
-  , localIdOccursIn
   , globalIdOccursIn
-  , localIdDoesNotOccurIn
-  , localIdsDoNotOccurIn
   , localVarsDoNotOccurIn
   , countFreeOccurances
   -- * Internal
@@ -47,7 +39,7 @@ import Clash.Core.Type                  (Type (..))
 import Clash.Core.Var
   (Id, IdScope (..), TyVar, Var (..), isLocalId)
 import Clash.Core.VarEnv
-  (VarEnv, VarSet, emptyVarEnv, unionVarEnvWith, unitVarSet, unitVarEnv)
+  (VarEnv, emptyVarEnv, unionVarEnvWith, unitVarEnv)
 
 -- | Gives the free type-variables in a Type, implemented as a 'Fold'
 --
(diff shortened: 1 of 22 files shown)


clash-lib/src/Clash/Core/HasFreeVars.hs
```diff
@@ -0,0 +1,50 @@
...
+
+instance HasFreeVars Type where
+  {-# INLINE freeVarsOf #-}
+  freeVarsOf = tyFVsOfTypes . Identity
+
+instance (Foldable f, HasFreeVars a) => HasFreeVars (f a) where
+  {-# INLINE freeVarsOf #-}
+  freeVarsOf = foldMap freeVarsOf
+
+-- TODO I wonder if Clash.Core.FreeVars should have it's functions changed to
```
christiaanb, line 40: I mean... if all uses of  `tyFVsOfTypes` and `localFVsOfTerms` are only within this module, then I see no reason to keep `tyFVsOfTypes` and `localFVsOfTerms`.

alex-mckenna: Yeah, good point

clash-lib/src/Clash/Core/FreeVars.hs
```diff
@@ -139,13 +116,6 @@ localVarsDoNotOccurIn
 localVarsDoNotOccurIn vs e =
   getAll (Lens.foldMapOf freeLocalVars (All . (`notElem` vs)) e)
 
--- | Check whether a local identifier occurs free in a term
-localIdOccursIn
-  :: Id
-  -> Term
-  -> Bool
-localIdOccursIn v e = getAny (Lens.foldMapOf freeLocalIds (Any . (== v)) e)
```
christiaanb, line 147: I think it'd be worthwhile to use our benchmark suite to determine the performance effects of this change. As in, I feel the short-circuiting of `All`s monoidal merging, `&&`, would be faster than building up a varset using `unionVarSet` and then checking whether the element exists. But perhaps laziness saves us once again.

clash-lib/src/Clash/Unique.hs
```diff
@@ -421,6 +422,13 @@ subsetUniqSet
   -> Bool
 subsetUniqSet (UniqSet e1) (UniqSet e2) = IntMap.null (IntMap.difference e1 e2)
 
+-- | Check whether A and B are disjoint
+disjointUniqSet
+  :: UniqSet a
+  -> UniqSet a
+  -> Bool
+disjointUniqSet (UniqSet e1) (UniqSet e2) = IntMap.disjoint e1 e2
```
alex-mckenna, line 430: TODO: It seems `IntMap.disjoint` doesn't exist in the version of `containers` used by GHC 8.6. The definition + relevant copyright notice should be added to a `Data.IntMap.Strict.Extra` module with a note saying it can be removed when GHC 8.6 is no longer supported

christiaanb (approved):

PR #1919, opened 2021-09-07
Remove unused definitions

alex-mckenna:
## Still TODO:


alex-mckenna:
One thing of interest in this PR: the `inlineHO` transformation was amongst the unused definitions I found. Apparently when it was introduced in 9285201 it wasn't enabled (but there was a comment in `Strategy.hs`). Over time that comment disappeared, but the transformation was seemingly never actually in use

christiaanb:
Yeah, it's fine to remove it. At some point in time I was undecided between inlining and specialising. The idea was to measure compile time differences and then add a heuristic to inline a HO function early, or specialise it at the end.

christiaanb (approved):

PR #1920, opened 2021-09-07
Remove Type/Kind distinction

alex-mckenna:
The distinction between `Type` and `Kind` is not particularly relevant since GHC has had `Type :: Type` for a while now. This PR somewhat tentatively removes the synonyms for kinds and replaces `typeKind` with an instance of `InferCoreType` for `Type`.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Core/HasType.hs
+++ b/clash-lib/src/Clash/Core/HasType.hs
@@ -18,6 +18,7 @@ module Clash.Core.HasType
   , piResultTys
   ) where
 
+import Data.Foldable (foldl')
 import qualified Data.Text as Text (isInfixOf)
 import Data.Text.Prettyprint.Doc (line)
 import GHC.Stack (HasCallStack)
@@ -29,12 +30,13 @@ import Clash.Core.Name (Name(nameOcc))
 import Clash.Core.Pretty
 import Clash.Core.Subst
 import Clash.Core.Term (Term(..), IsMultiPrim(..), PrimInfo(..), collectArgs)
-import Clash.Core.TyCon (TyConMap)
+import Clash.Core.TyCon (TyCon(SuperKindTyCon), TyConMap, tyConKind)
 import Clash.Core.Type
 import Clash.Core.TysPrim
 import Clash.Core.Var (Var(varType))
 import Clash.Core.VarEnv
 import Clash.Debug (debugIsOn)
+import Clash.Unique (lookupUniqMap')
 import Clash.Util (pprPanic)
 import qualified Clash.Util.Interpolate as I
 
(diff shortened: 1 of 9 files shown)


clash-lib/src/Clash/Core/TyCon.hs
```diff
@@ -43,7 +43,7 @@ data TyCon
   = AlgTyCon
   { tyConUniq   :: {-# UNPACK #-} !Unique
   , tyConName   :: !TyConName   -- ^ Name of the TyCon
-  , tyConKind   :: !Kind        -- ^ Kind of the TyCon
+  , tyConKind   :: !Type        -- ^ Type of the TyCon
```
leonschoorl, line 46: "Kind" in the comment was correct?

clash-lib/src/Clash/Core/TyCon.hs
```diff
@@ -52,15 +52,15 @@ data TyCon
   | FunTyCon
   { tyConUniq   :: {-# UNPACK #-} !Unique
   , tyConName   :: !TyConName      -- ^ Name of the TyCon
-  , tyConKind   :: !Kind           -- ^ Kind of the TyCon
+  , tyConKind   :: !Type           -- ^ Type of the TyCon
```
leonschoorl, line 55: "Kind" in the comment was correct?

clash-lib/src/Clash/Core/TyCon.hs
```diff
@@ -52,15 +52,15 @@ data TyCon
...
   , tyConArity  :: !Int            -- ^ Number of type arguments
   , tyConSubst  :: [([Type],Type)] -- ^ List of: ([LHS match types], RHS type)
   }
   -- | Primitive TyCons
   | PrimTyCon
   { tyConUniq    :: {-# UNPACK #-} !Unique
   , tyConName    :: !TyConName  -- ^ Name of the TyCon
-  , tyConKind    :: !Kind       -- ^ Kind of the TyCon
+  , tyConKind    :: !Type       -- ^ Type of the TyCon
```
leonschoorl, line 63: "Kind" in the comment was correct?

leonschoorl (commented):
Even GHC still has `type Kind = Type`
And uses that as the kind of Tycon's.

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -106,6 +108,42 @@ instance InferType Term where
...
+    go (LitTy SymTy{}) = typeSymbolKind
+    go (AnnType _ ty) = go ty
+    go (tyView -> FunTy _ res)
+      | let resTy = go res
+      , isSuperKind tcm resTy = resTy
+
+      | otherwise = liftedTypeKind
+
+    go (tyView -> TyConApp tc args) =
+      foldl' kindFunResult (tyConKind (lookupUniqMap' tcm tc)) args
```
christiaanb, line 126:
```suggestion
      piResultTys tcm (tyConKind (lookupUniqMap' tcm tc)) args
```
Now that it's available. GHC does the same: https://github.com/ghc/ghc/blob/62d720db4f6a53014400a608baf5c56555258eee/compiler/GHC/Core/Type.hs#L2776

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -106,6 +108,42 @@ instance InferType Term where
...
+
+    go (AppTy fun arg) =
+      kindFunResult (go fun) arg
```
christiaanb, line 129: Do what GHC does: https://github.com/ghc/ghc/blob/62d720db4f6a53014400a608baf5c56555258eee/compiler/GHC/Core/Type.hs#L2783-L2790

clash-lib/src/Clash/Core/HasType.hs
```diff
@@ -106,6 +108,42 @@ instance InferType Term where
...
+
+    go ConstTy{} =
+      error "inferCoreTypeOf: Naked ConstTy"
+
+kindFunResult :: Type -> Type -> Type
```
christiaanb, line 134: This can be dropped once we use `piResultTys`

christiaanb (commented):
Use `piResultTys` in `inferCoretypeOf`, just like GHC does.

alex-mckenna:
@leonschoorl 

> Even GHC still has `type Kind = Type`
> And uses that for clarity(?) in various places, for example as the kind of Tycon's.

I did think of this, but I think there's an argument that it's more a legacy thing which we don't need to adopt. Back when GHC had

```
Term : Type : Kind : Sort
```

This distinction was the only thing to do, and there was a nice finite hierarchy. But these days what actually exists is

```
Term : Type : Kind : Kind : ...
```

which if you squint is really just

```
Term : Type : Type : ...
```

I don't think there's any _inherent_ benefit to mirroring GHC here on the grounds that

  * you still have an infinite hierarchy in there, as every kind has an inferrable kind
  * having a different name for the base of that hierarchy is odd, every level has the same representation / operations
  * you end up with things which used to accept both having the now useless `KindOrType` synonym which implies the potential for different behaviour if you give it a so-called `Kind` where no such potential actually exists

Differing viewpoints welcome, I'm still slightly on the fence about this PR. If anyone has strong and reasonable objections about lack of clarity etc. from this I can close this PR and open a new one with just @christiaanb's suggestions which remove `kindFunResult`.
 
While I'm writing I'll also respond to

> "Kind" in the comment was correct?

Yes, but I thought it somewhat misleading to drop the name `Kind` from code and keep it in the comments. Although I can see on second glance that I should have also changed the field name :eyes:

leonschoorl:
Yes, I was going of the `tyConKind` field name.
You intent to remove the "kind" terminology entirely with this PR?

alex-mckenna:
I don't like this PR anymore, closing in favour of a different (to be created) one

PR #1922, opened 2021-09-13
Change Data.Text.Prettyprint.Doc to Prettyprinter

alex-mckenna:
The new version of `prettyprinter` (1.7.1) adds a `DEPRECATED` pragma to the old-style modules for the library. To prevent CI failing, now is the time to switch over to the new-style module names.

## Still TODO:


alex-mckenna:
Closed in favour of #1923

PR #1925, opened 2021-09-13
Don't loop on hidden recursive data type

### LINKED ISSUE #1921: Problem recursing on singleton lists
Below, I have two different version of a linear feedback shift register (LFSR). The goal is to synthesize
```haskell
{-# LANGUAGE DerivingVia
           , DerivingStrategies
           , LambdaCase
           , AllowAmbiguousTypes
           , ApplicativeDo
           , StandaloneDeriving
           , StandaloneKindSignatures #-}

import Clash.Prelude
import Control.Lens
import Data.Default
import Data.Singletons.Prelude
import Data.Singletons.TypeLits as TL

topEntity :: Clock System -> Reset System -> Enable System -> Signal System (Unsigned 8)
topEntity = exposeClockResetEnable fibonacciLFSR8

-- Straightforward newtype
type LFSRState :: Nat -> Type
newtype LFSRState n = LFSRState { runLFSRState :: BitVector n }
  deriving (NFDataX, AutoReg) via (BitVector n)
instance KnownNat n => Default (LFSRState n) where
  def = LFSRState (fromIntegral 1)
```
This first version passes the taps of the LFSR via a term-level list, and synthesizes just fine.

```haskell
fibonacciLFSR8 :: HiddenClockResetEnable dom => Signal dom (Unsigned 8)
fibonacciLFSR8 = fibonacciLFSRTerm @8 [3,4,5,7]

fibonacciLFSRTerm
  :: forall (n :: Nat) dom
   . KnownNat n
  => HiddenClockResetEnable dom
  => [Int] -> Signal dom (Unsigned n)
fibonacciLFSRTerm taps =
  let lfsr = autoReg def (LFSRState <$> lfsr')
      lfsr' = do lfsrState <- lfsr
                 -- shift the bit register by one, and then replace
                 -- the bit on the end by xor of the taps (via go)
                 return $
                   shiftL (runLFSRState lfsrState) 1
                   & ix 0
                   .~ go lfsrState taps
  in unpack <$> lfsr'
                    
  where
    go :: forall (n :: Nat)
        . KnownNat n
       => LFSRState n -> [Int] -> Bit
    go b@(LFSRState bs) = \case
      -- If there is only one tap left, return that bit
      (a:[]) -> bs ^?! ix a
      -- XOR a tapped bit with the remaining taps
      (a:as) -> xor (bs ^?! ix a) (go b as)
      -- This should never happen (we can't have no taps)
      [] -> error "A no-tap LFSR is ill-defined"
```
This second version (which was my original attempt) passes the taps through a type-level list. It runs just fine in clashi, but fails to synthesize:

```haskell
fibonacciLFSR8 :: HiddenClockResetEnable dom => Signal dom (Unsigned 8)
fibonacciLFSR8 = fibonacciLFSRType @('[3,4,5,7]) @8

fibonacciLFSRType
  :: forall (taps :: [Nat]) (n :: Nat) dom
   . SingI taps
  => KnownNat n
  => HiddenClockResetEnable dom
  => Signal dom (Unsigned n)
fibonacciLFSRType =
  let lfsr = autoReg def (LFSRState <$> lfsr')
      lfsr' = do lfsrState <- lfsr
                 -- shift the bit register by one, and then replace
                 -- the bit on the end by xor of the taps (via go)
                 return $
                   shiftL (runLFSRState lfsrState) 1
                   & ix 0
                   .~ go lfsrState (sing :: Sing taps)
  in unpack <$> lfsr'
                    
  where
    go :: forall (n :: Nat) (indices :: [Nat])
        . SingI indices
       => KnownNat n
       => LFSRState n -> SList indices -> Bit
    go b@(LFSRState bs) = \case
      -- If there is only one tap left, return that bit
      (SCons a SNil) -> withKnownNat a
                        $ bs ^?! ix (fromIntegral $ TL.natVal a)
      -- XOR a tapped bit with the remaining taps
      (SCons a as) -> withSingI as $ withKnownNat a
                        $ xor (bs ^?! ix (fromIntegral $ TL.natVal a)) (go b as)
      -- This should never happen (we can't have no taps)
      SNil -> error "A no-tap LFSR is ill-defined"
```

Operationally I can obviously just use the first version. But it bothers me that this second version fails to synthesize (the memory usage blew up beyond 32 GB, to the point where I had to power cycle my machine). I observed this on both clash-1.4.2 and clash-1.4.3. Is this expected behavior?

christiaanb:
The recursive occurrence was hiding behind a type family. So we should always expand type families when checking whether a
data type is recursive.

Fixes #1921

### DIFF
--- a/tests/shouldwork/Issues/T1921.hs
+++ b/tests/shouldwork/Issues/T1921.hs
@@ -1,0 +1,64 @@
+{-# LANGUAGE CPP
+           , DerivingStrategies
+           , GeneralizedNewtypeDeriving
+           , LambdaCase
+           , AllowAmbiguousTypes
+           , ApplicativeDo
+           , StandaloneDeriving #-}
+
+module T1921 where
+
+import Clash.Prelude
+import Control.Lens
+import Data.Default
+#if MIN_VERSION_singletons(3,0,0)
+import Prelude.Singletons
+import GHC.TypeLits.Singletons as TL
+#else
+import Data.Singletons.Prelude
+import Data.Singletons.TypeLits as TL
+#endif
+
+topEntity :: Clock System -> Reset System -> Enable System -> Signal System (Unsigned 8)
+topEntity = exposeClockResetEnable fibonacciLFSR8
+
+-- Straightforward newtype
+newtype LFSRState n = LFSRState { runLFSRState :: BitVector n }
+  deriving newtype (NFDataX, AutoReg)
+instance KnownNat n => Default (LFSRState n) where
+  def = LFSRState (fromIntegral 1)
+
+fibonacciLFSR8 :: HiddenClockResetEnable dom => Signal dom (Unsigned 8)
+fibonacciLFSR8 = fibonacciLFSRType @('[3,4,5,7]) @8
+
+fibonacciLFSRType
+  :: forall (taps :: [Nat]) (n :: Nat) dom
+   . SingI taps
+  => KnownNat n
+  => HiddenClockResetEnable dom
+  => Signal dom (Unsigned n)
+fibonacciLFSRType =
+  let lfsr = autoReg def (LFSRState <$> lfsr')
+      lfsr' = do lfsrState <- lfsr
+                 -- shift the bit register by one, and then replace
+                 -- the bit on the end by xor of the taps (via go)
+                 return $
+                   shiftL (runLFSRState lfsrState) 1
+                   & ix 0
+                   .~ go lfsrState (sing :: Sing taps)
+  in unpack <$> lfsr'
+
+  where
+    go :: forall (n :: Nat) (indices :: [Nat])
+        . SingI indices
+       => KnownNat n
+       => LFSRState n -> SList indices -> Bit
+    go b@(LFSRState bs) = \case
+      -- If there is only one tap left, return that bit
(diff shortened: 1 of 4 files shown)


tests/shouldwork/Issues/T1921.hs
```diff
@@ -0,0 +1,60 @@
...
+import Data.Singletons.Prelude
+import Data.Singletons.TypeLits as TL
+
+topEntity :: Clock System -> Reset System -> Enable System -> Signal System (Unsigned 8)
+topEntity = exposeClockResetEnable fibonacciLFSR8
+
+-- Straightforward newtype
+type LFSRState :: Nat -> Type
+newtype LFSRState n = LFSRState { runLFSRState :: BitVector n }
+  deriving (NFDataX, AutoReg) via (BitVector n)
```
alex-mckenna, line 23: I think we should remove `DerivingVia` and backport this to 1.4 in case there's another 1.4.x release

alex-mckenna (approved):

tests/shouldwork/Issues/T1921.hs
christiaanb: Yeah, I removed it as part of getting it through CI

PR #1928, opened 2021-09-14
Keep casts of the form `x ~ Integer`

### LINKED ISSUE #1927: Use of satSucc triggers "Not in normal form: application of a Lambda-expression"
Consider the following simple example (sorry for not being able to get it more minimal, I tried but failed):

```haskell
topEntity
  :: Signal System (Vec 1 (Signed 1))
  -> Signal System (Maybe (Vec 1 (Signed 1, Signed 1)))
  -> Clock System -> Reset System -> Enable System
  -> Signal System (Maybe (Vec 1 (Signed 1, Signed 1)))
topEntity angles xs c r e = exposeClockResetEnable (machine @8 angles xs) c r e 

machine
  :: forall numStages pairs coord angle dom counter
   . HiddenClockResetEnable dom
  => (1 <= numStages)
  => (KnownNat numStages, KnownNat pairs)
  => (Default coord, NFDataX coord)
  => (Default angle, NFDataX angle)
  => counter ~ (Index numStages)
  => Signal dom (Vec pairs angle)
  -> Signal dom (Maybe (Vec pairs (coord, coord)))
  -> Signal dom (Maybe (Vec pairs (coord, coord)))
machine thetas coords = mealy transition def $ liftA2 (liftM2 zip) coords (return <$> thetas)
  where
    transition :: (counter, Vec pairs ((coord, coord), angle))
               -> Maybe (Vec pairs ((coord, coord), angle))
               -> ( (counter, Vec pairs ((coord, coord), angle))
                  , Maybe (Vec pairs (coord, coord))
                  )
    transition (k, xyzs) maybeInputs = case (k == 0, maybeInputs)
                                         of (True, Nothing) -> ((0, xyzs), Just $ map fst xyzs)
                                            (True, Just xyzs0) -> ((0, xyzs0), Just $ map fst xyzs)
                                            (False, _) -> ((satSucc SatWrap k, xyzs), Nothing)
```

This triggers the following error:
```
    Clash.Netlist(914): Not in normal form: application of a Lambda-expression

    k[6989586621679046945][LocalId] ▷
      GHC.Integer.Type.Integer[3674937295934324784]
      ~
      Clash.Sized.Internal.Index.Index[8214565720323789603] 8
```

Now, remarkably, the following modifications *do* allow this design to compile:

- Fixing `numStages` as a static value in the type signature of `machine`, rather than passing it in via TypeApplication.
- Replacing `satSucc SatWrap k` with `k`, `k+1`, or even -- to my great surprise -- `satAdd SatWrap k 1` or `satAdd SatWrap 1 k`.

The fact that I can use `satAdd SatWrap 1 k` but not `satAdd SatSucc k` feels like a bug, so I'm reporting it here. Apologies again for not being able to reduce it further... simpler examples didn't seem to give the same error, which makes me worry that there's something more complicated going on with the `Vec`s or `Maybe`s that's helping induce Clash's confusion.

christiaanb:
Really a work-around until we properly handle all casts.

Fixes #1927

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
@@ -659,6 +659,8 @@ isSizedCast (TyConApp tc1 _) (TyConApp tc2 _) = do
         ,tc2 `hasKey` naturalTyConKey &&
           tc1Nm == "Clash.Sized.Internal.Unsigned.Unsigned"
         ])
+-- XXX: a work-around for issue #1927, the real fix is to handle all casts
+isSizedCast (TyVarTy {}) (TyConApp tc2 _) = return (tc2 `hasKey` integerTyConKey)
 isSizedCast _ _ = return False
 
 hasPrimCo :: Coercion -> C2C (Maybe Type)
--- a/changelog/2021-09-14T20_09_01+02_00_fix1927
+++ b/changelog/2021-09-14T20_09_01+02_00_fix1927
@@ -1,0 +1,1 @@
+FIXED: Don't error out on `counter ~ (Index numStages)` like equalities [#1927](https://github.com/clash-lang/clash-compiler/issues/1927)
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -533,6 +533,7 @@ runClashTest = defaultMain $ clashTestRoot
         , clashLibTest "T779" def{hdlTargets=[Verilog]}
         , outputTest "T1881" def{hdlSim=False}
         , runTest "T1921" def{hdlTargets=[Verilog], hdlSim=False}
+        , runTest "T1927" def{hdlTargets=[Verilog], hdlSim=False}
         ] <>
         if compiledWith == Cabal then
           -- This tests fails without environment files present, which are only
--- a/tests/shouldwork/Issues/T1927.hs
+++ b/tests/shouldwork/Issues/T1927.hs
@@ -1,0 +1,35 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+module T1927 where
+
+import Clash.Prelude
+import Control.Monad
+
+topEntity
+  :: Signal System (Vec 1 (Signed 1))
+  -> Signal System (Maybe (Vec 1 (Signed 1, Signed 1)))
+  -> Clock System -> Reset System -> Enable System
+  -> Signal System (Maybe (Vec 1 (Signed 1, Signed 1)))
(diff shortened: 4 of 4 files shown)


clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -659,6 +659,8 @@ isSizedCast (TyConApp tc1 _) (TyConApp tc2 _) = do
         ,tc2 `hasKey` naturalTyConKey &&
           tc1Nm == "Clash.Sized.Internal.Unsigned.Unsigned"
         ])
+-- XXX: a work-around for issue #1927, the real fix is to handle all casts
+isSizedCast (TyVarTy {}) (TyConApp tc2 _) = return (tc2 `hasKey` integerTyConKey)
```
alex-mckenna, line 663: Does this also need a case for `Integer ~ x`?

christiaanb: Not for issue #1927 specifically.

alex-mckenna: Right, but would you trigger the same issue by changing `counter ~ (Index numStages)` to `(Index numStages) ~ counter` in the test? If so I think we should add the extra case to avoid weird (albeit unlikely) situations where someone's design fails unexpectedly unless they change the argument order to `(~)`

christiaanb: I haven't tried. I'm somewhat uncomfortable with this PR to begin with. I'd rather properly support all casts.

alex-mckenna: True, supporting all casts would be the better option. How far was #1064 from being finished?

christiaanb: It went in the right direction. But I would want to redo some aspects, especially with regards to `Tick`. So [`squashCollectApp`](https://github.com/clash-lang/clash-compiler/compare/castSignal-rebased#diff-1051fc7ced5b1944b216dd30a323e3c1fd9410a615136ff4c717089c899888c8R608-R716) was there in terms of handling casts I think, but `Ticks` still need more care: depending on the type of info (source location, HDL naming, etc.) the ticks either should or shouldn't be distributed towards the arguments as we "squash" the entire application.

martijnbastiaan:
Superseded by https://github.com/clash-lang/clash-compiler/pull/1064

PR #1931, opened 2021-09-16
New unsafeSimSynchronizer, use in outputVerifier

DigitalBrains1:
Warn when `Clash.Explicit.Testbench.outputVerifier` is used in HDL which
is not marked as being a test bench and is given different clock domains
as `testDom` and `circuitDom`. If `testDom` and `circuitDom` refer to
the same domain, the warning is not emitted.

To this purpose, a new primitive `unsafeSimSynchronizer` is added to
`Clash.Explicit.Testbench`, and it is used if a clock domain crossing is
needed in `outputVerifier`.

The code for `outputVerifier` is also cleaned up. It now declares it
needs a non-empty list of samples to verify, since it never worked with
an empty list. Common code is split off and shared rather than
duplicated, and it now uses `Bounded` and `SaturatingNum` to work with
the `Index` iterating the samples instead of writing out the same
functionality.

Rationale: `outputVerifier` does clock domain crossing in a manner that
is not suitable for actual hardware but is efficient for HDL simulation.
The code was not intended to be used outside of a test bench context,
but nonetheless it could be synthesized as part of hardware. If a
user does this, we should warn them that the resulting hardware is
improper.

The user can then add their own clock domain crossing in the circuit if
they would need this functionality. An imaginable use of
`outputVerifier` in a circuit is in combination with an ILA standing in
for the `assert` behavior.

## Still TODO:


DigitalBrains1:
I purposely did not write a Changelog entry for `unsafeSimSynchronizer`.

It is purely something I needed for implementing the desired behaviour in `outputVerifier`. But I didn't see a reason to keep it hidden, so I exposed it in the module. On the other hand, I found it so unsignificant I didn't want to enshrine it in the Changelog.

christiaanb (approved):

PR #1368, opened 2020-06-11
Treat Enable Specially in Generated HDL

alex-mckenna:
Closes #1171 

Enable lines are now treated similarly to clock and reset lines
in generated HDL, being given their own special type alias in
VHDL, and being annotated with a comment that it is an enable
line in all HDLs.

PR #1936, opened 2021-09-28
Properly recognize workfree enable signals

### LINKED ISSUE #1935: Register duplication
```haskell
-- one register
f
    :: Clock System
    -> Reset System
    -> Signal System (Unsigned 8)
f clk rst = x
  where
    x :: Signal System (Unsigned 8)
    x = withClockResetEnable clk rst enableGen register 4 (x+1)


-- two registers
g
    :: Clock System
    -> Reset System
    -> Signal System (Unsigned 8)
g clk rst = withClockResetEnable clk rst enableGen x
  where
    x :: SystemClockResetEnable => Signal System (Unsigned 8)
    x = register 4 (x+1)
```
`f` work normally.
But the `register` in `g` duplicated in the HDL.

leonschoorl:
`isWorkFreeClockOrResetOrEnable` tried to detect constant enable signals,
but we forgot to update this when we changed the representation of Enable in #1368.

This fixes the register duplication issue seen in #1935

Fixes #1935

### DIFF
--- a/clash-lib/src/Clash/Rewrite/WorkFree.hs
+++ b/clash-lib/src/Clash/Rewrite/WorkFree.hs
@@ -155,7 +155,7 @@ isWorkFreeClockOrResetOrEnable tcm e =
     case collectArgs e of
       (Prim p,_) -> Just (primName p == Text.showt 'removedArg)
       (Var _, []) -> Just True
-      (Data _, []) -> Just True -- For Enable True/False
+      (Data _, [_dom, Left (stripTicks -> Data _)]) -> Just True -- For Enable True/False
       (Literal _,_) -> Just True
       _ -> Just False
   else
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Rewrite/WorkFree.hs
```diff
@@ -155,7 +155,7 @@ isWorkFreeClockOrResetOrEnable tcm e =
     case collectArgs e of
```
alex-mckenna, line 155: I wonder if this `collectArgs` should also be `collectArgsTicks` where the ticks are ignored...
It's not causing any issues that we know of though, so it's more _thinking out loud_ than requesting changes

alex-mckenna (approved):

clash-lib/src/Clash/Rewrite/WorkFree.hs
christiaanb: The API in `castSignal` branch will no longer allows you to drop ticks silently.

alex-mckenna: Good point, I forgot about that. In that case I don't think we need to worry about this (for anyone else reading: `collectArgs` is gone in the cast branch anyway so this call will be changed later)

PR #1940, opened 2021-09-29
New function: runUntil

DigitalBrains1:
`runUntil`, a function to sample a signal until it returns a value that
satisfies the user-given test. It is a convenience function that, among
others, allow easy running of a `testBench` style function in Haskell
simulation, logging assertion failures to stderr

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Signal.hs
+++ b/clash-prelude/src/Clash/Explicit/Signal.hs
@@ -236,6 +236,7 @@ module Clash.Explicit.Signal
   , simulateB
   , simulateWithReset
   , simulateWithResetN
+  , runUntil
     -- ** lazy versions
   , simulate_lazy
   , simulateB_lazy
@@ -279,7 +280,8 @@ import           Clash.Signal.Internal
 import           Clash.Signal.Internal.Ambiguous
   (knownVDomain, clockPeriod, activeEdge, resetKind, initBehavior, resetPolarity)
 import qualified Clash.Sized.Vector
-import           Clash.XException               (NFDataX, deepErrorX, fromJustX)
+import           Clash.XException
+  (NFDataX, deepErrorX, fromJustX, seqX, ShowX(..))
 
 {- $setup
 >>> :set -XDataKinds -XTypeApplications -XFlexibleInstances -XMultiParamTypeClasses -XTypeFamilies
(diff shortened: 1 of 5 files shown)


clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -859,4 +861,70 @@ sampleWithResetN
...
+-- result message.
+runUntil
+  :: forall dom a
+   . (KnownDomain dom, NFDataX a, ShowX a)
+  => (a -> Bool)
+  -- ^ Condition checking function, should return @True@ to finish run
+  -> Signal dom a
+  -- ^ 'Signal' we want to sample for the condition
+  -> IO ()
+runUntil check s = value `seqX` putStrLn msg
```
martijnbastiaan, line 922: Perhaps `sampleUntil`? Given that we already have `sample` and `sampleN`.

alex-mckenna: Oh that's nice, +1

DigitalBrains1: But `sample` and `sampleN` return samples, `runUntl` doesn't return them. I agree that it does sample the signal, to be able to do the testing, but if it were called `sampleUntil` I would think it would return the list of samples until the test was true.

martijnbastiaan: Although I guess that'd mean it'd have to return `[a]`, so perhaps this is better after all.

alex-mckenna: It could be worth having both? I can see a `sampleUntil` that returns the values also being useful

DigitalBrains1: I always have Haskell's `Prelude` imported qualified under `P` in my `clashi` sessions, and then it's as simple as
```
>>> P.takeWhile not $ sample testBench
```
to get what I imagine would look like
```
>>> sampleUntil id testBench
```
with that new `sampleUntil`. To me, the former is just as good and I wouldn't need the latter.

DigitalBrains1: (although I suppose if I were to design `sampleUntil`, I would include the first sample passing the check as well, so it would not be exactly equivalent)

DigitalBrains1: I suddenly realise I've completely forgotten to add some nice invocations to the `runUntil` example. To wit,
```
>>> mapM_ runUntil [ testBenchA, testBenchB ]
>>> sequence_ [ runUntil testBenchA, runUntil testBenchB ]
```
The latter for when the clock domain of the two testbenches is not the same.

DigitalBrains1: For a single test bench, `P.takeWhile not $ sample testBench` is pretty okay already, it's when you have multiple test benches that `runUntil` becomes a prety nice shorthand. So I had always intended to steer the reader towards that kind of use, and then totally forgot to put it in once I was actually writing the doc...

alex-mckenna: > I always have Haskell's `Prelude` imported qualified under `P` in my `clashi` sessions

To take a harsh stance: if you need to have `Prelude` from `base` imported in `clashi`, `Clash.Prelude` isn't doing it's job well enough. For me the pattern of `P.takeWhile not $ sample testBench` motivates a `sampleUntil` function being included because that's friendlier to beginners who may not be aware there's a second prelude they can import

DigitalBrains1: We are doing enough already... I was not aware, but you can just do
```
> takeWhile not $ sample testBench
```
:-)

alex-mckenna: Hmm okay, so the situation isn't as bad as I thought. In that case I guess it's not so bad to not have `sampleUntil`, although maybe it would still be a nice touch given `sampleN` is basically just a synonym for `take` after `sample` and `sampleUntil` is a synonym for `takeWhile (not . p)` for some `p` after `sample`

I leave it to your own preference :slightly_smiling_face:

DigitalBrains1: I think a `sampleUntil` with the documentation pointers you suggested on Slack are a good addition.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -859,4 +861,88 @@ sampleWithResetN
...
+--   -> 'Signal' 'System' Int
+-- topEntity = id
+--
+-- testBench
+--   :: 'Signal' 'System' Bool
+-- testBench = done
+--  where
+--   testInput = 'Clash.Explicit.Testbench.stimuliGenerator' clk rst $('Clash.Sized.Vector.listToVecTH' [1 :: Int .. 10])
+--   expectedOutput =
+--     'Clash.Explicit.Testbench.outputVerifier'' clk rst $('Clash.Sized.Vector.listToVecTH' $ [1 :: Int .. 9] P.++ [42])
```
alex-mckenna, line 896: If the qualified import of `Prelude` is just for `P.++` I'd be inclined to replace it with `<>` which is already exported from `Clash.Prelude`

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -859,4 +861,88 @@ sampleWithResetN
 sampleWithResetN nReset nSamples f =
   take nSamples (sampleWithReset nReset f)
 
+-- | Simulate a component until it matches a condition
+--
+-- It prints a message of the form
+--
+-- "Signal sampled for N cycles until value X"
```
alex-mckenna, line 868: I wonder if this should be in a `@` block and without quotation marks, so it's clearer at a glance that this is console output

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -859,4 +861,88 @@ sampleWithResetN
...
+--
+-- A common usage is with a test bench using
+-- 'Clash.Explicit.Testbench.outputVerifier'. __NB__: Since this uses
+-- 'Clash.Explicit.Testbench.assert', when using @clashi@, read the
+-- note at "Clash.Explicit.Testbench#assert-clashi".
+--
+-- __NB__: This function is not synthesizable
```
alex-mckenna, line 875: I think the "A common usage ..." sentence should be moved under the "Example with test bench heading" below and the notes as standalone paragraphs here. In other places where we have notes they each sit by themselves instead of inlined in a paragraph, i.e.

> Some paragraph of text documenting something.
>
> __NB__: This is a note, take note.
> __NB__: This is another note, still important.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -859,4 +861,88 @@ sampleWithResetN
...
+-- testBenchB :: Signal DomB Bool
+-- @
+--
+-- @
+-- > 'sequence_' [ runUntil id testBenchA, runUntil id testBenchB ]
+-- @
+
+-- The "value `seqX`" bit is to have any invocations of 'trace', such as in
+-- our 'assert' (, 'outputVerifier', ...) functions, print before printing the
+-- result message.
```
alex-mckenna, line 931: I would _probably_ move this closer, i.e.

```haskell
runUntil check s =
  -- Ensure invocations of 'trace' are printed before the result message
  value `seqX` 
```

clash-prelude/src/Clash/Explicit/Signal.hs
DigitalBrains1: The code is intended for human consumption, it's to show people what the output of `runUntil` looks like given a test bench that they should be able to quickly fathom fully. On the one hand it might be something people need to look up before they understand, on the other hand it's a neat "trick" to teach people in the process, to use the `Semigroup` instance of `List` to avoid needing imports. So I agree it's a good change.

clash-prelude/src/Clash/Explicit/Signal.hs
DigitalBrains1: Ooh, by moving the "A common usage ..." inside that paragraph, that indeed allows me to move the __NB__ to its own paragraph because it is neatly part of the example then. I wasn't happy with it being part of the paragraph, but it couldn't stand on its own either as part of the main text.

alex-mckenna (approved):

PR #1941, opened 2021-09-29
Assorted doc fixes

DigitalBrains1:
~~The documentation on the HaskellPrelude imports was missing in 2 of 4
places it was supposed to be. Also updated `Clash.Prelude.Safe` to
mention `even` and `odd` just like `Clash.Prelude` does, and added some
markup.~~
The documentation on the HaskellPrelude imports was missing in 2 of 4
places it was supposed to be. Rewritten to explain rather than
exhaustively list, also avoiding the need to update across all files
(Clash.Prelude.Safe was out-of-date already).

Added the `hasBlackBox` annotation to the example in the tutorial
because we should encourage our users to add that annotation.
Reformatted the import list in the example and added an import for
`hasBlackBox`.

In the example for `Clash.Explicit.Testbench.tbClockGen`, a signal
called `done'` (note prime) was defined as the logical inverse of the
`done` signal. Renamed it to `notDone` to avoid confusion.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Prelude.hs
+++ b/clash-prelude/src/Clash/Explicit/Prelude.hs
@@ -178,6 +178,15 @@ import Clash.XException
 >>> let windowD3 = windowD @2
 -}
 
+{- $hiding
+"Clash.Explicit.Prelude" re-exports most of the Haskell "Prelude" with the
+exception of those functions that the Clash API defines to work on 'Vec' from
+"Clash.Sized.Vector" instead of on lists as the Haskell Prelude does. In
+addition, for the 'Clash.Class.Parity.odd' and 'Clash.Class.Parity.even'
+functions a type class called 'Clash.Class.Parity.Parity' is available at
+"Clash.Class.Parity".
+-}
+
 -- | Give a window over a 'Signal'
 --
 -- @
--- a/clash-prelude/src/Clash/Explicit/Prelude/Safe.hs
+++ b/clash-prelude/src/Clash/Explicit/Prelude/Safe.hs
@@ -140,6 +140,14 @@ import Clash.XException
 >>> let rP clk rst en = registerB clk rst en (8::Int,8::Int)
 -}
 
+{- $hiding
+"Clash.Explicit.Prelude.Safe" re-exports most of the Haskell "Prelude" with the
+exception of those functions that the Clash API defines to work on 'Vec' from
+"Clash.Sized.Vector" instead of on lists as the Haskell Prelude does. In
+addition, for the 'Clash.Class.Parity.odd' and 'Clash.Class.Parity.even'
+functions a type class called 'Clash.Class.Parity.Parity' is available at
+"Clash.Class.Parity".
+-}
 
 -- | Create a 'register' function for product-type like signals (e.g.
 -- @('Signal' a, 'Signal' b)@)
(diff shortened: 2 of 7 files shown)


clash-prelude/src/Clash/Explicit/Prelude.hs
```diff
@@ -178,6 +178,19 @@ import Clash.XException
...
+exception of the following: @(++)@, @(!!)@, @concat@, @drop@, @even@, @foldl@,
+@foldl1@, @foldr@, @foldr1@, @head@, @init@, @iterate@, @last@, @length@,
+@map@, @odd@, @repeat@, @replicate@, @reverse@, @scanl@, @scanr@, @splitAt@,
+@tail@, @take@, @unzip@, @unzip3@, @zip@, @zip3@, @zipWith@, @zipWith3@.
+
+It instead exports the identically named functions defined in terms of 'Vec' at
+"Clash.Sized.Vector". For the 'Clash.Class.Parity.odd' and
+'Clash.Class.Parity.even' function a type class called
+'Clash.Class.Parity.Parity' is available at "Clash.Class.Parity".
+-}
```
leonschoorl, line 192: Should this (and the others) also document the changed `&&` and `||` from https://github.com/clash-lang/clash-compiler/pull/1900 ?

It's the same function, so maybe not?

DigitalBrains1: It's the same function, so in that sense I don't think it needs doc. However, if you look at [HaskellPrelude](https://hackage.haskell.org/package/clash-prelude-1.5.0/candidate/docs/Clash-HaskellPrelude.html#v:-38--38-), you'll see that it documents everything it re-exports, but omits the docs for the redefinitions `&&`, `||` and `not`. That's not really very nice, but seems pretty low priority.

clash-prelude/src/Clash/Explicit/Prelude/Safe.hs
```diff
@@ -140,6 +140,19 @@ import Clash.XException
...
+"Clash.Explicit.Prelude.Safe" re-exports most of the Haskell "Prelude" with the
+exception of the following: @(++)@, @(!!)@, @concat@, @drop@, @even@, @foldl@,
+@foldl1@, @foldr@, @foldr1@, @head@, @init@, @iterate@, @last@, @length@,
+@map@, @odd@, @repeat@, @replicate@, @reverse@, @scanl@, @scanr@, @splitAt@,
+@tail@, @take@, @unzip@, @unzip3@, @zip@, @zip3@, @zipWith@, @zipWith3@.
+
+It instead exports the identically named functions defined in terms of
+'Vec' at "Clash.Sized.Vector". For the 'Clash.Class.Parity.odd' and
+'Clash.Class.Parity.even' function a type class called
+'Clash.Class.Parity.Parity' is available at "Clash.Class.Parity".
```
alex-mckenna, line 153: This paragraph feels like it's missing the first sentence, I think it should be more explicit that common list functions are replaced by common vector functions.

alex-mckenna (changes requested):

clash-prelude/src/Clash/Explicit/Prelude/Safe.hs
DigitalBrains1: Changed based on discussion.

PR #1947, opened 2021-10-07
Suggestion: store memory as literal string

DigitalBrains1:
When you want to generate the contents of a memory, rather than read it
in from a pre-generated file, one could use Template Haskell and
`qRunIO` to create it on the fly while compiling or loading into the
interpreter. Our documentation refers to this as something
["for those of us who like to live on the edge"](https://hackage.haskell.org/package/clash-prelude-1.5.0/candidate/docs/Clash-Sized-Fixed.html#creatingdatafiles).
It is brittle for many reasons.

Additionally, our `Vec` has performance issues when they get long, for a
pretty small value of "long". And lists don't work either.

So here is another idea. Using Template Haskell, we convert a list of
`BitPack`able values to the string contents of a file such as `romFile`
consumes. We don't write it to a file, though, but store it as a literal
string in our Haskell module. `romString` then is the variant of
`romFile` that uses this string while simulating in Haskell, and the
black box definition for `romString` uses the include mechanism to
write a file called `memXYZ.bin` next to the generated HDL, whereupon
the mechanism in HDL is identical to `romFile`'s mechanism.

When I create a `MemString` that holds 4 MiB of 32-bit data, the
resulting data file is 33 MiB (ASCII binary digits and newline), and so
is the object file for the Haskell module containing that `MemString`.
As can be seen from the `BlackBoxContext`, the `memStringContents` is
stored as `unpackCString# "literal"#`, and it is an efficient storage
method (well, apart from the binary digits bit of it, which we might
improve upon later by implementing hexadecimal digits for memory files).

If people like this idea, I can work it out further, this is a first
draft to gauge enthousiasm.

## Still TODO:

### DIFF
--- a/clash-lib/clash-lib.cabal
+++ b/clash-lib/clash-lib.cabal
@@ -256,6 +256,7 @@ Library
                       Clash.Normalize.Util
 
                       Clash.Primitives.DSL
+                      Clash.Primitives.ROM.File
                       Clash.Primitives.Types
                       Clash.Primitives.Util
 
--- a/clash-lib/prims/vhdl/Clash_Explicit_ROM_File.primitives
+++ b/clash-lib/prims/vhdl/Clash_Explicit_ROM_File.primitives
@@ -54,4 +54,69 @@ end block;
 -- romFile end"
     }
   }
+, { "BlackBox" :
+    { "name" : "Clash.Explicit.ROM.File.romString#"
+    , "kind" : "Declaration"
+    , "type" :
+
+"romString#
+  :: ( KnownNat m           --            ARG[0]
+     , KnownDomain dom      --            ARG[1]
+     )
+  => Clock dom              -- clk,       ARG[2]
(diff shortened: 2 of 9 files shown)


DigitalBrains1:
> Like a `ByteString`

Well, I'd first have to see how Clash reacts to a `ByteString`. I picked `String` because I knew that Clash would not stumble over such an argument to a primitive.

> I don't think other synthesis/simulation tools mind the large constants

In that case I could rewrite it to include it in the generated HDL file, yes.

DigitalBrains1:
> I don't think HDL synthesis tools will do that for their constants.

Hehe, no, but they might try to load the .vhdl file as UTF-32 into a buffer that for some arcane computer science reason can't grow large, never hitting any problems because people never thought to put the full contents of all the BlockRAMs in their FPGA into their source files ;-)

DigitalBrains1:
This sprouted into the accepted PR #2041

PR #1949, opened 2021-10-10
Bump dependency versions

martijnbastiaan:
Related: https://github.com/commercialhaskell/stackage/issues/6217


phadej:
I see the 

```
    • Couldn't match expected type ‘GHC.Exts.Item
                                      aeson-2.0.1.0:Data.Aeson.Types.Internal.Object’
                  with actual type ‘(a0, Value)’
```

failure.

I will add the `IsList (KeyMap v)` instance in the next minor aeson release. It would be nice if you can report it though, so I don't forget.

PR #1953, opened 2021-10-12
Relax upper bound of prettyprinter

martijnbastiaan:
Related: https://github.com/commercialhaskell/stackage/issues/6228#issuecomment-940579749


martijnbastiaan:
@alex-mckenna Given you wrote the comment: couldn't we do CPP to work around this? Debian / nix can still use older packages if they like. Or am I missing something?

PR #1972, opened 2021-10-25
Remove blackboxes for GHC.Types.I# and GHC.Types.W#

### LINKED ISSUE #1966: Remove blackboxes for `GHC.Types.I#` and `GHC.Types.W#`
As noted in https://github.com/clash-lang/clash-compiler/pull/1948#discussion_r729205266 we should remove the blackboxes for `GHC.Types.I#` and `GHC.Types.W#`.
Clash should be able to handle them as regular data constructors, without having to special-case them.

alex-mckenna:
As mentioned in PR #1948, it is not necessary to have boxed literal types like Int and Word as blackboxes. We can instead render their data constructors directly.

Closes #1966

I have a feeling this PR needs to be benchmarked, locally I think I see some slowdown in the testsuite.

## Still TODO:


leonschoorl (approved):
LGTM

PR #1976, opened 2021-10-27
Clash Hedgehog Generators

### LINKED ISSUE #1962: Better pretty printing for `TyCon`
Reading traces of `TyCon` / `TyConMap` is a bit less pleasant than I'd like. I think it would be better if instead of just printing out the name, we instead printed out something that looked more like the output of `:i` in GHCi, or the signatures you would write in something like twelf. Some examples:

## An `AlgTyCon`

```haskell
data Foo a = A Int Char | B a a

newtype Quux = Quux (Foo Double)

data Bar = Bar { barA :: Int, barB :: Word }
```

would render as

```haskell
data Foo :: Type -> Type
Bar :: forall a. Int -> Char -> Foo a
Baz :: forall a. a -> a -> Foo a

newtype Quux :: Type
Quux :: Foo Double -> Quux

data Bar :: Type
Bar :: Int -> Word -> Bar
barA :: Bar -> Int
barB :: Word -> Int
```

## A `FunTyCon`

Show a line for the name and kind, and one line for every equation, e.g.

```haskell
type family Foo (a :: Nat) :: Type where
  Foo 0 = Int
  Foo 1 = Maybe (Foo 0)
```

would render as

```haskell
type family Foo :: Nat -> Type
type instance Foo 0 = Int
type instance Foo 1 = Maybe (Foo 0)
```

## A `PrimTyCon`

Show the name and the kind, e.g.

```haskell
Type
Nat
Int
~#
```

would render as

```haskell
Type :: Type
Nat :: Type
Int :: Type
~# :: forall a. forall b. a -> b -> Type
```

~The prefix "prim" is used to differentiate between a `PrimTyCon` and an `AlgTyCon` with no constructors.~

alex-mckenna:
This PR introduces the `clash-{prelude,lib}-hedgehog` libraries, providing some basic generators for types in `clash-prelude` and `clash-lib`. Currently this is capable of

  * generation of Clash-specific types (`BitVector`, `Signed`, `Vec`, `Signal` etc.)
  * generation of Clash core types (`TyConMap`, `Type`, `DataCon`, `Term` etc.)

Currently there are some shortcomings with the `clash-lib` generators, and they are unable to generate all the programs we would like to see. The main problems are a lack of type unification in `clash-lib`, and the inability to get primitives. In a later PR we should attempt to support loading primitives and `TyConMap` from given modules (so i.e. random generation can occur in a scope that has `Clash.Prelude` imported).

Also in this PR:

  * Implement `HasType` for `TyCon` so `coreKindOf` can be used on type constructors
  * Add the `PromotedDataCon` type constructor so promoted constructors do not appear as `PrimTyCon`

Closes #1962

## Still TODO:

### DIFF
--- a/changelog/2021-10-27T09_49_20+02_00_clash_hedgehog.md
+++ b/changelog/2021-10-27T09_49_20+02_00_clash_hedgehog.md
@@ -1,0 +1,1 @@
+ADDED: New `clash-{prelude,lib}-hedgehog` packages which provide generators for types in `clash-prelude` and `clash-lib`.
--- a/clash-prelude-hedgehog/src/Clash/Hedgehog/Sized/BitVector.hs
+++ b/clash-prelude-hedgehog/src/Clash/Hedgehog/Sized/BitVector.hs
@@ -1,0 +1,82 @@
+{-|
+Copyright   : (C) 2021, QBayLogic B.V.
+License     : BSD2 (see the file LICENSE)
+Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
+
+Random generation of BitVector.
+-}
+
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.KnownNat.Solver #-}
+
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE RankNTypes #-}
+
+module Clash.Hedgehog.Sized.BitVector
+  ( genDefinedBit
+  , genBit
+  , genDefinedBitVector
+  , genBitVector
+  , SomeBitVector(..)
+  , genSomeBitVector
+  ) where
+
+import GHC.Natural (Natural)
+import GHC.TypeNats
+import Hedgehog (MonadGen, Range)
+import qualified Hedgehog.Gen as Gen
+
+import Clash.Promoted.Nat
+import Clash.Sized.Internal.BitVector
+import Clash.Sized.Vector (v2bv)
+import Clash.XException (errorX)
+
+import Clash.Hedgehog.Sized.Vector (genVec)
+
+-- | Generate a bit which is guaranteed to be defined.
+-- This will either have the value 'low' or 'high'.
+--
+genDefinedBit :: (MonadGen m) => m Bit
+genDefinedBit = Gen.element [low, high]
+
+-- | Generate a bit which is not guaranteed to be defined.
+-- This will either have the value 'low' or 'high', or throw an 'XException'.
+--
+genBit :: (MonadGen m) => m Bit
+genBit = Gen.element [low, high, errorX "X"]
+
+-- | Generate a bit vector where all bits are defined.
+--
+genDefinedBitVector :: (MonadGen m, KnownNat n) => m (BitVector n)
+genDefinedBitVector =
+  Gen.frequency
+    [ (60, fmap v2bv (genVec genDefinedBit))
+    , (20, Gen.constant minBound)
+    , (20, Gen.constant maxBound)
+    ]
+
+-- | Generate a bit vector where some bits may be undefined.
+--
+genBitVector :: (MonadGen m, KnownNat n) => m (BitVector n)
+genBitVector =
+  Gen.frequency
+    [ (55, fmap v2bv (genVec genBit))
+    , (15, Gen.constant minBound)
+    , (15, Gen.constant maxBound)
+    , (15, Gen.constant undefined#)
+    ]
+
+data SomeBitVector atLeast where
+  SomeBitVector :: SNat n -> BitVector (atLeast + n) -> SomeBitVector atLeast
+
+genSomeBitVector
+  :: forall m atLeast
+   . (MonadGen m, KnownNat atLeast)
+  => Range Natural
+  -> (forall n. KnownNat n => m (BitVector n))
+  -> m (SomeBitVector atLeast)
+genSomeBitVector rangeBv genBv = do
+  numExtra <- Gen.integral rangeBv
+
+  case someNatVal numExtra of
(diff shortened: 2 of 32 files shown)


clash-hedgehog/src/Clash/Hedgehog/Sized/Index.hs
```diff
@@ -0,0 +1,20 @@
...
+  ) where
+
+import GHC.TypeLits (KnownNat)
+import Hedgehog (MonadGen, Range)
+import qualified Hedgehog.Gen as Gen
+
+import Clash.Sized.Internal.Index
+
+genIndex :: (MonadGen m, KnownNat n) => Range (Index n) -> m (Index n)
+genIndex = Gen.integral
```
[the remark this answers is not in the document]
alex-mckenna: For `Index` / `Signed` / `Unsigned` this isn't true because users control the `Range` used

clash-hedgehog/src/Clash/Hedgehog/Core/Literal.hs
```diff
@@ -0,0 +1,98 @@
...
+genInt64Literal =
+  Int64Literal <$> (toInteger <$> Gen.int64 Range.linearBounded)
+
+genWord64Literal :: forall m. MonadGen m => m Literal
+genWord64Literal =
+  Word64Literal <$> (toInteger <$> Gen.word64 Range.linearBounded)
+
+genStringLiteral :: forall m. MonadGen m => m Literal
+genStringLiteral =
+  StringLiteral <$> Gen.string (Range.linear 10 50) Gen.latin1
```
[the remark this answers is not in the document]
alex-mckenna: Oh, I had this on `latin1` for no real reason. I've changed it to `Gen.unicode` because it should be any valid Haskell string literal. I don't think there's a need to specify the encoding in the name because there's no intent of limiting to certain encodings

alex-mckenna:
Note to self, I want to pull out the alpha-equivalent `Pat` stuff and make it into it's own PR once #1990 is merged

clash-hedgehog/src/Clash/Hedgehog/Sized/Index.hs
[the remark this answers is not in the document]
martijnbastiaan: The Range just changes the extreme values from a grow/shrink perspective. So say you've got a range from 1 to 2 million, it would be _incredibly_ unlikely to actually generate that 2 million. However this is usually what you want while testing systems.

clash-prelude-hedgehog/src/Clash/Hedgehog/Sized/BitVector.hs
```diff
@@ -0,0 +1,82 @@
...
+-- | Generate a bit vector where some bits may be undefined.
+--
+genBitVector :: (MonadGen m, KnownNat n) => m (BitVector n)
+genBitVector =
+  Gen.frequency
+    [ (55, fmap v2bv (genVec genBit))
+    , (15, Gen.constant minBound)
+    , (15, Gen.constant maxBound)
+    , (15, Gen.constant undefined#)
+    ]
```
martijnbastiaan, line 67: This is what I'd like to see for the other generators too. I don't think it's merge blocking, but while testing I've been almost always been interested in testing extreme/interesting values (that's typically where all the bugs hid!). On the flip side, I have never been actually interested in a generator that picks values at random (after shrinking the range with some size multiplier). Still, I have used them out of sheer laziness. To safe lazy bums like me from themselves, I think our simplest generators should therefore be more likely to pick extreme values, in order to encourage usage.

martijnbastiaan: My proposal would therefore be to merge this and add an issue explaining the enhancement.

hie.yaml
```diff
@@ -21,6 +21,8 @@ cradle:
       config: { cradle: {none: }}
     - path: "./clash-ghc/src-bin-881"
       config: { cradle: {none: }}
+    - path: "./clash-hedgehog"
+      config: { cradle: {cabal: [{path: "./clash-hedgehog/src", component: "lib:clash-hedgehog"}]}}
```
martijnbastiaan, line 25: Is this still up-to-date?

martijnbastiaan (approved):
LGTM

Is the plan to keep them private (off Hackage) for now, and move them to a separate repo if they do or..?

alex-mckenna:
> Is the plan to keep them private (off Hackage) for now, and move them to a separate repo if they do or..?

I want to keep them in the same repo as `clash-prelude` and `clash-lib`, since they require API compatibility I've set them up to have the same version numbers anyway. As for hackage, `clash-prelude-hedgehog` should be released, I'm not sure about `clash-lib-hedgehog` though since that's more for clash devs than clash users.

changelog/2021-10-27T09_49_20+02_00_clash_hedgehog.md
```diff
@@ -0,0 +1 @@
+ADDED: A new `clash-hedgehog` package which provides generators for types in `clash-prelude` and `clash-lib` (optionally turned on, intended for clash developers).
```
martijnbastiaan, line 1:
```suggestion
ADDED: A new `clash-prelude-hedgehog` package which provides generators for types in `clash-prelude`.
```

martijnbastiaan (commented):
OK! I'll squat the names on Hackage. Don't forget to update:

https://github.com/clash-lang/clash-compiler/blob/a5174f61cb3d3a5cacce92de13fc4135dde286dd/.ci/gitlab/publish.yml#L6-L8

https://github.com/clash-lang/clash-compiler/blob/a5174f61cb3d3a5cacce92de13fc4135dde286dd/.ci/gitlab/publish.yml#L23-L25

PR #1978, opened 2021-10-29
Various small fixes

leonschoorl:
## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -501,7 +501,7 @@ loadModules startAction useColor hdl modName dflagsM idirs = do
                   No top-level function called 'topEntity' or 'testBench' found,
                   nor any function annotated with a 'Synthesize' or 'TestBench'
                   annotation. If you want to synthesize a specific binder in
-                  #{show modName}, use '-main-is=myTopEntity'.
+                  #{show modName}, use '-main-is myTopEntity'.
                 |]
               _ ->
                 allImplicit
--- a/clash-prelude/src/Clash/Signal/Bundle/Internal.hs
+++ b/clash-prelude/src/Clash/Signal/Bundle/Internal.hs
@@ -1,9 +1,9 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE TemplateHaskell #-}
 
-module Clash.Signal.Bundle.Internal where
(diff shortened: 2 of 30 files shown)


clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -501,7 +501,7 @@ loadModules startAction useColor hdl modName dflagsM idirs = do
                   No top-level function called 'topEntity' or 'testBench' found,
                   nor any function annotated with a 'Synthesize' or 'TestBench'
                   annotation. If you want to synthesize a specific binder in
-                  #{show modName}, use '-main-is=myTopEntity'.
+                  #{show modName}, use '-main-is myTopEntity'.
```
martijnbastiaan, line 504: Oof, telling users to do a thing that doesn't work is the worst kind of helpful message :see_no_evil:

clash-prelude/src/Clash/Signal/Bundle/Internal.hs
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE TemplateHaskell #-}
 
-module Clash.Signal.Bundle.Internal where
+module Clash.Signal.Bundle.Internal (deriveBundleTuples) where
```
martijnbastiaan, line 4: Restricting imports on Internal modules is a bit of an anti-pattern IMO

martijnbastiaan:
Nice batch of changes once more, almost double digits now!

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
leonschoorl: Or is it the best kind of unhelpful message? :wink:

martijnbastiaan: Hah

clash-prelude/src/Clash/Signal/Bundle/Internal.hs
leonschoorl: Is it?

When I did this I was thinking of it as telling GHC that all the helper functions in this module aren't used else where, so it's free to inline them in `deriveBundleTuples` and doesn't have to keep their core around.

leonschoorl:
I've added some more stuff @alex-mckenna came up with

martijnbastiaan (approved):
Lookin good :)

PR #1981, opened 2021-11-01
Add `-fclash-partial-evaluator` flag

alex-mckenna:
The partial evaluator is now exposed to users through a command line flag for clash. This means as the evaluator becomes more stable it can be enabled (experimentally) and should produce increasingly more meaningful results.

## Still TODO:


martijnbastiaan (approved):
LGTM, why hasn't this been merged yet? :)

PR #1982, opened 2021-11-03
Tdp blockram configurations

lmbollen:
Different configurations for TDP Ram, for each port you can configure the write mode (WriteFirst, ReadFirst or NoChange) and whether or not to use additional output registers.

trueDualPortBlockRam takes a TDPConfig argument that contains these parameters, default set to WriteFirst and not using additional output registers.

This branch also implements conflicts for write collisions and BlackBox functions that generate the HDLs based on the configuration.

## Still TODO:

- [x] Make sure the haskell model adapts to the three different configurations: WriteFirst, ReadFirst and NoChange.
- [x] Make sure the different configurations generate the correct HDL
- [ ] Check that Vivado and Quartus both correctly infer the configurations. 
- [x] Make sure the Haskell model and HDLs have the same behavior by using the tests implemeting in the clash-testsuite under DualBlockRam

Future goals:

- [ ] Add reset signals to the primitive.
- [ ] Implement reset behavior in Haskell.
- [ ] Write tests with test whether the reset behavior is the same.
- [x] Add optional output registers

We might want to postpone writing new tests until we can use e.g. Hedgehog to test the implementations.


lmbollen:
I'm pretty sure this isn't ready to merge.

DigitalBrains1:
For other blockRAMs, we implement Read First and then provide [`readNew`](https://hackage.haskell.org/package/clash-prelude-1.5.0/candidate/docs/Clash-Explicit-BlockRam.html#v:readNew) to adapt that into Write First. A similar adapter could be written for No Change.

Has this option been explored? It could very well be that not all synthesis tools will infer the correct type of blockRAM (or maybe only with flags set to non-default values), but I'd like to explore this option. Also note that even if it is not correctly inferred, it might lead to the exact same result. That is, a particular FPGA might very well have a blockRAM that really only does Read First and implement other options with regular LEs around the blockRAM. So in that case it doesn't matter if the synthesis tool recognises the adapter or not, since it's the exact same end result.

DigitalBrains1 (commented):
This is a review of the first commit. It is titled "Added writemode configs + fixed conflict logic" but I couldn't find a change to the conflict logic and I thought it didn't need fixing in the first place, so could you elaborate what you fixed?

Also, you renamed a whole bunch of A's to Slow and B's to Fast, but there's also still a bunch of A's and B's. I can see how it is more explicit to call them Slow and Fast, but this mixed-mode signalling is rather confusing ;-). Could you make sure it's all consistent?

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1282,9 +1303,9 @@ trueDualPortBlockRam#, trueDualPortBlockRamWrapper ::
 
   -> (Signal domA a, Signal domB a)
   -- ^ Outputs data on /next/ cycle. If write enable is @True@, the data written
-  -- will be echoed. If write enable is @False@, the read data is returned. If
-  -- port enable is @False@, it is /undefined/.
-trueDualPortBlockRam# wmA wmB clkA enA weA addrA datA clkB enB weB addrB datB
+  -- will be echoed. If write enable is @False@, the read data is returned.
+trueDualPortBlockRam# wmA wmB clkA enA weA addrA datA
+ clkB enB weB addrB datB
```
DigitalBrains1, line 1315: I'd prefer the continuation of the line to be indented by four spaces rather than one.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
```diff
@@ -114,9 +114,10 @@ fifoMem
   -> Signal wdom (Maybe a)
   -> Signal rdom a
 fifoMem wclk rclk wen ren full raddr waddr wdataM =
-  fst $ trueDualPortBlockRam WriteFirst WriteFirst
+  fst $ trueDualPortBlockRam tdpDefault
     rclk wclk portA portB
  where
+
```
DigitalBrains1, line 121: This blank line is unneeded.

clash-prelude/src/Clash/Prelude/BlockRam.hs
```diff
@@ -901,5 +900,5 @@ trueDualPortBlockRam inA inB =
   -- ^ Outputs data on /next/ cycle. When writing, the data written
   -- will be echoed. When reading, the read data is returned.
 trueDualPortBlockRam wmA wmB inA inB =
-  E.trueDualPortBlockRam wmA wmB hasClock hasClock inA inB
+  E.trueDualPortBlockRam E.tdpDefault hasClock hasClock inA inB
```
DigitalBrains1: This shouldn't even compile, both before and after the commit? The implicit version doesn't get `wmA` and `wmB`. CPP mucking stuff up again, I bet you only built with multiple hidden :-).

tests/shouldwork/LoadModules/BlockRam/DualBlockRamDefinitions.hs
```diff
@@ -0,0 +1,158 @@
...
+  --Operations
+  Signal domA (RamOp 30 ThisOrThat) ->
+  Signal domB (RamOp 30 ThisOrThat) ->
+
+  --Output
+  ( Signal domA ThisOrThat
+  , Signal domB ThisOrThat )
+
+topEntity = trueDualPortBlockRam
+
```
DigitalBrains1: Please don't call this thing `topEntity`. That's a name we ascribe special meaning to. One of them not being polymorphic as this function is.

tests/Main.hs
```diff
@@ -696,8 +696,78 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "BlockRamTest" def{hdlSim=False}
         , runTest "Compression" def
         , runTest "DelayedReset" def
-        , runTest "DualBlockRam0" def{verilate=SimOnly}
-        , runTest "DualBlockRam1" def{verilate=SimOnly}
+        , runTest "DualBlockRam" def
+          { clashFlags = ["-main-is", "testBench_WF_WF_AB"]
+          , buildTargets = BuildSpecific ["testBench_WF_WF_AB"]
+          , verilate=SimOnly}
```
DigitalBrains1, line 702: That's quite interesting, that you need both `BuildSpecific` and `-main-is`. Is this is a bug in the `LoadModules` stuff? I'll investigate.

What's wrong with verilator?

tests/shouldwork/LoadModules/BlockRam/DualBlockRamDefinitions.hs
```diff
@@ -137,8 +137,8 @@ opsA11 = twice (RamRead 11 :> Nil) ++ replicate d20 RamNoOp
 opsB11 = RamRead 11 :> (fmap RamRead addrsB)
 
 -- Test12
-opsA12 = twice (RamRead 12 :> Nil) ++ (replicate d20 RamNoOp)
-opsB12 = RamRead 12 :> (replicate d10 RamNoOp)
+opsA12 = twice (RamRead 1 :> Nil) ++ (fmap RamRead addrsA)
+opsB12 = RamRead 1 :>  (fmap RamRead addrsB)
```
DigitalBrains1, line 141: This test no longer matches its description in the comments above.

clash-prelude/src/Clash/Prelude/BlockRam.hs
```diff
@@ -875,7 +875,10 @@ trueDualPortBlockRam ::
...
   ( HasCallStack
   , KnownNat nAddrs
-  , HiddenClockResetEnable dom1
+  , HiddenClock dom1
+  , HiddenEnable dom1
+  , HiddenClock dom2
+  , HiddenEnable dom2
   , HiddenClockResetEnable dom2
   , NFDataX a
   , BitPack a
```
DigitalBrains1, line 884: You forgot to remove `HiddenClockResetEnable dom2`.

tests/shouldwork/LoadModules/BlockRam/DualBlockRam.hs
```diff
@@ -0,0 +1,95 @@
...
+testBench_RF_NC_BC = testBench ReadFirst NoChange rst10 rst7 simOutA_RF_NC_BC simOutB_RF_NC_BC
+testBench_NC_WF_AB = testBench NoChange WriteFirst rst20 rst10 simOutA_NC_WF_AB simOutB_NC_WF_AB
+testBench_NC_WF_BC = testBench NoChange WriteFirst rst10 rst7 simOutA_NC_WF_BC simOutB_NC_WF_BC
+testBench_NC_RF_AB = testBench NoChange ReadFirst rst20 rst10 simOutA_NC_RF_AB simOutB_NC_RF_AB
+testBench_NC_RF_BC = testBench NoChange ReadFirst rst10 rst7 simOutA_NC_RF_BC simOutB_NC_RF_BC
+testBench_NC_NC_AB = testBench NoChange NoChange rst20 rst10 simOutA_NC_NC_AB simOutB_NC_NC_AB
+testBench_NC_NC_BC = testBench NoChange NoChange rst10 rst7 simOutA_NC_NC_BC simOutB_NC_NC_BC
+
+-- Expected output generation for all configurations with Template Haskell
+processSimOutput x = replace (0 ::Integer) undefined# $ tail x
```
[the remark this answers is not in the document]
DigitalBrains1: Ah no, I also needed an additional fix in PR #2081: I no longer reset `outputVerifier` to get it running right from the get go. You also enable the reset on `outputVerifier` again. It's almost as if you took an old version of the code rather than the version that is in `master`. Wait... that's it, isn't it? You took an "ours" merge strategy on that file!

tests/shouldwork/LoadModules/BlockRam/DualBlockRamDefinitions.hs
DigitalBrains1: Also, an unneeded blank line.

martijnbastiaan:
We've decided to close this; writing generic, cross-vendor true dual port blockrams is pretty much impossible. We'll add vendor specific ones to `clash-cores` instead.

PR #1984, opened 2021-11-05
clash-lib: Add enableToBit

basile-henry:
Enable used to be a Bool in the DSL, so we could use `boolToBit`.
However it now has its own type in the DSL (Enable domainName), so we
need a new conversion function in order to convert it to a Bool.
This should work for all backends as they all currently normalize Enable
to a Bool.

I have tested this on the `1.4` branch and it worked well (for the VHDL backend, I didn't test the others).
If this is accepted, could we backport/release it in `1.4`? It shouldn't be a breaking change.

## Still TODO:


martijnbastiaan (approved):

martijnbastiaan:
Implementation LGTM. A changelog entry would be nice.

@alex-mckenna You were more closely involved in the discussion on the FP slack. Could you give it a glance too?

alex-mckenna:
Copyright notice! Otherwise LGTM too

PR #1986, opened 2021-11-07
Remove `Hashable Term` and `Hashable Type`

martijnbastiaan:
Hashable states that for an instance to be lawful it should observe that
`a == b` implies `hash a == hash b`. Note that without this law, the
resulting hash cannot be used in probabilistic data structures such as
`HashMap`s. Because `Eq Term` and `Eq Type` are defined such that they
account for alpha equivalency, an instance for `Hashable Term` is
non-trivial - and indeed unable to be derived with `Generic`. Hence, the
instance removed in this commit did not observe the proper laws.

While this commit chooses to remove the instance, this isn't strictly
necessary:

 * Krzysztof Maziarz et al, have shown that a lawful instance is possible
   with reasonable complexity bounds[1]. However, its implementation is
   non-trivial, and Clash's only valid use of the instance is in its
   netlist stage, where it is easily replaced by a reliance on `Ord`.

 * `Eq` and `Ord` could be defined such that they do *not* account for
   alpha equivalency. But, this would make them harder to use in various
   containers. (*~~This will be proposed as an alternative to this PR soon.~~* nvm, next weekend it is.)

[1] https://arxiv.org/abs/2105.02856

## CI failure
It currently fails CI, as the netlist needs term equality (not alpha equality) for it to work. Alternatively, `AnnType ann0 t` and `AnnType ann1 t` could be considered not alpha equivalent when `ann0 /= ann1`:

https://github.com/clash-lang/clash-compiler/blob/365cbde57dfeff67f4f971de9a076d5dc8ac1a72/clash-lib/src/Clash/Core/Subst.hs#L823-L828

In any case, I want to emphasize that the status quo is **completely broken**. It is not an option to keep `Hashable` as it is - it will make Clash probabilistically miss type annotations.

## Still TODO:

  - [x] Pass CI

### DIFF
--- a/clash-lib/src/Clash/Core/Subst.hs
+++ b/clash-lib/src/Clash/Core/Subst.hs
@@ -11,6 +11,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
@@ -72,6 +73,7 @@ import qualified Data.List                 as List
 import qualified Data.List.Extra           as List
 import           Data.Ord                  (comparing)
 import           GHC.Stack                 (HasCallStack)
+import           GHC.SrcLoc.Extra          ()
 
 import           Clash.Core.HasFreeVars
 import           Clash.Core.Pretty         (ppr, fromPpr)
@@ -834,12 +836,8 @@ acmpType' = go
   go env (AppTy s1 t1) (AppTy s2 t2) =
     go env s1 s2 `thenCompare` go env t1 t2
   go _ (LitTy l1) (LitTy l2) = compare l1 l2
-  go env (AnnType _ t1) (AnnType _ t2) =
-    -- XXX: maybe ignore annotations, like we ignore ticks, i.e.
-    --
-    -- go env (AnnType t1) t2 = go env t1 t2
-    -- go env t1 (AnnType t2) = go env t1 t2
-    go env t1 t2
+  go env (AnnType a1 t1) (AnnType a2 t2) =
+    compare a1 a2 `thenCompare` go env t1 t2
   go _ t1 t2 = compare (getRank t1) (getRank t2)
 
   getRank :: Type -> Word
@@ -863,7 +861,7 @@ eqType = go
     tv1 == tv2 && go (varType tv1) (varType tv2) && go t1 t2
   go (AppTy s1 t1) (AppTy s2 t2) = go s1 s2 && go t1 t2
   go (LitTy l1) (LitTy l2) = l1 == l2
-  go (AnnType _ t1) (AnnType _ t2) = go t1 t2
+  go (AnnType a1 t1) (AnnType a2 t2) = a1 == a2 && go t1 t2
   go _ _ = False
 
 -- | Alpha equality for terms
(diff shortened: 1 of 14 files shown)


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -939,10 +937,8 @@ acmpTerm' inScope = go (mkRnEnv inScope)
     go env e1 e2 `thenCompare`
     acmpType' env l1 l2 `thenCompare`
     acmpType' env r1 r2
-  -- Look through ticks for aeq
-  go env (Tick _ e1) e2 = go env e1 e2
-  go env e1 (Tick _ e2) = go env e1 e2
-
+  go env (Tick t1 e1) (Tick t2 e2) =
```
christiaanb, line 940: I'm not a big fan of this change, given how carelessly we handle ticks.

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -298,17 +298,17 @@ recToLetRec (TransformContext is0 []) e = do
     -- corresponding (sub)field from the target variable.
     --
     -- TODO: See [Note: Breaks on constants and predetermined equality]
-    eqApp tcm v args (collectArgs . stripTicks -> (Var v',args'))
+    eqApp tcm v args (collectArgs . stripAllTicks -> (Var v',args'))
```
christiaanb, line 301: Perhaps add an additional line to the comment why we need `stripAllTicks`

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -361,7 +361,7 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
       specBndrs :: [Either Id TyVar]
       specBndrs = map (Lens.over Lens._Left (normalizeId tcm)) specBndrsIn
       specAbs :: Either Term Type
-      specAbs = either (Left . (`mkAbstraction` specBndrs)) (Right . id) specArg
+      specAbs = either (Left . stripAllTicks . (`mkAbstraction` specBndrs)) (Right . id) specArg
```
christiaanb, line 364: Could you add a comment as to why we need this `stripAllTicks` here.

christiaanb (approved):

martijnbastiaan:
Cool, thanks for the work @alex-mckenna. Although I agree with the alpha-eq change in general, given that this PR is mostly there for backporting (making sure Clash remains on Stackage) I feel nervous about pushing it there. On the other hand, any new issue reports leading from that would be much easier to debug than when it's batched up in a 1.6 upgrade..

@christiaanb Does this comment still apply:

> I'm not a big fan of this change, given how carelessly we handle ticks.

Or was that resolved on Slack?

christiaanb:
> 
> @christiaanb Does this comment still apply:
> 
> > I'm not a big fan of this change, given how carelessly we handle ticks.
> 
> Or was that resolved on Slack?

I’m fine with it as long as there is that comment for stripAllTicks

martijnbastiaan:
@christiaanb @alex-mckenna Could you the comments for `stripAllTicks`? I'm relatively sure it's okay in the `recToLetRec` case, but I'm unconvinced (either way) when specializing.

alex-mckenna:
I don't really like these comments, but this is also 100% my fault for not adding the comments when I made the branch pass CI for you. I'll change the comments tomorrow

martijnbastiaan:
Me neither, but it's the best I could manage :sweat_smile:

Issue #1983, opened 2021-11-03
Using `signed` signals for bit indexing in Verilog output confuses Yosys


gergoerdi:
From the following Clash code:

```
import Clash.Prelude

topEntity :: Index 4 -> BitVector 4
topEntity = bit . fromIntegral
```

Clash 1.4.6 generates the following Verilog:

```
/* AUTOMATICALLY GENERATED VERILOG-2001 SOURCE CODE.
** GENERATED BY CLASH 1.4.6. DO NOT MODIFY.
*/
`timescale 100fs/100fs
module topEntity
    ( // Inputs
      input [1:0] idx


      // Outputs
    , output reg [3:0] result
    );
  // symbiflow/bug/calculator/clash.hs:4:1-9
  wire signed [63:0] wild;
  // symbiflow/bug/calculator/clash.hs:4:1-9
  wire signed [63:0] c$wild_app_arg;
  wire  c$din;

  assign wild = $signed(c$wild_app_arg);

  assign c$wild_app_arg = $unsigned({{(64-2) {1'b0}},idx});

  assign c$din = (1'b1);

  // replaceBit start
  always @(*) begin
    result = 4'b0000;
    result[(wild)] = c$din;
  end
  // replaceBit end


endmodule


```

Using a signed signal for indexing into the bits of `result` is strange (what is the -3rd bit?), and it exposes a bug in Yosys: https://github.com/YosysHQ/yosys/issues/3051

If I manually change the generated Verilog output by changing the following declarations and definitions:

```
  wire [63:0] wild;
  assign wild = $unsigned(c$wild_app_arg);
```

then Yosys synthesizes the correct circuit.

Of course, I understand that this is a Yosys bug and maybe Clash is generating copasetic Verilog, but in case that my intuition is right and `signed` isn't how bits should be indexed, would it make sense to change Clash so that it works nicely with the open source Yosys synthesizer?

christiaanb:
I guess we could index with `Word` instead of `Int`

leonschoorl:
Just look it up and the Verilog standard is very permissive about these things.  
It says reading from out or range bits should return an `x` and writing to them should have not have no effect.

But I agreed we shouldn't be indexing bits with `signed` indices, especially when that makes our generated code unusable with Yosys.

alex-mckenna:
Yosys have a candidate fix for this, so #1993 is closed. If this doesn't end up being a good fix we can re-open later

PR #1993, opened 2021-11-09
Change bit indexing to use `Word` parameter in `replaceBit#` (#1983)

gergoerdi:
This is more of a proof-of-concept for #1983 than a fully tested fix. With it applied, I can see that it is an effective workaround for https://github.com/YosysHQ/yosys/issues/3051, at least in the simple case. I haven't tried anything more complex yet; a good test case should be https://github.com/gergoerdi/clash-calculator which is where I initially noticed this.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
@@ -2022,11 +2022,11 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
         op :: KnownNat n => BitVector n -> Int -> Proxy n -> (Integer,Integer)
         op u i _ = (toInteger m, toInteger v)
           where Bit m v = (BitVector.index# u i)
-  "Clash.Sized.Internal.BitVector.replaceBit#" -- :: :: KnownNat n => BitVector n -> Int -> Bit -> BitVector n
+  "Clash.Sized.Internal.BitVector.replaceBit#" -- :: :: KnownNat n => BitVector n -> Word -> Bit -> BitVector n
     | Just (_, n) <- extractKnownNat tcm tys
     , [ _
       , PrimVal bvP _ [_, Lit (NaturalLiteral mskBv), Lit (IntegerLiteral bv)]
-      , valArgs -> Just [Literal (IntLiteral i)]
+      , valArgs -> Just [Literal (WordLiteral i)]
       , PrimVal bP _ [Lit (WordLiteral mskB), Lit (IntegerLiteral b)]
       ] <- args
     , primName bvP == "Clash.Sized.Internal.BitVector.fromInteger#"
@@ -2039,7 +2039,7 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
       where
         op :: KnownNat n => BitVector n -> Int -> Bit -> Proxy n -> (Integer,Integer)
         -- op bv i b _ = (BitVector.unsafeMask res, BitVector.unsafeToInteger res)
-        op bv i b _ = splitBV (BitVector.replaceBit# bv i b)
+        op bv i b _ = splitBV (BitVector.replaceBit# bv (fromIntegral i) b)
   "Clash.Sized.Internal.BitVector.setSlice#"
   -- :: SNat (m+1+i) -> BitVector (m + 1 + i) -> SNat m -> SNat n -> BitVector (m + 1 - n) -> BitVector (m + 1 + i)
     | mTy : iTy : nTy : _ <- tys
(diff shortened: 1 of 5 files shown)


christiaanb:
This commit shows that `Enum` could do with an overhaul to use `Word`; I wonder whether it’s worth submitting a proposal to the CLC. Or are there enumerable things that can have a negative amount?

alex-mckenna:
> Or are there enumerable things that can have a negative amount?

Unless I'm missing your meaning, `Int`?

gergoerdi:
> > Or are there enumerable things that can have a negative amount?
> 
> Unless I'm missing your meaning, `Int`?

But you can still *enumerate* `Int`s, e.g. `0, 1, -1, 2, -2, `.... Or by a raw coercion, `0, 1, ..., maxBound, minBound, minBound + 1, ..., -1`.

christiaanb:
No, I do mean that it should be `toEnum :: Word -> a` and `fromEnum :: a -> Word`

leonschoorl:
While it would be convenient for us if the Enum class would be changed to use `Word`s.
It wouldn't really be an improvement for Haskell in general, `toEnum 42 :: Bool` is just as wrong as `toEnum (-1) :: Bool`.
And `(-1) :: Word` isn't even an compile error, just a warning.
Also there is the added confusion around `fromEnum (-1 ::Int) == (maxBound :: Word)`

christiaanb:
Yeah, I know that `toEnum` and `fromEnum` are explicitly conversions from and to `Int`: it even says so in their documentation and in the Haskell report.

But I always read them more in the spirit of: "Class Enum defines operations on sequentially ordered types.", where `toEnum a` picks the `a`'th (0-based) element in the sequence; and `fromEnum a` gives the (0-based) position of `a` in the sequence.

martijnbastiaan:
I'd say if they'd consider a breaking change like that, they should consider using `Integer`/`Natural`.

gergoerdi:
> I haven't tried anything more complex yet; a good test case should be https://github.com/gergoerdi/clash-calculator which is where I initially noticed this.

Unfortunately, `clash-calculator` still doesn't work fully with this patch -- the next hurdle is https://github.com/YosysHQ/yosys/issues/3078 but that's outside the world of Clash (and might turn out to be a bug somewhere else in Symbiflow instead of Yosys anyway).

gergoerdi:
> I wonder whether it’s worth submitting a proposal to the CLC. Or are there enumerable things that can have a negative amount?

Isn't that, realistically, just a roundabout way of saying this won't get into Clash?

christiaanb:
> > I wonder whether it’s worth submitting a proposal to the CLC. Or are there enumerable things that can have a negative amount?
> 
> Isn't that, realistically, just a roundabout way of saying this won't get into Clash?

No, we’ll merge this

martijnbastiaan:
@gergoerdi Is there anything that still needs to happen for this PR? I'd like to get this in the 1.6 release (02-02-2022) if possible :)

DigitalBrains1:
Some musing.

While discussing my GHC issue report for `fromEnum :: Natural`, a GHC developer [expressed this viewpoint](https://gitlab.haskell.org/ghc/ghc/-/issues/20291#note_374325):
> IMHO `Enum` should not been trying to do more then desugar `[x .. y]`, `[x..]` and `[x,y...z]` notations. `fromEnum` and `toEnum` are handy helpers to specify the deriving, but IMO they didn't need to belong to the class.

This is one developer's opinion, and I disagree on the "only desugaring" when I observe that `Enum` also has `succ` and `pred` and those are pretty good IMHO. But it _is_ the case that the minimal complete definition of `Enum` is providing `fromEnum` and `toEnum`, and the thought that these purely exist to facilitate writing (or deriving) `Enum` instances does resonate with me. I wonder whether `fromEnum` and `toEnum` were ever even meant to be _used_ on an instance rather than only _defined_ for quick and easy instance building (after which you _use_ the other functions).

The takeaway is this: maybe we are using `Enum` in a manner different from its intent in functions such as:
https://github.com/clash-lang/clash-compiler/blob/a5174f61cb3d3a5cacce92de13fc4135dde286dd/clash-prelude/src/Clash/Prelude/ROM.hs#L51-L62

martijnbastiaan:
@alex-mckenna Could you pretty please have a look whether we can include this for 1.6? :)

alex-mckenna:
@kloonbot run_ci 9e64b68

Let's see what kloonbot says

alex-mckenna (commented):
On the whole, I think this is fine. I would like to see a regression test if possible, although I think that means adding `Yosys` as a vendor tool in the testsuite. If I make a PR for that, can you make a test for this @gergoerdi?

gergoerdi:
I'm not sure how I feel about this change, now that https://github.com/YosysHQ/yosys/issues/3051 has a candidate fix. If using signed types to index bits is valid Verilog/VHDL, and the practical consideration of working around a Yosys bug no longer applies, then is there an advantage (other than maybe decreasing surprises) to this change?

alex-mckenna:
Hmm yes, if Yosys have a fix then I agree we don't think we really need this since it _is_ allowed in Verilog / VHDL. I don't think there's much to be said for less surprise given how many Haskell APIs index with `Int` when `Word` is arguably more correct anyway...

PR #1994, opened 2021-11-09
Show days when pretty printing durations

### LINKED ISSUE #1989: reportTimeDiff formats time inconveniently
So this is a pretty minor issue, but just for the record, here's some output from my compilation:
```
GHC: Parsing and optimising modules took: 13.392s
GHC: Loading external modules from interface files took: 0.004s
GHC: Parsing annotations took: 0.009s
Clash: Parsing and compiling primitives took 0.106s
GHC+Clash: Loading modules cumulatively took 17.499s
Clash: Compiling Quasisystolic.MLP2.topEntity
Clash: Normalization took 23h59m53015.564s
Clash: Netlist generation took 17.250s
Clash: Total compilation took 23h59m53063.423s
```
I don't understand the formatting strings used to specify `Clock.formatTime`, but lumping 15 hours into the seconds field seems a bit ridiculuos.

alex-mckenna:
When showing time stats for parts of Clash, if something takes longer
than a day, it shows as 23h59mXs, where X is some large number of
seconds. When a can be measured in days, it should instead show a
message like '2d7h30m45s' instead.

Closes #1989

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Util.hs
+++ b/clash-lib/src/Clash/Util.hs
@@ -263,11 +263,11 @@ pkgIdFromTypeable :: Typeable a => a -> String
 pkgIdFromTypeable = tyConPackage . typeRepTyCon . typeOf
 
 reportTimeDiff :: UTCTime -> UTCTime -> String
-reportTimeDiff start end =
+reportTimeDiff end start =
   Clock.formatTime Clock.defaultTimeLocale fmt
     (Clock.UTCTime (toEnum 0) (fromRational (toRational diff)))
  where
-  diff = Clock.diffUTCTime start end
+  diff = Clock.diffUTCTime end start
   fmt  | diff >= 86400
        = "%-Dd%-Hh%-Mm%-S%03Qs"
        | diff >= 3600
--- a/clash-lib/src/Clash/Util.hs
+++ b/clash-lib/src/Clash/Util.hs
@@ -268,7 +268,9 @@ reportTimeDiff start end =
     (Clock.UTCTime (toEnum 0) (fromRational (toRational diff)))
  where
   diff = Clock.diffUTCTime start end
-  fmt  | diff >= 3600
+  fmt  | diff >= 86400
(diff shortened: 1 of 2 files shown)


clash-lib/src/Clash/Util.hs
```diff
@@ -268,7 +268,9 @@ reportTimeDiff start end =
     (Clock.UTCTime (toEnum 0) (fromRational (toRational diff)))
  where
   diff = Clock.diffUTCTime start end
-  fmt  | diff >= 3600
+  fmt  | diff >= 86400
+       = "%-Dd%-Hh%-Mm%-S%03Qs"
```
leonschoorl, line 272: According to the [docs](https://hackage.haskell.org/package/time-1.12/docs/Data-Time-Format.html#v:formatTime)
`%D` for `Day` (which includes `UTCTime`) means: `%m/%d/%y`

leonschoorl: You might hope to use `%d` directly and be good up to a month, but that doesn't work either.
Because `0 :: UTCTime` is 1858-11-17 00:00:00 UTC

Ideally we should be using `NominalDiffTime` instead of `UTCTime`, but sadly that's only supported in `time >= 1.9`.
And we can only use `time >= 1.9` starting with `ghc >= 8.8`. (`time` itself can compile on older GHC versions, but since it's also a dependency of `ghc` we have to use the `time` version GHC ships with)
Which makes it no good for clash-1.4, and also not for `master` as it supports ghc-8.6.

leonschoorl (changes requested):

clash-lib/src/Clash/Util.hs
christiaanb: We use `utcDiffTime` which returns a `NominalDiffTime` where the `D` modifier denotes Days (of week). Note that that’s not day-of-the-week.

christiaanb (approved):
LGTM!

PR #2000, opened 2021-11-12
Keep recursion info in BindingMap

alex-mckenna:
In GenerateBindings, the recursion information about global bindings
is currently discarded by Clash. However, much like the recursion info
for let bindings, this can be used to make the partial evaluator more
efficient - as non-recursive global bindings can always be inlined
without consuming fuel.

## Still TODO:


clash-lang-builder (approved):

alex-mckenna:
Oh no, the machines have gained sentience

Issue #1458, opened 2020-07-18
Synchronous RAM simulation is too strict in address argument


gergoerdi:
To demonstrate this problem, I create a block RAM with 8 cells, and then try to address it with a 4-bit signal:

```
foo :: (HiddenClockResetEnable dom) => Signal dom (Unsigned 4) -> Signal dom (Unsigned 8)
foo addr = mux (register False $ addr .<. 8) ram (pure 0xff)
  where
    ram = blockRam1 NoClearOnReset (SNat @8) 0 addr (pure Nothing)
```

Since the read from RAM is guarded by a multiplexer that checks the address (from the previous cycle, since the RAM is synchronous), I would expect this to work in simulation. However, it fails seemingly because `blockRam1`'s simulation tries to interpret the out-of-bounds read address:

```
sim = L.tail $ sampleN 15 $
      let addr = fromList [0..15] in
      bundle (addr, foo addr) :: Signal System (Unsigned 4, Unsigned 8)
```

```
> mapM_ print sim
(1,255)
(2,0)
(3,0)
(4,0)
(5,0)
(6,0)
(7,0)
(8,0)
(9,255)
*** Exception: index out of bounds in call to: Data.Sequence.index 8
CallStack (from HasCallStack):
  error, called at libraries/containers/containers/src/Data/Sequence/Internal.hs:2056:7 in containers-0.6.2.1:Data.Sequence.Internal
```

Of course, one can work around this by using a dummy address when it doesn't matter:

```
foo' :: (HiddenClockResetEnable dom) => Signal dom (Unsigned 4) -> Signal dom (Unsigned 8)
foo' addr = mux (register False sel) ram (pure 0xff)
  where
    sel = addr .<. 8
    addr' = mux sel addr (pure 0)

    ram = blockRam1 NoClearOnReset (SNat @8) 0 addr' (pure Nothing)
```

christiaanb:
Ah, so the actual issue is that we force the output elements of the Signal/Stream to WHNF using `seqX`, which only catches and (and the subsequently boxes) `XException`. But `Data.Sequence.index` throws `ErrorCall` exceptions on out-of-bounds indexing. Perhaps for the blockRam we want to use a version of `seqX` that also catches `ErrorCall` exceptions.

PR #2006, opened 2021-11-19
Fix asyncRam#: multiple clocks, undefineds, laziness, seqX

DigitalBrains1:
1. In Haskell simulation, the way read samples were produced, with
`unsafeSynchronizer`, was simply wrong. It would compress and duplicate
samples while traversing clock domains, which is not a correct model of
how it works.

The generated HDL was fine, it only affected Haskell simulation.

2. The Haskell model of `asyncRAM#` treated an _undefined_ write enable
as an asserted enable. But an _undefined_ value in Haskell can
correspond to any value whatsoever in HDL, so HDL simulation might or
might not write. With this commit, the `XException` of the write enable
is written as the value in the RAM, since it could have either been
written to or not been written to. On the next read of that address, it
will return the `XException`.

This issue did not propagate to `asyncRam` and `asyncRamPow2`, since
there, the same condition also causes the write address to be undefined,
and this is properly handled by the primitive.

3. The Haskell model threw an `XException` when an out-of-bounds write
address was used, halting simulation, as the whole `Signal` evaluated to
`XException`, not just the sample. Now, every address in RAM is
filled with an `XException`, to model that an OOB address might get
treated as some actual address in hardware.

4. The `asyncRAM#` primitive was also too strict in most of its inputs.
Combinatorially feeding the read output to the write-side inputs would
lock up the simulation, while it is a valid circuit. This problem did
not propagate to the `asyncRam` and `asyncRamPow2` functions, which are
lazy enough because the signals go through `Signal`'s `fmap` and `<*>`.

5. Several `seq(X)`s have been added to avoid memory leaks.

Additionally, documentation for memory components was harmonized and
corrected.

By the way, I did not copy the multi-clock `Signal` traversal from `veryUnsafeSynchronizer`. Rather, I only took the concept of a relative timekeeping from it, which allows using `Int` as the value doesn't grow as time passes in the simulation. I constructed the multi-clock `Signal` traversal for `asyncRam` from scratch to gain a proper insight into its operation. In the end, it turned out to be nearly identical to `veryUnsafeSynchronizer` apart from the sign of `relTime` and the way signals propagate across the domain boundary (the latter is deliberate and matches simulation of the generated HDL).

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/RAM.hs
+++ b/clash-prelude/src/Clash/Explicit/RAM.hs
@@ -35,16 +35,17 @@ import GHC.Stack             (HasCallStack, withFrozenCallStack)
 import GHC.TypeLits          (KnownNat)
 import qualified Data.Sequence as Seq
 
-import Clash.Explicit.Signal
-  (unbundle, unsafeSynchronizer, KnownDomain, andEnable)
+import Clash.Explicit.Signal (unbundle, KnownDomain, andEnable)
 import Clash.Promoted.Nat    (SNat (..), snatToNum, pow2SNat)
 import Clash.Signal.Internal (Clock (..), Signal (..), Enable, fromEnable)
+import Clash.Signal.Internal.Ambiguous (clockPeriod)
 import Clash.Sized.Unsigned  (Unsigned)
-import Clash.XException      (errorX, maybeIsX, fromJustX)
+import Clash.XException      (errorX, maybeIsX, fromJustX, seqX)
 
 -- | Create a RAM with space for 2^@n@ elements
 --
--- * __NB__: Initial content of the RAM is 'undefined'
+-- * __NB__: Initial content of the RAM is /undefined/, reading it will throw an
+-- 'Clash.XException.XException'
 --
 -- Additional helpful information:
 --
@@ -76,7 +77,8 @@ asyncRamPow2 = \wclk rclk en rd wrM -> withFrozenCallStack
 
 -- | Create a RAM with space for @n@ elements
 --
--- * __NB__: Initial content of the RAM is 'undefined'
+-- * __NB__: Initial content of the RAM is /undefined/, reading it will throw an
+-- 'Clash.XException.XException'
 --
 -- Additional helpful information:
 --
@@ -111,7 +113,8 @@ asyncRam = \wclk rclk gen sz rd wrM ->
 
 -- | RAM primitive
 asyncRam#
-  :: ( HasCallStack
+  :: forall wdom rdom n a
+   . ( HasCallStack
      , KnownDomain wdom
      , KnownDomain rdom )
   => Clock wdom
@@ -132,31 +135,57 @@ asyncRam#
   -- ^ Value to write (at address @w@)
   -> Signal rdom a
   -- ^ Value of the @RAM@ at address @r@
-asyncRam# wclk rclk en sz rd we wr din =
-    unsafeSynchronizer wclk rclk dout
+asyncRam# !_ !_ en sz rd we wr din = dout
   where
-    rd'  = unsafeSynchronizer rclk wclk rd
     ramI = Seq.replicate
-              (snatToNum sz)
+              szI
               (withFrozenCallStack (errorX "asyncRam#: initial value undefined"))
-    en' = fromEnable (andEnable en we)
-    dout = go ramI rd' en' wr din
+    en0 = fromEnable (andEnable en we)
+    dout = if rPeriod == wPeriod
+           then goSingle ramI rd en0 wr din
+           else go 0 ramI rd en0 wr din
+    rPeriod = snatToNum (clockPeriod @rdom) :: Int
+    wPeriod = snatToNum (clockPeriod @wdom) :: Int
     szI = snatToNum sz :: Int
 
-    go :: Seq.Seq a -> Signal wdom Int -> Signal wdom Bool
-       -> Signal wdom Int -> Signal wdom a -> Signal wdom a
-    go !ram (r :- rs) (e :- es) (w :- ws) (d :- ds) =
-      let ram' = upd ram e (fromEnum w) d
+    goSingle :: Seq.Seq a -> Signal rdom Int -> Signal wdom Bool
+       -> Signal wdom Int -> Signal wdom a -> Signal rdom a
+    goSingle !ram (r :- rs) ~(e :- es) wt@(~(w :- ws)) dt@(~(d :- ds)) =
+      let ram0 = upd ram e w d
           o    = ram `safeAt` r
-      in  o :- go ram' rs es ws ds
+      in  o :- (o `seqX` wt `seq` dt `seq` goSingle ram0 rs es ws ds)
 
-    upd ram we' waddr d = case maybeIsX we' of
+    -- Given
+    --   tR = absolute time of next active edge of read clock
+    --   tW = absolute time of next active edge of write clock
+    -- relTime is defined as relTime = tW - tR
+    --
+    -- Put differently, relative time 0 points at the next active edge of the
+    -- read clock, and relTime points at the next active edge of the write
+    -- clock.
+    go :: Int -> Seq.Seq a -> Signal rdom Int -> Signal wdom Bool
+       -> Signal wdom Int -> Signal wdom a -> Signal rdom a
+    go   relTime !ram rt@(~(r :- rs)) et@(~(e :- es)) wt@(~(w :- ws))
+         dt@(~(d :- ds))
+      | relTime < 0 = let ram0 = upd ram e w d
+                      in wt `seq` dt `seq`
+                         go (relTime + wPeriod) ram0 rt es ws ds
+      | otherwise   = let o = ram `safeAt` r
+                      in o :- (o `seqX` go (relTime - rPeriod) ram rs et wt dt)
+
+    upd ram we0 waddr d = case maybeIsX we0 of
       Nothing -> case maybeIsX waddr of
-        Nothing -> fmap (const (seq waddr d)) ram
-        Just wa -> safeUpdate wa d ram
+        Nothing -> -- Put the XException from `waddr` as the value in all
+                   -- locations of `ram`.
+                   seq waddr d <$ ram
+        Just wa -> -- Put the XException from `we` as the value at address
+                   -- `waddr`.
+                   safeUpdate wa (seq we0 d) ram
       Just True -> case maybeIsX waddr of
-        Nothing -> fmap (const (seq waddr d)) ram
-        Just wa -> safeUpdate wa d ram
+        Nothing -> -- Put the XException from `waddr` as the value in all
+                   -- locations of `ram`.
+                   seq waddr d <$ ram
+        Just wa -> d `seqX` safeUpdate wa d ram
       _ -> ram
 
     safeAt :: HasCallStack => Seq.Seq a -> Int -> a
@@ -174,8 +203,9 @@ asyncRam# wclk rclk en sz rd we wr din =
       if (0 <= i) && (i < szI) then
         Seq.update i a s
       else
-        withFrozenCallStack
-          (errorX ("asyncRam: write address " ++ show i ++
-                   " not in range [0.." ++ show szI ++ ")"))
+        let d = withFrozenCallStack
+                  (errorX ("asyncRam: write address " ++ show i ++
+                           " not in range [0.." ++ show szI ++ ")"))
+        in d <$ s
     {-# INLINE safeUpdate #-}
 {-# NOINLINE asyncRam# #-}
--- a/changelog/2021-11-19T16_19_41+01_00_asyncram_multi_clk_fix
+++ b/changelog/2021-11-19T16_19_41+01_00_asyncram_multi_clk_fix
@@ -1,0 +1,2 @@
+FIXED: `asyncRam` with different read and write clocks produced the wrong results in Haskell simulation.([#2006](https://github.com/clash-lang/clash-compiler/pull/2006))
+FIXED: `Clash.Explicit.RAM.asyncRam#` incorrectly treated an _undefined_ write enable as asserted. It now causes an _undefined_ value to be written instead. This problem did not propagate to the other `asyncRam` functions, where the same condition would simultaneously lead to an undefined write address, which would be handled correctly.([#2006](https://github.com/clash-lang/clash-compiler/pull/2006))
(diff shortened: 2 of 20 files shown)


DigitalBrains1:
Before this patch, `asyncRam#` was not lazy enough in the write-side inputs. For the local `goSingle` function, we only need lazy matching on the write-side inputs to deal with combinatorially feeding the output of `asyncRam#` to its write-side inputs. I'll explain the factors involved a bit more here.

First off, note that for `Signal`, the functions `fmap` and `<*>` are lazy in their second argument. This means that any combinatorial function involving those already lazily matches the `Signal` constructor via that route. This is also the explanation for the fact that `asyncRam` and `asyncRamPow2` already worked fine before this patch when combinatorially feeding the output to the write-side inputs, because `asyncRam` uses `<$>`.

Then, let's go over the inputs of `asyncRam#` one by one:
* `rd`: Since this is a combinatorial read operation, feeding the output back to `rd` is a proper combinatorial loop in the circuit and that makes it an invalid circuit. This input need not match the `Signal` constructor lazily.
* `we`: Actually, this uses `Applicative Signal`! It uses `andEnable` to combine the `en` input and the `we` input, which conceptually serve the exact same purpose (gating the write side). And `andEnable` uses `.&&.` which uses `Applicative`. However, it felt more consistent to also lazily match on the pattern for this input and not rely on this detail.
* `wr`: This one properly needs the lazy pattern match. It would lock up otherwise.
* `din`: Same as `wr`.

So for `goSingle`, we can match strictly for `rd`.

For `go`, I'm not so sure, because the pattern matching is done before deciding whether the next step in the simulation is the read side or the write side. So if we are currently processing samples for the write side and the read side is a sample from the _future_, I don't think the constructor for the read side should be matched strictly either. Looking at `veryUnsafeSynchronizer` I don't see a way to construct a stuck evaluation loop, but based on the observation that a future read sample could conceivably depend on the current sample in the write domain via some other component, I think we shouldn't match strictly on the `Signal` constructor in `rd`.

martijnbastiaan (approved):

clash-prelude/src/Clash/Explicit/RAM.hs
```diff
@@ -132,32 +135,55 @@ asyncRam#
...
+    -- read clock, and relTime points at the next active edge of the write
+    -- clock.
+    go :: Int -> Seq.Seq a -> Signal rdom Int -> Signal wdom Bool
+       -> Signal wdom Int -> Signal wdom a -> Signal rdom a
+    go   relTime !ram rt@(~(r :- rs)) et@(~(e :- es)) wt@(~(w :- ws))
+         dt@(~(d :- ds))
+      | relTime < 0 = let ram0 = upd ram e w d
+                      in go (relTime + wPeriod) ram0 rt es ws ds
+      | otherwise   = let o = ram `safeAt` r
+                      in o :- go (relTime - rPeriod) ram rs et wt dt
```
martijnbastiaan, line 173: Forgot to hit "start a review" :sweat_smile: 

```suggestion
                      in o :- (o `seqX` go (relTime - rPeriod) ram rs et wt dt)
```

The consumer could do this too, but in other memory blocks we take this precaution as well so it seems like a good fit here. If this is left out on purpose, could you add a comment as to why? :)

DigitalBrains1: I did the leakage assessment together with @leonschoorl , perhaps he can tell you why. I cannot. If you don't do this, do unused read outputs stay in the stream as thunks when you're further into the future?

I only looked at `blockRam#` after evaluating leakage of this function, and I noticed something else. I'm going to put that in a separate comment to take advantage of GitHub's suggestion view.

clash-prelude/src/Clash/Explicit/RAM.hs
```diff
@@ -132,32 +135,55 @@ asyncRam#
...
+
```
DigitalBrains1, line 174:
```suggestion
      | relTime < 0 = let ram0 = upd ram e w d
                      in wt `seq` dt `seq` go (relTime + wPeriod) ram0 rt es ws ds
      | otherwise   = let o = ram `safeAt` r
                      in o :- go (relTime - rPeriod) ram rs et wt dt

```

`blockRam#` does this:
```haskell
  go !ram o ret@(~(re :- res)) rt@(~(r :- rs)) et@(~(e :- en)) wt@(~(w :- wr)) dt@(~(d :- din)) = do
    o `seqX` (o :-) <$> (ret `seq` rt `seq` et `seq` wt `seq` dt `seq`
```
It explicitly `seq`s the whole `Signal`s, I think this means the constructor `:-` is evaluated. I think we should do that as well, but the way `blockRam#` does it is overkill. When we in `asyncRam#` are evaluating a write step, we should leave read well alone, and vice-versa. When we're in write, we necessarily evaluate `e`, and thereby we don't need to `seq` `et` anymore. However, `w` and `d` might go unevaluated (see other review comment for `w`).

And while making a read-side step, we evaluate `r`, so we don't need to `seq` `rt`.

clash-prelude/src/Clash/Explicit/RAM.hs
```diff
@@ -132,32 +135,55 @@ asyncRam#
...
+    upd ram e waddr d = case maybeIsX waddr of
+      Nothing -> case maybeIsX e of
+                   Just False -> ram
+                   _ ->  -- Put the XException from `waddr` as the value in all
+                         -- locations of `ram`.
+                         fmap (const (seq waddr d)) ram
+      Just wa -> case maybeIsX e of
+        Nothing -> -- Put the XException from `e` as the value at address
```
DigitalBrains1, line 182: I'd like to reorder this again, because we don't need to evaluate `waddr` if `e` is `False`. I ordered it to avoid duplicating code, but this code duplication provides a speed benefit if people have an expensive `waddr`.

clash-prelude/src/Clash/Explicit/RAM.hs
DigitalBrains1: I think we should use `defaultSeqX` here and for the data written to the memory, so people can do `-fsuper-strict` on it.

DigitalBrains1 (commented):
I'd like to make a new unit test for this for testing undefined inputs. Doing so in a HDL test bench is not the proper approach. To be continued...

clash-prelude/src/Clash/Explicit/RAM.hs
DigitalBrains1: Ah we can't do `defaultSeqX` because that requires the data to be `NFDataX` and it is not. So `seqX` it is.

DigitalBrains1: I've added the `seqX`. @leonschoorl , do you agree?

leonschoorl: @martijnbastiaan My reasoning for not having the ``o `seqX` `` went something like this:

Without that extra ``o `seqX` ``, if you feed that resulting signal into an register that isn't enabled (some of the time), then after taking of the head of this signal, the `o` can be garbage collected without having to be evaluated.

Now `asyncRam#` does something seemingly similar:
https://github.com/clash-lang/clash-compiler/blob/a076fa9e6fdcd0d07f547a1177e5bff28981809d/clash-prelude/src/Clash/Signal/Internal.hs#L1181-L1185
But that's subtly different.
There it forces `o` to reprevent to prevent the build up of a huge chain of thunks in the first argument of `go` that otherwise can happen when the register is held in reset.

---
Thinking about it some more and looking at `blockRam#`, forcing the `o` might be a good thing, if you see it as forcing the forcing the indexing into the (potentially large) contents for the ram.

`blockRam#` does `` o `seqX` (o :- ... )``, which make more sense to me than `` o :-  (o `seqX` ... ) ``, or is that no good because of the multiple clock domains here?

`blockRam#` also make sure to force the data being written into it, which also seems like a good idea, unless it's incompatible the the multiple domains here?

Does `asyncRam` suffer from https://github.com/clash-lang/clash-compiler/issues/1458 ? Which may need [`seqErrorX`](https://github.com/clash-lang/clash-compiler/commit/c3fe11fe2af4181c27e5de785b2c929c7fc2e856) instead of `seqX`.

As for the `seqX` vs `defaultSeqX`, you could just add the `NFDataX` constraint, (almost?) any other storage element has that anyway: `register`, `blockRam`, `delay`, ....

DigitalBrains1: > Thinking about it some more and looking at `blockRam#`, forcing the `o` might be a good thing, if you see it as forcing the forcing the indexing into the (potentially large) contents for the ram.

So the indexing operation itself seems pretty low computational overhead if the output is not used, but the same might not hold for the read address, which necessarily also will be evaluated. It's not really clear to me why you would want to force `o` to WHNF here, or to quote myself: _If you don't do this, do unused read outputs stay in the stream as thunks when you're further into the future?_ I don't know.

> `blockRam#` does `` o `seqX` (o :- ... )``, which make more sense to me than `` o :-  (o `seqX` ... ) ``, or is that no good because of the multiple clock domains here?

For the single clock domain version, that seems fine, but for the multiple clock domain version I'm not so sure indeed. Producing the constructor first seems like the safer option.

> `blockRam#` also make sure to force the data being written into it, which also seems like a good idea, unless it's incompatible the the multiple domains here?

I [do that](https://github.com/clash-lang/clash-compiler/pull/2006/commits/4c7f11326268bbdfec5561ecf5183bf63afed0f0#diff-42184b38e50f74675cd2af9c3fec2656e3036f76967a5c98a846df0bb37dbf46R188) as well.

> Does `asyncRam` suffer from https://github.com/clash-lang/clash-compiler/issues/1458 ? Which may need [`seqErrorX`](https://github.com/clash-lang/clash-compiler/commit/c3fe11fe2af4181c27e5de785b2c929c7fc2e856) instead of `seqX`.

No, out-of-bounds indexing evaluates the _whole `Signal`_ to `XException` rather than a sample in the `Signal`. I think this is wrong (as we discussed on Slack), but for now, I'd say it is handled by `seqX`. However, I'll add a unit test that tests this case, so CI will punish us if we ever make that mistake (for instance when we change OOB to produce `ErrorCall`).

> As for the `seqX` vs `defaultSeqX`, you could just add the `NFDataX` constraint, (almost?) any other storage element has that anyway: `register`, `blockRam`, `delay`, ....

Ah, this comment made me realise I had forgotten the `backport 1.4` label. Since I'd like to backport this fix to 1.4, I don't think we should add constraints in this PR. Perhaps in a separate PR that is not backported.

DigitalBrains1: > No, out-of-bounds indexing evaluates the _whole `Signal`_ to `XException` rather than a sample in the `Signal`. 

I'm only half right. This is what happens for OOB write address and asserted write enable. And as discussed on Slack, it is defensible to stop the simulation in such a case. For OOB read address, it's currently the sample that is an `XException`, not the `Signal`.

Anyway, I added a unit test that tests for #1458.

DigitalBrains1: >> `blockRam#` does `` o `seqX` (o :- ... )``, which make more sense to me than `` o :-  (o `seqX` ... ) ``, or is that no good because of the multiple clock domains here?

>For the single clock domain version, that seems fine [edit] Hmmm not so sure anymore, man this is intricate [/edit]

Yes, it's fine, I just tested it with a loop that would get stuck otherwise.

martijnbastiaan (commented):
Commenting to shut up the PR bot. I won't have the time to review this PR properly, unfortunately. My suggestion would be to take out the controversial bits (if any) and put them in a separate PR - just so we can get the important stuff in.

christiaanb (approved):
LGTM

I think I would like one more improvement which this PR didn't touch:

I think `safeUpdate` from `asyncRam#` should be like `safeUpdate` from `blockRam#`: https://github.com/clash-lang/clash-compiler/blob/9e7b9d3d335ae9d1115a6990c59c7cefdac70839/clash-prelude/src/Clash/Explicit/BlockRam.hs#L1072-L1080

In that it should _NOT_ return an undefined `Seq a`, but instead return a `Seq a` filled with undefineds. This way, if you never read an invalidly written value, you can continue as normally. That is, worst case, the out-of-bounds error is delayed; but it isn't masked.

DigitalBrains1:
Right, that was a concious decision not to touch that now. Because we discussed stuff on Slack, and there are two aspects to that:

1. Actually having an asserted write enable and a defined out-of-bounds write address is smelly code. @martijnbastiaan pointed out that it's not a bad idea to just error out the simulation here and force the user to rewrite their code to at the very least deassert the write enable in this case. We explicitly do it differently for an asserted write enable and an _undefined_ write address, as this might be the version of the hardware that has the desired least amount of reset lines to registers and can still lead to a perfectly valid design. So there we do fill the memory with undefineds.
2. There are a few places where we evaluate the whole `Signal` to `XException`, this is one of them. Note I'm not saying evaluating a sample in the `Signal` to `XException`. I feel evaluating the whole signal to `XException` is a mistake. It suggests to the user that they can somehow deal with this situation by not looking at undefined samples, since that is what `XException` is for. So I feel those few places where we evaluate the whole `Signal` to `XException` should actually be changed to evaluate to `ErrorCall` instead. But that is better dealt with in a different PR.


Note that we evaluate OOB `Vector` indexing to `ErrorCall` so the simulation comes to a halt whenever a user does that kind of out-of-bounds addressing:
https://github.com/clash-lang/clash-compiler/blob/74531b5993284d3a4858fa9842d8ce34e8bc3832/clash-prelude/src/Clash/Sized/Vector.hs#L1359-L1368
I feel that situation is pretty comparable. We do that for `(!!)` as well, and for `Bits (BitVector n)` (both read and update).

However, both Martijn and I are also okay with changing an OOB write to fill the whole memory with undefineds if that is still your preference, and that could be added to this PR.

christiaanb:
Having `XException` as a value for `Signal` is very undesirable. An exception for a single clock tick should not invalidate the complete future. One reason I want to avoid bailing out "early" is because a simulation might be long running. It's somewhat annoying to halt the simulation after an hour to show some out-of-bounds exception even if it will never affect the rest of the circuit ever.

In `Vec` it's somewhat unavoidable to have the entire value be XException (we should make a PR so that the ErrorCall becomes an XException) because we're polymorphic in the element type. But for `asyncRam#` we can avoid it. Basically, the only function that should create `XException/ErrorCall` values of type `Signal` is `fromList :: [a] -> Signal dom a`. So if there's any other locations where we know we're explicitly creating an `XException/ErrorCall` value of type `Signal` we should check whether we can ensure that the elements of the `Signal` are of value `XException/ErrorCall`.

PR #2007, opened 2021-11-19
First active edge at least one clock period from start

### LINKED ISSUE #2001: First active edge of clock in testbench not half of clock period
When generating a verilog testbench and a clock signal with a period of 20 ns the time to the first rising edge of the clock is set to 3 ns instead of the expected 10 ns. I have included the relevant part of the verilog code below. This results in problems with a post-synthesis simulation, as there is too little setup time before the first clock edge. Preferably this time to the first rising edge would be half of the clock period, or adjustable through a setting.

Verilog snippet:
// tbClockGen begin
  // pragma translate_off
  reg  clk_0;
  // 1 = 0.1ps
  localparam half_period = (200000 / 2);
  always begin
    // Delay of 1 mitigates race conditions (https://github.com/steveicarus/iverilog/issues/160)
    #1 clk_0 =  0 ;
    #30000 forever begin        //<------------- Where does this number come from?
      if (~ (~ c$result_rec)) begin
        $finish;
      end
      clk_0 = ~ clk_0;
      #half_period;
      clk_0 = ~ clk_0;
      #half_period;
    end
  end

christiaanb:
Fixes #2001

### DIFF
--- a/changelog/2021-11-19T17_41_41+01_00_fix2001
+++ b/changelog/2021-11-19T17_41_41+01_00_fix2001
@@ -1,0 +1,1 @@
+FIXED: Clash now generates clock generators that ensure that the amount of time between simulation start and the first active edge of the clock is equal to (/or longer than/) the period of the clock. The first active edges of the clocks do still occur simultaneously. [#2001](https://github.com/clash-lang/clash-compiler/issues/2001)
--- a/.ci/gitlab/common.yml
+++ b/.ci/gitlab/common.yml
@@ -3,7 +3,7 @@
   timeout: 2 hours
   stage: build
   variables:
-    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-$GHC_VERSION
+    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-$GHC_VERSION-3
     GIT_SUBMODULE_STRATEGY: recursive
     TERM: xterm-color
   retry:
--- a/clash-lib/prims/systemverilog/Clash_Explicit_Testbench.primitives
+++ b/clash-lib/prims/systemverilog/Clash_Explicit_Testbench.primitives
@@ -75,7 +75,7 @@ localparam ~GENSYM[half_period][0] = (~PERIOD[0]0 / 2);
 always begin
   ~RESULT = ~IF~ACTIVEEDGE[Rising][0]~THEN 0 ~ELSE 1 ~FI;
   `ifndef VERILATOR
-  #30000 forever begin
+  #~LONGESTPERIOD0 forever begin
     if (~ ~ARG[1]) begin
       $finish;
     end
@@ -92,7 +92,7 @@ end
 `ifdef VERILATOR
   `systemc_interface
   CData ~SYM[1](vluint32_t half_period, bool active_rising, bool result_rec) {
-    static vluint32_t init_wait = 30000;
+    static vluint32_t init_wait = ~LONGESTPERIOD0;
     static vluint32_t to_wait = 0;
     static CData clock = active_rising ? 0 : 1;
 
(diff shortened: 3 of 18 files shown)


changelog/2021-11-19T17_41_41+01_00_fix2001
```diff
@@ -0,0 +1 @@
+FIXED: Clash now generates clock generators that ensure there the amount of time between simulation start and the first active edge of the clock is equal /or longer/ than the period of the clock. The first active edges of the clock remain synchronized. [#2001](https://github.com/clash-lang/clash-compiler/issues/2001)
```
DigitalBrains1, line 1:
```suggestion
FIXED: Clash now generates clock generators that ensure that the amount of time between simulation start and the first active edge of the clock is equal /or longer/ than the period of the clock. The first active edges of the clock remain synchronized. [#2001](https://github.com/clash-lang/clash-compiler/issues/2001)
```

christiaanb:
For reviewers: with the change to the DDR test the Clash simulation and the HDL simulation now actually correspond. Before this change, it used to be that in the Clash simulation, outputverifier would trigger an assertion; where this assertion wasn't triggered in HDL simulation. Now neither the Clash simulation nor the HDL simulation trigger an assertion.

alex-mckenna:
@christiaanb I unchecked the copyright notice TODO, I can see non-updated notices in at least `clash-lib`

christiaanb:
All the files that have a copyright notice were originally published in 2021, and the copyright header mentions 2021; so it's all good. I rechecked the copyright notice TODO

leonschoorl (commented):
Why was there a difference in the `resetGen` prims behaviour between synchronous and asynchrous domains?
And why is that not longer needed?

christiaanb:
The difference has always bogus, we used to have:
```
, { "BlackBox" :
    { "name"      : "Clash.Signal.Internal.asyncResetGen"
    , "kind" : "Declaration"
    , "type" :
"asyncResetGen :: Reset domain 'Asynchronous"
    , "template" :
"-- pragma translate_off
~RESULT <= '1',
           '0' after 2000 ps;
-- pragma translate_on"
    }
  }
, { "BlackBox" :
    { "name"      : "Clash.Signal.Internal.syncResetGen"
    , "kind" : "Declaration"
    , "type" :
"syncResetGen :: ( domain ~ 'Dom n clkPeriod
                 , KnownNat clkPeriod )
              => Reset domain 'Synchronous"
    , "template" :
"-- pragma translate_off
~RESULT <= '1',
           '0' after (2999 ps + ~LIT[1]0 ps);
-- pragma translate_on"
    }
  }
```
then in #428 we changed to:
```
, { "BlackBox" :
    { "name"      : "Clash.Signal.Internal.asyncResetGen"
    , "workInfo"  : "Always"
    , "kind" : "Declaration"
    , "type" :
"asyncResetGen :: Reset domain 'Asynchronous"
    , "template" :
"-- asyncResetGen begin
-- pragma translate_off
~RESULT <= '1',
           '0' after 3001 ps;
-- pragma translate_on
-- asyncResetGen end"
    }
  }
, { "BlackBox" :
    { "name"      : "Clash.Signal.Internal.syncResetGen"
    , "workInfo"  : "Always"
    , "kind" : "Declaration"
    , "type" :
"syncResetGen :: ( domain ~ 'Dom n clkPeriod
                 , KnownNat clkPeriod )
              => Reset domain 'Synchronous"
    , "template" :
"-- syncResetGen begin
-- pragma translate_off
~RESULT <= '1',
           '0' after (2999 ps + ~LIT[1] ps);
-- pragma translate_on
-- syncResetGen end"
    }
  }
```
so that the de-assertion always happened after 3000ps (which is time all rising edges start (before this PR)).

Then in #527 those definitions got merged:
```
, { "BlackBox" :
    { "name"      : "Clash.Signal.Internal.resetGen"
    , "workInfo"  : "Always"
    , "kind" : "Declaration"
    , "type" :
"resetGen :: KnownDomain dom      => Reset dom"
    , "template" :
"-- resetGen begin
-- pragma translate_off
~RESULT <= ~IF ~ISACTIVEHIGH[0] ~THEN '1' ~ELSE '0' ~FI,
           ~IF ~ISACTIVEHIGH[0] ~THEN '0' ~ELSE '1' ~FI after ~IF~ISSYNC[0]~THEN(2999 ps + ~PERIOD[0] ps)~ELSE3001 ps~FI;
-- pragma translate_on
-- resetGen end"
    }
  }
```
By which time we should've already dropped the difference. Because both `3001ps` and `2999ps + PERIOD[0]ps` deassert before the second active edge; i.e. we cannot observe their difference.

leonschoorl (approved):
LGTM

PR #2009, opened 2021-11-22
JSON compliant blackboxes

oliverbunting:
This PR:

- Adds support for formatting multi-line strings in black-box files as arrays of strings.
  Existing black-boxes using multi-line strings remain supported.
- Re-formats _all_ primitives in clash with multi-line strings into this form
- Adds a setting to the project vscode settings, associating `.primitives` files as `.jsonc`

Backwards compatibility is preserved as I'm aware there are other repos out that would be broken by a breaking change in schema. 

The motivation for this is "Developer UX". By making these files valid JSON, syntax highlighting / editor support / auto format become available using any tool which operates on JSON. It also significantly reduces the signal to noise ratio in the editors error bar

In changing the formatting of the black-boxes, `template`, `comments`, and `type` fields have been modified. However, I could find no use of `type` or `comments`, so I guess they're advisory only?

## Still TODO:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -27,5 +27,8 @@
     80,
     90
   ],
-  "editor.tabSize": 2
+  "editor.tabSize": 2,
+  "files.associations": {
+    "*.primitives": "jsonc"
+  }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -29,6 +29,6 @@
   ],
   "editor.tabSize": 2,
   "files.associations": {
-    "*.primitives": "jsonc"
+    "*.primitives": "json"
   }
 }
--- a/clash-lib/prims/common/Clash_Explicit_Signal.primitives
+++ b/clash-lib/prims/common/Clash_Explicit_Signal.primitives
@@ -3,11 +3,12 @@
     , "workInfo"  : "Never"
     , "kind"      : "Expression"
     , "type" :
-"veryUnsafeSynchronizer
-  :: Int                         -- ARG[0]
-  -> Int                         -- ARG[1]
-  -> Signal dom1 a               -- ARG[2]
-  -> Signal dom2 a"
+      [  "veryUnsafeSynchronizer"
+      , "  :: Int                         -- ARG[0]"
+      , "  -> Int                         -- ARG[1]"
+      , "  -> Signal dom1 a               -- ARG[2]"
+      , "  -> Signal dom2 a"
+      ]
     , "template"  : "~ARG[2]"
     }
   }
--- a/clash-lib/prims/common/Clash_Explicit_Testbench.primitives
+++ b/clash-lib/prims/common/Clash_Explicit_Testbench.primitives
@@ -3,14 +3,15 @@
     , "workInfo"  : "Never"
     , "kind"      : "Expression"
     , "type" :
-"unsafeSimSynchronizer
-  :: forall dom1 dom2 a
-   . ( KnownDomain dom1         -- ARG[0]
-     , KnownDomain dom2 )       -- ARG[1]
-  => Clock dom1                 -- ARG[2]
-  -> Clock dom2                 -- ARG[3]
-  -> Signal dom1 a              -- ARG[4]
-  -> Signal dom2 a"
+      [ "unsafeSimSynchronizer"
+      , "  :: forall dom1 dom2 a"
+      , "   . ( KnownDomain dom1         -- ARG[0]"
+      , "     , KnownDomain dom2 )       -- ARG[1]"
+      , "  => Clock dom1                 -- ARG[2]"
+      , "  -> Clock dom2                 -- ARG[3]"
+      , "  -> Signal dom1 a              -- ARG[4]"
+      , "  -> Signal dom2 a"
+      ]
     , "template"  : "~ARG[4]"
     , "warning"   : "Clash.Explicit.Testbench.unsafeSimSynchronizer is not safely synthesizable!"
     }
--- a/clash-lib/prims/common/Clash_Signal_BiSignal.primitives
+++ b/clash-lib/prims/common/Clash_Signal_BiSignal.primitives
@@ -3,13 +3,14 @@
     , "workInfo" : "Never"
     , "kind" : "Declaration"
     , "type" :
-"veryUnsafeToBiSignalIn
-  :: ( HasCallStack                -- ARG[0]
-     , KnownNat n                  -- ARG[1]
-     , Given (SBiSignalDefault ds) -- ARG[2]
-     )
-  => BiSignalOut ds d n            -- ARG[3]
-  -> BiSignalIn ds d n"
+      [ "veryUnsafeToBiSignalIn"
+      , "  :: ( HasCallStack                -- ARG[0]"
+      , "     , KnownNat n                  -- ARG[1]"
+      , "     , Given (SBiSignalDefault ds) -- ARG[2]"
+      , "     )"
+      , "  => BiSignalOut ds d n            -- ARG[3]"
+      , "  -> BiSignalIn ds d n"
+      ]
     , "template" : "~DEVNULL[~ARG[3]]"
     }
   }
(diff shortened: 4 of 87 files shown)


oliverbunting:
@martijnbastiaan I'm unsure of the process here, but this PR is ready for review / comments

christiaanb:
@kloonbot run_ci 5ce50b1d09950aa843dd413e319943900cd98d24

.vscode/settings.json
```diff
@@ -27,5 +27,8 @@
     80,
     90
   ],
-  "editor.tabSize": 2
+  "editor.tabSize": 2,
+  "files.associations": {
+    "*.primitives": "jsonc"
```
martijnbastiaan, line 32: Should this just be `json`? I don't believe Aeson supports `jsonc` (although that would be nice).

martijnbastiaan (approved):
> However, I could find no use of `type` or `comments`, so I guess they're advisory only?

That's correct.

----------------------

I like this change. I've always been irritated by either having to completely disable any syntax highlighting, or having to suffer through the stupid error messages and accidental auto-formats of my IDE. I particularly dislike the [patch I introduced a while ago](https://github.com/clash-lang/clash-compiler/commit/a1ad57cc46a4875f7cc33ff8a90102a0df7708e6#diff-2224cd8024706aef61b5f069f271f5c8d3f8a96c328f0ec4de919958f37cefb7) to preprocess our "JSON" files to support multi-line strings. I therefore think this is a step in the right direction.

I know @christiaanb is a bit more conservative than me when it comes to sweeping changes  like these, as they make code archeology more difficult. As for me, I like big diffs and I cannot lie.

.vscode/settings.json
oliverbunting: @martijnbastiaan yes, it should. This is a result of habit. I'll change

Also :rofl: 
> As for me, I like big diffs and I cannot lie.

martijnbastiaan:
@kloonbot run_ci 0806e4a

leonschoorl:
I get what you're trying to do with this PR.
But I personally don't see this as an improvement.

My editor doesn't bother me with errors about these files, I wasn't aware of any noise there.
And I feel all the `, "` and their (mis)alignment adds extra visual noise in any editor.
Editing files in the JSON compliant multiline syntax is more effort for people without editors that help them with JSON autoformatting.

For me the JSON part isn't actually the hard part of the `.primitives` files, it's the HDL.
And we the current syntax I can use you a simpler editor, one that doesn't do full parsing, but more regex matching style syntax highlighting, and get pretty decent syntax highlight for the HDL with the JSON:
prims-syntax-highlighting
Which is more valuable to me then syntax highlighting of the JSON.

I'm not against clash supporting these kinds of multiline JSON strings. If you'd like to use it for your own primitives.
But I feel changing the clash bundled primitives to this new format is not an improvement.

I'm also not super happy with this illegal JSON we're using currently.
But see moving to an format that natively support "clean" multiline strings as the preferred solution.
Maybe YAML

martijnbastiaan:
I can see your point @leonschoorl, but I can't really agree. It seems like any editor worth its salt _should_ throw a fight upon seeing our primitive files: it's neither HDL nor JSON. The fact that we'd let editors that only do simple regex parsing dictate our file formats doesn't sit right with me. 

In any case, I wonder why we couldn't just use JSON for meta information, while using plain `.vhdl`/`.verilog`/`.common`/`.commonverilog` for HDL implementations. I'd imagine something like:

- clash-lib/prims/Clash/Class/Exp
  - expIndex#.prim-meta
  - expIndex#.prim-vhdl
  - expIndex#.prim-verilog

where `expIndex#.meta` would look something like:

```json
{
  "common":{
    "kind": "Expression",
    "warning": "....."
  }
}
```

whatever we do, I think we should offer tooling to our users to automate the transition.

alex-mckenna:
I'm not sure if I mean this as a serious suggestion or not, but moving to YAML means we could move that meta-information into a so-called frontmatter section, i.e.

```yaml
---
name: Some.Qualified.name
kind: Declaration
---
assign ~GENSYM[is_this_actually_a_good_idea][0] = ~IF~ISGOODIDEA~THENtrue~ELSEfalse~FI;
```

christiaanb:
But that'd mean a yaml file for every primitive wouldn't it?

alex-mckenna:
> I'm not sure if I mean this as a serious suggestion or not

Yeah, there's a lot to not like, it's why I left the disclaimer

alex-mckenna:
Given the change should be backwards compatible, I'm inclined to say allow it but don't change the primitives in `clash-lib`. I think I agree with @leonschoorl that changing all our primitives will make it worse for us editing them later

oliverbunting:
> But that'd mean a yaml file for every primitive wouldn't it?

It could support lists of primitives, in the same way you currently do in json

```yaml
- Blackbox:
    name: Thing
    kind: "Declaration"
    template: |
       foo
       bar
- Blackbox:         
    name: Thing2
    kind: "Declaration"
    template: |
       baz
       bork
```

This conversion could be done entirely automatically from the json in this PR.

martijnbastiaan:
I'm not particularly fond of YAML. It strikes me as being designed by a PHP cabal:

```
foo: 2.3
11: 2.3.5
011: yes
088: yess
```

would roughly be parsed as:

```
"foo": 2.3
11: "2.3.5"
9: true
"088": "yess"
```

A lot of these "mistakes" I've made when configuring GitLab's CI system and I remember them being rather frustrating.

Chances are this is not a problem for Haskell's implementation as it usually knows what it wants to decode - and would therefore never mistake `yes` for a boolean where a string is expected.

martijnbastiaan:
```
martijn@thinklap:~$ cat foo.yaml 
a: true
martijn@thinklap:~$ cat Test.hs
{-# LANGUAGE DeriveAnyClass #-}
{-# LANGUAGE DeriveGeneric #-}
{-# LANGUAGE TypeApplications #-}

import Data.Yaml
import GHC.Generics (Generic)

data Test = Test { a :: String } 
  deriving (Generic, FromJSON, ToJSON, Show)

main :: IO ()
main = print =<< decodeFileThrow @_ @Test "foo.yaml"
martijn@thinklap:~$ stack exec --package yaml ghc Test.hs
martijn@thinklap:~$ ./Test 
Test: AesonException "Error in $.a: expected String, but encountered Boolean"
```

Meh :(

alex-mckenna:
You meant `"true"`, maybe this is just YAML being type safe :wink:

(I do agree with @martijnbastiaan though)

martijnbastiaan:
As much as I dislike YAML's quirks, I dislike the status-quo much more. Also I _think_ any time a quirk is used it should result in a parse error - nothing should go unnoticed. As long as the conversion is automated (I'm willing to write a tool for this) I'm fine with it.

Opinions?

martijnbastiaan:
@oliverbunting I've added a `json-to-yaml` utility here: https://github.com/clash-lang/clash-compiler/commit/0016262239c0366e7f4d431251103024394cef5b. Example usage:

```
cabal run json-to-yaml -fjson-to-yaml -- --delete clash-lib/**/*.primitives
```

I'm fine with moving to YAML this way. Upgrading Clash should be a matter of just running this utility. I propose we keep supporting both for a few releases.

alex-mckenna:
> Given the change should be backwards compatible, I'm inclined to say allow it but don't change the primitives in clash-lib.

My feeling is still summed up by this comment from _past Alex_. Supporting standard-compliant JSON and YAML are good additions (nice YAML converter btw @martijnbastiaan), but we should retain backwards compatibility for some reasonable amount of time. I do agree though that if we use some common format, we should respect its standard.

Perhaps a nice middle ground is to continue to interpret `.primitives` the same way, but also allow `.json` and `.yaml` for standards-compliant JSON and YAML primtiives files? This may trip up users upgrading old projects from when `.primitives` were called `.json` (but was not standards-compliant).

If we decide we want to change the primitives in `clash-lib`, that should be done in a separate PR. It feels like a distinct - and more contentious - point of discussion to me since how write our own `.primitives` files are written doesn't affect users.

alex-mckenna:
But to be clear: if the `.primitives` files are reverted then I think this is merge-worthy

martijnbastiaan:
To keep track of updates, I've added the json-to-yaml utility to a separate PR: https://github.com/clash-lang/clash-compiler/pull/2037.

oliverbunting:
@martijnbastiaan Nice yaml widget.

I don't see the need for both `*.primitives.yaml` and `*.primitives.json`. I propose this PR is closed, and support for existing `.primitives` is extended with only `.primitives.yaml` in a new PR.

martijnbastiaan:
That works for me too. Thanks!

PR #2010, opened 2021-11-22
Use pointer equality to shortcut `eqTerm`

martijnbastiaan:
This speeds up the testsuite by ~10%. On my laptop using `-p VHDL.clash`,
I see 758.23s before the patch, and 682.40s after. When synthesizing our
I2C testcase, I see the following normalization times:

|      | Before patch | After patch | B/A  |
|------|--------------|-------------|------|
| Bit  | 0.428        | 0.311       | 1.38 |
| Byte | 0.223        | 0.166       | 1.34 |
| Top  | 0.017        | 0.012       | 1.42 |

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Util/Eq.hs
+++ b/clash-lib/src/Clash/Util/Eq.hs
@@ -1,0 +1,46 @@
+{-|
+  Copyright  :  (C) 2021     , QBayLogic B.V.
+  License    :  BSD2 (see the file LICENSE)
+  Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+  Utilities related to the 'Eq' type class.
+-}
+{-# LANGUAGE MagicHash #-}
+
+module Clash.Util.Eq
+  ( fastEq
+  , fastEqBy
+  ) where
+
+import GHC.Exts (isTrue#, reallyUnsafePtrEquality#)
+
+-- | Compare two values using pointer equality. If that fails, use 'Eq' to
+-- determine equality. Note that this function will only shortcut for values
+-- that are the same, but will always use 'Eq' for values that differ.
+--
+-- Values are evaluated to WHNF before comparison. This function can therefore
+-- not be used if any of its arguments is expected to be bottom.
+fastEq :: Eq a => a -> a -> Bool
+fastEq = fastEqBy (==)
+
+-- | Compare two values using pointer equality. If that fails, use given function
+-- to determine equality. Note that this function will only shortcut for values
+-- that are the same, but will always use the given function for values that
+-- differ.
+--
+-- Values are evaluated to WHNF before comparison. This function can therefore
+-- not be used if any of its arguments is expected to be bottom.
+fastEqBy :: (a -> a -> Bool) -> a -> a -> Bool
+fastEqBy f a1 a2
+  | a1 `pointerEq` a2 = True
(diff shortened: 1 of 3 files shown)


christiaanb (approved):
LGTM

clash-lib/src/Clash/Util/Eq.hs
```diff
@@ -0,0 +1,39 @@
...
+
+module Clash.Util.Eq
+  ( fastEq
+  , fastEqBy
+  ) where
+
+import GHC.Exts (Int(I#), reallyUnsafePtrEquality#)
+
+-- | Compare two values using pointer equality. If that fails, use 'Eq' to
+-- determinte equality. Note that this function will only shortcut for values
```
alex-mckenna, line 11: `%s/determinte/determine/`

martijnbastiaan: tnx!

clash-lib/src/Clash/Util/Eq.hs
```diff
@@ -0,0 +1,39 @@
...
+
+{-# NOINLINE pointerEq #-}
+-- | Compares two values by comparing their positions on the heap. This function
+-- will return 'True' if the values are the same object, 'False' otherwise. Note
+-- that 'False' does *not* mean that the values are *not* the same. Values are
+-- evaluated to WHNF before comparison.
+--
+-- Note: copied from @unordered-containers@.
+pointerEq :: a -> a -> Bool
+pointerEq !x !y = I# (reallyUnsafePtrEquality# x y) == 1
```
alex-mckenna, line 39: I would be moderately surprised if these bangs were necessary, especially as `ptrEq` in `unordered-containers` doesn't force the arguments. I'm not in favour of adding bangs where they're not needed because they're presence implies someone discovered they _are_ needed. Do they actually make a difference?

alex-mckenna: I also wonder why you're not just using `isTrue#`, it even returns a `Bool`!

martijnbastiaan: `unordered-containers` is careful to insert bang patterns close to the call site, but didn't want to make it that fragile. Also, it looks like I copied the one from `atomic-primops`, not `unordered-containers`.

As for the necessity, I'm _fairly_ sure they are necessary (either at call site, or in the function), but let's measure!

martijnbastiaan: > I also wonder why you're not just using isTrue#, it even returns a Bool!

What is this magic!

alex-mckenna: `Bool` is a sneaky type, it lives in `ghc-prim`

martijnbastiaan: I'm getting a fairly consistent 0.351s / 0.184s / 0.014s after removing the bang patterns, so it looks like that resolving to WHNF does give you the same pointers slightly more often. Even though the performance benefits are marginal, I'd like to keep it in (until it gives us problems).

martijnbastiaan: I'll use `isTrue#` though!

leonschoorl (approved):
Cool stuff!

PR #2011, opened 2021-11-23
Add flag to generate verilator shims

alex-mckenna:
Clash can now be called with `-fclash-verilator` which generates the required C++ shim for simulating a top entity.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver.hs
+++ b/clash-lib/src/Clash/Driver.hs
@@ -56,6 +56,11 @@ import           Data.Text.Lazy                   (Text)
 import qualified Data.Text.Lazy                   as Text
 import           Data.Text.Lazy.Encoding          as Text
 import qualified Data.Text.Lazy.IO                as Text
+#if MIN_VERSION_prettyprinter(1,7,0)
+import           Prettyprinter (pretty)
+#else
+import           Data.Text.Prettyprint.Doc (pretty)
+#endif
 import           Data.Text.Prettyprint.Doc.Extra
   (Doc, LayoutOptions (..), PageWidth (..) , layoutPretty, renderLazy)
 import qualified Data.Time.Clock                  as Clock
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Driver.hs
```diff
@@ -740,6 +755,50 @@ createHDL backend modName seen components domainConfs top topName = flip evalSta
...
+  pure ((file, digest) : filesAndDigests)
+
+-- | Create a shim for using verilator, which loads the entity and steps
+-- through simulation until finished.
+--
+pprVerilatorShim :: Id.Identifier -> Doc
+pprVerilatorShim (Id.toText -> topNm) =
+  -- Extra newlines are aggressively inserted so the quasiquoter doesn't wrap
+  -- the outlines lines in the file. It doesn't matter for code inside main,
+  -- but is fatal for the #include directives.
```
martijnbastiaan, line 775: Perhaps use:

```
import           Data.String.Interpolate      (i)
import           Data.String.Interpolate.Util (unindent)

unindent [i|
```

This quasiquoter was really made for pretty error messages.

martijnbastiaan (commented):
Wouldn't know whether the C++ is correct, but otherwise LGTM. In time we should move this out of the clash compiler, into something that only uses the Manifest files.

clash-lib/src/Clash/Driver.hs
alex-mckenna: That didn't work, I'm not going to put any more time into trying to layout this file

martijnbastiaan (approved):

alex-mckenna:
I don't like this PR anymore, closing in favour of a different (to be created) one

gergoerdi:
Can someone explain how this is different from Clashilator? Wouldn't it be easier to add missing features to it?

alex-mckenna:
So the intent of the current work with `verilator` is being able to run the regression testsuite through `verilator` for (System)Verilog. Since these are largely testbenches with no need for fine-control over the simulation, or probing / driving signals etc. `clashilator` is somewhat of an overkill compared to the tiny C++ shim being used. For anything more complex than _I want to simulate this until I hit `$finish`_, `clashilator` is still a much better choice because you can interact with the simulation directly from Haskell.

FWIW, the changes to Clash being done for this benefit `clashilator` equally, i.e. changes to black boxes to make them verilate correctly or (hopefully) later changes to Clash to remove warnings from `verilator` about things which slow down simulation.

PR #2013, opened 2021-11-28
Use `doctest-parallel` for Prelude's doctests

### LINKED ISSUE #1622: [clash-prelude] doctests:  executable not found
For the Stackage Nightly build:

```
Test suite failure for package clash-prelude-1.2.5
    doctests:  executable not found
```

The rest of the tests all seem to pass.

martijnbastiaan:
This PR is mostly driven by yet another erratic doctest failure on [one of my branches](https://gitlab.com/clash-lang/clash-compiler/-/jobs/1826708505), caused by doctest not properly isolating tests between modules.

---------------

From the project's README, this has a couple of advantages:

 * It runs tests in parallel
 * It runs tests against compiled code, instead of reinterpreting
   your whole project
 * It isolates examples in modules, ensuring your tests don't
   accidentally rely on each other
 * It parses cabal files to discover modules, no need for custom setup
   anymore!

The project lists two disadvantges, of which only one is relevant for
clash-compiler:

 * Use of conditionals in a cabal file as well as CPP flags will be
   ignored

**This patch speeds up doctest from 183 seconds to 9 seconds on my
desktop, running a (12 core) 3900X.** Note that this isn't a 1:1
comparison as this PR also moves some test to a module, avoiding some
pointless reevaluation.

Fixes #1622.

--------------

For those interested, this is how it scales on my machine:

| Threads | Time (s) |
|---------|----------|
| 1       | 43.362   |
| 2       | 22.973   |
| 4       | 13.623   |
| 8       | 10.143   |
| 12      | 9.672    |
| 12+     | ~9.6     |

I.e., after 8 cores there's no benefits anymore. I'm not sure what's causing this, as all cores _seem_ busy, but it does correspond to previous experience with multithreading on Haskell/GHC.

## Still TODO:

  - [x] Publish `doctest-parallel` on Hackage
  - [X] CI `nix-build`: should be fixed by publishing package on Hackage
  - [x] ~~CI 8.6: a dependency of `doctest-parallel`, `cabal-install-parsers`, forces a `Cabal` >= 3.0. This fails on 8.6 if you force it to keep its "default" Cabal version. Perhaps `doctest-parallel` can use a parser from `Cabal` directly.~~
  - [x] ~~CI 8.8: no clue why this shows up for this PR~~
  - [X] ~~CI 8.10 / 9.0: `error while loading shared libraries: libHSaeson-pretty-0.8.9-79ec45a030a798b6d6f88a443ae792fa5b6e3d11e89793ec28d088768c7775b4-ghc9.0.1.so: cannot open shared object file: No such file or directory` :shrug: . It is in `dist.tar.zstd`, and the `.conf` file refers to the correct dirs. I have no clue why this happens.~~
  - [x] ~~Check copyright notices are up to date in edited files.~~ Most are _tiny_ changes so I don't really feel like I should update the copyright notices.


leonschoorl:
Even without the parallelization, going from 183 to 43 is already great

martijnbastiaan:
I've dropped `cabal-install-parsers` in `doctest-parallel-0.2` which seem to have fixes all the weird CI issues. I fail to see how it could have caused the error though, but oh well.

clash-prelude/src/Clash/Sized/Internal/Vector.hs
```diff
@@ -0,0 +1,64 @@
+{-|
+Copyright  :  (C) 2021, QBayLogic B.V.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+This module exists purely to speed up doctests of "Clash.Sized.Vector". It should
+never be imported by user code.
+-}
```
alex-mckenna, line 8: I remember looking at this file and the other new `*.Internal` modules for `doctests` and noticing that the definitions used to be in `$setup` but could actually be moved to the doctest of a single test. I think this would make the tests for the modules quicker without needing a new module like this which I think is a _very_ bad idea.

clash-prelude/tests/doctests.hs
```diff
@@ -1,14 +1,7 @@
-{-# LANGUAGE CPP #-}
-
 module Main where
 
-import Build_doctests (flags, pkgs, module_sources)
-import Data.Foldable (traverse_)
-import Test.DocTest (doctest)
+import Test.DocTest (mainFromCabal)
+import System.Environment (getArgs)
```
alex-mckenna, line 4: Alphabetical please

nix/sources.json
```diff
@@ -1,4 +1,16 @@
 {
+    "doctest-parallel": {
+        "branch": "main",
+        "description": "Test interactive Haskell examples",
+        "homepage": null,
```
alex-mckenna, line 5: I think you should change this from `null` to `"https://github.com/martijnbastiaan/doctest-parallel"`

clash-prelude/src/Clash/XException/Internal.hs
```diff
@@ -53,6 +53,9 @@ import           GHC.Show            (appPrec)
 import           GHC.Stack           (HasCallStack)
 import           System.IO.Unsafe    (unsafeDupablePerformIO)
 
+-- $setup
+-- >>> import Clash.Prelude
```
alex-mckenna, line 57: Should this module have `:m -Prelude`?

clash-prelude/clash-prelude.cabal
```diff
@@ -365,14 +360,13 @@ test-suite doctests
...
-  x-doctest-options: -fobject-code
-
   if !flag(doctests)
     buildable: False
   else
     build-depends:
       base,
-      doctest >= 0.16.1 && < 0.19,
+      filepath,
+      doctest-parallel >= 0.2 && < 0.3,
```
alex-mckenna, line 369: Alphabetical please

alex-mckenna (changes requested):
Really nice PR @martijnbastiaan, just needs the smallest bit of tidying up IMO. The only thing I really have a problem with is the new `*.Internal` modules which are for `doctest` only and not ever intended to be imported. Having these end up in `exposed-modules` is a big mistake IMO because it adds "documentation noise" / confusion to users (especially new users).

I think they all fall under _the code was only used for a single test but incorrectly placed in `$setup`_, in which case moving them back to their original modules but not in `$setup` should hopefully be enough to avoid the slowdown they were causing.

alex-mckenna:
Also a question: Given we currently run `doctests` on the GitLab shared runners, what kind of speedup can we expect there? I'm not sure how many cores those machines have but I'd (naïvely) assume 2 or 4...?

martijnbastiaan:
Thanks for the comments! I'll look into it hopefully tonight.

> Also a question: Given we currently run doctests on the GitLab shared runners, what kind of speedup can we expect there?

The VHDL/Verilog test cases are slower than doctest on shared runners now, so I don't expect a speedup. But, given that the number of shared runners is virtually unlimited, this should alleviate our local runners during rush hour. Still, the real profit is when running locally on a beefy machine - i.e., the test/fix loop is _much_ tighter now.

clash-prelude/tests/doctests.hs
martijnbastiaan: Are there any linting tools that:

* Check this
* Work during compile time (i.e., generate a warning WITHOUT running something after the fact)

That you are aware of?

christiaanb (commented):
It might be faster, but then it seems to run fewer examples:

`HEAD of master`
```
$ cabal run clash-prelude:doctests
Examples: 947  Tried: 947  Errors: 0  Failures: 0
```
This PR:
```
$ cabal run clash-prelude:doctests
Examples: 626  Tried: 626  Errors: 0  Unexpected output: 0
```
So there's several possibilities:
1. HEAD of master ran tests double
2. The difference is due to the removal of all the `>>> :m X.Y.Z` and other `>>> ...` statements to setup the environment for testing
3. We accidentally test less in this PR

martijnbastiaan:
Yeah, it doesn't count `$setup` as examples anymore. The parser/test collection logic is exactly the same between `doctest` and `doctest-paralell`, so I'd say there's only an incredibly low chance that it actually tests less.

clash-prelude/tests/doctests.hs
alex-mckenna: Not that I'm aware of unfortunately, but I don't think it would be too hard to hack together a check like we have the check for EOL whitespace. It would just have to

  1. `grep` for any line starting with `import`
  2. remove any word `qualified` that appears
  3. check the module names are sorted

So if either of us finds the time we can hack a check together in `bash` and add it to the CI checks. I don't think I'll have time this week, so if you don't can you make an `enhancement` issue for it?

martijnbastiaan: Yeah, the reason I asked is that I really "hate" after-the-fact checks. The EOL checks are alright, because it's just an editor setting. Time to test [`styleless-haskell`](https://gitlab.com/martijnbastiaan/styleless-haskell/-/blob/master/src/Plugin/StylelessHaskell/Checks/Alphabetization.hs)? :wink:

alex-mckenna: `haskell-language-server` has `retrie` integration, perhaps we could configure that to do the rewrite in-editor?

martijnbastiaan: yes pls

nix/sources.json
martijnbastiaan: Ah I just used `niv add`. So my guess is that this is some cabal field I missed in `doctest-parallel`

christiaanb:
Instead of trying to write everything like a review suggestion, I just made a commit.
The commit removes the `Clash.Sized.Internal.Vector` module and moves the definitions used by the doctests back to the relevant tests.

The haddock for e.g. `vfold`  now look like:
image

where the ` doctests setup`  expands to:
image

nix/sources.json
martijnbastiaan: Which is to say, I'm going to ignore it until a new `doctest-parallel` release :smiley_cat:

clash-prelude/src/Clash/XException/Internal.hs
martijnbastiaan: It doesn't seem to ambiguous definitions, so not necessarily :smile:

PR #2019, opened 2021-12-08
Support `verilator` in `clash-testsuite`

alex-mckenna:
This PR adds support for using `verilator` in (System)Verilog tests in the testsuite. This largely consists of changes to black boxes which `verilator` does not support to include C++ definitions which can be used for simulation instead.

## Still TODO:


tests/src/Test/Tasty/Verilator.hs
```diff
@@ -0,0 +1,119 @@
...
+          copyFile memFile (dir </> "memory.list")
+
+        let topExe = dir </> "obj_dir" </> ("V" <> top)
+
+        case expectFail of
+          Nothing -> run optionSet (verilated dir topExe) progressCallback
+          Just exit -> run optionSet (failingVerilated dir topExe exit) progressCallback
+
+    | otherwise =
+        pure (testPassed "Ignoring test: Use --verilator to run")
```
martijnbastiaan, line 109: Same comment as `--no-ghdl`

tests/src/Test/Tasty/Clash.hs
```diff
@@ -399,6 +425,63 @@ runTest1 modName opts@TestOptions{..} path target =
...
+
+  -- HACK: We want to run verilator and simulator tests independently if they
+  -- are both going to be run, otherwise failures from whichever comes first
+  -- will mean the second is skipped. In lieu of a better way to sequence tests
+  -- we can sequence tests for multiple simulators, then drop the first test
+  -- tree for all but the first simulator (as it will be copies of clash gen).
+  --
+  -- TODO: Since tasty doesn't provide one, we should really provide a better
+  -- set of combinators for describing test dependencies. That way we can have
+  -- some more principled way of having a test structure like
```
martijnbastiaan, line 451: It's definitely Tasty's weakest point. Basically we already work our way around it by passing down names so we can construct the absolute name of the test at test location. It would be some much more intuitive if you could just do something like: `TestGroupSeq [ ... ]`. But alas..

clash-lib/src/Clash/Driver.hs
```diff
@@ -383,7 +388,14 @@ generateHDL reprs domainConfs bindingsMap hdlState primMap tcm tupTcm typeTrans
...
 
-      writeManifest manPath manifest0{fileNames=fileNames1}
+      -- If we are generating (System)Verilog, we could potentially verilate
+      -- the results. Clash can output a C++ shim for doing this automatically.
+      fileNames2 <-
+        case hdl of
+          VHDL -> pure fileNames1
+          _ -> writeVerilatorShim hdlDir topNm fileNames1
+
+      writeManifest manPath manifest0{fileNames=fileNames2}
```
martijnbastiaan, line 398: I'd love for this to become a separate executable someday. It feel icky to have this all dumped in `Driver`.

martijnbastiaan (changes requested):
LGTM overall

.ci/bindist/linux/debian/focal/buildinfo.json
```diff
@@ -1,5 +1,5 @@
 {
-  "docker": "docker.pkg.github.com/clash-lang/clash-compiler/bindist-debian-focal:2021-05-14",
+  "docker": "docker.pkg.github.com/clash-lang/clash-compiler/bindist-debian-focal:2021-12-08",
```
martijnbastiaan, line 2: You probably need to redo this after https://github.com/clash-lang/clash-compiler/pull/2043..

PR #2026, opened 2021-12-28
Bump dependency version bounds

martijnbastiaan:
Removed version bounds for internal packages, as their dependency bounds
are often imposed by clash-{prelude,lib,ghc} anyway, and they cause
mismatches between what a user compiles against versus us / CI.


martijnbastiaan:
@christiaanb Agree with my reasoning on deleting the version bounds on private pkgs?

christiaanb (approved):

Issue #2020, opened 2021-12-08
warnNonSynthesizable triggers error even when code is unused


acairncross:
I'm using `warnNonSynthesizable` expecting it to error only when actually using (synthesizing) the code labelled as non-synthesizable, e.g. if it's possible/easy to statically determine that the code shouldn't be synthesized then it won't be, but instead it triggers an error. Is this expected behaviour, and if it is, is there another way to annotate non-synthesizable code that would work for this use case?

```haskell
module Test where

import Clash.Prelude
import Clash.Annotations.Primitive

unsupported :: String -> a
unsupported msg =
  error $ msg <> ": unsupported"
{-# ANN unsupported (warnNonSynthesizable "unsupported") #-}

topEntity
  | True = 42
  | otherwise = unsupported "hello"
```

With Clash 1.4:
```
GHC: Parsing and optimising modules took: 0.367s
GHC: Loading external modules from interface files took: 0.000s
GHC: Parsing annotations took: 0.000s

<no location info>: error:
    Clash error call:
    No BlackBox definition for 'Test.unsupported' even though this value was annotated with 'HasBlackBox'.
    CallStack (from HasCallStack):
      error, called at src/Clash/Primitives/Util.hs:142:11 in clash-lib-1.4.6-EBw2mxbesS05MW7zTaSOaM:Clash.Primitives.Util
```

With Clash 1.2:
```
GHC: Parsing and optimising modules took: 0.370s
GHC: Loading external modules from interface files took: 0.000s
GHC: Parsing annotations took: 0.000s

<no location info>: error:
    Clash error call:
    No BlackBox definition for 'Test.unsupported' even though this value was annotated with 'WarnNonSynthesizable', implying it has a BlackBox.
    CallStack (from HasCallStack):
      error, called at src/Clash/Primitives/Util.hs:137:11 in clash-lib-1.2.5-H3gfsjD1Tn39TYN6Isgnmk:Clash.Primitives.Util
```

alex-mckenna:
This error is expected: when you use `warnNonSynthesizable` it implicitly enables `hasBlackBox` which makes Clash throw an error for any black boxes without definitions provided. You can provide this definition by creating a `.primitives` file like those found [here](https://github.com/clash-lang/clash-compiler/tree/master/clash-lib/prims), or by using something like [`InlinePrimitive`](https://hackage.haskell.org/package/clash-prelude-1.4.6/docs/Clash-Annotations-Primitive.html#t:Primitive).

[...]

PR #2027, opened 2021-12-28
Refactor `LoadInterfaceFiles` to prevent O(n^2) binder loading

martijnbastiaan:
In particular, this line:

    let declM = filter ((== nameFun) . Name.nameOccName . IfaceSyn.ifName) decls

caused linear lookups for every binder used in a design. This refactor
makes Clash load a module once (through the GHC API) and use an index
for subsequent lookups. Additionally, annotations are now loaded once
per module instead of once per binder.

This speeds up the testsuite by ~10% on my machine.

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
@@ -6,29 +6,41 @@
 -}
 
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TupleSections #-}
 
 module Clash.GHC.LoadInterfaceFiles
   ( loadExternalExprs
   , loadExternalBinders
   , getUnresolvedPrimitives
   , LoadedBinders(..)
-  , mergeLoadedBinders
-  , emptyLb
   )
 where
 
 -- External Modules
 import           Control.Monad.IO.Class      (MonadIO (..))
+import           Control.Monad               (forM_, join)
+import           Control.Monad.State.Strict
+  (StateT, gets, MonadState (get), MonadTrans (lift), execStateT)
+import           Control.Monad.Trans.State.Strict (modify)
+import           Control.Monad.Extra         (unlessM)
 import qualified Data.ByteString.Lazy.UTF8   as BLU
 import qualified Data.ByteString.Lazy        as BL
+import qualified Data.Sequence               as Seq
+import           Data.Sequence               (Seq)
 import           Data.Either                 (partitionEithers)
-import           Data.List                   (elemIndex, foldl')
+import           Data.Foldable               (foldl')
+import           Data.List                   (elemIndex)
 import qualified Data.Text                   as Text
 import           Data.Maybe                  (isNothing, mapMaybe, catMaybes)
+import           Data.Map.Strict             (Map)
+import qualified Data.Map.Strict             as Map
+import           Data.Set                    (Set)
+import qualified Data.Set                    as Set
 import           Data.Word                   (Word8)
 
 -- GHC API
@@ -58,7 +70,6 @@ import qualified GHC.IfaceToCore as TcIface
 import qualified GHC.Tc.Utils.Monad as TcRnMonad
 import qualified GHC.Tc.Types as TcRnTypes
 import qualified GHC.Types.Unique.FM as UniqFM
-import qualified GHC.Types.Unique.Set as UniqSet
 import qualified GHC.Types.Var as Var
 import qualified GHC.Unit.Types as UnitTypes
 #else
@@ -85,7 +96,6 @@ import qualified TcIface
 import qualified TcRnMonad
 import qualified TcRnTypes
 import qualified UniqFM
-import qualified UniqSet
 import qualified Var
 #endif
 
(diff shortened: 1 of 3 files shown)


clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -103,49 +113,127 @@ import qualified Clash.Util.Interpolate              as I
...
-  , lbReprs=concat (map lbReprs lbs)
-  }
+type LoadedBinderT m a = StateT LoadedBinders m a
+
+-- | Stores modules with easy binder lookup
+type DeclCache = Map GHC.Module (Maybe (Map GHC.Name IfaceSyn.IfaceDecl))
+
+-- | Insert a list of keys and values into a 'Map'
+mapInsertAll :: Ord k => Map k a -> [(k, a)] -> Map k a
+mapInsertAll = foldl' (\m (k, v) -> Map.insert k v m)
```
alex-mckenna, line 138: You only use this in one place, it may as well be inlined

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -103,49 +113,127 @@ import qualified Clash.Util.Interpolate              as I
...
+  go :: Var.Var -> Either Var.Id (CoreSyn.CoreBndr, Int)
+  go v = case Id.isClassOpId_maybe v of
+    Just cls -> Right (v, goClsOp v cls)
+    Nothing -> Left v
+
+  goClsOp :: Var.Var -> GHC.Class -> Int
+  goClsOp v c =
+    case elemIndex v (Class.classAllSelIds c) of
+      Nothing -> error [I.i|
+        Internal error: couldn't find class-method
```
alex-mckenna, line 158: s/class-method/class method/

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -103,49 +113,127 @@ import qualified Clash.Util.Interpolate              as I
...
+        lb{lbUnlocatable=Set.insert bndr lbUnlocatable}
+    Just expr -> do
+      -- Add current expression and its class ops
+      let (fvs, clsOps) = bndrsInExpr expr
+      modify $ \lb@LoadedBinders{..} ->
+        lb { lbBinders=Map.insert bndr expr lbBinders
+           , lbClassOps=mapInsertAll lbClassOps clsOps }
+
+      -- Load all free variables - if not yet loaded
+      forM_ fvs $ \v -> do
```
alex-mckenna, line 189: No need for a `do` here

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -103,49 +113,127 @@ import qualified Clash.Util.Interpolate              as I
...
+-- load binders found in the optionally supplied expression.
+addBndrM ::
+  GHC.GhcMonad m =>
+  HDL ->
+  CoreSyn.CoreBndr ->
+  Maybe CoreSyn.CoreExpr ->
+  LoadedBinderT m ()
+addBndrM hdl bndr exprM =
+  case exprM of
+    Nothing -> do
```
alex-mckenna, line 178: No need for a `do` here

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -168,145 +256,132 @@ loadIface foundMod = do
...
-    clsOps1 = map goClsOp clsOps0
-
-  loaded1 <- mergeLoadedBinders <$> mapM (loadExprFromIface hdl) fvs2
-
-  loadExternalExprs'
-    hdl
-    (mergeLoadedBinders [loaded0, loaded1, emptyLb{lbClassOps=clsOps1}])
-    (foldl' UniqSet.addListToUniqSet visited0 [collectLbBinders loaded1, map fst clsOps0])
-    (es ++ map snd (lbBinders loaded1))
+  flip execStateT emptyLb $ do
```
alex-mckenna, line 263: No need for a `do` here

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -168,145 +256,132 @@ loadIface foundMod = do
...
+    mapM_ (loadExprFromIface hdl) bndrs
+
+-- Given a list of binds, recursively load all its binders, primitives, and
+-- type classes it is using. (Exported function.)
+loadExternalExprs :: GHC.GhcMonad m => HDL -> [CoreSyn.CoreBind] -> m LoadedBinders
```
alex-mckenna, line 268: Given the [recent mention of loading primitives later](https://github.com/clash-lang/clash-compiler/issues/2020#issuecomment-989669791) and the fact this part of the code base is about loading GHC things, it feels strange to me on re-read that `HDL` is given as an argument to these things. I think we should open an issue around

  1. loading primitives later in the compilation process (ideally in `clash-lib` not `clash-ghc`, since there is nothing inherently GHC-specific about loading Clash primitives)
  2. Removing the `HDL` parameter from functions that load haskell modules / interfaces etc.

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -168,145 +256,132 @@ loadIface foundMod = do
...
+loadExternalExprs hdl binds0 = do
+  flip execStateT initLb $ do
```
alex-mckenna, line 270: Not one, but _two_ redundant `do`. I suppose this is what I get for having HLint in my IDE setup

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
```diff
@@ -168,145 +256,132 @@ loadIface foundMod = do
...
+    (,reprs') <$> getAnnTargetName_maybe ann_target
+
+  go :: (Name.Name, [DataReprAnn]) -> Maybe DataRepr'
+  go (_name, []) = Nothing
+  go (_name,  [repr]) = Just $ dataReprAnnToDataRepr' repr
+  go (name, reprs')   =
+    error $ $(curLoc) ++ "Multiple DataReprAnn annotations for same type: \n\n"
+                      ++ (Outputable.showPpr DynFlags.unsafeGlobalDynFlags name)
+                      ++ "\n\nReprs:\n\n"
+                      ++ show reprs'
```
alex-mckenna, line 384: Can we not use the `[i| ... |]` quote here?

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -159,6 +159,7 @@ import           Clash.Annotations.BitRepresentation.Internal
   (DataRepr', dataReprAnnToDataRepr')
 
 import           Clash.Signal.Internal
+import Data.Foldable (toList)
```
alex-mckenna, line 162: This should be grouped with the other imports from `base`

martijnbastiaan: That's what I get for using an IDE ;)

alex-mckenna: :exploding_head:

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
martijnbastiaan: TBH, I like the `do`. It makes it easier to extend later on. But I'll remove it if it bothers you/hlint 👍

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
martijnbastiaan: Agreed. I'll open an issue, but it's not something I want addressed in this PR (as you also imply, I think).

christiaanb: The argument is provided because we load the primitives from `ANN` pragma's and it could be the case the primitive is only defined for a certain language.

There are reasons that primitives could only specified for certain backends
1. As a work-around for a certain backend
2. You have some custom logic and you (a non-clash-core-developer but just an advanced user) only care about one language

martijnbastiaan: Filtering could still happen after binder loading though. That way you could potentially make this backend agnostic and cache the process (however useless that may be). In any case I don't thing it's too important.

christiaanb: I needs to happen in `clash-ghc` because that's where `GHC2Core` lives, which uses the information to determine whether to turn a GHC Core `Var` into a Clash Core `Var` or `Prim`.

clash-ghc/src-ghc/Clash/GHC/LoadInterfaceFiles.hs
martijnbastiaan: Alright, I've moved it to be a local definition. I've kept it as a separate thing, because I think `mapInsertAll m xs` is a bit more parsable than `foldl' (\m (k, v) -> Map.insert k v m) theMap xs`

PR #2041, opened 2022-01-20
Efficient constants for initializing memories

DigitalBrains1:
Define a new data type, `MemBlob`, and provide new memory primitive
variants that can use a `MemBlob` for their initial contents.

Currently, we provide two ways of initializing memories: with a `Vec`
and with a file. In some cases, `Vec` can be prohibitively inefficient;
depending on how the `Vec` is constructed, this can already happen with
a few tens of memory locations. Files allow for very large and arbitrary
contents, but generating the contents of the file in the same
compilation run as generating the HDL is very brittle and is a great way
to shoot oneself in the foot. This is why our current documentation
refers to this as "living on the edge".

This new mechanism dubbed `MemBlob` provides the user with Template
Haskell constructing functions which store the data alongside the
compiled Haskell code with very little overhead (max 7%, often no
overhead at all). The Template Haskell functions are the only way of
constructing a concrete `MemBlob`.

The `MemBlob` is a new Clash netlist type, and the template language for
primitives can refer to such an argument with ~CONST for its value, and
~LENGTH and others for its type information, just like with a normal
`Vec`. Primitives for the memory functions are very similar to the
existing primitives, but can even be slightly simpler because the data
is always of type `BitVector m`.

A basic functionality test for `asyncRom` is also included as I had
written the code anyway to debug a silly issue with `asyncRomBlob`.

## Still TODO:

### DIFF
--- a/clash-lib/prims/verilog/Clash_Explicit_BlockRam_Blob.primitives
+++ b/clash-lib/prims/verilog/Clash_Explicit_BlockRam_Blob.primitives
@@ -1,0 +1,52 @@
+[ { "BlackBox" :
+    { "name" : "Clash.Explicit.BlockRam.Blob.blockRamBlob#"
+    , "kind" : "Declaration"
+    , "type" :
+"blockRamBlob#
+  :: KnownDomain dom           --       ARG[0]
+  => Clock dom                 -- clk,  ARG[1]
+  -> Enable dom                -- en,   ARG[2]
+  -> MemBlob n m               -- init, ARG[3]
+  -> Signal dom Int            -- rd,   ARG[4]
+  -> Signal dom Bool           -- wren, ARG[5]
+  -> Signal dom Int            -- wr,   ARG[6]
+  -> Signal dom (BitVector m)  -- din,  ARG[7]
+  -> Signal dom (BitVector m)"
+    , "outputReg" : true
+    , "template" :
+"// blockRamBlob begin
+reg ~TYPO ~GENSYM[~RESULT_RAM][1] [0:~LENGTH[~TYP[3]]-1];
+
+reg ~TYP[3] ~GENSYM[ram_init][3];
+integer ~GENSYM[i][4];
+initial begin
+  ~SYM[3] = ~CONST[3];
+  for (~SYM[4]=0; ~SYM[4] < ~LENGTH[~TYP[3]]; ~SYM[4] = ~SYM[4] + 1) begin
+    ~SYM[1][~LENGTH[~TYP[3]]-1-~SYM[4]] = ~SYM[3][~SYM[4]*~SIZE[~TYPO]+:~SIZE[~TYPO]];
+  end
+end
+~IF ~ISACTIVEENABLE[2] ~THEN
+always @(~IF~ACTIVEEDGE[Rising][0]~THENposedge~ELSEnegedge~FI ~ARG[1]) begin : ~GENSYM[~RESULT_blockRam][5]~IF ~VIVADO ~THEN
+  if (~ARG[2]) begin
+    if (~ARG[5]) begin
+      ~SYM[1][~ARG[6]] <= ~ARG[7];
+    end
+    ~RESULT <= ~SYM[1][~ARG[4]];
+  end~ELSE
+  if (~ARG[5] & ~ARG[2]) begin
+    ~SYM[1][~ARG[6]] <= ~ARG[7];
+  end
+  if (~ARG[2]) begin
+    ~RESULT <= ~SYM[1][~ARG[4]];
+  end~FI
+end~ELSE
+always @(~IF~ACTIVEEDGE[Rising][0]~THENposedge~ELSEnegedge~FI ~ARG[1]) begin : ~SYM[5]
+  if (~ARG[5]) begin
+    ~SYM[1][~ARG[6]] <= ~ARG[7];
+  end
+  ~RESULT <= ~SYM[1][~ARG[4]];
+end~FI
+// blockRamBlob end"
+    }
+  }
+]
--- a/clash-lib/prims/verilog/Clash_Explicit_ROM_Blob.primitives
+++ b/clash-lib/prims/verilog/Clash_Explicit_ROM_Blob.primitives
@@ -1,0 +1,37 @@
+[ { "BlackBox" :
+    { "name" : "Clash.Explicit.ROM.Blob.romBlob#"
+    , "kind" : "Declaration"
+    , "type" :
+"romBlob#
+  :: KnownDomain dom  --       ARG[0]
+  => Clock dom        -- clk,  ARG[1]
+  -> Enable dom       -- en,   ARG[2]
+  -> MemBlob n m      -- init, ARG[3]
+  -> Signal dom Int   -- rd,   ARG[4]
+  -> Signal dom (BitVector m)"
+    , "outputReg" : true
+    , "template" :
+"// romBlob begin
+reg ~TYPO ~GENSYM[ROM][1] [0:~LENGTH[~TYP[3]]-1];
+
+reg ~TYP[3] ~GENSYM[rom_init][3];
+integer ~GENSYM[i][4];
+initial begin
+  ~SYM[3] = ~CONST[3];
+  for (~SYM[4]=0; ~SYM[4] < ~LENGTH[~TYP[3]]; ~SYM[4] = ~SYM[4] + 1) begin
+    ~SYM[1][~LENGTH[~TYP[3]]-1-~SYM[4]] = ~SYM[3][~SYM[4]*~SIZE[~TYPO]+:~SIZE[~TYPO]];
+  end
+end
+~IF ~ISACTIVEENABLE[2] ~THEN
+always @(~IF~ACTIVEEDGE[Rising][0]~THENposedge~ELSEnegedge~FI ~ARG[1]) begin : ~GENSYM[~COMPNAME_rom][5]
+  if (~ARG[2]) begin
+    ~RESULT <= ~SYM[1][~ARG[4]];
+  end
+end~ELSE
+always @(~IF~ACTIVEEDGE[Rising][0]~THENposedge~ELSEnegedge~FI ~ARG[1]) begin : ~SYM[5]
+  ~RESULT <= ~SYM[1][~ARG[4]];
+end~FI
+// romBlob end"
+    }
+  }
+]
--- a/clash-lib/src/Clash/Backend/SystemVerilog.hs
+++ b/clash-lib/src/Clash/Backend/SystemVerilog.hs
@@ -23,6 +23,7 @@ import           Control.Monad                        (forM,liftM,zipWithM)
 import           Control.Monad.State                  (State)
 import           Data.Bifunctor                       (first)
 import           Data.Bits                            (Bits, testBit)
+import qualified Data.ByteString.Char8                as B8
 import           Data.Coerce                          (coerce)
 import           Data.Function                        (on)
 import           Data.HashMap.Lazy                    (HashMap)
@@ -53,6 +54,7 @@ import           Clash.Backend
 import           Clash.Backend.Verilog
   (bits, bit_char, encodingNote, exprLit, include, noEmptyInit, uselibs)
 import           Clash.Driver.Types                   (ClashOpts(..))
+import           Clash.Explicit.BlockRam.Internal     (unpackNats)
 import           Clash.Netlist.BlackBox.Types         (HdlSyn (..))
 import           Clash.Netlist.BlackBox.Util
   (extractLiterals, renderBlackBox, renderFilePath)
@@ -133,6 +135,7 @@ instance Backend SystemVerilogState where
     Vector {} -> pure UserType
     RTree {} -> pure UserType
     Product {} -> pure UserType
+    MemBlob {} -> pure UserType
     BiDirectional _ ty -> hdlHWTypeKind ty
     Annotated _ ty -> hdlHWTypeKind ty
     _ -> pure PrimitiveType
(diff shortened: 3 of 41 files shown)


DigitalBrains1:
And literally today, the day I submit the PR for review, I suddenly think: could it have been much simpler?

The whole purpose of this exercise is to circumvent the issue of unacceptably slow HDL generation because we are given a `Vec` that happens to be constructed inefficiently. In this PR this is done by side-stepping `Vec` completely, creating a new netlist type, and new primitives that take a `MemBlob` instead of a `Vec` but are otherwise almost the same.

Could we also have done something broadly like:

Instead of
```haskell
topEntity = romBlob $(memBlobTH @8 Nothing [1 .. 16])
```
we do
```haskell
blobVecTH
  :: [...]
  -> ExpQ
blobVecTH care es = [| blobToVec ($(memBlobTH care es)) |]

blobToVec :: MemBlob n m -> Vec n (BitVector m)
blobToVec = unsafeFromList . unpackMemBlob


topEntity = rom $(blobVecTH @8 Nothing [1 .. 16])
```
and we make `clash-lib` do what amounts to constant folding over the special `blobToVec` so the `MemBlob` gets turned into a `Vec`, but by lucky chance this `Vec` is not exhibiting the problematic behaviour? Because not all `Vec`s are created equal, the problems turn up only in certain constructions.

That would allow to dispense with all the extra primitives, and make it possible to use `blobVecTH` directly with all primitives that take a constant `Vec n a`.

martijnbastiaan:
Template Haskell will always generate data in its normal form, i.e. no thunks/calculations left.

> `Vec` can be prohibitively inefficient

If this purely refers to the normalization process, then yeah, using Template Haskell would probably work :sweat:. I was under the impression that your work related to memory usage. I.e., that 2 MB of ROM would blow up to some silly like 200.

DigitalBrains1:
> I was under the impression that your work was related to memory usage. I.e., that 2 MB of ROM would blow up to some silly like 200.

No, it's about time spent generating HDL. It's spent normalising, if I'm not misremembering :-). And I think Haskell simulation was also fine. But I can't quickly reproduce the issue anymore. It showed in these testbenches: https://github.com/clash-lang/clash-compiler/blob/d087788ded8721c60a912b05015bdb28e5d36aaa/tests/shouldwork/Cores/Xilinx/Floating.hs , which is why I switched to `romFile` for those. That turns out to be a great way to shoot oneself in the foot as noted. However, if I quickly change one to use a large `Vec`, it generates HDL just fine, so something changed that particular case for the better.

DigitalBrains1:
@christiaanb suspected the problem with some `Vec`s might be that GHC opted to generate a maximally lazy form when compiling. And if I'm not making stuff up on the spot, that this then exhibited a quadratic blowup while normalising. Exactly why some structures did and others didn't wasn't looked into. `listToVecTH` also stores the elements fully evaluated, yet it exhibited the issue. So it's probably not tied to that. Hmmm, come to think of it, was the large amount of time spent compiling Haskell? Was it `listToVecTH`? If only I could reproduce the issue :-).

christiaanb:
It is normalization of large vector literals that takes a long time. Because in certain circumstances GHC decides to make the vector maximally lazy by creating top-level binders for all the `Cons`-constructors, and all the elements, basically:
```
xs = Cons x xs1
x = 1
xs1 = Cons x1 xs2
x = 2
xs2 = Cons x2 xs3
etc.
```

clash-lib/prims/verilog/Clash_Explicit_ROM_Blob.primitives
```diff
@@ -0,0 +1,37 @@
...
+reg ~TYPO ~GENSYM[ROM][1] [0:~LENGTH[~TYP[3]]-1];
+
+reg ~TYP[3] ~GENSYM[rom_init][3];
+integer ~GENSYM[i][4];
+initial begin
+  ~SYM[3] = ~CONST[3];
+  for (~SYM[4]=0; ~SYM[4] < ~LENGTH[~TYP[3]]; ~SYM[4] = ~SYM[4] + 1) begin
+    ~SYM[1][~LENGTH[~TYP[3]]-1-~SYM[4]] = ~SYM[3][~SYM[4]*~SIZE[~TYPO]+:~SIZE[~TYPO]];
+  end
+end
```
christiaanb, line 24:
```suggestion
reg ~TYPO ~GENSYM[ROM][1] [0:~LENGTH[~TYP[3]]-1] = ~CONST[3];
```

clash-lib/prims/verilog/Clash_Explicit_BlockRam_Blob.primitives
```diff
@@ -0,0 +1,52 @@
...
+reg ~TYPO ~GENSYM[~RESULT_RAM][1] [0:~LENGTH[~TYP[3]]-1];
+
+reg ~TYP[3] ~GENSYM[ram_init][3];
+integer ~GENSYM[i][4];
+initial begin
+  ~SYM[3] = ~CONST[3];
+  for (~SYM[4]=0; ~SYM[4] < ~LENGTH[~TYP[3]]; ~SYM[4] = ~SYM[4] + 1) begin
+    ~SYM[1][~LENGTH[~TYP[3]]-1-~SYM[4]] = ~SYM[3][~SYM[4]*~SIZE[~TYPO]+:~SIZE[~TYPO]];
+  end
+end
```
christiaanb, line 27:
```suggestion
reg ~TYPO ~GENSYM[~RESULT_RAM][1] [0:~LENGTH[~TYP[3]]-1] = ~CONST[3];
```

clash-lib/src/Clash/Backend/SystemVerilog.hs
```diff
@@ -492,6 +501,22 @@ funDec ty@(RTree n elTy) | typeSize elTy > 0 = Just A.<$>
...
+  "function" <+> "automatic" <+> tName <+> tName <> "_from_lv" <> parens ("logic" <+> ranges <+> "i") <> semi <> line <>
+  indent 2
+    ("for" <+> parens ("int n = 0" <> semi <+> "n <" <+> int n <> semi <+> "n=n+1") <> line <>
+      indent 2 (tName <> "_from_lv" <> brackets "n" <+> "=" <+> "i[n]" <> semi)) <> line <>
+  "endfunction" <> line
+  where
+    tName  = tyName ty
+    ranges = brackets (int 0 <> colon <> int (n-1)) <>
+             brackets (int (m - 1) <> colon <> int 0)
+
```
christiaanb, line 519: I don't think we want to convert `MemBlob`s to logic vectors in BlackBox templates. We only ever want to use them as literals for memory-like functions. So this can be deleted.

Instead, you probably want to explicitly error out on MemBlob in the to/from SLV conversions: https://github.com/clash-lang/clash-compiler/blob/d087788ded8721c60a912b05015bdb28e5d36aaa/clash-lib/src/Clash/Backend/SystemVerilog.hs#L1268-L1290

clash-lib/src/Clash/Backend/SystemVerilog.hs
```diff
@@ -358,6 +361,12 @@ tyDec ty@(Product _ _ tys) | typeSize ty > 0 = Just A.<$> prodDec
         Just y' -> Just A.<$> (pure y' <+> x <> semi)
     tName    = tyName ty
     selNames = map (\i -> tName <> "_sel" <> int i) [0..]
+tyDec ty@(MemBlob n m) | typeSize ty > 0 = Just A.<$>
+  "typedef" <+> "logic"
+            <+> brackets (int (m - 1) <> colon <> int 0)
+            <+> tyName ty
+            <+> brackets (int 0 <> colon <> int (n - 1))
+            <> semi
```
christiaanb, line 369: I would remove this and instead add the following to `normalizeType`: https://github.com/clash-lang/clash-compiler/blob/d087788ded8721c60a912b05015bdb28e5d36aaa/clash-lib/src/Clash/Backend/SystemVerilog.hs#L292

```haskell
normaliseType (MemBlob n m) = Vector n (BitVector m)
```

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -781,6 +790,42 @@ funDec _ syn t@(Vector _ elTy) = Just
...
+    "begin" <> line <>
+      indent 2
+        ("for i in result'range loop" <> line <>
+            indent 2
+              ( "result" <> parens "i" <+> ":=" <+> "islv" <> parens ("i * " <> int m <+> "to (i+1) * " <> int m <+> "- 1") <> semi) <> line <>
+         "end" <+> "loop" <> semi <> line <>
+         "return" <+> "result" <> semi
+        ) <> line <>
+    "end" <> semi
+  )
```
christiaanb, line 827: This can be removed if we just treat `MemBlobn n m` as an alias for `Vector n (BitVector m)`

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -564,6 +567,12 @@ tyDec hwty = do
                  <+> parens (hsep (punctuate comma variantNames))
                  <> semi
 
+    MemBlob _ m ->
+      "type" <+> tyName hwty
+             <+> "is array (integer range <>) of std_logic_vector"
+             <> parens (int (m - 1) <+> "downto 0")
+             <> semi
+
```
christiaanb, line 575:
```suggestion
    MemBlob n m -> tyDec (Vector n (BitVector m))

```
Treat `MemBlob n m` as an alias for `Vector n (BitVector m)`. There's really no point to treat it any differently than an alias since the whole point of this exercise is to speed up compile time (and perhaps memory use in simulation)

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -1133,6 +1179,11 @@ tyName' rec0 (filterTransparent -> t) = do
                             , if rec0 then showt n `TextS.append` "_" else ""
                             , elTy']
       Ap $ makeCached (t, rec0) nameCache (return nm)
+    MemBlob n m -> do
+      return $ TextS.concat [ "memblob_of_"
+                            , if rec0 then showt n `TextS.append` "_" else ""
+                            , "std_logic_vector_"
+                            , showt m]
```
christiaanb, line 1186: This can be removed. And I would add the following clause to `filterTransparent` https://github.com/clash-lang/clash-compiler/blob/d087788ded8721c60a912b05015bdb28e5d36aaa/clash-lib/src/Clash/Backend/VHDL.hs#L1210
```
  MemBlob n m -> Vector n (BitVector m)
```

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -1191,6 +1242,7 @@ normaliseType enums@(RenderEnums e) hwty = case hwty of
   RTree _ _     -> hwty
   Product _ _ _ -> hwty
   Sum _ _       -> if e then hwty else BitVector (typeSize hwty)
+  MemBlob _ _   -> hwty
```
christiaanb, line 1245:
```suggestion
  MemBlob n m   -> Vector n (BitVector m)
```

clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -1223,6 +1275,7 @@ filterTransparent hwty = case hwty of
   Sum _ _           -> hwty
   CustomSum _ _ _ _ -> hwty
   FileType          -> hwty
+  MemBlob _ _       -> hwty
```
christiaanb, line 1278:
```suggestion
  MemBlob n m       -> Vector n (BitVector m)
```

clash-prelude/src/Clash/Explicit/BlockRam/Internal.hs
```diff
@@ -0,0 +1,181 @@
...
+unpackMemBlob
+  :: forall n m
+   . MemBlob n m
+  -> [BitVector m]
+unpackMemBlob MemBlob{..}
+  = map (BV 0) $
+      unpackNats (natToNum @n) (natToNum @m) runsB endsB
+ where
+  runsB = unsafePerformIO $ B.unsafePackAddressLen memBlobRunsLen memBlobRuns
+  endsB = unsafePerformIO $ B.unsafePackAddressLen memBlobEndsLen memBlobEnds
```
christiaanb, line 60:
```suggestion
unpackMemBlob
  :: forall n m
   . MemBlob n m
  -> [BitVector m]
unpackMemBlob = unsafePerformIO . unpackMemBlob'

unpackMemBlob'
  :: forall n m
   . MemBlob n m
  -> IO [BitVector m]
unpackMemBlob' MemBlob{..}
  = runsB <- B.unsafePackAddressLen memBlobRunsLen memBlobRuns
    endsB <- B.unsafePackAddressLen memBlobEndsLen memBlobEnds
    return $ map (BV 0) $
      unpackNats (natToNum @n) (natToNum @m) runsB endsB
```

clash-prelude/src/Clash/Explicit/BlockRam/Blob.hs
```diff
@@ -0,0 +1,370 @@
...
+  -> Signal dom Bool
+  -- ^ Write enable
+  -> Signal dom Int
+  -- ^ Write address @w@
+  -> Signal dom (BitVector m)
+  -- ^ Value to write (at address @w@)
+  -> Signal dom (BitVector m)
+  -- ^ Value of the blockRAM at address @r@ from the previous clock cycle
+blockRamBlob# !_ gen content@MemBlob{} = \rd wen waS wd -> runST $ do
+  ramStart <- newListArray (0,szI-1) $ unpackMemBlob content
```
christiaanb, line 169:
```suggestion
  bvList <- unsafeSTToIO (unpackMemBlob' content)
  ramStart <- newListArray (0,szI-1) bvList
```

clash-lib/src/Clash/Backend/SystemVerilog.hs
DigitalBrains1: I don't know, why not give users the option? People are creative; they might think of a completely new purpose for `MemBlob` in their fancy primitive, one where they want to look at the data differently with `toSLV`. And even though the actual `MemBlob` is only ever a `~CONST`, people might still conceivably want some `fromSLV` to make non-constant data the same shape as the `MemBlob`. They could write the relevant SystemVerilog conversion functions themselves, but we could also provide it to them here.

DigitalBrains1:
```haskell
toSLV t e = case t of
  MemBlob n m -> toSLV (Vector n (BitVector m)) e

fromSV (MemBlob n m) = fromSLV (Vector n (BitVector m))

simpleFromSLV (MemBlob n m) = simpleFromSLV (Vector n (BitVector m))

expFromSLV (MemBlob n m) = expFromSLV (Vector n (BitVector m))
```

christiaanb: Yeah, that works for me

PR #2048, opened 2022-01-23
Remove `-fclash-float-support`

### LINKED ISSUE #1757: Make -fclash-float-support the default and deprecate the flag
If you have a project that legitimately uses `Float` and forget to include the command-line parameter `-fclash-float-support`, the generated HDL will silently omit anything that works with `Float`s. It would be much better if Clash at least warned that something seems to go wrong.

martijnbastiaan:
Closes #1757

## Still TODO:

  - [x] ~~Check copyright notices are up to date in edited files~~ Such small changes..

### DIFF
--- a/Clash.hs
+++ b/Clash.hs
@@ -63,7 +63,7 @@ doHDL b src = do
   putStrLn $ "Loading dependencies took " ++ prepStartDiff
 
   generateHDL (buildCustomReprs reprs) domainConfs bindingsMap (Just b) primMap tcm tupTcm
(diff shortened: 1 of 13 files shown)


alex-mckenna (approved):

christiaanb:
Perhaps we should've updated this text as well: https://github.com/clash-lang/clash-compiler/blob/632a535596fc55f6e9996e6fe9506b685c7ab66e/clash-prelude/src/Clash/Tutorial.hs#L2186-L2208

Same for: https://github.com/clash-lang/clash-compiler/blame/master/docs/developing-hardware/flags.rst#L195-L199

alex-mckenna:
I'll add these corrections to #2047 later

PR #2050, opened 2022-01-25
Add bounds to dependencies for hackage upload

alex-mckenna:


christiaanb:
Ehm... if we're uploading to hackage, why not add bounds to all dependencies in the library stanza

PR #2060, opened 2022-01-29
Detect file changes even if Manifest changes

### LINKED ISSUE #2008: Manifest changes cause "target directory not empty" errors
Upgrading Clash can cause slightly annoying error messages as mentioned in this error:

https://github.com/clash-lang/clash-compiler/blob/88ced730c34214f3020fd01e2e68addb36473216/clash-lib/src/Clash/Driver.hs#L859-L873

We could improve this by defining a version of `Manifest` that just reads in files and their hashes:

https://github.com/clash-lang/clash-compiler/blob/88ced730c34214f3020fd01e2e68addb36473216/clash-lib/src/Clash/Driver/Manifest.hs#L173-L179

as this is the only information used to determine whether a target directory can be cleaned out or not.

martijnbastiaan:
Fixes #2008

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver/Manifest.hs
+++ b/clash-lib/src/Clash/Driver/Manifest.hs
@@ -1,5 +1,5 @@
 {-|
-Copyright  : (C) 2021, QBayLogic B.V.
+Copyright  : (C) 2021-2022, QBayLogic B.V.
 License    : BSD2 (see the file LICENSE)
 Maintainer : QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -124,6 +124,12 @@ instance FromJSON ManifestPort where
       <*> v .: "is_clock"
       <*> v .:? "domain"
 
+-- | Just the 'fileNames' part of 'Manifest'
+newtype FilesManifest = FilesManifest [(FilePath, ByteString)]
+
+instance FromJSON FilesManifest where
+  parseJSON = Aeson.withObject "FilesManifest" $ fmap FilesManifest . parseFiles
+
 -- | Information about the generated HDL between (sub)runs of the compiler
 data Manifest
   = Manifest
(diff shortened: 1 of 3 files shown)


clash-lib/src/Clash/Driver/Manifest.hs
```diff
@@ -223,13 +239,7 @@ instance FromJSON Manifest where
...
         <*> v .: "components"
         <*> (topComponent >>= (.: "name"))
-        <*> do
-              files <- v .: "files"
-              forM files $ \obj -> do
-                fName <- obj .: "name"
-                sha256 <- obj .: "sha256"
-                -- See Note [Failed hex digest decodes]
-                pure (fName, unsafeFromHexDigest sha256)
+        <*> parseFiles v
```
alex-mckenna, line 242: I think it might be better to change the field to a `FilesManifest` and then just use `parseJSON` here, that way if anything to do with how files + hashes are parsed you only need to change one place

alex-mckenna (approved):
LGTM, modulo that one change

clash-lib/src/Clash/Driver/Manifest.hs
martijnbastiaan: Makes sense

martijnbastiaan:
@alex-mckenna Changing the `Manifest` file to include `FilesManifest` amounted in an annoying number of changes across our codebase where I had to insert dumb wrap/unwrap code. I now do the wrap/unwrap once in the `FromJSON`/`ToJSON` instances, which keeps the changes local. Is that acceptable to you?

alex-mckenna:
Sure that works for now. There's manifest changes as part of #1932, so if I can do those before Wednesday I'll take care of it then

martijnbastiaan:
Hmm, not sure what you're saying. I'm proposing merging as-is, because people consuming `Manifest` files won't have to change their code. Still, the current code puts the burden of parsing the files part in one location, which is what I thought the core of your complaint was. Also, I didn't say it explicitly, but I find the wrapping/unwrapping kinda tedious in the first place.

If you're going to do it anyway for #1932, then there's no point in not doing it in this PR I think.

alex-mckenna:
> amounted in an annoying number of changes

I'm saying if it's neater (but more effort) to go through and change all this wrapping / unwrapping code then it can be done later, but I don't want to hold up this PR because it's not _really_ a problem

alex-mckenna:
> If you're going to do it anyway for #1932

That may not be before 1.6 though, depends what I have time to get around to. There are still open issues marked as 1.6 which I think should be prioritised

martijnbastiaan:
Wait, the current code doesn't even work in the first place. The whole point of `FilesManifest` is that it parses a manifest file, but only the `files` portion. I'm going to revert it to the original proposal 😅.

martijnbastiaan:
@alex-mckenna I've added a test to prevent people making the same mistake I did. Feel free to change it however you like after 1.6.

PR #2065, opened 2022-01-30
Remove fakedata hack

alex-mckenna:
The upstream fakedata has been updated to support aeson-2, so now
we can use that instead of the horrible URL-in-local-packages hack.


DigitalBrains1:
Kill it with fire!

PR #2067, opened 2022-02-01
Use `CACHE_FALLBACK_KEY` on GitLab CI

martijnbastiaan:
We've been seeing a lot of dependency rebuilds, likely due to parallel
CI runs overwriting each other's cache. Using a branch-specific cache,
plus a fallback on master hopefully mitigates this.

I've also taken the opportunity to remove the `cabal-store` part of the
cache key, as it is used to cache more than just Cabal.

### DIFF
--- a/.ci/gitlab/common.yml
+++ b/.ci/gitlab/common.yml
@@ -3,6 +3,7 @@
   timeout: 2 hours
   stage: build
   variables:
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
That's the thing with CI changes. You'll find out in a day or two whether it actually works.

PR #2068, opened 2022-02-02
Change RewriteMonad to use RWST from `transformers`

alex-mckenna:
The RewriteMonad as implemented was the same as the CPS version
of the transformer from transformers-0.5.6.0 onwards. Since we now
also want a monad based in IO so we can use `async`, it makes more
sense to use the transformers implementation (preferring the more
space efficient CPS version if available) instead of rewriting all
instances to use IO.

This commit also removes the MonadFail instance for RewriteMonad.
It was only used in one place, where a potentially failing pattern
match was used. This is replaced with a call to `either` which throws
a more meaningful error message on failures.

For users using `transformers-0.5.6.2` or newer (i.e. GHC 8.10 or newer),
there is effectively no change in the implementation as the manually
implemented version was CPS'd anyway.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Rewrite/Types.hs
+++ b/clash-lib/src/Clash/Rewrite/Types.hs
@@ -11,6 +11,7 @@
 
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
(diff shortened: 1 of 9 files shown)


alex-mckenna:
Note: this is cherry-picked from my concurrent normalization branch. If I can't get that ready I still want this in before 1.6 because it will make backports more tedious

clash-lib/src/Clash/Rewrite/Types.hs
```diff
@@ -162,83 +170,63 @@ normalizeUltra = clashEnv . Lens.to (opt_ultra . envOpts)
...
-  return a = R $ \ _ s w -> (a, s, w)
-  {-# INLINE return #-}
-  m >>= k  =
-    R $ \ r s w -> case unR m r s w of
-      (a,s',w') -> unR (k a) r s' w'
-  {-# INLINE (>>=) #-}
+  -> IO (a, RewriteState extra, Any)
+runR m = RWS.runRWST (unR m)
 
+#if MIN_VERSION_transformers(0,5,6)
```
christiaanb, line 189:
```suggestion
#if !MIN_VERSION_transformers(2,3,0)
```
The CPS instances will be added in the 2.3 release of `mtl`; this future-proofs us when we bump the `mtl` bounds in `clash-lib.cabal`

christiaanb (approved):
The only difference is that `state` is less strict in `transformers`.

Ours:
```haskell
state f = R $ \_ s w -> case f s of (a,s') -> (a,s',w)
{-# INLINE state #-}
```
Theirs:
```haskell
state :: (Monad m) => (s -> (a, s)) -> RWST r w s m a
state f = RWST $ \ _ s w -> let (a, s') = f s in return (a, s', w)
{-# INLINE state #-}
```
I think we're fine though

PR #2073, opened 2022-02-03
Mergify add 1.6 backport

martijnbastiaan:


alex-mckenna:
@martijnbastiaan I think we should probably turn off the _require up to date_, since it made this need an unnecessary rebase

martijnbastiaan:
Jup.

PR #2074, opened 2022-02-03
Concurrent normalization

alex-mckenna:
## Still TODO:

  - [ ] Make sure the flag to turn it on/off works


alex-mckenna:
> I recall Alex mentioning he couldn't reliably measure speedups, is that now fixed by using a work queue?

Part of that was a lack of programs to try it on where there was a significant branching factor when normalizing top-level binders. For the benchmarks / examples we have I was finding that actually not much was running at the same time (since so many entities are also so small this doesn't help). For many examples we have the number of children for a binder is typically 1, which means in the worst case there is no concurrency at all. When there are multiple children, it typically doesn't go higher than 2 or 3 in a design so the actual concurrency you observe is somewhat limited

The point is I don't think this can really be answered without a benchmark program where normalizing a binder can result in many child binders needing to be normalized as a result (i.e. an entity to normalize where it's pretty much guaranteed it will be able to find enough work for all cores on a machine). I think ideally this would be two benchmarks:

  * one where every child in the tree of things to normalize is distinct. This lets us see just the effect of concurrency without the work queue being able to share any normalization results
  * one where children can be repeated in the tree. This lets us see the effect of the work queue in further reducing the time needed to normalize the entire design

vmchale:
@martijnbastiaan sort of? It runs faster instead of slower, but not always drastically better. 

Let me gather more data.

leonschoorl (commented):
Is there a way to easily disable concurrent normalization?
I'd think that disabling it might be useful when trying debug the compiler.
Because with it enabled the various debug messages that clash can generate might be interleaved from different things being normalized in parallel.

I tried running with `+RTS -N1` or `+RTS -maxN1` but that doesn't seem to work.

vmchale:
comparison


Ok so: the work queue means that concurrent normalization isn't pathologically slow, but it isn't particularly impressive either.

alex-mckenna:
> Is there a way to easily disable concurrent normalization?

For easy debugging you would likely have to compile without `-threaded`. This is already somewhat of an implicit recommendation for when you need to use `gdb` with Haskell

clash-lib/src/Clash/Normalize.hs
```diff
@@ -108,54 +114,75 @@ runNormalization
...
+  -- ^ Synchronization on stdout
   -> [Id]
   -- ^ topEntities
   -> NormalizeSession a
   -- ^ NormalizeSession to run
   -> IO a
-runNormalization env supply globals typeTrans peEval eval rcsMap topEnts =
-  runRewriteSession rwEnv rwState
-  where
-    -- TODO The RewriteEnv should just take ClashOpts.
```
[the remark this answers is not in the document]
alex-mckenna: No, `RewriteEnv` now contains `ClashEnv` which has the opts in it

alex-mckenna:
One thing we should probably do is make we record transformation history sensibly though. Actually debugging the normalization process becomes somewhat more hectic if rewrites are seen (e.g. through the medium of `-fclash-debug-info AppliedTerm`) in the other they are performed.

Perhaps that means waiting util an entity is normalized, and only then printing out the rewrite steps it took (with locked access to `stdout`). That way even if entities aren't printed in the other they're normalized, you still see all their rewrite steps together without steps from other entities interspersed.

martijnbastiaan:
> Ok so: the work queue means that concurrent normalization isn't pathologically slow, but it isn't particularly impressive either.

Okay, that's good - hopefully it means our examples are just not benefitting from concurrent normalization. An example of a more realistic codebase would be Christiaan's [Contranomy](https://github.com/christiaanb/contranomy), perhaps try there?

(GitHub please copy threaded conversation from GitLab next pls.)

christiaanb:
Other "larger" public code bases I can think of are:
1. https://github.com/cbiffle/cfm (I started a port to Clash 1.4 over here https://github.com/cbiffle/cfm/pull/2, perhaps port over to Clash 1.6 first)
2. https://github.com/gergoerdi/clash-spaceinvaders
3. https://github.com/gergoerdi/clash-compucolor2

vmchale:
Unfortunately it seems to be a little slower on Contranomy too! 

Not concurrent:

```
time                 7.020 s    (6.929 s .. 7.067 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 6.929 s    (6.882 s .. 6.960 s)
std dev              48.11 ms   (31.76 ms .. 60.14 ms)
```

Concurrent:

```
time                 8.506 s    (8.338 s .. 8.615 s)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 8.438 s    (8.375 s .. 8.479 s)
std dev              61.79 ms   (24.56 ms .. 84.55 ms)
```

So you can see it's slower but nothing crazy (with work queues)

alex-mckenna:
Yeah, this is fairly similar to my experiences pre Vanessa's improvements, i.e. work queue and non-parallel GC (which I guess doesn't matter here anyway). At least contranomy is large enough that the std dev. doesn't go ridiculous

alex-mckenna:
It may be what we want to do is make a new flag, `-fclash-concurrent-normalization` which turns on concurrent normalization _only_. The concurrent topentity compilation remains always on. We document this flag with the caveat it's only worth it for designs where normalization is sufficiently wide that cores have enough to do

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -185,24 +204,36 @@ applyDebug
...
+
+    MVar.withMVar ioLockV $ \() ->
+      traceWhen (hasDebugInfo TryTerm name opts) ("Tried: " ++ name ++ " on:\n" ++ before)
+
+    countersV <- Lens.use transformCounters
 
-    Monad.when (dbg_countTransformations opts && hasChanged) $ do
-      transformCounters %= HashMap.insertWith (const succ) (Text.pack name) 1
+    Monad.when (dbg_countTransformations opts && hasChanged) $
+      MVar.putMVar countersV (HashMap.insertWith (const succ) (Text.pack name) 1 counters)
```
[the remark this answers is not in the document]
leonschoorl: @vmchale Now it's only putting the counters MVar back when there is no change: https://github.com/clash-lang/clash-compiler/blob/6866bcfccbc2bf241394b8bf3be7542cd1782715/clash-lib/src/Clash/Rewrite/Util.hs#L230-L235

Which will create a deadlock when there is a change, because the MVar isn't fill anymore.
Like I said above, I think it's better to not empty the MVar in the first place in `applyDebug`

(Try running the compiler with `-fclash-debug-history`, to see this deadlock)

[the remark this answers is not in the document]
vmchale: Ok, I think I got something sensible!

clash-lib/src/Clash/Normalize.hs
```diff
@@ -108,54 +114,75 @@ runNormalization
...
+    -> MVar (VarSet, [(Id, Binding Term)])
+    -> NormalizeSession ()
+normalizeStep q binds = do
+  res <- Monad.liftIO $ MS.tryPopR q
+  case res of
+      Just id' -> do
+        (bound, _) <- MVar.readMVar binds
+        unless (id' `elemVarSet` bound) $ do
+          -- immediately mark this work as being done
+          MVar.modifyMVar_ binds (pure . first (`extendVarSet` id'))
```
martijnbastiaan, line 174: You need to hold on to the binds-lock until line 174, or you risk race conditions, right?

clash-lib/src/Clash/Normalize.hs
```diff
@@ -192,10 +219,20 @@ normalize' nm = do
...
+            normV <- Lens.use (extra.normalized)
+
+            toNormalize <-
+              MVar.withMVar normV $ \norm ->
+                let prevNorm = mapVarEnv bindingId norm
+                    toNormalize = filter (`notElemVarSet` topEnts)
+                                $ filter (`notElemVarEnv` extendVarEnv nm nm prevNorm) usedBndrs
+                 in pure toNormalize
+
+            -- traceM ("normalize: end: " <> nmS)
```
martijnbastiaan, line 232: leftover

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: I'm also a bit suspicious of immediately marking it as done, isn't this misleading?

martijnbastiaan (changes requested):

alex-mckenna:
To save you looking around for everything, adding a flag is basically:

  * add the new configuration option to [`ClashOpts`](https://github.com/clash-lang/clash-compiler/blob/master/clash-lib/src/Clash/Driver/Types.hs#L291)
  * add the new option to the `NFData`, `Eq` and `Hashable` instances, and the `defClashOpts` further down in the same file
  * Define the new flag in [`ClashFlags.hs`](https://github.com/clash-lang/clash-compiler/blob/master/clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs)
  * Update the documentation for [`compiler flags`](https://github.com/clash-lang/clash-compiler/blob/master/docs/developing-hardware/flags.rst)

Then you can use the new flag by getting the `ClashEnv` from the monad and looking at it's `ClashOpts` inside.

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: > I think if you don't immediately mark it as "done" then two threads might work on it at once.

But I don't think you do it immediately, you need to

1. Get the lock for `binds`
2. Get the value of `binds`
3. Check for the id
4. Mark it as done
5. Release the lock for `binds`

Right now, I think this is happening:

1. Get the lock for `binds`
2. Get the value of `binds`
3. Release the lock for `binds`
4. Check for the id
5. Get the lock for `binds`
6. Mark it as done
7. Release the lock for `binds`

Because you acquire and release the lock twice, a competing thread can get hold of the value of `binds` and incorrectly conclude that it should _not_ skip.

The second part of my complaint is probably just a naming issue, like @alex-mckenna indicates :)

vmchale: Oh wow you're completely right, thank you for noticing! 

I've added a comment as well.

clash-lib/src/Clash/Normalize.hs
```diff
@@ -108,53 +114,85 @@ runNormalization
...
+      Just id' -> do
+        (bound, pairs) <- MVar.takeMVar binds
+        if not (id' `elemVarSet` bound)
+          then do
+            -- mark that we are attempting to normalize id'
+            MVar.putMVar binds (bound `extendVarSet` id', pairs)
+            pair <- normalize' id' q
+            MVar.modifyMVar_ binds (pure . second (pair:))
+          else
+            MVar.putMVar binds (bound, pairs)
```
martijnbastiaan, line 187: This still has the race condition. Modifying `binds` should be an atomic operation.

martijnbastiaan (changes requested):

clash-lib/src/Clash/Normalize.hs
vmchale: I don't follow - the `takeMVar` ... `putMVar` combination holds the `MVar`, no?

We do want the `MVar.modifyMVar_ binds (pure . second (pair:))` to be separate.

martijnbastiaan: Ah sorry, you're right! I confused myself.

martijnbastiaan (approved):

clash-lib/src/Clash/Normalize/Types.hs
```diff
@@ -28,25 +29,25 @@ import Clash.Rewrite.Types    (Rewrite, RewriteMonad)
 -- | State of the 'NormalizeMonad'
 data NormalizeState
   = NormalizeState
-  { _normalized          :: BindingMap
+  { _normalized          :: MVar BindingMap
```
christiaanb, line 32: This is insufficient because [`inlineWorkFree`](https://github.com/clash-lang/clash-compiler/blob/f39fb9805e27745177939eb269427ac482648a53/clash-lib/src/Clash/Normalize/Transformations/Inline.hs#L637-L711) [calls](https://github.com/clash-lang/clash-compiler/blob/f39fb9805e27745177939eb269427ac482648a53/clash-lib/src/Clash/Normalize/Transformations/Inline.hs#L705) [`normalizeTopLvlBndr`](https://github.com/clash-lang/clash-compiler/blob/f39fb9805e27745177939eb269427ac482648a53/clash-lib/src/Clash/Normalize/Util.hs#L419-L458)

So why is this bad? Take the following example:
```
top x = (g x, f x)

f x = h + x
{-# NOINLINE f #-}
g x = h * x
{-# NOINLINE g #-}

h = expensive_workfree_expression
```
After normalizing `top`, we normalize `f` and `g` concurrently. And both will do an `inlineWorkFree` of `h`. They could then both start normalizing `h` before committing it to the cache. Thus I think `normalizeTopLvlBndr` should do something akin:

```
normalizeTopLvlBndr isTop nm (Binding nm' sp inl pr tm _) = do
  normalizedV <- Lens.use (extra.normalized)
  cache <- takeMVar normalizedV
  case lookup nm cache of
    Just vMVar ->
       putMVar normalizedV cache
       readMVar vMVar
    Nothing ->
      tmp <- newEmptyMVar
      putMVar normalizedV (extendVarEnv nm tmp cache)
      ... continue normalization ...
      putMVar tmp value
      return value
```
I doesn't save us much time, because the thread arriving second still has to wait for the first. But the further along the first thread is, the more time we save.

martijnbastiaan:
So a couple of things:

 1. We haven't been able to verify any (meaningful) speedups on public codebases. I'll go around and ask some commercial partners to see if they observe any speedups. If not, I don't think we should merge this PR given that it adds quite a bit of complexity. 
 2. We should finish the flag work. Thanks @leonschoorl.
 3. We should clean up the commit history.

No use in doing (2) and (3) before (1) though.

changelog/2022-03-21T11_09_46-05_00_concurrent_normalization
```diff
@@ -0,0 +1 @@
+CHANGED: Add concurrent normalization flag [#2074](https://github.com/clash-lang/clash-compiler/pull/2074)
```
DigitalBrains1, line 1: Drive by review: Please explicitly name the flag (`-fclash-concurrent-normalization`) in the Changelog, otherwise people still have to hunt for it.

DigitalBrains1: As well as a brief description of what it does. The Changelog should be an understandable stand-alone unit. It doesn't have to be extensive, but it should be informative.

clash-lib/src/Clash/Normalize/Types.hs
vmchale: Done! That is clever.

It slows things down a little bit, however, perhaps due to more `MVar`s or perhaps because I convert to lists in order to `mapM` over a `VarEnv`.

changelog/2022-03-21T11_09_46-05_00_concurrent_normalization
DigitalBrains1: Also, it should be `ADDED`, not `CHANGED`. It would probably be good to document somehwere the keywords we expect here.

DigitalBrains1: I started documenting changelog best practices in #2169.

martijnbastiaan:
We've only been able to measure slight slowdowns, and never speedups. Let's keep the branch around, but I doubt we're going to merge this as is.

PR #2078, opened 2022-02-04
Clean up `clash-{prelude,lib}-hedgehog` flags

martijnbastiaan:
@alex-mckenna Do you need the old `debug` flag?


alex-mckenna:
> Do you need the old debug flag?

No

alex-mckenna (approved):

PR #2081, opened 2022-02-09
Minor improvements to TDP blockRAM

DigitalBrains1:
- Rename `NoOp` to `RamNoOp`
- Add `Show` to `RamOp`
- Include TDP BRAM in the several Preludes
- Improve documentation
- Fix test benches: they didn't run the final test because the test
length was hardcoded to the wrong value; now it is computed.

`NoOp` was added later, and its name breaks a pattern. The type itself and its constructors are all prefixed with `Ram`, except for `NoOp`. This also makes it annoying to import unqualified, because it might clash with another thing that also does nothing ;-). So it is renamed to fit with the rest. And all are added to the Preludes, making it available unqualified in most Clash designs by default.

## Still TODO:


christiaanb (approved):
LGTM

PR #2083, opened 2022-02-09
Use tdpRAM for `asyncFIFOSynchronizer`

christiaanb:
So that the generated circuit is nearly guaranteed to map to the dual-ported RAMs found on most FPGAs.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Synchronizer.hs
+++ b/clash-prelude/src/Clash/Explicit/Synchronizer.hs
@@ -16,6 +16,7 @@ Synchronizer circuits for safe clock domain crossings
 
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise       #-}
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.KnownNat.Solver #-}
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.Extra.Solver    #-}
 
 {-# OPTIONS_HADDOCK show-extensions #-}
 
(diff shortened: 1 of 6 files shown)


DigitalBrains1 (approved):
LGTM overall. It would be nice if we had a test, but perhaps we should do that later when there's more time.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
```diff
@@ -97,27 +97,62 @@ dualFlipFlopSynchronizer clk1 clk2 rst en i =
...
+  -> ( ( BitVector (addrSize + 1)
+       , BitVector (addrSize + 1)
+       , Bool )
+     , ( Bool
+       , BitVector addrSize
+       , BitVector (addrSize + 1)
+       )
+     )
+readPtrCompareT (bin, ptr, flag) (s_ptr, inc) =
+  ((bin', ptr', flag'), (flag, addr, ptr))
```
DigitalBrains1, line 144: Didn't you mention wanting to use one register less in the read-addressing, because a TDP BRAM has registered read inputs whereas `asyncRam` does not? Have you tested you still get the next element when `empty` goes from `True` to `False`?

DigitalBrains1 (changes requested):
Quick experimentation in `clashi` seem to indicate problems (no data on data out). We should make a test bench for this and see that it passes.

[edit]
PEBKAC. I accidentally didn't supply enough input on the write side, and that made the read side undefined, as well it should. There was no problem.
[/edit]

DigitalBrains1:
I'll write a test bench, I'm waiting for review of my own 1.6 PR's so I have time to kill.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
DigitalBrains1: Those primes are too easy to miss, IMHO. I used to do `f`, `f'`, `f''`, `f'3`, `f'4`. But now I do `f`, `f0`, `f1`, ... At least they don't look like a speck on your laptop screen. Note: I have many specks on my laptop screen ;-P.

christiaanb: Yeah, I do that as well these days, but I didn't want to change the overall style of the module.

DigitalBrains1:
Phew. I created a test bench, and even though there were differences between the `asyncRam` implemenation and the TDP BRAM implementation, they were all in the data out when the FIFO was empty. Arguably that shouldn't even be part of the test; on the other hand, if it ever changes that might be an interesting fact that should deserve some attention before the change is accepted.

I force-pushed to include this test bench. If you think it's not too bad, we can merge.

DigitalBrains1 (approved):

DigitalBrains1:
Ah. There is no disclaimer that the synchroniser might still be unsafe without cross-domain timing requirements on the synchronising flipflops for the pointers... It would be good to warn people about this.

christiaanb:
@DigitalBrains1 I've added the following lines w.r.t. the timing constraints: https://github.com/clash-lang/clash-compiler/pull/2083/files#diff-009f54843e228a7f1eafeaa90e2fb5a27a75d70cc3f2afd02790c357a356e01fR202-R204

DigitalBrains1:
Yes, that looks fine! I've added another test, and I changed a code comment:

https://github.com/clash-lang/clash-compiler/pull/2083/files#diff-009f54843e228a7f1eafeaa90e2fb5a27a75d70cc3f2afd02790c357a356e01fR179
```diff
--- a/clash-prelude/src/Clash/Explicit/Synchronizer.hs
+++ b/clash-prelude/src/Clash/Explicit/Synchronizer.hs
@@ -176,7 +176,7 @@ writePtrCompareT addrSize@SNat (bin, ptr, flag) (s_ptr, inc) =
 
   flag' = isFull addrSize ptr' s_ptr
 
--- FIFO full: when next pntr == synchonized {~wptr[addrSize:addrSize-1],wptr[addrSize-1:0]}
+-- FIFO full: when next pntr == synchronized {~wptr[addrSize:addrSize-1],wptr[addrSize-2:0]}
 isFull
   :: forall addrSize
    . (2 <= addrSize)
```

This mistake was in the tutorial as well and was fixed in 882fbf252.

Feel free to merge!

PR #2101, opened 2022-02-21
Fix Haskell/HDL simulation mismatch for `trueDualPortRam`

christiaanb:
1. The Haskell simulation now respects the port enable, just like the HDL blackbox
2. We now accurately model the concurrent active edges of both ports. ~~by running both the slow and fast port for 1 cycle, followed by the regular cadence of `ceil(fast/slow)` fast port cycles followed by one slow port cycle.~~
3. Read/Write are now only check when both ports are activated concurrently by an active edge
4. We also implement invalidation of a memory location on a Write/Write conflict

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/BlockRam.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam.hs
@@ -1317,8 +1317,8 @@ trueDualPortBlockRamModel ::
 
   (Signal domSlow a, Signal domFast a)
 trueDualPortBlockRamModel labelA !_clkA enA weA addrA datA labelB !_clkB enB weB addrB datB =
-  ( deepErrorX ("trueDualPortBlockRam: " <> labelA <> ": First value undefined") :- outA
-  , deepErrorX ("trueDualPortBlockRam: " <> labelB <> ": First value undefined") :- outB )
+  ( startA :- outA
+  , startB :- outB )
  where
   (outA, outB) =
     go
@@ -1327,9 +1327,13 @@ trueDualPortBlockRamModel labelA !_clkA enA weA addrA datA labelB !_clkB enB weB
       tA -- ensure 'go' hits fast clock first
       (bundle (enA, weA, fromIntegral <$> addrA, datA))
       (bundle (enB, weB, fromIntegral <$> addrB, datB))
+      startA startB
 
   tA = snatToNum @Int (clockPeriod @domSlow)
   tB = snatToNum @Int (clockPeriod @domFast)
+
+  startA = deepErrorX $ "trueDualPortBlockRam: " <> labelA <> ": First value undefined"
+  startB = deepErrorX $ "trueDualPortBlockRam: " <> labelB <> ": First value undefined"
 
   initElement :: Int -> a
   initElement n =
@@ -1402,6 +1406,7 @@ trueDualPortBlockRamModel labelA !_clkA enA weA addrA datA labelB !_clkB enB weB
     Int ->
     Signal domSlow (Bool, Bool, Int, a) ->
     Signal domFast (Bool, Bool, Int, a) ->
+    a -> a ->
     (Signal domSlow a, Signal domFast a)
   go conflict0 ram0 relativeTime as0 bs0 =
     if relativeTime <= 0 then goSlow else goFast
@@ -1410,11 +1415,11 @@ trueDualPortBlockRamModel labelA !_clkA enA weA addrA datA labelB !_clkB enB weB
     (enB_, weB_, addrB_, datB_) :- bs1 = bs0
 
     -- 1 iteration here, as this is the slow clock.
-    goSlow = out1 `seqX` (out1 :- as2, bs2)
+    goSlow _ prevB | enA_ = out1 `seqX` (out1 :- as2, bs2)
      where
       (wrote, !ram1) = writeRam weA_ addrA_ datA_ ram0
       out0 = fromMaybe (ram1 `Seq.index` addrA_) wrote
-      (as2, bs2) = go Nothing ram1 (relativeTime + tA) as1 bs0
+      (as2, bs2) = go Nothing ram1 (relativeTime + tA) as1 bs0 out1 prevB
       out1 =
         case conflict0 of
           Just Conflict{cfWrite=IsDefined True} ->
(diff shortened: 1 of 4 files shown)


DigitalBrains1:
My test is indeed merge-ready.

I copied the three clock domains from the TDP BRAM tests, but I was thinking about a larger difference in clock frequencies so pointers definitely propagate to the other domain while the origin domain still hasn't ticked again, but I don't want to add yet another clock domain because it is diluting the number of tests run. Right now we have three domains which is 7 combinations occurring in the 100 tests totally run. If I add another, that's 13 combinations, that's very few tests per domain combination.

However, as a compromise, I've manually run the Hedgehog test with domain `A` at 60 MHz. And it passed 30,000 tests, so it's probably okay.

christiaanb:
@martijnbastiaan This patch is now ready for review

DigitalBrains1:
> We now more accurately model the concurrent first rising edge of both ports, by running both the slow and fast port for 1 cycle, followed by the regular cadence of ceil(fast/slow) fast port cycles followed by one slow port cycle.

Actually, the number of rising edges of the fast port between two rising edges of the slow port can alternate between floor(fast/slow) and ceil(fast/slow), catching up on the accumulating "error" of the rounding over time. But the only thing wrong is this description, the actual code is fine at a glance.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1341,24 +1328,25 @@ trueDualPortBlockRamModel !_clkA enA weA addrA datA !_clkB enB weB addrB datB =
...
                 "\nAddress error message: " <> msg)
 
-  getConflict :: Bool -> Int -> Bool -> Int -> Maybe Conflict
-  getConflict enableA addrA_ enableB addrB_ =
+  getConflict :: Bool -> Bool -> Bool -> Int -> Bool -> Int -> Maybe Conflict
+  getConflict enA_ enB_ wenA addrA_ wenB addrB_ =
     -- If port A or port B is writing on (potentially!) the same address,
     -- there's a conflict
-    if sameAddr then maybeConflict else Nothing
+    if sameAddr then Just conflict else Nothing
```
martijnbastiaan, line 1335: Having `Maybe Conflict` instead of just `Conflict` is now an optimization step, right? That is, `Conflict` could signal there is no conflict by virtue of having all its fields set to `False`.

To be clear, I'm not opposed to this, I actually like it. I'm asking for my own understanding.

clash-prelude/tests/Clash/Tests/AsyncFIFOSynchronizer.hs
```diff
@@ -963,6 +970,149 @@ test6W7 = testW sync6W7
...
+     )
+fifoOperations racts wacts = (bundle (rAllDone, rel), bundle (wAllDone, wel))
+ where
+  (rdata, rempty, wfull) =
+    asyncFIFOSynchronizer d3 wclk rclk noWRst noRRst (toEnable wen)
+      (toEnable ren) rinc wDataM
+  rclk = clockGen @rdom
+  wclk = clockGen @wdom
+  noRRst = unsafeFromHighPolarity @rdom (pure False)
+  noWRst = unsafeFromHighPolarity @wdom (pure False)
```
martijnbastiaan, line 1053: Does it make sense to not reset at all? If so, could you add a comment why?

clash-prelude/tests/Clash/Tests/AsyncFIFOSynchronizer.hs
```diff
@@ -963,6 +970,149 @@ test6W7 = testW sync6W7
...
+  NamedTest { namedTest=test0, testName = name }
+ where
+  name = testNameShowsPrec 0 (shows 'fifoFunctionalTest) wdom rdom ""
+  test0 = do
+    racts <- fmap P.concat . forAll . Gen.list (Range.linear 1 10) . Gen.list (Range.linear 0 20) $ Gen.enumBounded
+    wacts <- fmap P.concat . forAll . Gen.list (Range.linear 1 10) . Gen.list (Range.linear 0 20) $ Gen.enumBounded
+    let rels = catMaybes . P.map snd . P.takeWhile (not . fst) . sample . fst $
+                 fifoOperations @wdom @rdom racts wacts
+        wels = catMaybes . P.map snd . P.takeWhile (not . fst) . sample . snd $
+                 fifoOperations @wdom @rdom racts wacts
```
martijnbastiaan, line 1102: This simulates `fifoOpperations` twice. I doubt that is what we want, given that the arguments to it are exactly the same. I'd split it up, something like:

```haskell
let
  results = fifoOperations @wdom @rdom racts wacts
  catValids = catMaybes . P.map snd . P.takeWhile (not . fst)
  rels = catValids (fst (sample results))
  wels = catValids (snd (sample results))
```

clash-prelude/tests/Clash/Tests/AsyncFIFOSynchronizer.hs
```diff
@@ -963,6 +970,149 @@ test6W7 = testW sync6W7
...
+   . Monad m
+  => DomProxy wdom
+  -> DomProxy rdom
+  -> NamedTest (PropertyT m ())
+fifoFunctionalTest wdom@DomProxy rdom@DomProxy =
+  NamedTest { namedTest=test0, testName = name }
+ where
+  name = testNameShowsPrec 0 (shows 'fifoFunctionalTest) wdom rdom ""
+  test0 = do
+    racts <- fmap P.concat . forAll . Gen.list (Range.linear 1 10) . Gen.list (Range.linear 0 20) $ Gen.enumBounded
```
martijnbastiaan, line 1097: Yeet this into a separate `genTestInputs` functions (maybe with a better name)

martijnbastiaan (approved):
I general LGTM, I've left a few suggestions.

It would be _wonderful_ if we could reuse the Hedgehog tests to test against actual HDL, but let's wait for proper cosimulation to land, instead of going through a bunch of TH pain (if it's possible at all). ~~@DigitalBrains1 I trust you validated the changes with HDL simulation?~~ Disregard that :-)

clash-prelude/src/Clash/Explicit/BlockRam.hs
christiaanb: I guess it is an "optimization" step. Or at least, `Nothing` indicates "no conflict".

clash-prelude/tests/Clash/Tests/AsyncFIFOSynchronizer.hs
DigitalBrains1: Ah yes, of course, good point.

clash-prelude/tests/Clash/Tests/AsyncFIFOSynchronizer.hs
martijnbastiaan: Good question, I don't really know. I always assumed we do it to get rid of undefined startup values - even though we don't actually test with domains that don't have defined startup values :|. My question was triggered by it being different from all the other testbenches. 

> Also, Xilinx even recommends avoiding resets in FPGA's if possible.

That's true, but they do recommend designing it in such a way that you can still reset your system. I.e., a pipeline might use resetless registers, but should nonetheless still converge to all known values while asserting the reset by virtue of registers in front of the pipeline.

I agree with your general thesis, we don't typically test "interesting" properties of the reset. I believe this to be an artifact of the painfulness that is writing Clash HDL testbenches..

DigitalBrains1: My `asyncRam` test benches also don't reset, and I changed the TDP BRAM test benches to no longer reset as the logic that was there to handle the first two samples was odd. So it's not the first instance, although those tests are recent!

Issue #2112, opened 2022-03-01
`fromJustX` should use `deepErrorX`


jonfowler:
Not sure what other peoples opinions are, but personally I think it is best practice to ensure the spine of a structure is always defined (avoiding the need to use lazy patterns everywhere). `fromJustX` doesn't follow this practice as it uses `errorX` as opposed to `deepErrorX`.

martijnbastiaan:
Makes sense to me, lazy patterns are a pain.

PR #2113, opened 2022-03-01
change fromJustX to use deepErrorX

jonfowler:
Does what it says in the title. Issue #2112

### DIFF
--- a/clash-prelude/src/Clash/Explicit/BlockRam/Blob.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam/Blob.hs
@@ -62,7 +62,7 @@ import Clash.Signal.Internal (Clock, Signal(..), (.&&.))
 import Clash.Sized.Internal.BitVector (Bit(..), BitVector(..))
 import Clash.Sized.Internal.Unsigned (Unsigned)
 import Clash.XException
-  (maybeIsX, deepErrorX, defaultSeqX, fromJustX, XException (..), seqX)
+  (maybeIsX, deepErrorX, defaultSeqX, fromJustX, NFDataX, XException (..), seqX)
 
 -- $setup
 -- >>> :set -XTemplateHaskell
@@ -89,6 +89,7 @@ blockRamBlob
   :: forall dom addr m n
    . ( KnownDomain dom
      , Enum addr
+     , NFDataX addr
      )
   => Clock dom
   -- ^ 'Clock' to synchronize to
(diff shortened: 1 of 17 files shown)


martijnbastiaan:
Could you add a changelog entry? See `changelog/README.md`.

clash-prelude/src/Clash/Explicit/BlockRam/Blob.hs
```diff
@@ -88,7 +88,9 @@ import Clash.XException
 blockRamBlob
   :: forall dom addr m n
    . ( KnownDomain dom
+     , KnownNat m
```
DigitalBrains1, line 91: This `KnownNat m` should not be necessary. If for some reason type checking does want this, it should be done by pattern matching on the `MemBlob` constructor instead. But it seems odd that the introduction of `NFDataX` would necessitate this addition. I can look at it tomorrow.

clash-prelude/src/Clash/Explicit/BlockRam/Blob.hs
```diff
@@ -126,6 +128,7 @@ blockRamBlob = \clk gen content rd wrM ->
 blockRamBlobPow2
   :: forall dom m n
    . ( KnownDomain dom
+     , KnownNat m
```
DigitalBrains1, line 131: Same as above.

clash-prelude/src/Clash/Prelude/BlockRam/Blob.hs
```diff
@@ -59,7 +60,9 @@ blockRamBlob
   :: forall dom addr m n
    . ( HiddenClock dom
      , HiddenEnable dom
+     , KnownNat m
```
DigitalBrains1, line 63: Same as above.

clash-prelude/src/Clash/Prelude/BlockRam/Blob.hs
```diff
@@ -91,6 +94,7 @@ blockRamBlobPow2
   :: forall dom m n
    . ( HiddenClock dom
      , HiddenEnable dom
+     , KnownNat m
```
DigitalBrains1, line 97: More of the same.

DigitalBrains1 (changes requested):

clash-prelude/src/Clash/Explicit/BlockRam/Blob.hs
DigitalBrains1: Right, I get it! Yes, the change does add a `KnownNat`, because the `NFDataX` instance for `BitVector` adds that constraint. It's late :-).

jonfowler: Done, I had just blindly followed the types!

DigitalBrains1 (approved):

DigitalBrains1:
I've made the changelog a bit more verbose and added a strongly related change that I had planned to include for 1.6 but ran out of time for.

If another reviewer agrees with my changes, is it okay if I force push onto your branch to reorganise the commits into a form suitable for merging? I don't like to force push onto other people's work without discussing it first :-).

DigitalBrains1:
@kloonbot run_ci badf8e1

DigitalBrains1:
@kloonbot run_ci ee7f005

martijnbastiaan:
@kloonbot run_ci fe2a53d

martijnbastiaan:
Thanks @jonfowler

PR #2115, opened 2022-03-02
Improve documentation of memory functions

DigitalBrains1:
## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/BlockRam.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam.hs
@@ -6,13 +6,13 @@ Copyright  :  (C) 2013-2016, University of Twente,
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
-BlockRAM primitives
+Block RAM primitives
 
 = Using RAMs #usingrams#
 
 We will show a rather elaborate example on how you can, and why you might want
-to use 'blockRam's. We will build a \"small\" CPU+Memory+Program ROM where we
-will slowly evolve to using blockRams. Note that the code is /not/ meant as a
+to use block RAMs. We will build a \"small\" CPU + Memory + Program ROM where we
+will slowly evolve to using block RAMs. Note that the code is /not/ meant as a
 de-facto standard on how to do CPU design in Clash.
 
 We start with the definition of the Instructions, Register names and machine
@@ -36,7 +36,7 @@ data Instruction
   | Load MemAddr Reg
   | Store Reg MemAddr
   | Nop
-  deriving (Eq,Show)
+  deriving (Eq, Show)
 
 data Reg
   = Zero
@@ -46,10 +46,10 @@ data Reg
   | RegC
   | RegD
   | RegE
-  deriving (Eq,Show,Enum)
+  deriving (Eq, Show, Enum, Generic, NFDataX)
 
 data Operator = Add | Sub | Incr | Imm | CmpGt
-  deriving (Eq,Show)
+  deriving (Eq, Show)
 
 data MachCode
   = MachCode
@@ -63,53 +63,60 @@ data MachCode
   , jmpM    :: Maybe Value
   }
 
-nullCode = MachCode { inputX = Zero, inputY = Zero, result = Zero, aluCode = Imm
-                    , ldReg = Zero, rdAddr = 0, wrAddrM = Nothing
-                    , jmpM = Nothing
-                    }
+nullCode =
+  MachCode
+    { inputX = Zero
+    , inputY = Zero
+    , result = Zero
+    , aluCode = Imm
+    , ldReg = Zero
+    , rdAddr = 0
+    , wrAddrM = Nothing
+    , jmpM = Nothing
+    }
 @
 
 Next we define the CPU and its ALU:
 
 @
 cpu
-  :: Vec 7 Value
-  -- ^ Register bank
-  -> (Value,Instruction)
-  -- ^ (Memory output, Current instruction)
+  :: Vec 7 Value          -- ^ Register bank
+  -> (Value,Instruction)  -- ^ (Memory output, Current instruction)
   -> ( Vec 7 Value
      , (MemAddr, Maybe (MemAddr,Value), InstrAddr)
      )
-cpu regbank (memOut,instr) = (regbank',(rdAddr,(,aluOut) '<$>' wrAddrM,fromIntegral ipntr))
-  where
-    -- Current instruction pointer
-    ipntr = regbank 'Clash.Sized.Vector.!!' PC
+cpu regbank (memOut, instr) =
+  (regbank', (rdAddr, (,aluOut) '<$>' wrAddrM, fromIntegral ipntr))
+ where
+  -- Current instruction pointer
+  ipntr = regbank 'Clash.Sized.Vector.!!' PC
 
-    -- Decoder
-    (MachCode {..}) = case instr of
-      Compute op rx ry res -> nullCode {inputX=rx,inputY=ry,result=res,aluCode=op}
-      Branch cr a          -> nullCode {inputX=cr,jmpM=Just a}
-      Jump a               -> nullCode {aluCode=Incr,jmpM=Just a}
-      Load a r             -> nullCode {ldReg=r,rdAddr=a}
-      Store r a            -> nullCode {inputX=r,wrAddrM=Just a}
-      Nop                  -> nullCode
+  -- Decoder
+  (MachCode {..}) = case instr of
+    Compute op rx ry res -> nullCode {inputX=rx,inputY=ry,result=res,aluCode=op}
+    Branch cr a          -> nullCode {inputX=cr,jmpM=Just a}
+    Jump a               -> nullCode {aluCode=Incr,jmpM=Just a}
+    Load a r             -> nullCode {ldReg=r,rdAddr=a}
+    Store r a            -> nullCode {inputX=r,wrAddrM=Just a}
+    Nop                  -> nullCode
 
-    -- ALU
-    regX   = regbank 'Clash.Sized.Vector.!!' inputX
-    regY   = regbank 'Clash.Sized.Vector.!!' inputY
-    aluOut = alu aluCode regX regY
+  -- ALU
+  regX   = regbank 'Clash.Sized.Vector.!!' inputX
+  regY   = regbank 'Clash.Sized.Vector.!!' inputY
+  aluOut = alu aluCode regX regY
 
-    -- next instruction
-    nextPC = case jmpM of
-               Just a | aluOut /= 0 -> ipntr + a
-               _                    -> ipntr + 1
+  -- next instruction
+  nextPC =
+    case jmpM of
+      Just a | aluOut /= 0 -> ipntr + a
+      _                    -> ipntr + 1
 
-    -- update registers
-    regbank' = 'Clash.Sized.Vector.replace' Zero   0
-             $ 'Clash.Sized.Vector.replace' PC     nextPC
-             $ 'Clash.Sized.Vector.replace' result aluOut
-             $ 'Clash.Sized.Vector.replace' ldReg  memOut
-             $ regbank
+  -- update registers
+  regbank' = 'Clash.Sized.Vector.replace' Zero   0
+           $ 'Clash.Sized.Vector.replace' PC     nextPC
+           $ 'Clash.Sized.Vector.replace' result aluOut
+           $ 'Clash.Sized.Vector.replace' ldReg  memOut
+           $ regbank
 
 alu Add   x y = x + y
 alu Sub   x y = x - y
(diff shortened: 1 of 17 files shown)


clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -6,14 +6,14 @@ Copyright  :  (C) 2013-2016, University of Twente,
...
+Block RAM primitives
 
 = Using RAMs #usingrams#
 
 We will show a rather elaborate example on how you can, and why you might want
-to use 'blockRam's. We will build a \"small\" CPU+Memory+Program ROM where we
-will slowly evolve to using blockRams. Note that the code is /not/ meant as a
-de-facto standard on how to do CPU design in Clash.
+to use 'blockRam's. We will build a \"small\" CPU + Memory + Program ROM where
+we will slowly evolve to using 'blockRam's. Note that the code is /not/ meant as
```
alex-mckenna, line 15: I think I would distinguish between "block RAMs" the concept and `blockRam` the function, i.e. either write

> to using block RAMs.

or

> to using 'blockRam'.

If we do this, a function would never need to be pluralized, which feels slightly odd to me

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -247,62 +256,62 @@ replace it with the string @\"undefined\"@ in the first few leading outputs.
...
 especially as the memories we need for our application get bigger. The
-'blockRam' function will be translated to such a /Block RAM/.
+'blockRam' function will be translated to such a /block RAM/.
 
-One important aspect of Block RAMs have a /synchronous/ read port, meaning that,
-unlike the behavior of 'Clash.Prelude.RAM.asyncRam', given a read address @r@
-at time @t@, the value @v@ in the RAM at address @r@ is only available at time
-@t+1@.
+One important aspect of block RAMs is that they have a /synchronous/ read port,
+meaning that, unlike the behavior of 'Clash.Prelude.RAM.asyncRam', given a read
```
alex-mckenna, line 264: * I would just say "meaning " instead of "meaning that, " with the extra comma. Having those two commas gives unnecessary emphasis on the words between them

  * I would take the oppourtunity to re-work the end of this sentence too. We give names to the address and value (`r` and `v`) but the only important bit is time IMO. Something like this reads better to me:

    > ... meaning, unlike an asyncRam, a value read at time t becomes available at cycle t + 1.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -247,62 +256,62 @@ replace it with the string @\"undefined\"@ in the first few leading outputs.
...
+address @r@ at time @t@, the value @v@ in the RAM at address @r@ is only
+available at time @t+1@.
 
 For us that means we need to change the design of our CPU. Right now, upon a
 load instruction we generate a read address for the memory, and the value at
 that read address is immediately available to be put in the register bank.
-Because we will be using a BlockRAM, the value is delayed until the next cycle.
-We hence need to also delay the register address to which the memory address
-is loaded:
+Because we will be using a block RAM, the value is delayed until the next cycle.
```
alex-mckenna, line 271: This isn't really a correct way to start a sentence with "Because". Generally speaking I would avoid it, and instead write something like

> We will be using a block RAM, so the value is delayed until the next cycle. Thus, we will need to also delay the register address to which the memory address is loaded:

A more proper way still starting with "Because" would be

> Because values are delayed until the next cycle when using a block RAM, we need to delay the register address to which the memory address is loaded:

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -317,17 +326,17 @@ system3
...
+  (rdAddr,dout,ipntr) = 'Clash.Explicit.Prelude.mealyB' clk rst en cpu2 (('Clash.Sized.Vector.replicate' d7 0),Zero) (memOut,instr)
+  instr  = 'Clash.Explicit.Prelude.asyncRom' instrs '<$>' ipntr
 @
 
 We are, however, not done. We will also need to update our program. The reason
 being that values that we try to load in our registers won't be loaded into the
 register until the next cycle. This is a problem when the next instruction
-immediately depended on this memory value. In our case, this was only the case
-when the loaded the value @6@, which was stored at address @1@, into @RegB@.
+immediately depends on this memory value. In our case, this was only the case
```
alex-mckenna, line 338: > In our case, this was only the case

I would change the first case to "example" to avoid repetition

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -729,12 +738,24 @@ prog2 = -- 0 := 4
...
 -}
 
--- | Create a blockRAM with space for @n@ elements
+-- | Create a block RAM with space for @n@ elements
 --
 -- * __NB__: Read value is delayed by 1 cycle
 -- * __NB__: Initial output value is /undefined/, reading it will throw an
 -- 'XException'
 --
+-- Additional helpful information:
```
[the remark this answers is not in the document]
alex-mckenna: Or possibly just "See Also" so the same heading can refer to snippets of documentation _and_ related functions etc.

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -729,12 +738,24 @@ prog2 = -- 0 := 4
...
+--
+-- * See "Clash.Explicit.BlockRam#usingrams" for more information on how to use a
+-- block RAM.
+-- * Use the adapter 'readNew' for obtaining write-before-read semantics like
+-- this: @'readNew' clk rst en ('blockRam' clk inits) rd wrM@.
+-- * A large 'Vec' for the initial content might be too inefficient, depending
```
alex-mckenna, line 753: `s/might/may/`

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -763,30 +774,42 @@ blockRam
   => Clock dom
   -- ^ 'Clock' to synchronize to
   -> Enable dom
-  -- ^ Global enable
+  -- ^ 'Enable' line
   -> Vec n a
-  -- ^ Initial content of the BRAM, also determines the size, @n@, of the BRAM.
+  -- ^ Initial content of the BRAM, also determines the size, @n@, of the BRAM
```
alex-mckenna, line 779: I would change "BRAM" to "block RAM" in documentation comments

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -847,24 +859,24 @@ blockRamU
   => Clock dom
   -- ^ 'Clock' to synchronize to
   -> Reset dom
-  -- ^ 'Reset' line to listen to. Needs to be held at least /n/ cycles in order
+  -- ^ 'Reset' line. Needs to be asserted at least /n/ cycles in order
```
alex-mckenna, line 862: _This_ needs to be asserted _for_ at least n cycles...

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -914,7 +926,7 @@ blockRamU#
   -> Signal dom a
   -- ^ Value to write (at address @w@)
   -> Signal dom a
-  -- ^ Value of the @blockRAM@ at address @r@ from the previous clock cycle
+  -- ^ Value of the BRAMblockRAM@ at address @r@ from the previous clock cycle
```
alex-mckenna, line 929: > BRAMblockRAM

_Is a normal block ram not blocky enough? Try new block RAM block RAM today_

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -926,8 +938,8 @@ blockRamU# clk en SNat =
 {-# NOINLINE blockRamU# #-}
 {-# ANN blockRamU# hasBlackBox #-}
 
--- | Version of blockram that is initialized with the same value on all
--- memory positions.
+-- | Version of 'blockRam' that is initialized with the same value on all
```
alex-mckenna, line 941: _A_ version of

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -939,10 +951,10 @@ blockRam1
   => Clock dom
   -- ^ 'Clock' to synchronize to
   -> Reset dom
-  -- ^ 'Reset' line to listen to. Needs to be held at least /n/ cycles in order
+  -- ^ 'Reset' line. Needs to be asserted at least /n/ cycles in order
```
alex-mckenna, line 954: This needs to be asserted for at least n cycles...

clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -1111,7 +1122,7 @@ blockRam# (Clock _) gen content = \rd wen waS wd -> runST $ do
 {-# ANN blockRam# hasBlackBox #-}
 {-# NOINLINE blockRam# #-}
 
--- | Create read-after-write blockRAM from a read-before-write one
+-- | Create read-after-write block RAM from read-before-write one
```
alex-mckenna, line 1125: Create _a_ read-after-write block RAM from _a_ read-before write one

clash-prelude/src/Clash/Explicit/BlockRam/File.hs
```diff
@@ -38,17 +38,18 @@ Such a file can be produced with 'memFile':
 writeFile "memory.bin" (memFile Nothing [7 :: Unsigned 9 .. 13])
 @
 
-We can instantiate a BlockRAM using the content of the above file like so:
+We can instantiate a block RAM using the content of the file above like so:
```
alex-mckenna, line 41: content**s** of the file above

clash-prelude/src/Clash/Explicit/BlockRam.hs
alex-mckenna: This comes up quite a bit, so I'll only comment it here for the first one

clash-prelude/src/Clash/Explicit/BlockRam.hs
DigitalBrains1: _BRAM BRAM BRAAAAAM_

DigitalBrains1:
GitHub is being really weird. I just force-pushed a rebased version and now it says closed, and it says both merged and not merged. I can't re-open, so I'll probably need to create a new PR, but that's something for another day.

DigitalBrains1:
I managed to utterly confuse poor GitHub by accidentally pushing the HEAD of master to this branch. Lesson: don't ever do that. GitHub closed the PR, but I'm now re-opening it because it is not done yet.

alex-mckenna:
It's alive!

clash-prelude/src/Clash/Explicit/BlockRam.hs
DigitalBrains1: I also moved the _also_ in the last sentence:

> Thus, we will also need to delay the register address to which the memory address is loaded:

Do you agree that's a more natural word order?

Or would you rather I say "I moved the _also_ also in the last sentence"? 😜

clash-prelude/src/Clash/Explicit/BlockRam.hs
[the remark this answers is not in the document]
DigitalBrains1: I made it a heading, and made it "See also:", no title case, ending in a colon.

clash-prelude/src/Clash/Explicit/BlockRam.hs
DigitalBrains1: Leaving it as is as discussed.

clash-prelude/src/Clash/Explicit/BlockRam.hs
alex-mckenna: That's _also_ fine :wink:

clash-prelude/src/Clash/Explicit/BlockRam.hs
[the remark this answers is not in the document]
alex-mckenna: _No_ title case? Why not?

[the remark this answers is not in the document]
DigitalBrains1: I don't see us use title case in headings? I checked `Clash.Explicit.BlockRam` as it was the file I was editing and `Clash.Tutorial` as I expected it to have a lot of headings. Neither use title case. I just followed what was already there.

I can also give a specific reason: I already think the "See also:" heading is pretty heavy, adding title case only increases its weight.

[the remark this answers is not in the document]
alex-mckenna: > I don't see us use title case in headings?

If it's more consistent for now then by all means keep it that way. If we consistently use sentence case for headings then we could also just adopt that as convention and remove that style question permanently.

I mentioned it some time recently, but perhaps it would be a good idea to have a style guide for documentation. I think it would be a good effort to make in terms of improving the quality of documentation if we set some rules for how it should be written.

DigitalBrains1:
All still was not well with GitHub, it wouldn't pick up the force-push. I temporary switched base and now it picked it up. Alex, I implemented all your suggestions. Can I merge this PR?

alex-mckenna (approved):
Looks good!

PR #2120, opened 2022-03-08
`~ISUNDEFINED` for `BitVector` and product types

### LINKED ISSUE #2117: -fclash-aggressive-x-optimization-blackboxes does not work on `deepErrorX` of `BitVector`
The following code generates a reset on the data when compiled with `clash -fclash-aggressive-x-optimization-blackboxes --vhdl`:
```haskell
import           Clash.Explicit.Prelude

topEntity
  :: Clock XilinxSystem
  -> Reset XilinxSystem
  -> Enable XilinxSystem
  -> Signal XilinxSystem (BitVector 8)
  -> Signal XilinxSystem (BitVector 8)
topEntity clk rst en = register clk rst en (deepErrorX "undefined reset")
```
vhdl:
```vhdl
  -- register begin
  result_1_register : process(clk)
  begin
    if rising_edge(clk) then
      if rst =  '1'  then
        result_1 <= std_logic_vector'("--------"); -- OFFENDING LINE
      elsif en then
        result_1 <= i;
      end if;
    end if;
  end process;
  -- register end
```
Replacing `BitVector` for `Unsigned` does not have this problem.

Presumably this is because of a BitVector creates a mask not a direct `errorX` and presumably similar problems will exist with records where every field is `errorX` but the structure is defined. The intuitive fix would be to pack the data and check whether it is all undefined as opposed to relying on the Haskell representation.

DigitalBrains1:
Fixes #2117

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
@@ -474,10 +474,8 @@ renderElem b (IF c t f) = do
                       _            -> 0
 
       (IsUndefined n) ->
-        let (e, _, _) = bbInputs b !! n in
-        case (xOpt, e) of
-          (True, BlackBoxE _ _ _ _ (N.BBTemplate [Err _]) _ _) -> 1
-          _ -> 0
+        let (e, _, _) = bbInputs b !! n
+        in if xOpt && checkUndefined e then 1 else 0
 
       (IsActiveEnable n) ->
         let (e, ty, _) = bbInputs b !! n in
@@ -1147,3 +1145,16 @@ onBlackBox
   -> r
 onBlackBox f _ (N.BBTemplate t) = f t
 onBlackBox _ g (N.BBFunction n h t) = g n h t
+
+-- | Is the value of the 'Expr' fully undefined?
+checkUndefined :: Expr -> Bool
+checkUndefined = \case
+  BlackBoxE _ _ _ _ (N.BBTemplate [Err _]) _ _ -> True
+  BlackBoxE "Clash.Sized.Internal.BitVector.fromInteger#" [] [] [] _ bbCtx _
+    | [sz, mask, _val] <- bbInputs bbCtx
+    , (Literal _ (NumLit sz0), _, True) <- sz
+    , (Literal _ (NumLit mask0), _, True) <- mask
+    , mask0 == 2^sz0 - 1
+    -> True
+  DataCon (Product _ _ _) _ es -> and (map checkUndefined es)
+  _ -> False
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -474,10 +475,19 @@ renderElem b (IF c t f) = do
...
-          _ -> 0
+        let
+          (e, _, _) = bbInputs b !! n
+          checkUndefined e0 = case e0 of
+            BlackBoxE _ _ _ _ (N.BBTemplate [Err _]) _ _ -> 1
+            BlackBoxE "Clash.Sized.Internal.BitVector.fromInteger#" [] [] [] _ bbCtx _
+              | [sz, mask, _val] <- bbInputs bbCtx
+              , (Literal _ (NumLit sz0), _, True) <- sz
+              , (Literal _ (NumLit mask0), _, True) <- mask
+              , mask0 == (1 `shiftL` fromInteger sz0) - 1
```
martijnbastiaan, line 486: Perhaps

```suggestion
              , mask0 == 2^sz0 - 1
```

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -474,10 +475,19 @@ renderElem b (IF c t f) = do
                       _            -> 0
 
       (IsUndefined n) ->
-        let (e, _, _) = bbInputs b !! n in
-        case (xOpt, e) of
-          (True, BlackBoxE _ _ _ _ (N.BBTemplate [Err _]) _ _) -> 1
-          _ -> 0
+        let
+          (e, _, _) = bbInputs b !! n
+          checkUndefined e0 = case e0 of
```
martijnbastiaan, line 480: Perhaps

```suggestion
          checkUndefined = \case
```

martijnbastiaan: I think `checkUndefined` is a very nice function to have on top-level too! I'm sure it'll come in handy somewhere else in the future.

tests/shouldwork/BlackBox/T2117.hs
```diff
@@ -0,0 +1,83 @@
...
+  :: NFDataX a
+  => a
+  -> Signal System Bool
+testUndefined !_ = pure True
+{-# NOINLINE testUndefined #-}
+{-# ANN testUndefined (InlinePrimitive [VHDL] $ unindent [i|
+  [ { "BlackBox" :
+      { "name"      : "T2117.testUndefined"
+      , "kind"      : "Expression"
+      , "template"  : "~IF ~ISUNDEFINED[1] ~THEN true ~ELSE false ~FI"
```
martijnbastiaan, line 61: Clever!

martijnbastiaan (approved):
LGTM

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
DigitalBrains1: Done!

martijnbastiaan (approved):

PR #2066, opened 2022-01-31
Black Boxes for `toEnum` / `fromEnum`

alex-mckenna:
This PR introduces black boxes for `toEnum` / `fromEnum` on the numeric types in `Clash.Sized.Internal`, suppressing the seemingly spurious integer black box warnings previously seen in #1918. As an added bonus, a commit is included which propagates the `-Werror` flag from GHC to Clash, allowing primitive warnings to be thrown as `ClashException` when desired.

Fixes #2046

## Still TODO:

  - [x] Write a changelog entry (see changelog/README.md)
  - [x] Check copyright notices are up to date in edited files
  - [x] Add a test based on `toEnum`, the current test uses `fromEnum` only

[comment]: # (Depending on the change, some of these points may not be necessary. If so, leave the boxes unchecked so it is easier for a reviewer to see what has been omitted.)

PR #2127, opened 2022-03-21
Use Int instead of Int# in toEnum / fromEnum prims

### LINKED ISSUE #2126: Clash 1.6.2 reports internal bug when generating VHDL
> <no location info>: error:
>     Clash error call:
>     Clash.Normalize.Transformations.Case(442): Report as bug: caseCon
> error: case 1# of
>       GHC.Types.I#[3891110078048108562]
>         (c$sel[170698662] :: GHC.Prim.Int#[3674937295934324764]) ->
>         c$sel[170698662][LocalId]
>     CallStack (from HasCallStack):
>       error, called at src/Clash/Normalize/Transformations/Case.hs:442:3 in
> clash-lib-1.6.2-DM9JJDZux0KEcYQd3mJFPd:Clash.Normalize.Transformations.Case
>

alex-mckenna:
In the primitives for conversion to / from Enum introduced in 0613f21,
the definitions are converting to / from Int# instead of Int as the
type of the functions demands. This is not correct, and means that
conversion to the enum type will never fire (as the argument is always
a boxed int instead of an unboxed int), and the conversion from the
enum type to Int will result in an ill-typed expression being output
which can lead to

```haskell
case 1# of
  I# x -> ...
```

which are obviously bogus.

Fixes #2126.

## Still TODO:


alex-mckenna:
Given the enum primitives from #2066 are completely broken, I think when this is merged we should immediately release 1.6.3

christiaanb (approved):

alex-mckenna:
I've added a regression test that uses the same types as the existing round-tripping one, but instead uses `fromEnum` then attempts to do something which pattern matches on the result. So for all the primitives from #2066 we can see that none produce an `Int#` instead of an `Int`.

tests/shouldwork/Issues/T2046C.hs
```diff
@@ -0,0 +1,21 @@
+module T2046C where
+
+import Clash.Prelude
+import T2046BType
+
+inBit :: (Enum a) => a -> Bool
+inBit x = case fromEnum x of
+        0 -> True
+        1 -> True
+        _ -> False
```
martijnbastiaan, line 10: Wouldn't GHC optimize this out? I think we might need a `NOINLINE`.

alex-mckenna: No, it doesn't

martijnbastiaan: > GHC won't constant fold because the fromEnum it would pick is NOINLINE anyway (since it's a clash prim)
> So the simplifier only gets as far as
> ``` case Clash.Whatever.fromEnum ... of```
> and gives up

PR #2129, opened 2022-03-22
Remove leading / trailing '\\' in SDC signal names

alex-mckenna:
In SDC files, escaped identifiers should not be rendered with
leading trailing backslashes as this causes faults in vendor tools
like Quartus.

## Still TODO:

### DIFF
--- a/changelog/2022-03-22T11_36_15+01_00_sdc_escaped_ids.md
+++ b/changelog/2022-03-22T11_36_15+01_00_sdc_escaped_ids.md
@@ -1,0 +1,1 @@
+FIXED: Clash no longer escapes extended identifiers when rendering SDC files
--- a/changelog/2022-03-22T11_36_15+01_00_sdc_escaped_ids.md
+++ b/changelog/2022-03-22T11_36_15+01_00_sdc_escaped_ids.md
@@ -1,1 +1,2 @@
-FIXED: Clash no longer escapes extended identifiers when rendering SDC files
+FIXED: Clash no longer escapes extended identifiers when rendering SDC files.
+CHANGED: Clash now always generates non-extended identifiers for port names, so that generated names play nicer with different vendor tools.
--- a/clash-lib/src/Clash/Driver/Types.hs
+++ b/clash-lib/src/Clash/Driver/Types.hs
@@ -31,6 +31,7 @@ import           Data.Maybe                     (isJust)
 import           Data.Set                       (Set)
 import qualified Data.Set                       as Set
 import           Data.Text                      (Text)
+import qualified Data.Text as Text              (dropAround)
 
 #if MIN_VERSION_prettyprinter(1,7,0)
 import           Prettyprinter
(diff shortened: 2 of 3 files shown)


DigitalBrains1:
The code does what we intended it to do and it makes Quartus happy. Quartus runs fine. However, it still leaves Vivado unsolved, I just discovered. It is rather more strict, as I get the following error from Vivado 2020.2:
```
[Vivado 12-2270] Clock names may not contain tcl special characters: '"{};$# - Skipping 'c$arg' ["/home/peter/src/clash/experiments/vhdl/Intel.topEntity/topEntity.sdc":1]
```

Perhaps instead of this PR we should implement @martijnbastiaan 's suggestion of avoiding extended identifiers for ports altogether?

alex-mckenna:
Perhaps, or is this more fitting when hdlSyn is Vivado?

alex-mckenna:
I guess the question really is _how much faith do we have in the lattice tools_

DigitalBrains1:
Looking at it more broadly, vendor tools aren't that happy about extended identifiers in certain specific places. We can probably avoid all the hullabaloo by using only basic identifiers in ports, since ports are most likely to crop up in places like SDC files.

This is all VHDL-centric, I don't know what crops up in these places with Verilog.

Would it be much work to only use basic identifiers in entity ports?

martijnbastiaan:
> I guess the question really is how much faith do we have in the lattice tools

[Ah. Ah. Ah.](https://www.youtube.com/watch?v=jwoWBrT-FyU)

martijnbastiaan:
My general feel is that issues with extended identifiers keep cropping up, while their only benefit seems to be that they more closely correspond to Haskell identifiers. I don't think anyone except Clash compiler developers care about that though.

> Would it be much work to only use basic identifiers in entity ports?

`s/Id.make/Id.makeBasic/g` on https://github.com/clash-lang/clash-compiler/blob/a85df82b6a9454f36014f4e5a65277a49c866df6/clash-lib/src/Clash/Netlist/Util.hs I think.

alex-mckenna:
@martijnbastiaan Yeah something like that. The two of you have convinced me though, ports should just be basic IDs since extended IDs just get you into the mire of _everything that can go wrong in every language where you might refer to the interface of an entity_. I'll tack on a new commit + updated changelog later

DigitalBrains1:
Users can be exposed to signal names. Most importantly I think during timing analysis, finding the critical path and manually assisted placing to reach your target speed. But also when looking at waveforms in an HDL simulator and whatnot.

I agree we should do basic identifiers for ports, but I do think other people besides Clash compiler developers care about the names.

alex-mckenna:
True, but for that kind of analysis you should just be using `-fclash-no-escaped-ids` anyway, since the extended IDs are very... _generated by a tool_

martijnbastiaan (approved):
LGTM. Let's wait for @DigitalBrains1's approval though.

DigitalBrains1 (approved):
Yes, this works fine both in Quartus and Vivado. I tested the case where Clash generates unfriendly identifiers itself (like `\c$arg\` which became `carg`) as well as the case where a `Synthesize` annotation specified a VHDL extended identifier. I picked `\CLOCK-50\`, which ends up as a `{CLOCK-50}` Tcl argument in the SDC. It all worked.

I did not test anything relating to (System)Verilog, I only looked at VHDL.

Note that it's still possible to use a `Synthesize` annotation to generate ports that lead to invalid Tcl, but those port names would also be rejected by Vivado even if we generated the Tcl syntactically correct. It does limit the user's options in Quartus, but this PR is good enough for me as it is now. We don't generate invalid stuff ourselves anymore.

PR #2134, opened 2022-03-23
better RTS options for executables, makes benchmarks more sensible

vmchale:
Stumbled upon this while working on https://github.com/clash-lang/clash-compiler/pull/2074

This makes benchmarks run much faster. It brings them into agreement with the ghc-options set in the `cabal.project` file. 

## Still TODO:

### DIFF
--- a/clash-ghc/clash-ghc.cabal
+++ b/clash-ghc/clash-ghc.cabal
@@ -68,7 +68,7 @@ flag use-ghc-paths
 executable clash
   Main-Is:            src-ghc/Batch.hs
   Build-Depends:      base, clash-ghc
-  GHC-Options:        -Wall -Wcompat -threaded -rtsopts
+  GHC-Options:        -Wall -Wcompat -threaded -rtsopts -with-rtsopts=-A128m
   if flag(dynamic)
     GHC-Options: -dynamic
   extra-libraries:    pthread
@@ -77,7 +77,7 @@ executable clash
 executable clashi
   Main-Is:            src-ghc/Interactive.hs
   Build-Depends:      base, clash-ghc
-  GHC-Options:        -Wall -Wcompat
+  GHC-Options:        -Wall -Wcompat -rtsopts -with-rtsopts=-A128m
   if flag(dynamic)
     GHC-Options: -dynamic
   extra-libraries:    pthread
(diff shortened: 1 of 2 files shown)


alex-mckenna (approved):
Looks good. One question: since we've added these here do we still need the ones in the `cabal.project` or can they be removed?

vmchale:
We still do! The `ghc-options` in our `cabal.project` are to control GHC when it's building our stuff, while the `ghc-options` passed in here are for the executable that is produced.

alex-mckenna (approved):
> We still do! The ghc-options in our cabal.project are to control GHC when it's building our stuff, while the ghc-options passed in here are for the executable that is produced.

In that case, have another green tick for the PR

clash-ghc/clash-ghc.cabal
```diff
@@ -68,7 +68,7 @@ flag use-ghc-paths
 executable clash
   Main-Is:            src-ghc/Batch.hs
   Build-Depends:      base, clash-ghc
-  GHC-Options:        -Wall -Wcompat -threaded -rtsopts
+  GHC-Options:        -Wall -Wcompat -threaded -rtsopts -with-rtsopts=-A128m
```
martijnbastiaan, line 71: No love for `clashi`? :)

alex-mckenna: Hmm good point. I would guess it doesn't normally matter for `clashi` until you do something like `:vhdl` which kicks of normalization etc.

alex-mckenna: So it should probably have it for `clashi` too I guess

alex-mckenna: If it's not tons of work, maybe just throw these things into a `common` stanza, seeing how we probably want them to be the same for `clash` and `clashi`

alex-mckenna: And we can just `import: rts-config`

martijnbastiaan (approved):

vmchale:
Ah, thanks!

alex-mckenna:
Reworded commit message just to do _something_ so GitHub actions would re-run

martijnbastiaan:
@vmchale Our commit message typically look like (but it's not a law or anything):

> Added X to Y

> Changed `foo` to include `bar`

> Improved error handling in `wibble`

with a longer explanation in the commit description, possibly ending with `Closes #11111` or `Fixes #11111`.

DigitalBrains1:
Note that the squash merge reset the commit message to what it was before, ignoring the new message. An interesting interaction between a squash merge and a _later_ force-push.

PR #2136, opened 2022-03-23
Fix and document Altera PLL's

DigitalBrains1:
Commit d325557750 introduced an off-by-one error in the number of clocks
to generate for `alteraPll`.

Additionally, the internal `locked` signal for `alteraPll` is changed to
be of type `Bit` like in `altpll`. The borrowing of the type of the
`Reset` input probably came about out of convenience back when resets
were still a simple type so it did not matter. Now it looked odd but
worked fine nonetheless.

Also adjusted the type in `CondAssignment` in `altpll`, which showed the
same pattern.

## Still TODO:

### DIFF
--- a/changelog/2022-03-20T12_59_08+01_00_fix_alterapll
+++ b/changelog/2022-03-20T12_59_08+01_00_fix_alterapll
@@ -1,0 +1,1 @@
+FIXED: The `alteraPll` primitive was unusable since commit d325557750 (release v1.4.0), it now works again. [#2136](https://github.com/clash-lang/clash-compiler/pull/2136)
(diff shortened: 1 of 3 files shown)


alex-mckenna:
This is what you get for recently doing a documentation PR, I actually thought about the documentation while reviewing :wink:

DigitalBrains1:
Okay, you motivated me to write more documentation for this module. [How's this?](https://clash-lang.gitlab.io/-/clash-compiler/-/jobs/2245713654/artifacts/hadocs/clash-prelude/Clash-Intel-ClockGen.html) _(note: build artifact of CI, will disappear next time CI runs for this PR)_

alex-mckenna:
This looks much better to me, but I think @leonschoorl or @christiaanb should give it a quick skim too in case I've missed something obvious (or worse, something subtle)

PR #2140, opened 2022-03-24
Improve `clash-prelude-hedgehog` generators

### LINKED ISSUE #2138: Misleading functions in `clash-prelude-hedgehog`
E.g., `genSigned` can generate results outside of its given range:

https://github.com/clash-lang/clash-compiler/blob/690fbea81758444d752e750964141d3f790677c8/clash-prelude-hedgehog/src/Clash/Hedgehog/Sized/Signed.hs#L28-L34

I think this is misleading at best. We probably want to inspect the given range and generate _its_ extreme values.

alex-mckenna:
This PR improves the generation of extreme values, and ensures that values generated better fit within the `Range` specified by callers.

Fixes #2138

## Still TODO:

### DIFF
--- a/changelog/2022-03-24T11_33_42+01_00_hedgehog_gen_extreme_values.md
+++ b/changelog/2022-03-24T11_33_42+01_00_hedgehog_gen_extreme_values.md
@@ -1,0 +1,1 @@
+FIXED: Extreme values are now generated from the input range instead of the type's bounds [#2138](https://github.com/clash-lang/clash-compiler/issues/2138)
--- a/clash-prelude-hedgehog/src/Clash/Hedgehog/Sized/Index.hs
+++ b/clash-prelude-hedgehog/src/Clash/Hedgehog/Sized/Index.hs
@@ -29,8 +29,8 @@ genIndex :: (MonadGen m, KnownNat n) => Range (Index n) -> m (Index n)
 genIndex range =
   Gen.frequency
     [ (60, Gen.integral range)
-    , (20, Gen.constant minBound)
-    , (20, Gen.constant maxBound)
+    , (20, Gen.constant (Range.lowerBound 99 range))
+    , (20, Gen.constant (Range.upperBound 99 range))
     ]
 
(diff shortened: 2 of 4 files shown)


leonschoorl (commented):
The first commit looks good to me.

But I'm unsure  about what you wrote in [Remove lowerBound extreme values from unsigned types](0624f1482384e1205663831a628437a1e9e1512a).

I don't have any experience with hedgehog specifically, but it's my understanding that the shrinking you mentioned only comes into effect after a test failure is detected.
So if the implementation being tested has a flaw that only shows up for an input of `0`, and no other flaws, then the bias in shrinking toward `0` doesn't help in finding that flaw.
Because it doesn't increase the probability of generating a `0` in the first place.
Or am I misunderstanding how this works?

DigitalBrains1:
Hedgehog will generate random values starting with a size parameter of 0, then stepwise increasing to 99 for following tests. So I think it's very likely to pick 0, because due to size 0 it will pick the first of `Gen.frequency` and then with high likelihood pick the `origin` from `Gen.range`. You could do some `bounds` invocations with a size 0 to see which values is picked from.

When Hedgehog calls something `Range.linear`, I think it's actually exponentially distributed. The linearity is not the distribution of actual samples, but how it scales relative to the size parameter. It took me some time to come to realise this. I tend to pick a much higher high bound than I think is a useful test, to increase the likelihood of high values occurring in the tests. There are also size-indifferent generators but those don't shrink, and shrinking is very nice.

martijnbastiaan:
Perhaps superfluously, but the origin != lower bound for signed types. So Hedgehog will start out with a range of `[0, 0]`, then perhaps `[-5, 5]`, then `[-10, 10]`, until it ends up in `[minBound, maxBound]`. Like @DigitalBrains1 mentions, it is therefore incredibly likely that `0` ends up in the tested values. Given that for unsigned types lower bound == origin, we do not need to increase the number of times it picks lower bound; Hedgehog is already biased towards it.

PR #2143, opened 2022-03-24
Only store Stack snapshots on CI

alex-mckenna:


alex-mckenna:
This is really just a backport of #2114, but Mergify doesn't like to backport things which change the CI workflow on GHA

PR #2153, opened 2022-03-29
Prim eval fixes

christiaanb:
Some fixes to the primitive evaluator

## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
@@ -1,6 +1,6 @@
 {-|
-  Copyright   :  (C) 2017, Google Inc.,
-                     2021, QBayLogic B.V.
+  Copyright   :  (C) 2017-2022, Google Inc.,
+                     2021     , QBayLogic B.V.
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -49,7 +49,7 @@ import           Clash.Core.Util
 import           Clash.Core.Var
 import           Clash.Core.VarEnv
 import           Clash.Debug
-import qualified Clash.Normalize.Primitives as NP (undefined)
+import qualified Clash.Normalize.Primitives as NP (removedArg, undefined, undefinedX)
 import           Clash.Unique
 import           Clash.Util                              (curLoc)
 
@@ -166,7 +166,7 @@ stepApp x y m tcm =
                     in  ghcPrimStep tcm (forcePrims m) p [] [Suspend (Var i), Suspend (Var j)] m1
 
               (e':es)
-                | primName p `elem` undefinedPrims
+                | primName p `elem` (undefinedXPrims ++ undefinedPrims)
                 -- The above primitives are (bottoming) values, whose arguments
                 -- are never used anywhere in the rest of the compiler. So
                 -- instead of pushing a PrimApply frame on the stack to evaluate
@@ -204,8 +204,9 @@ stepTyApp x ty m tcm =
       let tys = fst $ splitFunForallTy (primType p)
        in case compare (length args) (length tys) of
             EQ -> case lefts args of
-                    [] | primName p `elem` [ "Clash.Normalize.Primitives.removedArg"
-                                           , "Clash.Normalize.Primitives.undefined" ] ->
+                    [] | primName p `elem` fmap primName [ NP.removedArg
+                                                         , NP.undefined
+                                                         , NP.undefinedX ] ->
                             ghcUnwind (PrimVal p (rights args) []) m tcm
 
                        | otherwise ->
@@ -321,6 +322,8 @@ apply _tcm (Lambda x' e) x m =
   subst  = extendIdSubst subst0 x' (Var x)
   subst0 = mkSubst $ extendInScopeSet (mScopeNames m) x
 apply tcm pVal@(PrimVal (PrimInfo{primType}) tys vs) x m
+  | isUndefinedXPrimVal pVal
+  = setTerm (TyApp (Prim NP.undefinedX) ty) m
   | isUndefinedPrimVal pVal
   = setTerm (TyApp (Prim NP.undefined) ty) m
  where
@@ -337,6 +340,8 @@ instantiate _tcm (TyLambda x e) ty m =
   subst0 = mkSubst iss0
   iss0   = mkInScopeSet (freeVarsOf e `unionUniqSet` freeVarsOf ty)
 instantiate tcm pVal@(PrimVal (PrimInfo{primType}) tys []) ty m
+  | isUndefinedXPrimVal pVal
+  = setTerm (TyApp (Prim NP.undefinedX) (piResultTys tcm primType (tys ++ [ty]))) m
   | isUndefinedPrimVal pVal
   = setTerm (TyApp (Prim NP.undefined) (piResultTys tcm primType (tys ++ [ty]))) m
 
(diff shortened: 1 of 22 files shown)


tests/shouldwork/Basic/T1591.hs
```diff
@@ -32,8 +32,7 @@ data NextLineMask this that =
   } deriving (Generic)
 
 nextLineMask :: forall this that. (KnownNat this, KnownNat that) => Index (CountStates this that) -> NextLineMask this that
-nextLineMask =
-  errorX ""
+nextLineMask x = NextLineMask x
```
christiaanb, line 35: This change is needed because otherwise we get this failure: https://gitlab.com/clash-lang/clash-compiler/-/jobs/2269702690

Why does this failure pop up you might ask. That's because https://github.com/clash-lang/clash-compiler/pull/2153/files#diff-64d9f9192b0238fa29af4ce5d35293874d555046eb4708c9e1e9f56fa992c617R466-R467 transforms:
```haskell
case errorX "" of {}
```
to
```haskell
Clash.Normalize.Primitives.undefinedX
```
Which are in turn treated like values in the primitive evaluator here https://github.com/clash-lang/clash-compiler/pull/2153/files#diff-6a8afb744be8dbdbd693897804e34547af6fbd7a6e44bbe4a0fb79b028fd1986R169 which then results in needing to calculate the type `ty` in https://github.com/clash-lang/clash-compiler/pull/2153/files#diff-6a8afb744be8dbdbd693897804e34547af6fbd7a6e44bbe4a0fb79b028fd1986R319-R320

The ultimate cause of the error by the way is because we throw away casts.

tests/shouldwork/TopEntity/PortGeneration.hs
```diff
@@ -55,7 +53,7 @@ data Passthrough a b = Passthrough a b
 topEntity1 :: "in1" ::: Signal System Int
            -> "in2" ::: Signal System Bool
            -> "out" ::: Signal System Int
-topEntity1 = undefined
+topEntity1 _ _ = pure (deepErrorX "out")
```
christiaanb, line 56: We need this change, and the other changes in this file, for the same reasons as we need them for `tests/shouldwork/Basic/T1591.hs`

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -479,6 +478,12 @@ undefinedPrims =
   , "GHC.Real.underflowError"
   ]
 
+undefinedXPrims :: [T.Text]
+undefinedXPrims =
+  [ "Clash.Normalize.Primitives.undefinedX"
+  , "Clash.XException.errorX"
```
alex-mckenna, line 484: Something that just occurs to me: since `errorX` is just `throw (XException ...)` it presumably always gets left alone by the simplifier, but what if a user writes `throw (XException "Boom")` directly? Does this then fail to be identified as an _undefined X_ primitive and get identified as a normal undefined primitive?

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -263,29 +263,38 @@ caseCon' ctx@(TransformContext is0 _) e@(Case subj ty alts) = do
...
       -- means the entire case-expression is _|_
       (Prim pInfo,[_],ticks)
         | primName pInfo `elem` [ Text.showt 'NP.undefined
-                                , "Clash.GHC.Evaluator.undefined"
-                                , "EmptyCase"] ->
+                                , Text.showt 'NP.undefinedX ] ->
         let e1 = mkApps (mkTicks (Prim pInfo) ticks) [Right ty]
         in changed e1
+      -- WHNF of subject is _|_, in the form of `errorX`: that means that
+      -- the entire case-expression is evaluates to _|_
```
alex-mckenna, line 297: This hasn't changed, but reading this case back it seems odd that we silently discard the ticks when it's `errorX` since we keep the ticks for other forms of error. Is this intentional?

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
```diff
@@ -165,8 +165,11 @@ stepApp x y m tcm =
                         (m1,j) = newLetBinding tcm m0 a1
                     in  ghcPrimStep tcm (forcePrims m) p [] [Suspend (Var i), Suspend (Var j)] m1
 
-              (e':es) ->
-                Just . setTerm e' $ stackPush (PrimApply p (rights args) [] es) m
+              (e':es)
+                | primName p `elem` (undefinedXPrims ++ undefinedPrims)
+                -> ghcUnwind (PrimVal p (rights args) (map Suspend (e':es))) m tcm
```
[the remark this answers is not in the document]
christiaanb: My thought was: why would we want to spend the time evaluating some complex expression forming a error message or callstack when we're never going to use those things in the rest of the evaluator.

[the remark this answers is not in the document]
christiaanb: I'll add a note

clash-lib/src/Clash/Normalize/Transformations/Case.hs
christiaanb: Seems like an oversight when the code was originally introduced https://github.com/clash-lang/clash-compiler/commit/cedd0eeafe051e515f17403bf96e65ff979f0f54, I'll fix it.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -1585,6 +1599,11 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
     | [Lit (NaturalLiteral n), _] <- args
     -> reduce (Literal (NaturalLiteral n))
 
+  "GHC.TypeNats.someNatVal"
+    | [Lit (NaturalLiteral n)] <- args
+    -> let resTy = getResultTy tcm ty tys
+        in reduce (mkSomeNat tcm n resTy)
```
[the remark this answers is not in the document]
christiaanb: I'll add it as a new issue.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -1790,6 +1809,27 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
            val = unpack (toBV i :: BitVector 64)
         in reduce (mkDoubleCLit tcm val resTy)
 
+  "Clash.Class.BitPack.Internal.xToBV"
+    | isSubj
+    , Just (nTy, kn) <- extractKnownNat tcm tys
+    , [ _, (Suspend arg) ] <- args
+    , eval <- Evaluator ghcStep ghcUnwind ghcPrimStep ghcPrimUnwind
+    , mach1@Machine{mStack=[],mTerm=argWHNF} <-
+        whnf eval tcm True (setTerm arg (stackClear mach))
```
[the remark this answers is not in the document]
christiaanb: I've added a note reference now

clash-lib/src/Clash/Core/Util.hs
christiaanb: There's no eval rule for `throw` right now, so the whole evaluator will reduce to `Nothing`. I'll add a new issue so this doesn't get lost.

clash-ghc/src-ghc/Clash/GHC/Evaluator.hs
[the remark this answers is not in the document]
alex-mckenna: > unwind the stack with the with the primitive value

alex-mckenna (approved):

clash-lib/src/Clash/Core/Util.hs
christiaanb: https://github.com/clash-lang/clash-compiler/issues/2158

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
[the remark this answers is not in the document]
christiaanb: I don't think it's needed. I'm pretty sure we get an unfolding for https://hackage.haskell.org/package/base-4.16.1.0/docs/src/GHC-TypeLits.html#someNatVal

PR #1980, opened 2021-11-01
Keep recursive and non-recursive let in Clash Core

alex-mckenna:
In GHC core, the difference between recursive and non-recursive binders is preserved. In Clash, this difference is not preserved in let expressions / the bindings map, which leads to sometimes needing to calculate the connected components of the bindings.

However, since Letrec is used a lot (i.e. in normalization) it is painful to simply differentiate everywhere immediately. As a compromise, this is kept as a pattern synonym so existing code should work as it did before. In the future we can migrate these cases until only `Let NonRec{}` and `Let Rec{}` are used in code. Until this time, the assumption is made that `NonRec` is _always_ non-recursive, but `Rec` holds recursive and _potentially_ recursive bindings.

The parts of the compiler where it is easier to change the definitions now have been changed.

## Still TODO:

  - [x] Write a changelog entry (see changelog/README.md)
  - [x] Check copyright notices are up to date in edited files

[comment]: # (Depending on the change, some of these points may not be necessary. If so, leave the boxes unchecked so it is easier for a reviewer to see what has been omitted.)

PR #2164, opened 2022-04-06
Preserve non-recursive let in normalization

alex-mckenna:
In #1980, Clash started to keep the distinction between recursive
and non-recursive let expressions in Core (like GHC). However, for
convenience the old Letrec constructor is still used in some places
to avoid the need to fix every location at once.

The `inlineCleanup` transformation was one such place, however the
fix is relatively simple: when there is only one binder at the end
of cleanup and the original let expression was non-recursive, this
binding must also be non-recursive.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Core/Term.hs
+++ b/clash-lib/src/Clash/Core/Term.hs
@@ -29,6 +29,7 @@ module Clash.Core.Term
   , TmName
   , varToId
   , Bind(..)
+  , bindToList -- TODO Eventually all Letrec are removed and this isn't needed.
   , LetBinding
   , Pat (..)
   , patIds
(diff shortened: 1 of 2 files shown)


alex-mckenna:
This PR was badly named because it just used the name of the first commit.

The intent of this is to gradually preserve more of the non-recursive let expressions in normalization, so that the output of normalization does not clump everything into one large let expression. This should decrease the amount of work spent topologically sorting let binders since (1) a lot of bindings can stay in non-recursive binders which don't get top. sorted and (2) the remaining recursive let expressions are smaller so quicker to sort.

This will eventually involve some pain: netlist generation currently relies on the output of normalization having this single recursive let at the top of the term. As a result some transformations like ANF have the implicit invariant that they produce code in this style.

martijnbastiaan:
Closing due to inactivity - please reopen if there is an intention to go through with it.

PR #2167, opened 2022-04-07
Clash FFI

alex-mckenna:
This PR introduces the `clash-ffi` package, providing a means to interface with simulation tools over a standard interface (currently only VPI, the _Verilog Procedural Interface_). This package supersedes `clash-cosim`, which can be re-implemented on top of `clash-ffi` later if desired.

## Still TODO:

  - [ ] Implement parsers for reading bitstrings as `BitVector n` / `Vec n Scalar`
  - [ ] Move property, time and value under `Clash.FFI.VPI.Object`, expose operations from relevant modules with classes (similar to how VPI methods for getting child objects are already implemented)
  - [ ] Document public API
  - [ ] Add developer source notes for internal design choices
  - [ ] Integrate into CI / DevOps processes

### DIFF
--- a/clash-ffi/README.md
+++ b/clash-ffi/README.md
@@ -45,57 +45,57 @@ All interaction with a simulator follows the same general process:
 
 General Functions
 
-| VPI Definition            | Supported | Haskell API Function(s)             |
-| :---                      | :---:     | :---                                |
-| vpi_chk_error             | YES       |                                     |
-| vpi_compare_objects       | YES       |                                     |
-| vpi_control               | YES       |                                     |
-| vpi_flush                 | NO        |                                     |
-| vpi_get                   | YES       |                                     |
-| vpi_get_cb_info           | YES       |                                     |
-| vpi_get_data              | NO        |                                     |
-| vpi_get_delays            | NO        |                                     |
-| vpi_get_str               | YES       |                                     |
-| vpi_get_systf_info        | NO        |                                     |
-| vpi_get_time              | YES       |                                     |
-| vpi_get_userdata          | NO        |                                     |
-| vpi_get_value             | YES       |                                     |
-| vpi_get_vlog_info         | YES       |                                     |
-| vpi_handle                | YES       |                                     |
-| vpi_handle_by_index       | YES       |                                     |
-| vpi_handle_by_multi_index | YES       |                                     |
-| vpi_handle_by_name        | YES       |                                     |
-| vpi_handle_multi          | NO        |                                     |
-| vpi_iterate               | YES       |                                     |
-| vpi_mcd_close             | NO        |                                     |
-| vpi_mcd_flush             | NO        |                                     |
-| vpi_mcd_name              | NO        |                                     |
-| vpi_mcd_open              | NO        |                                     |
-| vpi_mcd_printf            | NO        |                                     |
-| vpi_mcd_vprintf           | NO        |                                     |
-| vpi_printf                | YES       |                                     |
-| vpi_put_data              | NO        |                                     |
-| vpi_put_delays            | NO        |                                     |
-| vpi_put_userdata          | NO        |                                     |
-| vpi_put_value             | YES       |                                     |
-| vpi_register_cb           | NO        |                                     |
-| vpi_register_systf        | NO        |                                     |
-| vpi_remove_cb             | NO        |                                     |
-| vpi_scan                  | YES*      |                                     |
-| vpi_vprintf               | NO        |                                     |
+| VPI Definition            | Supported | Haskell API Function(s)               |
+| :---                      | :---:     | :---                                  |
+| vpi_chk_error             | YES       | simulationError, simulationErrorLevel |
+| vpi_compare_objects       | YES       | compareHandles                        |
+| vpi_control               | YES       | controlSimulator                      |
+| vpi_flush                 | NO        |                                       |
+| vpi_get                   | YES       | getProperty                           |
+| vpi_get_cb_info           | YES       | callbackInfo                          |
+| vpi_get_data              | NO        |                                       |
+| vpi_get_delays            | NO        |                                       |
+| vpi_get_str               | YES       | getProperty                           |
+| vpi_get_systf_info        | NO        |                                       |
+| vpi_get_time              | YES       | simulationTime                        |
+| vpi_get_userdata          | NO        |                                       |
+| vpi_get_value             | YES       | receiveValue, unsafeReceiveValue      |
+| vpi_get_vlog_info         | YES       | simulatorInfo                         |
+| vpi_handle                | YES       | childHandle                           |
+| vpi_handle_by_index       | YES       | childHandle                           |
+| vpi_handle_by_multi_index | YES       | childHandle                           |
+| vpi_handle_by_name        | YES       | childHandle                           |
+| vpi_handle_multi          | NO        |                                       |
+| vpi_iterate               | YES       | iterate, iterateAll                   |
+| vpi_mcd_close             | NO        |                                       |
+| vpi_mcd_flush             | NO        |                                       |
+| vpi_mcd_name              | NO        |                                       |
+| vpi_mcd_open              | NO        |                                       |
+| vpi_mcd_printf            | NO        |                                       |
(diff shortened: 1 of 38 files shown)


clash-ffi/src/Clash/FFI/VPI/Control.hs
```diff
@@ -0,0 +1,90 @@
...
+
+foreign import ccall "vpi_user.h vpi_control"
+  c_vpi_control_restart :: CInt -> CInt -> CInt -> CInt -> IO Bool
+
+controlSimulator :: forall o. Control -> SimCont o ()
+controlSimulator control =
+  liftOrThrow $
+    case control of
+      Stop d ->
+        c_vpi_control_end 66 (diagnosticLevelToCInt d)
```
vmchale, line 74: You might like c2hs for the constants here (66, 67, ...). You can automatically marshal `vpiStop`... etc. as a Haskell sum type.

https://github.com/vmchale/libarchive/blob/master/src/Codec/Archive/Types/Foreign.chs#L54

A lot of these constants are hard-coded and seem fickle! Unless hsc2hs/c2hs choke on the header files, I think they'd make this more robust + easier on your end.

alex-mckenna: I didn't use `c2hs` because I wanted more control over failures to read. For some function `decode :: CInt -> MyType` with `c2hs` an input with no output means you get an `ErrorCall`, but in `clash-ffi` operations fail with more specific exception types. The reason being that sometimes errors are unavoidable and depending on the particular error you may want to recover, and this is somewhat tedious if all your errors become `ErrorCall`

alex-mckenna: This module is a bit rough anyway, I should have made a type for these constants like I did with `Time` / `TimeType` (another instance where you have some type, but also want to refer to just it's tag as a separate type)

vmchale: You can still use hsc2hs! Something like `#{const vpiStop}`?

alex-mckenna: True, I could also just `#include <vpi_user.h>` and use the macros directly from Haskell. That's probably the easiest from where I am now, and the case expressions would read nicer, e.g.

```haskell
case foo of
  vpiThing -> HaskellThing
  ...

Comment (alex-mckenna) [reply clash-ffi/src/Clash/FFI/VPI/Control.hs:74]: So I gave this a try and it didn't work because just including the C file isn't allowed when you're not using `c2hs` or something similar. I think `c2hs` constants might be nice if it works, or `c2hs` enum if it's possible to also use the pre-processor around those definitions (since the enum members depend on things like the Verilog spec version etc.)

I'm not sure if I want to rewrite things to use `c2hs` at this point, since it seems like a fair bit of work to change all the types / decoding functions across... I guess that's a debate for when I move this out of draft

**Comment by alex-mckenna** [comment]
I'm marking this as _ready to review_. I haven't done the CI changes yet, but I expect there to be changes for the Haskell (and once those are in I can ask just @martijnbastiaan to check the CI changes). @vmchale, you've already expressed an interest (and have some Haskell FFI experience), so I've requested a review from you too

File: clash-ffi/src/Clash/FFI/VPI/Object.hs [hunk 2]
```diff
@@ -0,0 +1,689 @@
...
+
+{-
+NOTE [use of Coercible in public API]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+While VPI has an object model, objects are always represented internally as a
+`vpiHandle`, which is a pointer to the object provided by the simulator. This
+means in `clash-ffi`, the base object type is a pointer, and any more specific
+object types are newtypes around this base object.
+
+Rather than duplicate the API for each object, we take a more lenaint approach.
```
vmchale, line 160: I think this should be 'lenient'

clash-ffi/README.md
```diff
@@ -0,0 +1,107 @@
...
+| vpi_handle_by_name        | YES       | childHandle                           |
+| vpi_handle_multi          | NO        |                                       |
+| vpi_iterate               | YES       | iterate, iterateAll                   |
+| vpi_mcd_close             | NO        |                                       |
+| vpi_mcd_flush             | NO        |                                       |
+| vpi_mcd_name              | NO        |                                       |
+| vpi_mcd_open              | NO        |                                       |
+| vpi_mcd_printf            | NO        |                                       |
+| vpi_mcd_vprintf           | NO        |                                       |
+| vpi_printf                | YES       | simPutStr, simPutStrLn                |
```
vmchale, line 76: I think this has a trailing space!

vmchale (commented):
This looks unobjectionable, but I don't know how to test it! 

Also there's trailing lines in all your files.

clash-ffi/src/Clash/FFI/VPI/Object/Value/Parse.hs
```diff
@@ -0,0 +1,237 @@
...
+  str <- IO.liftIO (FFI.peekCString bin)
+  let is = [size - 1, size - 2 .. 0]
+
+  let go acc (i, x) =
+        case x of
+          '0' -> pure (replaceBit i 0 acc)
+          '1' -> pure (replaceBit i 1 acc)
+          'x' -> pure acc
+          'z' -> pure acc
+          _   -> Sim.throw (InvalidBitString BinStrFmt str callStack)
```
martijnbastiaan, line 74: `replaceBit` is a fairly expensive operation, I think you're better off shifting + reversing in the end. You could then use lookup tables ([like this](https://hackage.haskell.org/package/haskus-binary-1.0/docs/Haskus-Format-Binary-Bits-Reverse.html)) to be fairly speedy. A `TODO` suffices though, no need for premature optimization.

martijnbastiaan: This comment would apply to quite a number of parsing functions, so maybe a source note?

martijnbastiaan (approved):
Really impressive stuff, thank you @alex-mckenna. Like @vmchale I'm unsure how to properly review this; my guess is that we'll just have to use it and discover all the simulator/protocol idiosyncrasies that way.

martijnbastiaan:
PS: I'm not sure what your testing strategy would be. The parse functions can be tested pretty much on their own, but otherwise I'd just say plug the example you made into CI.

alex-mckenna:
> I'd just say plug the example you made into CI

I'm not a fan of this for two reasons:

  * an FFI application would need it's own stanza in the `.cabal` file, which means examples / tests would be released on hackage as though they were "normal" artifacts. I personally consider releasing such example / tutorial type things in this way as bad-style from a package maintenance POV
  * the "interesting" example I had relied on a model that could be simulated indefinitely, which means we would need to know when to kill the simulation, or the test runner would never stop

I wonder if it is a good idea to release  just the bindings as a standalone thing, or whether we want to have any/all cosim applications as artifacts of this package. It's easier to test an application using the bindings than the bindings themselves

martijnbastiaan:
Like discussed IRL, I'm unsure about the ergonomics of using this as it's currently structured.  Is the idea that every program wishing to use `clash-ffi` should define its own `foreign-library` (and compile everything in `clash-ffi` again), also exporting `ffiMain`?

If so, I wonder if we can:

 * Export the types and functions in a "normal" library
 * Create a single `ffiMain` in `clash-ffi` that uses [hint](https://hackage.haskell.org/package/hint) to load a "real" entry point, possibly passed in through environment variables. From the looks of it, this is similar to how cocotb operates.

Have you got any insight here?

martijnbastiaan:
Okay, gotcha. As discussed offline, right now we only now how to use `clash-ffi` from within itself (so to speak). Of course, we'd like to use it as a standalone library for as much as possible, but we simply don't know what that would look like. Hence, creating a normal `library` stanza and exporting modules from there (like I suggested) is foolish because we have no idea whether that would work. Working this out should be part of working out how to write a coroutine scheduler for it (`clash-cotest`).

martijnbastiaan:
Thanks for all the work on this!

PR #2172, opened 2022-04-19
add scanl1 to Clash prelude

vmchale:
This adds `scanl1` for vectors.

### DIFF
--- a/changelog/2022-04-19T12_26_48-05_00_add_scanl1
+++ b/changelog/2022-04-19T12_26_48-05_00_add_scanl1
@@ -1,0 +1,1 @@
+CHANGED: Before this release, `scanl1` was re-exported from the Haskell Prelude. Clash's Prelude now exports a `Vec` specialized version.
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -77,7 +77,7 @@ module Clash.Sized.Vector
     -- ** Specialized folds
   , dfold, dtfold, vfold
     -- * Prefix sums (scans)
-  , scanl, scanr, postscanl, postscanr
+  , scanl, scanl1, scanr, scanr1, postscanl, postscanr
   , mapAccumL, mapAccumR
     -- * Stencil computations
   , stencil1d, stencil2d
@@ -127,8 +127,9 @@ import Prelude                    hiding ((++), (!!), concat, concatMap, drop,
                                           foldl, foldl1, foldr, foldr1, head,
                                           init, iterate, last, length, map,
                                           repeat, replicate, reverse, scanl,
-                                          scanr, splitAt, tail, take, unzip,
-                                          unzip3, zip, zip3, zipWith, zipWith3)
+                                          scanl1, scanr, scanr1, splitAt, tail,
+                                          take, unzip, unzip3, zip, zip3, zipWith,
+                                          zipWith3)
 import qualified Data.String.Interpolate as I
 import qualified Prelude          as P
 import Test.QuickCheck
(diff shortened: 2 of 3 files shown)


vmchale:
btw what should I tag this branch with so that it works correctly?

changelog/2022-04-19T12_26_48-05_00_add_scanl1
```diff
@@ -0,0 +1 @@
+CHANGED: Add scanl1 for vectors to the prelude
```
martijnbastiaan, line 1:
```suggestion
CHANGED: Before this release, `scanl1` was re-exported from the Haskell Prelude. Clash's Prelude now exports a `Vec` specialized version.
```

Write changelogs with the user reading it in mind. (I know we haven't applied this mantra consistently in the past, but I always end up spending a bit of time rewriting changelogs upon release.)

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1111,6 +1112,11 @@ scanl f z xs = ws
     ws = z `Cons` zipWith (flip f) xs (init ws)
 {-# INLINE scanl #-}
 
+-- | 'scanl' with no seed value
+scanl1 :: KnownNat n => (a -> a -> a) -> Vec (n+1) a -> Vec (n+1) a
+scanl1 op v = scanl op (head v) (tail v)
```
martijnbastiaan, line 1117: Can't you do `(v :> vs)` or ``v `Cons` vs``? Or does that trip up the exhaustiveness checker?

martijnbastiaan (approved):
Overall LGTM

clash-prelude/src/Clash/Sized/Vector.hs
alex-mckenna: `Cons` should _never_ trip up the exhaustiveness check I think, the only reason `:>` does is because of nasty interaction between GADTs and pattern synonyms

alex-mckenna: Although maybe pre GHC 8.10 it might get stupid, there were some meaningful changes then. I always forget we support GHCs older than I use locally

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1111,6 +1112,11 @@ scanl f z xs = ws
     ws = z `Cons` zipWith (flip f) xs (init ws)
 {-# INLINE scanl #-}
 
+-- | 'scanl' with no seed value
+scanl1 :: KnownNat n => (a -> a -> a) -> Vec (n+1) a -> Vec (n+1) a
```
alex-mckenna, line 1116: I think if we add `scanl1` to `clash-prelude` we should also add `scanr1` for completeness

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1111,6 +1112,11 @@ scanl f z xs = ws
     ws = z `Cons` zipWith (flip f) xs (init ws)
 {-# INLINE scanl #-}
 
+-- | 'scanl' with no seed value
```
alex-mckenna, line 1115: I would improve this documentation. It doesn't need to be as complete as the ones with big descriptions and diagrams, but should have an example in a similar style to `zip`, i.e.

```
-- | 'zip' takes two vectors and returns a vector of corresponding pairs.
--
-- >>> zip (1:>2:>3:>4:>Nil) (4:>3:>2:>1:>Nil)
-- (1,4) :> (2,3) :> (3,2) :> (4,1) :> Nil
```

alex-mckenna: This way we at least have a `doctest` for the function as a very basic sanity-check

martijnbastiaan:
> btw what should I tag this branch with so that it works correctly?

Everything _should_ just work. But I think GitLab status reporting has been broken for a few days now, I'll look into it tomorrow morning. I think it's to do with the new-style GitHub tokens..

changelog/2022-04-19T12_26_48-05_00_add_scanl1
vmchale: Yes! Good point.

clash-prelude/src/Clash/Sized/Vector.hs
vmchale: Yes, fair enough.

clash-prelude/src/Clash/Sized/Vector.hs
vmchale: That works!

clash-prelude/src/Clash/Sized/Vector.hs
vmchale: Yes! I will add doctests.

alex-mckenna (approved):

martijnbastiaan:
CI has been fixed; I've squashed all your commits to make a force push that triggers CI.

martijnbastiaan (approved):

DigitalBrains1:
We usually also add a GitHub PR link or issue link to the Changelog, but it doesn't seem to really add interesting information here, so I suppose it's not a problem that it is not there.

martijnbastiaan:
I [wrote a script](https://github.com/clash-lang/clash-compiler/blob/073b01a720f37072f13b8d17bf9bdcfeb797d4a4/changelog/blame-pr.py) that acts like `git blame`, but instead it shows the PR. But yeah, would be handy-dandy if it was already in the changelog of course.

PR #2174, opened 2022-04-21
Disallow implicit nets in (System)Verilog

alex-mckenna:
Verilog allows nets to be implicitly defined in a module, assuming
they are of type `wire` by default. However, for Clash-generated
Verilog it should never be the case that a net is implicitly defined
in a design. We can disallow this by setting the default net type
to none.

According to IEEE 1364-2005:

    When the `default_nettype is set to none, all nets shall be
    explicitly declared. If a net is not explicitly declared,
    an error is generated.

This seems a much more sensible default for Clash.

## Still TODO:


martijnbastiaan (approved):
Agreed, this seems much more sensible.

PR #2177, opened 2022-04-21
add scanrPar, scanlPar

vmchale:
This adds a low-depth scan, see [Conal Elliott's paper](https://dl.acm.org/doi/10.1145/3110251) for more. This uses a different approach, however - Haskell's "dependent types" rather than type families. 

I added a new module because `.hs-boot` files don't allow pattern synonyms. The alternative would be to have two parts of `Clash.Sized.Vector` to avoid mutual module dependencies, which would be a larger diff.

## Still TODO:


clash-prelude/src/Clash/Sized/Vector/Par.hs
```diff
@@ -0,0 +1,85 @@
...
+scanrPar op v = scanrInductiveRTree op (v2t v)
+{-# INLINE scanrPar #-}
+
+scanlInductiveRTree
+  :: forall a n. KnownNat n
+  => (a -> a -> a)
+  -> RTree n a
+  -> Vec (2^n) a
+scanlInductiveRTree op tr =
+  -- I have to use sameNat and compareSNat both; otherwise it doesn't compile
```
rowanG077, line 61: This comment is true but doesn't really add why it doesn't compile. Just write something like: "sameNat and compareSNat are both required because type checker doesn't know `n <= 0 <=> n ~ 0`"

clash-prelude/src/Clash/Sized/Vector/Par.hs
```diff
@@ -58,7 +58,8 @@ scanlInductiveRTree
   -> RTree n a
   -> Vec (2^n) a
 scanlInductiveRTree op tr =
-  -- I have to use sameNat and compareSNat both; otherwise it doesn't compile
+  -- I have to use sameNat and compareSNat both; the proofs that n>=0 and n~0
```
alex-mckenna, line 61: To be pedantic, with the way you're using `compareSNat` you end up with a proof for `1 <= n`, instead of a proof for `n >= 0` since `SNatGT` has the type

```haskell
SNatGT :: forall a b. (b + 1) <= a => SNatLE a b
```

If you did want a proof `0 <= n` you would have to swap the arguments to `compareSNat`

vmchale: Corrected!

DigitalBrains1 (changes requested):
Thanks for the nice piece of engineering!

I think it would be better if these functions were added to `Clash.Sized.RTree` (based on `RTree`s then, not `Vec`s) instead of creating a new small module specially for them. Is this possible?

clash-prelude/src/Clash/Sized/Vector/Par.hs
```diff
@@ -0,0 +1,86 @@
...
+--
+-- >>> scanlPar (+) (1 :> 2 :> 3 :> 4 :> Nil)
+-- 1 :> 3 :> 6 :> 10 :> Nil
+scanlPar :: KnownNat n
+         => (a -> a -> a) -- ^ Must be commutative
+         -> Vec (2^n) a -> Vec (2^n) a
+scanlPar op v = scanlInductiveRTree op (v2t v)
+{-# INLINE scanlPar #-}
+
+-- | Low-depth (right) scan
```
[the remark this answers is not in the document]
DigitalBrains1: Thank you GitHub for changing the order for some reason. I'm trying to say that just like `scanlPar` this needs much more fish. Haddock to be precise.

clash-prelude/src/Clash/Sized/Vector/Par.hs
```diff
@@ -0,0 +1,86 @@
...
+  -> Vec (2^n) a
+scanrInductiveRTree op tr =
+  -- I have to use sameNat and compareSNat both; otherwise it doesn't compile
+  case (sameNat (Proxy @n) (Proxy @0), compareSNat (SNat @n) (SNat @0), tr) of
+    (Just Refl, _, LR x) -> x :> Nil
+    (_, SNatGT, BR x y) ->
+        let
+          x' = scanrInductiveRTree op x
+          y' = scanrInductiveRTree op y
+          l = y' !! (0::Int)
```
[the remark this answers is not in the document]
vmchale: It can't infer that `2^n` is at least 1, I guess because there's no `n>=0` constraint on type-level integers.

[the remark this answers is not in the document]
rowanG077: Yeah you can use `leToPlus` I tried this before a few days ago.

clash-prelude/src/Clash/Sized/Vector/Par.hs
```diff
@@ -0,0 +1,86 @@
...
+import Clash.Sized.RTree (RTree (..), v2t)
+import Data.Type.Equality ((:~:) (..))
+import Data.Proxy (Proxy (..))
+
+{- $setup
+>>> :m -Prelude
+>>> import Clash.Prelude
+-}
+
+-- | Low-depth (left) scan
```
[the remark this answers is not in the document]
vmchale: > Are you sure commutativity is needed, btw? At a glance I don't see any argument swaps.

Sorry, my mistake - associativity is what I needed.

martijnbastiaan:
I appreciate the effort to include the SVG, but to include a proprietary, paid-for, macOS only application file in a BSD2 project is not the way to go IMO. 

The diagram doesn't look to complicated, perhaps you could draw it in draw.io? Most of use VSCode nowadays, which would enable everyone to just edit it in [their IDEs](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio). It also exports to  PNG/SVG.

DigitalBrains1:
Martijn, did you see that there are already 31 other `.graffle` files in [clash-prelude/doc](https://github.com/clash-lang/clash-compiler/tree/ee87e5b278599a0dcdc5b60ba8c4623e87d510d7/clash-prelude/doc)? I've never looked at it before but it seems we already do this for 8 years, starting with 082426886.

martijnbastiaan:
Oh wow. I'd didn't know.

martijnbastiaan:
I stand by my comment _but_ if that's what we have there's no problem merging this PR.

alex-mckenna (approved):
On the whole looks good. Thinking out loud, I wonder if we might want to also expose `scanlInductiveRTree` and `scanrInductiveRTree` as scans on trees (since you can really scan anything that's foldable if you believe). If you think that's a good idea they should probably be renamed `tscanl` and `tscanr` to better fit with the naming convention of other `RTree` functions

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +544,63 @@ lazyT :: KnownNat d
...
+--
+-- >>> scanlPar (+) (1 :> 2 :> 3 :> 4 :> Nil)
+-- 1 :> 3 :> 6 :> 10 :> Nil
+--
+-- The operator must be associative.
+--
+-- <<doc/scanlPar.svg>>
+scanlPar :: KnownNat n
+         => (a -> a -> a) -- ^ Must be associative
+         -> Vec (2^n) a -> Vec (2^n) a
```
martijnbastiaan, line 558:
```suggestion
scanlPar :: 
  KnownNat n =>
  -- | Must be associative
  (a -> a -> a) ->
  Vec (2^n) a ->
  Vec (2^n) a
```

We've switched to [Ormolu's style](https://yairchu.github.io/posts/silly-haskell-formatting) oone or two releases back.

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +544,63 @@ lazyT :: KnownNat d
...
+
+-- | Low-depth (right) scan
+--
+-- >>> scanrPar (+) (1 :> 2 :> 3 :> 4 :> Nil)
+-- 10 :> 9 :> 7 :> 4 :> Nil
+--
+-- The operator must be associative.
+scanrPar :: KnownNat n
+         => (a -> a -> a) -- ^ Must be associative
+         -> Vec (2^n) a -> Vec (2^n) a
```
martijnbastiaan, line 570:
```suggestion
scanrPar :: 
  KnownNat n =>
   -- | Must be associative
  (a -> a -> a) ->
  Vec (2^n) a ->
  Vec (2^n) a
```

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +544,63 @@ lazyT :: KnownNat d
...
+scanrPar op v = scanrInductiveRTree op (v2t v)
+{-# INLINE scanrPar #-}
+
+scanlInductiveRTree
+  :: forall a n. KnownNat n
+  => (a -> a -> a)
+  -> RTree n a
+  -> Vec (2^n) a
```
martijnbastiaan, line 578:
```suggestion
scanlInductiveRTree ::
  forall a n.
  KnownNat n =>
  (a -> a -> a) ->
  RTree n a ->
  Vec (2^n) a
```

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +544,63 @@ lazyT :: KnownNat d
...
+          x' = scanlInductiveRTree op x
+          y' = scanlInductiveRTree op y
+          l = x' !! (length x'-1) -- 'last' doesn't work here
+        in x' ++ map (l `op`) y'
+
+scanrInductiveRTree
+  :: forall a n. KnownNat n
+  => (a -> a -> a)
+  -> RTree n a
+  -> Vec (2^n) a
```
martijnbastiaan, line 595:
```suggestion
scanrInductiveRTree ::
  forall a n.
  KnownNat n =>
  (a -> a -> a) ->
  RTree n a ->
  Vec (2^n) a
```

clash-prelude/src/Clash/Sized/RTree.hs
martijnbastiaan: It fits one line btw, I wouldn't mind losing the argument doc as it's explained in the function doc anyway.

martijnbastiaan (changes requested):
I think you still need to update `Clash.Prelude` and `Clash.Explicit.Prelude` to have them exported there right?

DigitalBrains1:
> I think you still need to update `Clash.Prelude` and `Clash.Explicit.Prelude` to have them exported there right?

No, they both export `RTree` fully.

Furthermore, could you change the type to

```
scanlPar :: (a -> a -> a) -> RTree n a -> RTree n a
```
? I suppose that overlaps with Alex's suggestion and they would be better named `tscanl` and `tscanr`. Or `tscanl1` if you want to emphasise they don't get the zero element separately. Incidentally, this might also fix your `head` and `last` issue perhaps?

I'm also not satisfied with the documentation yet. I appreciate the work that went into the nice picture (though I would have named `op` `f` to match the existing graphics), and it's a great addition, but the text still needs improvements, IMO. I can put some work into this myself tomorrow.

DigitalBrains1:
I see we don't have `head` and `last` for `RTree`. It seemed such an obvious feature I hadn't checked. Maybe add those? :-)

vmchale:
I don't think it's that productive to generalize `tscanl` and `tscanr` - they work for vectors quite sensibly but not sure about `Foldable` in general.

(I still need documentation for `tscanl` and `tscanr`).

alex-mckenna:
FWIW, when I mentioned `Foldable` it was in reference to "you can scan over any foldable thing _to produce a `Vec`_" (the part in italics I was maybe too implicit about). Producing an `RTree` instead like @DigitalBrains1 suggested and exposing you have now is as generalised as I was aiming for, I just wanted the `scan{l,r}` on `RTree` to be exported too.

Unrelated: I see a merge from this branch into itself in the commit history. Since we want to keep exposing the GADT constructors for `RTree` in a separate commit you should go back and tidy up the history here

DigitalBrains1:
> Unrelated: I see a merge from this branch into itself in the commit history. Since we want to keep exposing the GADT constructors for `RTree` in a separate commit you should go back and tidy up the history here

I think it needs to be squashed anyway. It might change several things, but I feel it can still fit in one commit, don't you?

vmchale:
I think squashing everything is the way to go.

vmchale:
@DigitalBrains1 yes! I expose `tscanr`, `tscanl` now. It's a little easier in some ways.

vmchale:
I think `tscanr` and `tscanl` are fine since there's so need to differentiate `tscanr1` for `RTree`s - we only have `tfold` and that's fine!

martijnbastiaan (approved):
Agreed with @DigitalBrains1 on the documentation, otherwise LGTM!

DigitalBrains1:
Should we change the name of the GADT constructors? It's kinda odd to export things named `X_`. I know we're trying to move away from primes, but maybe `LR'` and `BR'` are good names still. I'm not too enthousiastic about `LR0` and `BR0`. `LR#` and `BR#` are also options. Let's bikeshed some more!

rowanG077:
Why not just `Leaf` and `Branch`

DigitalBrains1:
Since these are part of `Prelude` I think it is good practice to retain something of the type name in the constructor names (to prevent confusion with other types of trees), so I'd prefer `RLeaf` and `RBranch`. Although frankly I don't know why this module is named `RTree`, isn't an R-tree a different structure?

Also, by changing the name this has gotten to be so "invasive" that I think it's best it gets its own commit like Alex also originally suggested. I'll work on the documentation this afternoon and add a changelog file too.

alex-mckenna:
> I'd prefer `RLeaf` and `RBranch`

To play devil's advocate: we didn't name the constructors of `Vec` `VNil` and `VCons`. I think `Leaf` and `Branch` are fine.

> Although frankly I don't know why this module is named RTree, isn't an R-tree a different structure?

An r-tree is a different structure (spatial). If I look at the docs I see `RTree` is "a perfect depth binary tree", so I can only assume the "R" stands for "Rerfect"

christiaanb:
> Since these are part of `Prelude` I think it is good practice to retain something of the type name in the constructor names (to prevent confusion with other types of trees), so I'd prefer `RLeaf` and `RBranch`. Although frankly I don't know why this module is named `RTree`, isn't an R-tree a different structure?
> 
> Also, by changing the name this has gotten to be so "invasive" that I think it's best it gets its own commit like Alex also originally suggested. I'll work on the documentation this afternoon and add a changelog file too.

I called it an RTree because of http://conal.net/papers/generic-parallel-functional/generic-parallel-functional.pdf#page=7, so in that paper we have the following definition
```
newtype Identity a = Identity a
newtype Compose g f a = Compose (g (f a))

type family RPow h n where
  RPow h 0 = Identity
  RPow h n = Compose h (RPow h (n-1))

type family LPow h n where
  LPow h 0 = Identity
  LPow h 0 = Compose (LPow h (n-1)) h
```
Where the `R` indicates we associate on the right, and the `L` indicates we associate on the left. So when we have a binary tree we can say:
```
data Pair a = Pair a a
type RTree n a = RPow Pair n a
```
Let's expand that for a concrete depth to get a better sense of what's going on:
```
RTree 2 a = RPow Pair 2 a
RTree 2 a = Compose Pair (RPow Pair 1) a
RTree 2 a = Compose Pair (Compose Pair (RPow Pair 0)) a
RTree 2 a = Compose Pair (Compose Pair Identity) a
RTree 2 a = Pair (Pair (Identity a))
```
So that's how I ended up with the `RTree` data type:
```
data RTree n a where
  LR_ :: a -> RTree 0 a
  BR_ :: RTree n a -> RTree n a -> RTree (n+1) a
```
Perhaps it should've been called `RBin` though, like in http://conal.net/papers/generic-parallel-functional/generic-parallel-functional.pdf#page=8

vmchale:
I've used `RBranch` and `RLeaf` in case people already have their own stuff (since this is in Prelude)

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +544,93 @@ lazyT :: KnownNat d
...
+tscanr ::
+  forall a n.
+  KnownNat n =>
+  (a -> a -> a) ->
+  RTree n a ->
+  RTree n a
+tscanr op tr =
+  case tr of
+    (RLeaf x) -> LR x
+    (RBranch x y) ->
```
christiaanb, line 631:
```suggestion
    RLeaf x -> LR x
    RBranch x y ->
```

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +544,93 @@ lazyT :: KnownNat d
       => RTree d a
       -> RTree d a
 lazyT = tzipWith (flip const) (trepeat ())
+
+-- | Low-depth (left) scan, see 'Clash.Sized.Vector.scanl1'. The generated
+-- circuit will use more gates but can be clocked faster.
```
christiaanb, line 549:
```suggestion
-- | Low-depth (left) scan, see 'Clash.Sized.Vector.scanl1'. The generated
-- circuit will do O(2*n) operations, but has a logic depth of O(log n);
-- i.e. it will use more gates but can be clocked faster.
```

christiaanb (approved):

DigitalBrains1:
What is the reason for retaining `scanlPar` and `scanrPar` when we have `tscanl` and `tscanr`? We don't provide `Vec` wrappers for the other functions in `Clash.Sized.RTree`. Why not have people use `v2t` and `t2v` themselves?

vmchale:
@DigitalBrains1 I could remove them I suppose - but they are what is most sensibly useful! Since vectors naturally lend themselves to an idea of scanning.

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -105,26 +112,26 @@ let populationCount' :: (KnownNat k, KnownNat (2^k)) => BitVector (2^k) -> Index
 -- * Only has elements at the leaf of the tree
 -- * A tree of depth /d/ has /2^d/ elements.
 data RTree :: Nat -> Type -> Type where
-  LR_ :: a -> RTree 0 a
-  BR_ :: RTree d a -> RTree d a -> RTree (d+1) a
+  RLeaf :: a -> RTree 0 a
+  RBranch :: RTree d a -> RTree d a -> RTree (d+1) a
```
martijnbastiaan, line 116: This warrants a `CHANGED` changelog entry, right?

christiaanb: Mmm… I think it’d sooner be ADDED rather than CHANGED, since the constructors weren’t exported before.

DigitalBrains1: Yes, I'm on it :-)

DigitalBrains1:
I've pushed additional commits on the mass of commits here already, but I've also prepared a version that I think is ready to be committed with a merge commit (don't squash or rebase!) here: [parallel-scan-peter](https://github.com/clash-lang/clash-compiler/tree/parallel-scan-peter). You could force-push that over this branch if you want. I'm not force-pushing here without discussing.

I'd like it if someone reviewed my documentation ([full changes here](https://github.com/clash-lang/clash-compiler/compare/c2ec97aaca498defe9b1df96523d45d57e1bd13b..parallel-scan)). I kinda feel like I had more text, but my dull head can't locate it in there.

clash-prelude/src/Clash/Sized/RTree.hs
vmchale: Ok, I have added this.

clash-prelude/src/Clash/Sized/RTree.hs
```diff
@@ -537,3 +545,125 @@ lazyT :: KnownNat d
       => RTree d a
       -> RTree d a
 lazyT = tzipWith (flip const) (trepeat ())
+-- | Extract the first element of an RTree
```
martijnbastiaan, line 548: I can't let it go :P. Enter pls.

martijnbastiaan (approved):
Great docs. Thanks @DigitalBrains1, and thanks @vmchale for the implementations :).

DigitalBrains1:
Actually, there was a real deficiency in the docs that I wrote with a rather dull brain. Before this sentence:

> Circuit size is \(\mathcal{O}(n \cdot \log n)\), but logic depth is \(\mathcal{O}(\log n)\).

I added

> When \(n\) is the number of elements, [...]

as this is not obvious at all, I could have meant tree depth, but I didn't.

While I needed to amend it anyway, I also changed "Extract the first element of an RTree" to "Extract th first element of a tree" to be more in line with the rest, same for `tlast`.

clash-prelude/src/Clash/Sized/RTree.hs
DigitalBrains1: Thanks for spotting that! I reordered the stuff a bit and introduced that omission. Or omitted that introduction, your pick.

PR #2181, opened 2022-04-28
Render sequential processes in VHDL

alex-mckenna:
The sequential processes represented by the `Seq` netlist type were
added for `SimIO`, and are only available in the (System)Verilog
backends. To generalize the `SimIO` machinery for producing arbitrary
Clash as seuqential HDL, we need to be able to support the features
in VHDL as well.

## Still TODO:


alex-mckenna:
There's a test in here but it deliberately fails. The reason being that Clash doesn't generate `variable` in VHDL (only `signal`) so the process with calls to `Clash.Explicit.SimIO.{readReg,writeReg}` doesn't behave the same as the Verilog output. I need to go back and make another PR to add support for `variable` and then rebase this later.

Other than that, this PR is in a reviewable state.

alex-mckenna:
Note (for anyone wondering why this is draft again): This PR _can_ produce sequential code in VHDL, but it has some problems that make it kind of pointless to merge. These mostly come from VHDL having restrictions on where signals / variables can be defined, and this PR doesn't respect that. When the other related netlist PRs are merged, it should be possible to just extract the changes to the VHDL backend and make any remaining fixes

martijnbastiaan:
Closing due to inactivity, see https://github.com/clash-lang/clash-compiler/issues/2751

PR #2182, opened 2022-05-01
Add Xilinx dual clock FIFO primitive

martijnbastiaan:
This PR adds a Xilinx dual clock FIFO implementation to `clash-cores`. 

## Comments for implementers:

  * There's no observable difference between the different "implementation mode"s AFAIK. Feel free to ignore for the Haskell model.
  * We don't care too much about having the Haskell model correspond 1:1 to the RTL for our client's project. Please de-prioritize.

<details>
  <summary>How to test</summary>

The easiest way to create a quick test bench (IMO) is to instantiate a new Clash project with `stack new fifotest clash-lang/simple`, add 

```
../clash-compiler/clash-cores/clash-cores.cabal
../clash-compiler/clash-prelude/clash-prelude.cabal
../clash-compiler/clash-lib/clash-lib.cabal
../clash-compiler/clash-ghc/clash-ghc.cabal
```

to `cabal.project` and run `cabal run clash -- Test.hs --vhdl -fclash-clear -fclash-no-cache && cat  vhdl/Test.topEntity/dcfifo*` where `Test.hs` is something like:

```haskell
module Test where

import Clash.Prelude

import Clash.Cores.Xilinx.DcFifo.Explicit

type WriteDom = System
type ReadDom = System

topEntity ::
  Clock WriteDom -> Clock ReadDom -> Reset ReadDom ->

  -- | Write data
  Signal WriteDom (BitVector 32) ->
  -- | Write enable
  Signal WriteDom Bool ->
  -- | Read enable
  Signal ReadDom Bool ->
  -- |
  ( Signal WriteDom ResetBusy
  , Signal WriteDom Full
  , Signal WriteDom (DataCount 5)

  , Signal ReadDom ResetBusy
  , Signal ReadDom Empty
  , Signal ReadDom (DataCount 5)
  , Signal ReadDom (BitVector 32)
  )
topEntity wClk rClk rRst = dcFifo defConfig{dcReadDataCount=False} wClk rClk rRst
```
</details>

## Questions for reviewers:

 * Dual clock FIFOs and single clock FIFOs differ enough in terms of signals/features that I've decided to preemptively namespace the dual clock FIFOs. All function names are still prefixed with `dc` though. Remove?
 * I've instantiated a very specific version of `fifo_generator` here (`13.2`) - pretty much linked to Vivado 2021.2. How can we deal with versioning?

Resources:

 * [FIFO Generator v13.2](https://docs.xilinx.com/v/u/en-US/pg057-fifo-generator) ([PDF](https://docs.xilinx.com/api/khub/documents/Ds0JjAYlvIFpRMxJMOrbRw/content?Ft-Calling-App=ft%2Fturnkey-portal&Ft-Calling-App-Version=3.11.23&download=true))

## Still TODO:

  - [x] Add Haskell model (@vmchale)
  - [x] Add Haskell tests (@vmchale)
  - [ ] Write documentation (@vmchale, @martijnbastiaan)
  - [ ] Write `checkDcConfig :: DcConfig -> Maybe Error` that returns a (human-readable) error message for misconfigurations.
  - [x] Make primitive return booleans (we probably want to implement [`boolFromBit`](https://github.com/clash-lang/clash-compiler/blob/ea114d8edd6a110f72d148203b9db2454cae8f37/clash-lib/src/Clash/Primitives/DSL.hs#L396-L405) analogous to [`boolToBit`](https://github.com/clash-lang/clash-compiler/blob/ea114d8edd6a110f72d148203b9db2454cae8f37/clash-lib/src/Clash/Primitives/DSL.hs#L357-L375), right now it can only be used in VHDL) (@vmchale)
  - [x] Add RTL tests (@vmchale, @DigitalBrains1?)
  - [x] ~~Made primitive a [multi result primitive](https://github.com/clash-lang/clash-compiler/pull/1560) for prettier HDL (@vmchale / @martijnbastiaan)~~ Attempting this triggers a bug, I'll post an issue.
  - [ ] Make `Clash.Cores.Xilinx.Floating` use functions in `Clash.Cores.Xilinx.Common`?
  - [x] ~~Write a changelog entry (see changelog/README.md)~~ `clash-cores` isn't released on Hackage


clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,258 @@
...
+  -- https://github.com/clash-lang/clash-compiler/blob/ea114d8edd6a110f72d148203b9db2454cae8f37/clash-prelude/src/Clash/Explicit/BlockRam.hs#L1276-L1280
+
+  go ::
+    Signal read Bool ->
+    Signal write Bool ->
+
+    FifoState n ->
+    Signal write Bool ->
+    Signal read Bool ->
+    Signal write (BitVector n) ->
```
martijnbastiaan, line 188: Perhaps:

```suggestion
    FifoState n ->
    Signal read Bool -> -- reset
    Signal read Bool -> -- read enabled
    Signal write Bool -> -- reset
    Signal write Bool -> -- write enabled
    Signal write (BitVector n) -> -- write data
```

Just to cluster the read/write domains. This makes it easier to see whether the pattern matches in `go` are correct.

martijnbastiaan: Maybe even:


```suggestion
    FifoState n ->
    Signal read Bool -> -- reset
    Signal read Bool -> -- read enabled
    
    Signal write Bool -> -- reset
    Signal write (Maybe (BitVector n)) -> -- write data
```

martijnbastiaan: Like the `Bit`/`Bool` thing, I wish we had a good way to deconstruct/construct `Maybe`s in the blackboxes. It would save us a little bit of computation on the Haskell side + annoying wrapper functions + `deepErrorX` everywhere.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,258 @@
...
+      (wRstBusy, preFull, preWCnt, rRstBusy, fifoEmpty, rCnt, rData) = go rstR rstWNext (FifoState mempty (rt-tWr)) wEna rEna wData
+
+  goWrite rstR (_ :- rstW) (FifoState q rt) wEna rEna wData =
+    (0 :- wRstBusy, full, wCnt, rRstBusy, fifoEmpty, rCnt, rData)
+    where
+      (wRstBusy, preFull, preWCnt, rRstBusy, fifoEmpty, rCnt, rData) = go rstR rstW (FifoState q' (rt-tWr)) wEna' rEna wData'
+      wCnt = sDepth q :- preWCnt
+      full = (if Seq.length q == rD then high else low) :- preFull
+      (en :- wEna') = wEna
+      (wDatum :- wData') = wData
```
martijnbastiaan, line 216: Keep it consistent:

```suggestion
      (wEna :- wEnas1) = wEnas0
      (wDat :- wDats1) = wDats0
```

Also, we avoid the usage of `'` in the Clash code base. Numbering is much clearer in case of multiple `'`.

martijnbastiaan: You can also move these to line `209` to avoid clutter

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,258 @@
...
+      q' =
+        if Seq.length q + 1 <= rD && en
+          then wDatum Seq.<| q
+          else q
+
+  sDepth = fromIntegral . Seq.length
+
+  goRead (rstR :- rstRNext) rstW (FifoState q rt) wEna rEna wData =
+    (wRstBusy, full, wCnt, (if rstR then 1 else 0) :- rRstBusy, fifoEmpty, rCnt, rData)
```
martijnbastiaan, line 225: If you'd use `bitToBool` in the blackbox, you could do:

```suggestion
    (wRstBusy, full, wCnt, rstR :- rRstBusy, fifoEmpty, rCnt, rData)
```

which also makes me think we can just move it out of the return type entirely.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,258 @@
...
+    where
+      rCnt = sDepth q :- preRCnt
+      fifoEmpty = (if Seq.length q == 0 then high else low) :- preEmpty
+      rData = nextData :- preRData
+      (wRstBusy, full, wCnt, rRstBusy, preEmpty, preRCnt, preRData) = go rstRNext rstW (FifoState q' (rt+tR)) wEna rEna' wData
+      (en :- rEna') = rEna
+      (q', nextData) =
+        if en && not rstR
```
martijnbastiaan, line 233: It's good practice IMO to handle the reset part -given that it overrides _everything_- of a function in it's own body. So:

```
  goRead (True :- rstRNext) rstW (FifoState q rt) wEna rEna wData =
    ..
```

This makes it more obvious what the IP is doing during reset, and there's no chance of accidentally forgetting something (like we did before :-)).

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,258 @@
...
+          then
+            case Seq.viewr q of
+              Seq.EmptyR -> (q, deepErrorX "FIFO empty")
+              qData Seq.:> qDatum -> (qData, qDatum)
+          else (q, deepErrorX "Enable off or resetting")
+
+  start :: FifoState n
+  start = FifoState Seq.empty 0
```
martijnbastiaan, line 241: usually called something like `initState`

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -0,0 +1,92 @@
...
+  [a] ->
+  -- | Stalls
+  [a] ->
+  [a]
+intersperseStalls (d:ds) s = d:s P.++ intersperseStalls ds s
+intersperseStalls [] _ = []
+
+prop_fifo :: Property
+prop_fifo = property $ do
+  xs <- fmap Just <$> forAll (replicateM 10 genData)
```
martijnbastiaan, line 36: As discussed on Slack (and for completeness here): it's better to start with:

```
n <- generate number of elements to produce bounded by [0, 10*size of fifo]
```

this makes sure you test overflow conditions.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -0,0 +1,92 @@
...
+  stallRead <- forAll (Gen.maybe (Gen.int (Range.linear 0 10)))
+  stallWrite <- forAll (Gen.maybe (Gen.int (Range.linear 0 10)))
```
martijnbastiaan, line 38: To maximize the chaos, you want to generate one int per driven element. By doing this you generate `n` cycles of stall for every element driven. Most programming errors occur around things like bursts of data, followed by nothing, etc.

martijnbastiaan: It's probably easiest to zip these generated ints with the elements you're driving. In that case you can simply count down in your drive/sample functions.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -0,0 +1,92 @@
...
+
+takeState ::
+  (Bool, [Bool]) ->
+  (ResetBusy, Empty, DataCount depth, BitVector n) ->
+  ((Bool, [Bool]), (Maybe (BitVector n), Bool))
+takeState (_,[]) _ = error "Unreachable."
+takeState (_, _:stalls) (1, _, _, _) = ((False, stalls), (Nothing, False))
+takeState (readLastCycle, True:stalls) (_, _, _, d) = ((False, stalls), (nextData, False))
+  where
+    nextData = if readLastCycle then Just d else Nothing
```
martijnbastiaan, line 59:
```suggestion
    nextData = readLastCycle `orNothing` d
```

I see this pattern a lot, so I wrote a little helper function for it. You can decide for yourself if you want to to use it :)

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -0,0 +1,92 @@
...
+  [a]
+intersperseStalls (d:ds) s = d:s P.++ intersperseStalls ds s
+intersperseStalls [] _ = []
+
+prop_fifo :: Property
+prop_fifo = property $ do
+  xs <- fmap Just <$> forAll (replicateM 10 genData)
+  stallRead <- forAll (Gen.maybe (Gen.int (Range.linear 0 10)))
+  stallWrite <- forAll (Gen.maybe (Gen.int (Range.linear 0 10)))
+  let iWrites = case stallRead of
```
martijnbastiaan, line 39:
```suggestion
  let iWrites = case stallWrite of
```
?

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -0,0 +1,92 @@
...
+        Nothing -> []
+        Just i -> P.replicate i Nothing
+  let readStalls = case stallWrite of
```
martijnbastiaan, line 42:
```suggestion
  let readStalls = case stallRead of
```
?

martijnbastiaan (commented):
Here's some inspiration for a dual clock test bench:

https://github.com/clash-lang/clash-compiler/blob/master/tests/shouldwork/Signal/DualBlockRam0.hs

Given that we don't have a cycle accurate Haskell model, you'd probably want to build a circuit similar to the one you've got in the unittests, and test your assumptions in HDL.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: That would be nice! But would it be a lot of work?

DigitalBrains1:
> I've instantiated a very specific version of fifo_generator here (13.2) - pretty much linked to Vivado 2021.2. How can we deal with versioning?

Looking at the revision history in the product guide, 13.2 has been the version since 2017-10-04. I doubt you can reasonably deal with changes to the FIFO core; that seems to quickly lead to separate Clash functions or modules for separate Vivado versions. How would you deal with a new configuration option, new signals in Clash? One Vivado version can generally generate only a single version of bundled IP; it seems simplest to just pin a version of `clash-cores` to a version range of Vivado. We'd have to watch out that we don't generate one IP for one version and the other IP for another version.

In general, I suggest tracking the latest version of Vivado available for all platforms. If people really need an old version of Vivado, it might be more of a hassle for them.

Vivado can upgrade an older version of IP to a (reasonably-identically configured?) new version, though. So if Vivado switches to 13.3, people can still synthesise by introducing another step in their workflow: using Vivado to upgrade the IP. It [seems](https://support.xilinx.com/s/question/0D52E00006iHkM7SAK/how-to-upgrade-ip-in-a-project-tcl-mode?language=en_US) they just need to add this to their Tcl:
```
upgrade_ip [get_ips *]
```

https://www.xilinx.com/products/intellectual-property/vivado-ip-versioning.html

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: Yes! I think I have it working.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
vmchale: Ok, I can trigger an overflow! 

I will have separate tests in such cases I think.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
vmchale: Right! I have variable stalls now.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
vmchale: Right now the test I have is simple - it just tests that everything that is read out is in order (and present in the input) - is that enough?

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+  , dcImplementation = Independent_Clocks_Block_RAM
+  , dcReadMode = Standard_FIFO
+  , dcReadDataCount = True
+  , dcWriteDataCount = True
+  }
+
+data FifoState depth n = FifoState
+  { hsQueue      :: Vec (2^depth) (Maybe (BitVector n))
+  , relativeTime :: Int
+  } deriving Show
```
martijnbastiaan, line 82: I think we had a slight miscommunication @vmchale :-). The types used in `dcFifo#` are good as they are: they will never be seen by Clash because of the BlackBox implementation. (See the ANN pragmas.) That means the internals of the FIFO can be implemented any way you like: `Seq` is a very good choice for a data structure as it has great big O complexities for appending/prepending/popping. Using `Vec` would be a step backwards here! Luckily you kept the commits separate, so it should be easy to revert.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -0,0 +1,139 @@
...
+-- | Generated stalls are small enough that we don't expect any data loss
+prop_noloss :: Property
+prop_noloss = property $ do
+  (xs, wrIn, rdStalls) <- forAll $ genScenario 12 12
+  throughFifo wrIn rdStalls === xs
+
+genData :: (KnownNat n, MonadGen m) => m (BitVector n)
+genData = Gen.enumBounded
+
+takeState ::
```
martijnbastiaan, line 86: What I was talking about is making a synthesizable version of this (and its counterpart), so we could build a synthesizable test. Randomization in HDL is hard, so I'd probably have the type signature read something like:

```haskell
type ReadLastCycle = Bool
type Stall = Bool
type ExpectedToRead n = BitVector n
type UnexpectedRead = Bool

fifoSampler ::
  Clock dom -> Reset dom -> Enable dom ->
  -- | Stall circuit? [note: probably coming from an LSFR, see `Clash.Examples`]
  Signal dom Stall ->
  -- | Signals from FIFO
  Signal dom (ResetBusy, Empty, DataCount depth, BitVector n) ->
  -- | Maybe output read from FIFO
  Signal dom (Maybe (BitVector n))
fifoSampler clk rst ena stalls inps =
  mealyB clk rst ena go False (stalls, inps)
 where
  go :: 
    ReadLastCycle ->
    (Stall, (ResetBusy, Empty, DataCount depth, BitVector n)) ->
    (ReadLastCycle, Maybe BitVector n)
  go = _
```

On the feed side you could generate a sequence to write to the FIFO (0, 1, 2..) so the type signature could look something like:

```haskell
type ReadLastCycle = Bool
type Stall = Bool
type ExpectedToRead n = BitVector n
type UnexpectedRead = Bool

fifoDriver ::
  Clock dom -> Reset dom -> Enable dom ->
  -- | Stall circuit? [note: probably coming from an LSFR, see `Clash.Examples`]
  Signal dom Stall ->
  -- | Signals from FIFO
  Signal dom (ResetBusy, Full, DataCount depth) ->
  -- | Maybe write input to FIFO
  Signal dom (Maybe (BitVector n))
fifoDriver clk rst ena stalls inps =
  mealyB clk rst ena go 0 (stalls, inps)
 where
  go :: 
    BitVector n ->
    (Stall, (ResetBusy, Full, DataCount depth)) ->
    (BitVector n, Maybe BitVector n)
  go = _
```

We can pair program a bit to get this working if you'd like.

martijnbastiaan:
> Vivado can upgrade an older version of IP to a (reasonably-identically configured?) new version, though. So if Vivado switches to 13.3, people can still synthesise by introducing another step in their workflow: using Vivado to upgrade the IP. It [seems](https://support.xilinx.com/s/question/0D52E00006iHkM7SAK/how-to-upgrade-ip-in-a-project-tcl-mode?language=en_US) they just need to add this to their Tcl:

Cool @DigitalBrains1, I've seen that before, it should just work. Another option would be to add the version to the `DcConfig`. This would let users override the version in case of need. Obviously, this would be dangerous due to possible API changes but that's on the user. Perhaps we could even test for versions in the BlackBox :thinking:. We could make an ADT to make it explicit that you're doing something unsupported:

```haskell
data Version
  = v13_3
  | v13_4
  | CustomVersion String
```

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+  DcConfig (SNat depth) ->
+
+  Clock write -> Clock read ->
+  -- | Asynchronous reset
+  Reset read ->
+
+  -- | Write data
+  Signal write (BitVector n) ->
+  -- | Write enable
+  Signal write Bool ->
```
martijnbastiaan, line 133: `Signal write (Maybe (BitVector n))`?

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+  -- | Read enable
+  Signal read Bool ->
+  XilinxFifo read write depth n
+dcFifo dcCfg wClk rClk rst writeData wEnable rEnable =
+  let (wRst, f, wCnt, rRst, e, rCnt, rData) = dcFifo# dcCfg wClk rClk rst writeData wEnable rEnable
+    in XilinxFifo wRst f wCnt rRst e rCnt rData
```
martijnbastiaan, line 139: I'm thinking now that we use `DSL.constructProduct` we don't need to do this in a wrapper function :thinking:.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+  goWrite (FifoState q rt) rstR rEna (_ :- rstW) wEnas0 wDats0 =
+    (0 :- wRstBusy, full, wCnt, rRstBusy, fifoEmpty, rCnt, rData)
+    where
+      (wRstBusy, preFull, preWCnt, rRstBusy, fifoEmpty, rCnt, rData) =
+        go (FifoState q' (rt-tWr)) rstR rEna rstW wEnas1 wDats1
+
+      wCnt = sDepth q :- preWCnt
+      full = (countJust q == rD) :- preFull
+      (wDat :- wDats1) = wDats0
+      (wEna :- wEnas1) = wEnas0
```
martijnbastiaan, line 232: Probably just move these to line 223

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+      q' =
+        if countJust q + 1 <= rD && wEna
```
martijnbastiaan, line 234:
```suggestion
        if countJust q < rD && wEna
```

martijnbastiaan: I'd probably rename `rD` to `maxDepth` or something too.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+    (wRstBusy, full, wCnt, 1 :- rRstBusy, fifoEmpty, rCnt, rData)
+    where
+      rData = deepErrorX "Reset" :- preRData
+      fifoEmpty = True :- preEmpty
+      rCnt = 0 :- preRCnt
+
+      (wRstBusy, full, wCnt, rRstBusy, preEmpty, preRCnt, preRData) =
+        go (FifoState emptyInternalState (rt+tR)) rstRNext rEnas1 rstW wEna wData
+
+      (_ :- rEnas1) = rEnas0
```
martijnbastiaan, line 250: Move this to line 240. If you need to lazy behavior use `~(_ :- rEnas)`

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,291 @@
...
+    (wRstBusy, full, wCnt, 0 :- rRstBusy, fifoEmpty, rCnt, rData)
+    where
+      rCnt = sDepth q :- preRCnt
+      fifoEmpty = (countJust q == 0) :- preEmpty
+      rData = nextData :- preRData
+
+      (wRstBusy, full, wCnt, rRstBusy, preEmpty, preRCnt, preRData) =
+        go (FifoState q' (rt+tR)) rstRNext rEnas1 rstW wEna wData
+
+      (rEna :- rEnas1) = rEnas0
```
martijnbastiaan, line 262: Same

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -2,6 +2,7 @@
   Copyright   :  (C) 2019, Myrtle Software Ltd.
                      2020-2021, QBayLogic B.V.
                      2021, Myrtle.ai
+                     2022, Google Inc
```
martijnbastiaan, line 5: Let's move changes to this module to a separate PR.

clash-lib/src/Clash/Core/TermLiteral/TH.hs
```diff
@@ -1,3 +1,12 @@
+{-|
+Copyright   :  (C) 2019, Myrtle Software Ltd,
+                   2021, QBayLogic B.V.
+                   2022, Google Inc
```
martijnbastiaan, line 4: Let's move changes to this module to a separate PR

clash-lib/src/Clash/Primitives/DSL.hs
martijnbastiaan: https://github.com/clash-lang/clash-compiler/pull/2202 Done!

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
martijnbastiaan: With https://github.com/clash-lang/clash-compiler/pull/2202 we can do this AND get rid of the wrapper!

clash-lib/src/Clash/Core/TermLiteral/TH.hs
martijnbastiaan: https://github.com/clash-lang/clash-compiler/pull/2203 Done!

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
vmchale: Ah! Okay, I will take a look + fix.

martijnbastiaan:
These signals are generated as `wire`, but should be generated as `reg`:

```
  reg  wr_full_bool;
  reg  rd_empty_bool;
  reg  wr_enable;
  reg  rd_enable;
  reg  wr_reset_busy_bool;
  reg  rd_reset_busy_bool;
```

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: Yes! At least it seems to work when I spin it up in Vivado...

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: Oh wow that is very nice.

vmchale:
@martijnbastiaan just pushed an "implementation" for more optional Xilinx FIFO features; haven't updated `BlackBoxes.hs` yet.

clash-cores/src/Clash/Cores/Xilinx/Common.hs
```diff
@@ -0,0 +1,85 @@
...
+      -version #{version} \\
+      -module_name {#{moduleNameString}}
+
+    set_property -dict [list \\
+#{renderedProperties}
+                       ] \\
+                      [get_ips {#{moduleNameString}}]
+
+    generate_target {synthesis simulation} [get_ips {#{moduleNameString}}]
+  |]
```
martijnbastiaan, line 75: We should generate @DigitalBrains1's datastructure here.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/BlackBoxes.hs
```diff
@@ -0,0 +1,224 @@
...
+  ipConfig = (defIpConfig "fifo_generator" "13.2" dcFifoName){properties = props}
+  TermLiteralSNat depth = dcDepth
+
+  props =
+    [ ("Fifo_Implementation",    show dcImplementation)
+    , ("Performance_Options",    show dcReadMode)
+    , ("Reset_Type",             "Asynchronous_Reset")
+    , ("Full_Flags_Reset_Value", "1")
+    , ("Input_Data_Width", show width)
+    , ("Output_Data_Width", show width)
```
martijnbastiaan, line 216: If we're going to align, align everything. (Or nothing.)

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+
+defConfig :: KnownNat depth => DcConfig (SNat depth)
+defConfig = DcConfig
+  { dcDepth = SNat
+  , dcImplementation = Independent_Clocks_Block_RAM
+  , dcReadMode = Standard_FIFO
+  , dcReadDataCount = True
+  , dcWriteDataCount = True
+  , dcOverflow = False
+  , dcUnderflow = False
```
martijnbastiaan, line 80: Hmm, I'm thinking maybe enable all signals by default. This way you explicitly need to disable them. This probably prevents unexpected behavior. What do you think?

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+data FifoState n = FifoState
+  { hsQueue      :: Seq.Seq (BitVector n)
+  , relativeTime :: Int
+  } deriving Show
```
martijnbastiaan, line 86: Add documentation

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+  XilinxFifo
+    { writeReset :: Signal write ResetBusy
+    , isFull :: Signal write Full
+    , isOverflow :: Signal write Bool
+    , writeCount :: Signal write (DataCount depth)
+    , readReset :: Signal read ResetBusy
+    , isEmpty :: Signal read Empty
+    , isUnderflow :: Signal read Bool
+    , readCount :: Signal read (DataCount depth)
+    , fifoData :: Signal read (BitVector n)
```
martijnbastiaan, line 98: Add documentation. Probably include what signals in the Xilinx docs the correspond to too.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+-- | Mock Xilinx FIFO, see [documentation](https://docs.xilinx.com/v/u/en-US/pg057-fifo-generator)
```
martijnbastiaan, line 101: You can leave out the word "mock". But it probably is a good idea to mention that the behavior does not map 1:1 to the RTL model Xilinx provides. (And the RTL model doesn't correspond 1:1 to the generated hardware!)

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+      (deepErrorX "No sample" :- rData)
+
+ where
+  rstSignalR = unsafeToHighPolarity rst
+  rstSignalW = unsafeSynchronizer rClk wClk $ unsafeToHighPolarity rst
+
+  -- reified depth
+  maxDepth = snatToNum @Int (powSNat (SNat @2) (SNat @depth))
+
+  -- TODO: how do we ensure it doesn't "space leak" in Haskell?
```
martijnbastiaan, line 150: That's a good question. Typically what we do is strategically add `seqX` statements. You should ensure that evaluating an element from one signal evaluates elements from all other signals too. This prevents thunk buildups when a user only uses one signal.

The blockram / register implementations should give you some inspiration.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+    , Signal read Empty
+    , Signal read Bool
+    , Signal read (DataCount depth)
+    , Signal read (BitVector n)
+    )
+  go st@(FifoState _ rt) rstR rEna rstW =
+    if rt < tWr
+      then goRead st rstR rEna rstW
+      else goWrite st rstR rEna rstW
+    -- TODO: goBoth case?
```
martijnbastiaan, line 176: I think we can let this slide, unless we want a 1:1 model to the behavioral RTL. But I doubt we want that, given that the behavioral model doesn't correspond to the generated hardware anyway.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
```diff
@@ -0,0 +1,246 @@
...
+
+      (wRstBusy, full, over, wCnt, rRstBusy, preEmpty, preUnder, preRCnt, preRData) =
+        go (FifoState mempty (rt+tR)) rstRNext rEnas1 rstW wData
+
+  goRead (FifoState q rt) (_ :- rstRNext) (rEna :- rEnas1) rstW wData =
+    (wRstBusy, full, over, wCnt, False :- rRstBusy, fifoEmpty, under, rCnt, rData)
+    where
+      rCnt = sDepth q :- preRCnt
+      fifoEmpty = (Seq.length q == 0) :- preEmpty
+      rData = nextData :- preRData
```
martijnbastiaan, line 215: You can move this up to the `goRead` line. If you need the lazy behavior use `~(wibble :- wobble)`. This will probably help you get a consistent naming scheme for the element / signals too.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Instances.hs
```diff
@@ -0,0 +1,31 @@
...
+
+module Clash.Cores.Xilinx.DcFifo.Instances where
+
+import Clash.Core.TermLiteral (TermLiteral (termToData))
+import Clash.Core.TermLiteral.TH (deriveTermToData)
+
+import Clash.Cores.Xilinx.DcFifo.Explicit
+
+-- XXX: We might want to move these in their expanded form to
+-- `Clash.Cores.Xilinx.DcFifo.Explicit`.
```
martijnbastiaan, line 22: Whatdo you think @vmchale ?

tests/src/Clash/Vivado.hs
```diff
@@ -0,0 +1,127 @@
...
+
+set_property top #{entity} [get_fileset sim_1]
+
+save_project_as sim -force
+
+launch_simulation -simset sim_1 -mode behavioral
+
+run -all
+
+quit
```
martijnbastiaan, line 115: This is currently hardcoded for this test case. Use @DigitalBrains1's scripts?

martijnbastiaan (commented):
@vmchale I still need to review the tests and the netlist stuff. Hopefully I'll get to that tonight.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: I think they are disabled by default with Xilinx's FIFO generator so this seems good to me - it will error in simulation!

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: Sure!

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: Sure!

vmchale: Wait is that really better? I just split it out into `where` clauses since the line was long.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Instances.hs
vmchale: I think having orphans is unfortunate but maybe not worth writing these by hand - I will make a `Clash.Cores.Xilinx.DcFifo` module that exports everything a user would want, I think this is a good solution.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
martijnbastiaan: Right, actually, I misread. You're constructing a signal here, not deconstructing it :see_no_evil:. Sorry.

clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
vmchale: Ok, this is done.

clash-cores/src/Clash/Cores/Xilinx/Common.hs
vmchale: Wouldn't this be an API-breaking change? I have pushed something so that we don't hardcode anything but it's not as thorough as Peter's scripts.

tests/src/Clash/Vivado.hs
vmchale: I have fixed it so that it doesn't need to hardcode anything but it is not as hygienic as Peter's scripts: https://github.com/clash-lang/clash-compiler/pull/2182#discussion_r899152751

clash-cores/src/Clash/Cores/Xilinx/Common.hs
martijnbastiaan: Yes it would, but that's fine for a new version of Clash. I'd say it's definitely an improvement over what we currently do!

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
martijnbastiaan: Agreed :)

clash-cores/src/Clash/Cores/Xilinx/DcFifo/Explicit.hs
vmchale: Ok, I think based on inspection `seqX` will not be needed? I also ran simulation a little bit and that didn't blow up with `-K1K` so I'm pretty confident there is no thunk leak.

clash-cores/src/Clash/Cores/Xilinx/Common.hs
vmchale: Oh, nevermind it looks like `renderTcl` is only used by the Xilinx FIFO, so we can do whatever we want. 

I will do something more hygienic, like what Peter did.

clash-cores/src/Clash/Cores/Xilinx/DcFifo/BlackBoxes.hs
```diff
@@ -0,0 +1,224 @@
...
+dcFifoBBF :: HasCallStack => BlackBoxFunction
+dcFifoBBF _isD _primName args _resTys = do
+  let
+    [  knownNatN, _knownDomainWrite, _knownDomainRead, _knownNatDepth
+     , _constraint1, _constraint2
+     , either error id . termToDataError -> dcConfig
+     , _wClk, _rClk, _rst, _wData
+     , _rEnable
+
+    -- TODO: Make this blackbox return multiple results, instead of a tuple. See:
```
vmchale, line 58: Is this done? The PR is merged!

(We return a product type atm).

martijnbastiaan: We can leave it for this PR

martijnbastiaan:
@vmchale Could you split this up into three PRs:

1. Support VHDL component declaration in blocks
2. Support Vivado in CI
3. New: Xilinx DCFifo (this PR has become a bit cluttered)

I've already started with (2) as I wanted to get Vivado running on CI:

  * https://github.com/clash-lang/clash-compiler/commits/add-vivado-support

That branch has CI setup with Vivado running on it. It's probably best to comment out some GHC versions in `.gitlab.yml` to save on CI resources while testing that. Enabling Vivado for *all* tests does lead some weird cycles:

https://github.com/UnkindPartition/tasty/pull/340

so it probably takes some debugging..

vmchale:
@martijnbastiaan I think it is sensible to merge Vivado support and the FIFO at the same time.

If no potential Vivado tests are in the main test group tasty doesn't consider `--no-vivado` to be a valid CLI option! 

```
Invalid option `--no-vivado'
```

vmchale:
This is based on https://github.com/clash-lang/clash-compiler/pull/2257 but CI is passing now.

vmchale:
I think this is also going to need to be squashed :(

DigitalBrains1:
> I think this is also going to need to be squashed :(

Martijn [suggested above](https://github.com/clash-lang/clash-compiler/pull/2182#issuecomment-1161927210) that you create a new PR for this. Anyway, I think this can be a single commit indeed, but we'll see in review.

vmchale:
Oh you're right @DigitalBrains1! I'll close this; wait til the Vivado support is merged so it can be sensibly reviewed.

PR #2184, opened 2022-05-02
Added the `~ISSCALAR` template

rowanG077:
Added the `~ISSCALAR` template which can be used to check if an argument is rendered to a scalar in HDL.

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
@@ -2,6 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2021-2022, QBayLogic B.V.
+                    2022     , LUMI GUIDE FIETSDETECTIE B.V.
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -23,7 +24,7 @@ import           Control.Exception               (throw)
 import           Control.Lens
   (use, (%=), _1, _2, element, (^?))
 import           Control.Monad                   (forM, (<=<))
-import           Control.Monad.State             (State, StateT (..), lift)
+import           Control.Monad.State             (State, StateT (..), lift, gets)
 import           Data.Bitraversable              (bitraverse)
 import           Data.Bool                       (bool)
 import           Data.Coerce                     (coerce)
@@ -54,7 +55,7 @@ import           Text.Read                       (readEither)
 import           Text.Trifecta.Result            hiding (Err)
 
 import           Clash.Backend
-  (Backend (..), Usage (..), AggressiveXOptBB(..))
+  (Backend (..), Usage (..), AggressiveXOptBB(..), RenderEnums(..))
 import           Clash.Netlist.BlackBox.Parser
 import           Clash.Netlist.BlackBox.Types
 import           Clash.Netlist.Types
@@ -67,6 +68,8 @@ import           Clash.Signal.Internal
   (ResetKind(..), ResetPolarity(..), InitBehavior(..), VDomainConfiguration (..))
 import           Clash.Util
 import qualified Clash.Util.Interpolate          as I
+
+import           Clash.Annotations.Primitive     (HDL(VHDL))
 
 inputHole :: Element -> Maybe Int
 inputHole = \case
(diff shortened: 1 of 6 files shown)


martijnbastiaan (changes requested):
Could you add it to this list? 

https://github.com/clash-lang/clash-compiler/blob/ea114d8edd6a110f72d148203b9db2454cae8f37/clash-prelude/src/Clash/Tutorial.hs#L1214-L1295

rowanG077:
@martijnbastiaan Fixed

alex-mckenna:
You should also note this is incomplete wrt some HDL. For instance, VHDL specifies that integers, reals and user-defined enums are also scalar types. Verilog treats integer and real as vector types (insofar as you can bit-select and part-select into them). By not including them we implicitly bias towards the verilog view of the world, but maybe we should take the more restrictive view and treat them as scalar in Clash?

This isn't something to solve now, but it should be documented

martijnbastiaan (approved):
Agreed with @alex-mckenna

alex-mckenna:
You're also missing copyright notices :eyes:

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -1270,6 +1271,8 @@ a general listing of the available template holes:
 * @~ISLIT[N]@: Is the @(N+1)@'th argument to the function a literal.
 * @~ISVAR[N]@: Is the @(N+1)@'th argument to the function explicitly not a
   literal.
+* @~ISSCALAR[N]@: Is the @(N+1)@'th argument to the function a 1-bit scalar. Note
+  that this means VHDL types like `integers`, `real` are not considered scalar.
```
alex-mckenna, line 1275: This is fine, but doesn't mention as much detail as I wanted. I think a note in `clash-lib/src/Clash/Netlist/BlackBox/Util.hs` like this suffices:

```haskell
{- NOTE [scalar netlist types]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In VHDL and (System)Verilog, there are some differences between the types
considered to be scalar. In Verilog, the `integer`, `real` and `time` types
can be indexed into, whereas in VHDL they are scalar types. This means when
writing Clash blackboxes, ~ISSCALAR may not be consistent with the targeted
HDLs notion of a scalar type.

Currently, Clash treats only `Bit` and `Bool` as scalar. This means it may
result in incorrect VHDL if ~ISSCALAR is used to determine whether a indexing
is used. If we extend this to include types VHDL considers scalar, then we
must also ensure that Clash internally also treats these types as scalar to        
avoid confusion.
-}
```

and then a reference to the note where the check is implemented:

```haskell
(IsScalar n) -> let (_,ty,_) = bbInputs b !! n
                      in case ty of
                        -- See NOTE [scalar netlist types]
                        Bit  -> 1
                        Bool -> 1
                        _    -> 0
```

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -476,6 +492,12 @@ renderElem b (IF c t f) = do
                       Literal {}   -> 1
                       BlackBoxE {} -> 1
                       _            -> 0
+      (IsScalar n) -> let (_,ty,_) = bbInputs b !! n
+                      in case ty of
+                        -- See NOTE [scalar netlist types]
```
christiaanb, line 497: The type of this function is:
```
renderElem
  :: HasCallStack
  => Backend backend
  => BlackBoxContext
  -> Element
  -> State backend (Int -> Text)
```
i.e. we get access to the `Backend backend => backend`. This means we get access to https://github.com/clash-lang/clash-compiler/blob/ee31c20e73516cffa396eb4720083acfbe10fd42/clash-lib/src/Clash/Backend.hs#L94 which allows us to differentiate between the HDL we're rendering for. This in turns means we can decide what is a scalar or not depending on the hdl we're generating, obviating the need for the note, and just implementing it properly.

christiaanb (changes requested):

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
rowanG077: I have implemented it. I'm no Verilog/VHDL star but from the specs I can read online in (System)Verilog only `Bit` and `Bool` are scalar. in VHDL in addition to those `integer`, `real`, enumeration and physical types are scalar. But we never generate `real` or physical types.

rowanG077:
I noticed the `inputHole` function which is used to validate a blackbox is missing quite a few cases. Is this by design? Or just forgotten to update over the years?

https://github.com/clash-lang/clash-compiler/blob/8a66d964dc7597e9e5d2c832e8ba30e550dde6c3/clash-lib/src/Clash/Netlist/BlackBox/Util.hs#L71-L80

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -430,13 +433,15 @@ renderElem b (Tag n) = do
 
 renderElem b (IF c t f) = do
   iw <- iwWidth
+  hdl <- hdlKind <$> get
```
alex-mckenna, line 436:
```haskell
gets hdlKind
```

alex-mckenna (approved):
This is looking good now I think, modulo that one nit-picking comment :wink:

alex-mckenna (approved):

PR #2185, opened 2022-05-02
Don't render "(end)generate" in (System)Verilog

alex-mckenna:
The `generate` and `endgenerate` pairs in (System)Verilog are not
required, and introduce an arbitrary restriction that they cannot
be nested. We can ignore this special case by simply not rendering
the keywords, as generate statements can be written as normal
concurrent statements.

## Still TODO:


martijnbastiaan (approved):
LGTM, given that it passes the SystemVerilog tests locally.

PR #2195, opened 2022-05-08
Test suite: fix topEntity inclusion Ram/RMultiTop

### LINKED ISSUE #2192: `clash-testsuite` ... `Ram/RMultiTop` includes top entity in test bench
Instead of creating a test bench named e.g. `vhdl/RMultiTop.testBench/testBench.vhdl` that tests a `vhdl/RMultiTop.topEntity/topEntity.vhdl`, the former includes a copy of the latter.

It is rather odd that this faulty version of the code ended up in the commit and I know what to do to fix it.

I don't know the repercussions of the fault. I only know that for instance an `Enable` input on a top entity that is driven by `enableGen` in the test bench might get the enable elided from the final HDL. This will never happen with a properly declared top entity / test bench relation. But since it is generally expected that test benches do not include a copy of the top entities they test, there might be other repercussions? Alternatively it's just about sharing of work and it has no functional differences (other than the `Enable`).

DigitalBrains1:
Instead of `testBench` testing `topEntity`, the former included a
functionally indentical copy of the latter.

Fixes #2192

## Still TODO:

### DIFF
--- a/tests/shouldwork/Signal/Ram/RMulti.hs
+++ b/tests/shouldwork/Signal/Ram/RMulti.hs
@@ -2,44 +2,52 @@
 --
 -- Test reading with a fast clock. Memory is initialized through the write port,
 -- but after that the write side is dormant.
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
 module RMulti where
 
 import Clash.Explicit.Prelude
 import Clash.Explicit.Testbench
-import qualified Prelude as P
 
 createDomain vSystem{vName="P10", vPeriod=10000}
 createDomain vSystem{vName="P20", vPeriod=20000}
 
-ram
-  :: Clock P20
+type Ram =
+     Clock P20
   -> Clock P10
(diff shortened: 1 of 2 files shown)


tests/shouldwork/Signal/Ram/RMulti.hs
```diff
@@ -2,44 +2,52 @@
 --
 -- Test reading with a fast clock. Memory is initialized through the write port,
 -- but after that the write side is dormant.
+
+{-# OPTIONS_GHC -Wno-orphans #-}
+
```
christiaanb, line 7: This shouldn't be needed.

DigitalBrains1: Depends on your definition of "needed". I like to compile HDL with `-Wall` and not have any warnings, and then it's needed, since we use `createDomain` below.

So instead of "this shouldn't be needed", I'd like to counter "we should change CI so this actually is needed" 😆

PR #2201, opened 2022-05-10
CI: Upgrade 8.10 to 8.10.7, Haddock to 9.0.2

### LINKED ISSUE #2200: Haddock before GHC 8.10.5 doesn't support dashes in anchors
We have this in our documentation:
https://github.com/clash-lang/clash-compiler/blob/6cf8edbdaf2a1865f449b788a6f1221064231fb2/clash-prelude/src/Clash/Explicit/Testbench.hs#L79
https://github.com/clash-lang/clash-compiler/blob/6cf8edbdaf2a1865f449b788a6f1221064231fb2/clash-prelude/src/Clash/Prelude/Testbench.hs#L59

With GHC before 8.10.5 (Haddock reports version 2.24.2 in 8.10.5), linking to that with a module-link doesn't render the link:
https://github.com/clash-lang/clash-compiler/blob/6cf8edbdaf2a1865f449b788a6f1221064231fb2/clash-prelude/src/Clash/Explicit/Signal.hs#L876-L877
renders [like this](https://hackage.haskell.org/package/clash-prelude-1.7.0/candidate/docs/Clash-Explicit-Signal.html#v:runUntil) (or if you're looking at this issue once the issue has been fixed, I suppose 1.6.2 will still show the [offending behaviour](https://hackage.haskell.org/package/clash-prelude-1.6.2/docs/Clash-Explicit-Signal.html#v:runUntil)).

Interestingly, page-relative links like
https://github.com/clash-lang/clash-compiler/blob/6cf8edbdaf2a1865f449b788a6f1221064231fb2/clash-prelude/src/Clash/Explicit/Testbench.hs#L227-L228
[render fine](https://hackage.haskell.org/package/clash-prelude-1.7.0/candidate/docs/Clash-Explicit-Testbench.html#v:outputVerifier). Probably because that is just a plain HTML link, not a Haddock module link.

The problem is the dash in the anchor, if you remove that it all works. However, updating Haddock to the version bundled with GHC 8.10.5 or newer _also_ fixes the issue, so I think we should update the version of GHC we use in `build_sdist.sh` in CI. If this is problematic, I guess we can also get rid of the dash.

DigitalBrains1:
Now that the ghc package documentation is on Hackage for 8.10.7, we can
upgrade our build version.

Furthermore, we now build Haddock documentation for Hackage with 9.0.2
now.

This fixes a Haddock bug we encountered:

Fixes #2200

## Still TODO:

  -  ~~Write a changelog entry (see changelog/README.md)~~

### DIFF
--- a/.ci/gitlab/common.yml
+++ b/.ci/gitlab/common.yml
@@ -1,5 +1,5 @@
 .common:
-  image: ghcr.io/clash-lang/clash-ci-$GHC_VERSION:2022-02-02
+  image: ghcr.io/clash-lang/clash-ci-$GHC_VERSION:2022-05-10
   timeout: 2 hours
   stage: build
   variables:
--- a/.ci/docker/Dockerfile
+++ b/.ci/docker/Dockerfile
@@ -107,7 +107,7 @@ RUN git clone https://github.com/verilator/verilator verilator \
 
 FROM builder AS build-ghc
 
-ARG ghcup_version="0.1.17.4"
+ARG ghcup_version="0.1.17.7"
 
 # Must be explicitly set
 ARG ghc_version
--- a/.ci/docker/build-and-publish-docker-image.sh
+++ b/.ci/docker/build-and-publish-docker-image.sh
@@ -7,7 +7,7 @@ NAME="clash-ci-"
 DIR=$(dirname "$0")
 now=$(date +%F)
 
-GHC_VERSIONS=(  "9.0.2"   "8.10.2"  "8.8.4"   "8.6.5")
+GHC_VERSIONS=(  "9.0.2"   "8.10.7"  "8.8.4"   "8.6.5")
 CABAL_VERSIONS=("3.4.0.0" "3.2.0.0" "3.2.0.0" "3.0.0.0")
 
 # We want to use docker buildkit so that our layers are built in parallel. This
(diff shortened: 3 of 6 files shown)


.ci/gitlab/common.yml
```diff
@@ -1,5 +1,5 @@
 .common:
-  image: ghcr.io/clash-lang/clash-ci-$GHC_VERSION:2022-02-02
+  image: ghcr.io/clash-lang/clash-ci-$GHC_VERSION:2022-05-10
```
alex-mckenna, line 2: There are other dates to change, grep for them in `.ci/` and `.github/`

DigitalBrains1: Ah okay, thanks!

.ci/gitlab/common.yml
```diff
@@ -1,9 +1,9 @@
...
   timeout: 2 hours
   stage: build
   variables:
-    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-$GHC_VERSION-3
+    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-$CI_JOB_IMAGE
```
martijnbastiaan, line 6: This needs the `-3` appended, due to GitLab not adding cache bust tokens to `CACHE_FALLBACK_KEY` :roll_eyes:

martijnbastiaan (changes requested):

.ci/gitlab/common.yml
leonschoorl: I'd assumed that I could get rid of this weirdness because it'll be using a new key.
But it's a global thing that stays around and only goes up?

martijnbastiaan: yes

martijnbastiaan (approved):

PR #2206, opened 2022-05-11
Allow different cabal-store directories in integration tests.

vmchale:
My cabal stores its files in `/development/.cabal` (for various SSD/Linux reasons). Thus I would like to be able to override the default `~/.HOME/.cabal/store`. This has the added benefit of allowing users on other platforms to run the test suite.


martijnbastiaan (approved):
Makes sense.

PR #2211, opened 2022-05-17
Bound tasty-hedgehog dependency

DigitalBrains1:
CI fails because of the deprecation of `testProperty` in
`tasty-hedgehog` 1.2.0.0. Setting an upper bound allows us to add
commits to 1.4 again.

## Still TODO:


alex-mckenna (approved):
This seems fine to me, since we don't care about 1.4 being in stackage anymore

PR #2230, opened 2022-05-31
Netlist track usage

alex-mckenna:
When producing sequential and concurrent code, there are restrictions for how the same declared signal can be assigned which depend on the HDL being targeted. In short, these are

  * VHDL: A `signal` can be assigned continuously, or procedurally with non-blocking assignment. A `variable` can only be assigned procedurally with blocking assignment.
  * Verilog: A `wire` can only be assigned continuously. A `reg` can only be assigned procedurally, but can be assigned both non-blocking and blocking.

When the backend produces HDL, it needs to know the type of a declaration to render it. With the `WireOrReg` type, this was easy for VHDL, but does not help differentiate between if a signal is intended to be a VHDL signal or variable. A more general type is now provided which provides the extra information to make this choice:

```haskell
data Blocking = Blocking | NonBlocking
data Usage    = Cont | Proc Blocking
```

Since a declaration can accept more than one type of assignment, the usage is not tracked with the declaration like `WireOrReg` was, but tracked in the assignments and a usage map is created for every component (tracking declared signals to their most restrictive use). When producing netlist, the current use of a signal should be checked to ensure that generated code only produces assignments that are valid for the backend being targeted.

## Still TODO:

  - [x] Write a changelog entry (see changelog/README.md)
  - [x] Check copyright notices are up to date in edited files

[comment]: # (Depending on the change, some of these points may not be necessary. If so, leave the boxes unchecked so it is easier for a reviewer to see what has been omitted.)

PR #2231, opened 2022-06-01
(Optionally) Sequential Black Boxes

alex-mckenna:
This PR adds the ability to define black boxes which can render in concurrent and sequential contexts. A new template tag `~ISSEQUENTIAL` is provided, which templates can use to provide alternative implementations for the two contexts.

**NOTE:** This branch is based off #2230, not `master`, so that should be merged first

## Still TODO:

  - [ ] Write tests showing concurrent and sequential output with the same result in simulation (for some primitives like `Vector.map` etc. which can be defined in both contexts)

### DIFF
--- a/clash-lib/clash-lib.cabal
+++ b/clash-lib/clash-lib.cabal
@@ -230,6 +230,7 @@ Library
                       Clash.Netlist.Id.SystemVerilog
                       Clash.Netlist.Id.Verilog
                       Clash.Netlist.Id.VHDL
+                      Clash.Netlist.Scope
                       Clash.Netlist.Types
                       Clash.Netlist.Util
 
--- a/clash-lib/src/Clash/Netlist/BlackBox.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox.hs
@@ -103,6 +103,9 @@ import           Clash.Unique                  (lookupUniqMap')
 import           Clash.Util
 import qualified Clash.Util.Interpolate        as I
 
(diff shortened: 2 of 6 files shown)


alex-mckenna:
Me and Leon had a discussion about this a while ago, so I think it makes sense to add what I remember from that here.

  * The templates which can be concurrent and sequential are going to get ridiculous, especially given things like `~GENSYM[i]` needing a unique `i` for every occurrence in a template (not just all the occurrences hit when the template is instantiated. It would probably be a better idea to allow concurrent and sequential versions of templates to be given separately.

  * For black boxes which only support either concurrent or sequential (but not both), Clash should be able to handle this in some useful way. This may be throwing an error in netlist generation when something is used incorrectly, or potentially floating these instantiations to somewhere they can appear (this would depend on the work in [`netlist-scopes`](https://github.com/clash-lang/clash-compiler/tree/netlist-scopes) being finished).

  * While the original plan I had for "getting sequential Clash as quickly as possible" was to extend `SimIO` to work with all HDL, it should be no more difficult in practice to make all netlist sequential by default. With `netlist-scopes` this becomes a case of starting with an empty sequential scope inside an empty concurrent scope so inherently concurrent things are still floated up correctly (instead of just an empty concurrent scope).

I think these were the main points. Anything I forgot @leonschoorl?

martijnbastiaan:
Closing due to inactivity, see https://github.com/clash-lang/clash-compiler/issues/2751

PR #2238, opened 2022-06-12
Add flag rendering to 'ClashOpts'

martijnbastiaan:
This adds a more structured way to define Clash command line flags to
`Clash.GHC.ClashFlags`. This comes with a couple of improvements over
the status quo:

  * Each boolean flag now defines two flags: `-fclash-foo` and
    `-fclash-no-foo`. This makes sure users can always unset a flag
    after setting it. This can be helpful in certain environments, and
    should help us once we'll support module-level flags.

  * Each flag backed by a `Maybe` can now be unset using `-fclash-no-foo`

Last but not least, `ClashOpts` can now be rendered to a set of flags.
This can help users of `clash-shake` to define a set of options using
`ClashOpts`, and then pass them to clash-the-binary.

## Still TODO:

  - [ ] Document `*suffix` functions
  - [ ] Split up PR (=> lens stuff to separate one) to reduce noise?

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
@@ -1,0 +1,399 @@
+{-|
+  Copyright   :  (C) 2022, QBayLogic B.V.
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+  Utilities to parse and render flags exposed on the Clash's command line.
+-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+
+module Clash.GHC.ClashFlag
+  ( ClashFlag(..)
+  , FlagName
+
+  -- * Simple flag builders
+  , boolFlag
+  , wordFlag
+  , intFlag
+  , readableFlag
+
+  , maybeStringFlag
+  , maybeIntFlag
+  , maybeWordFlag
+  , maybeReadableFlag
+
+  , maybeStringWithDefaultFlag
+
+  -- * Generic flag builders
+  , boolArg
+  , unsetArg
+  , wordSuffix, wordSuffixId
+  , intSuffix, intSuffixId
+  , strSuffix, strSuffixId
+
+  -- * Flag renderers
+  , renderNoArg
+  , renderSuffix
+  , renderSepSuffix
+
+  -- * Utilities
+  , deprecated
+  , isDefault
+  ) where
+
+-- base
+import Data.IORef (IORef, modifyIORef)
+
+-- ghc
+#if __GLASGOW_HASKELL__ >= 900
+import GHC.Driver.CmdLine
+#else
+import CmdLineParser
+#endif
+  (EwM, Flag (Flag, flagOptKind), liftEwM, OptKind (..), defFlag, addErr, deprecate)
+
+-- clash-lib
+import Clash.Driver.Types (ClashOpts(..), defClashOpts)
+
+-- lens
+import Control.Lens (Lens', (^.), set)
+import Text.Read (readMaybe)
+
+-- | Defines a way to parse command line flags, set their values in a 'ClashOpts',
+-- and render flags based on a 'ClashOpts'.
+data ClashFlag = ClashFlag
+  { cfFlags :: IORef ClashOpts -> [Flag IO]
+  , cfRender :: ClashOpts -> [String]
+  }
+
+type FlagName = String
+
+-- | Return two flags covering a single property: @-fclash-foo@ and @-fclash-no-foo@.
+boolFlag :: FlagName -> Lens' ClashOpts Bool -> ClashFlag
+boolFlag nm opt = ClashFlag
+  { cfFlags = \opts -> boolArg nm opts (set opt)
+  , cfRender = \opts -> if isDefault opt opts then [] else renderNoArg nm (opts ^. opt)
+  }
+
+-- | Return a flag setting a 'Word': @-fclash-foo=3@.
+wordFlag :: FlagName -> Lens' ClashOpts Word -> ClashFlag
+wordFlag nm opt = ClashFlag
+  { cfFlags = \ref -> wordSuffixId nm ref (set opt)
+  , cfRender = \opts ->
+      if isDefault opt opts then [] else renderSuffix nm (show (opts ^. opt))
+  }
+
+-- | Return a flag setting a 'Int': @-fclash-foo=3@.
+intFlag :: FlagName -> Lens' ClashOpts Int -> ClashFlag
+intFlag nm opt = ClashFlag
+  { cfFlags = \ref -> intSuffixId nm ref (set opt)
+  , cfRender = \opts ->
+      if isDefault opt opts then [] else renderSuffix nm (show (opts ^. opt))
+  }
+
+-- | Return two flags setting a 'Maybe' 'String': @-fclash-foo wibble@ and
+-- @-fclash-no-foo@.
+maybeStringFlag :: FlagName -> Lens' ClashOpts (Maybe String) -> ClashFlag
+maybeStringFlag nm opt = ClashFlag
+  { cfFlags = \opts ->
+         strSuffixId nm opts (set opt . Just)
+      ++ unsetArg nm opts (set opt)
+  , cfRender = \opts ->
+      if isDefault opt opts
+      then []
+      else case opts ^. opt of
+        Nothing -> renderNoArg nm False
+        Just s -> renderSepSuffix nm s
+  }
+
+
+-- | Return three flags setting a 'Maybe' 'String'. If no argument is supplied, a
+-- default value is set. For example, if the default is @simpleDefault@, then:
+--
+--   * @-fclash-foo@ sets the option to @Just "simpleDefault"@
+--   * @-fclash-foo=wibble@ sets the option to @Just "wibble"@
+--   * @-fclash-no-foo@ sets the option to @Nothing@
+--
+maybeStringWithDefaultFlag ::
+  FlagName ->
+  -- | Default
+  String ->
+  Lens' ClashOpts (Maybe String) ->
+  ClashFlag
+maybeStringWithDefaultFlag flagName dflt opt = ClashFlag
+  { cfFlags = \opts ->
+         anySuffix flagName opts parser (set opt)
+      ++ unsetArg flagName opts (set opt)
+  , cfRender = \opts ->
+      if isDefault opt opts
+      then []
+      else case opts ^. opt of
+        Nothing -> renderNoArg flagName False
+        Just s | s == dflt -> renderSuffix flagName ""
+        Just s -> renderSuffix flagName s
+  }
+ where
+  parser :: String -> EwM IO (Maybe (Maybe String))
+  parser ['='] = pure (Just (Just dflt))
+  parser ('=':v) = pure (Just (Just v))
+  parser s = do
+    addErr ("Could not parse: " <> s)
+    pure Nothing
+
+-- | Return two flags setting a 'Maybe' 'Int'.
+--
+--   * @-fclash-foo=3@ sets the option to @Just 3@
+--   * @-fclash-no-foo@ sets the option to @Nothing@
+--
+maybeIntFlag :: FlagName -> Lens' ClashOpts (Maybe Int) -> ClashFlag
+maybeIntFlag nm opt = ClashFlag
+  { cfFlags = \ref ->
+         intSuffixId nm ref (set opt . Just)
+      ++ unsetArg nm ref (set opt)
+
+  , cfRender = \opts ->
+      if isDefault opt opts
+      then []
+      else case opts ^. opt of
+        Nothing -> renderNoArg nm False
+        Just a -> renderSuffix nm (show a)
+  }
+
+-- | Return two flags setting a 'Maybe' 'Word'.
+--
+--   * @-fclash-foo=3@ sets the option to @Just 3@
+--   * @-fclash-no-foo@ sets the option to @Nothing@
+--
+maybeWordFlag :: FlagName -> Lens' ClashOpts (Maybe Word) -> ClashFlag
+maybeWordFlag nm opt = ClashFlag
+  { cfFlags = \ref ->
+         wordSuffixId nm ref (set opt . Just)
+      ++ unsetArg nm ref (set opt)
+
+  , cfRender = \opts ->
+      if isDefault opt opts
+      then []
+      else case opts ^. opt of
+        Nothing -> renderNoArg nm False
+        Just a -> renderSuffix nm (show a)
+  }
+
+maybeReadableFlag ::
+  forall a.
+  (Eq a, Read a, Show a) =>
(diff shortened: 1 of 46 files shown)


clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
...
+-- lens
+import Control.Lens (Lens', (^.), set)
+import Text.Read (readMaybe)
+
+-----------
+-- Types --
+-----------
+
+-- | Defines a way to parse command line flags, set their values in a 'ClashOpts',
+-- and render flags based on a 'ClashOpts'.
```
alex-mckenna, line 70: This is listed as three separate things, but I _think_ the parsing and setting are inherently related here. If that's the case it would be better to re-word and make it clear the two happen together / use bullet points

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
...
+data ClashFlag = ClashFlag
+  { cfFlags :: IORef ClashOpts -> [Flag IO]
+  , cfRender :: ClashOpts -> [String]
```
alex-mckenna, line 73: Document these fields, it's kind of weird to think of a flag corresponding to multiple flags until you remember / know the functions later add `-fclash-foo` and `-fclash-no-foo` flags

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+type FlagName = String
+
+--------------------------
+-- Simple flag builders --
+--------------------------
+
```
alex-mckenna, line 81: Marker comments like this in code are bad style IMO. They can end up in the wrong place, or have a label which is no longer relevant, and worse if a new function is added which could fit under more than one heading they make it harder to find things

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
...
+  parser ('=':v) = pure (Just (Just v))
+  parser s = do
+    addErr ("Could not parse: " <> s)
+    pure Nothing
+
+maybeIntFlag :: FlagName -> Lens' ClashOpts (Maybe Word) -> ClashFlag
+maybeIntFlag = maybeReadableFlag
+
+maybeWordFlag :: FlagName -> Lens' ClashOpts (Maybe Word) -> ClashFlag
+maybeWordFlag = maybeReadableFlag
```
alex-mckenna, line 158: I would just use `maybeReadableFlag` directly for these. Also `maybeIntFlag` has a `Maybe Word` which I assume should be `Maybe Int`

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
...
+  parser :: String -> EwM IO (Maybe a)
+  parser s = case readMaybe s of
+    Nothing -> addErr ("Could not parse " <> s)  >> pure Nothing
+    a -> pure a
+
+---------------------------
+-- Generic flag builders --
+---------------------------
+
+-- | Helper function....
```
alex-mckenna, line 208: Comments like this aren't particularly helpful, _especially_ as haddock comments

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
...
+  }
+ where
+  parser :: String -> EwM IO (Maybe a)
+  parser s = case readMaybe s of
+    Nothing -> addErr ("Could not parse " <> s)  >> pure Nothing
+    a -> pure a
+
+---------------------------
+-- Generic flag builders --
+---------------------------
```
alex-mckenna, line 206: See previous comment on markers

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
```diff
@@ -1,327 +1,328 @@
...
-import           Data.List.Split                (splitOn)
-import qualified Data.Set                       as Set
-import qualified Data.Text                      as Text
-import           Text.Read                      (readMaybe)
+import           Data.List (intercalate)
+import           Text.Read (readMaybe)
+
+-- unordered-containers
+import qualified Data.Set as Set
+import           Data.Set (Set)
```
alex-mckenna, line 44: I'm also not a fan of these marker comments for package providence. Especially when it's wrong like this

alex-mckenna: The main reason I'm against it is I don't see what benefit it brings. If you want to see if a package is unused there's `-Wunused-packages`, or `weeder` or any other tool which is going to be quicker and more correct than using these comments. If it's just because there's a lot of imports and it's difficult to read that's more a motivation to split up the module IMO

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
```diff
@@ -1,327 +1,328 @@
...
+
+  -- Deprecated/removed flags:
+  , deprecated "-fclash-float-support is always enabled from Clash 1.6 and onwards" $
+      boolFlag "float-support" voidLens
+
+  , deprecated "-fclash-clean has been removed" $
+      boolFlag "clean" voidLens
+
+  , deprecated "-fclash-nocache deprecated in favor of -fclash-no-cache" $
+      (boolFlag "nocache" (opt_cachehdl . iso' not not)){ cfRender = const [] }
```
alex-mckenna, line 152: I think much like deprecated APIs, when we deprecate a flag it should include a message saying it won't be a flag after version 1.X. Then when that version is released it gets removed and just becomes a normal "unknown flag" error

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
```diff
@@ -1,327 +1,328 @@
...
+      -- TODO: Should we add a deprecation warning for these flags? We could
+      --       just use 'maybeReadableFlag' and ditch this custom impl if we
+      --       once the warning has been up long enough.
+      "Xilinx" -> pure (Just Vivado)
+      "ISE"    -> pure (Just Vivado)
+      "Altera" -> pure (Just Quartus)
+      "Intel"  -> pure (Just Quartus)
+      _ -> do
+        addErr (s ++ " is an unknown hdl synthesis tool")
+        pure Nothing
```
alex-mckenna, line 188: While we're here: I think we should

  * make this comparison case-insensitive, since it costs us nothing but potentially makes Clash less annoying
  * recognize the names `vivado` and `quartus` since it's weird to me we use those names internally but don't accept them on the command line

clash-ghc/tests/Clash/Tests/GHC/ClashFlags.hs
```diff
@@ -0,0 +1,407 @@
...
+
+  gen_dbg_countTransformations = genBool (opt_debug . dbg_countTransformations)
+  gen_dbg_historyFile = genMaybe (opt_debug . dbg_historyFile) genSimpleString
+  gen_dbg_invariants = genBool (opt_debug . dbg_invariants)
+  gen_dbg_transformationInfo = genEnumBounded (opt_debug . dbg_transformationInfo)
+  gen_dbg_transformationsFrom = genMaybe (opt_debug . dbg_transformationsFrom) (Gen.word Range.constantBounded)
+  gen_dbg_transformationsLimit = genMaybe (opt_debug . dbg_transformationsLimit) (Gen.word Range.constantBounded)
+
+  -- TODO
+  gen_dbg_transformations _fuzz = pure mempty
```
alex-mckenna, line 304: What is there TODO?

clash-ghc/tests/Clash/Tests/GHC/ClashFlags.hs
```diff
@@ -0,0 +1,407 @@
...
+testCases =
+  case $(testGroupGenerator) of
+    TestGroup _ tree -> tree
+    _ -> error "Internal error"
+
+tests :: TestTree
+tests =
+  testGroup
+    "Clash.Tests.GHC.ClashFlags"
+    [ -- tasty-th doesn't account for Hedgehog tests
```
alex-mckenna, line 399: > `-- tasty-th doesn't account for Hedgehog tests`

Even if you prefix with `prop_` instead of `hprop_`, as they do in the hedgehog docs?

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -138,59 +132,60 @@ data TransformationInfo
...
   --
   -- Command line flag: -fclash-debug-transformations-limit=N
-  , dbg_historyFile :: Maybe FilePath
+  , _dbg_historyFile :: Maybe FilePath
   -- ^ Save information about all applied transformations to a history file
   -- for use with @clash-term@.
   --
   -- Command line flag: -fclash-debug-history[=FILE]
   } deriving (Generic, NFData, Show, Eq)
+makeLenses ''DebugOpts
```
alex-mckenna, line 177: I wonder if instead of changing the names to every field, it would be better to just call `makeLenses` with options, i.e. tell it to generate a lens `foo` from a field `dbg_foo`. If you do this here and for `ClashOpts` the diff will be much shorter

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -540,6 +536,14 @@ newtype SdcInfo = SdcInfo
...
   }
 
+data ClashEnv = ClashEnv
+  { envOpts        :: ClashOpts
+  , envTyConMap    :: TyConMap
+  , envTupleTyCons :: IntMap TyConName
+  , envPrimitives  :: CompiledPrimMap
+  , envCustomReprs :: CustomReprs
+  } deriving (Generic, NFData)
+
```
alex-mckenna, line 546: Why did this move down here? You don't do anything with it after which would need the new lenses in scope

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
```diff
@@ -0,0 +1,387 @@
+{-|
+  Copyright   :  (C) 2021, QBayLogic B.V.
```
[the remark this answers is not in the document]
martijnbastiaan: Nope, just living in the past :see_no_evil:

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
martijnbastiaan: > Also maybeIntFlag has a Maybe Word which I assume should be Maybe Int

Woops

> I would just use maybeReadableFlag directly for these.

This is actually wrong in another way: ints are passed as `-fclash-foo=3`, but `maybeReadableFlag` uses `-fclash-foo 3`.

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
martijnbastiaan: Yeah it's a `TODO`

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
martijnbastiaan: Agreed. Can this go in a separate PR? I don't want to battle out in which version things need to deprecate now :smile:

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
martijnbastiaan: `Vivado` and `Quartus` are recognized. It only takes this branch if the `readMaybe` fails. Agreed on the lower case stuff, but this is true for all the "readable" flags. Let's postpone until we have a way to do those case-insensitively too?

clash-ghc/tests/Clash/Tests/GHC/ClashFlags.hs
martijnbastiaan: make it fuzz :-)

clash-ghc/tests/Clash/Tests/GHC/ClashFlags.hs
martijnbastiaan: Yeah, it turns it into a `QuickCheck` test.

clash-lib/src/Clash/Driver/Types.hs
martijnbastiaan: `TemplateHaskell` :sparkles:

clash-lib/src/Clash/Driver/Types.hs
martijnbastiaan: Don't we add `dbg_` to prevent naming collisions?

But even if we don't, `lens` is already complicated enough as it is, I think bringing in custom settings for smaller diffs isn't a good idea. The underscore-to-no-underscore pattern is pretty well established.

martijnbastiaan:
Thanks for the review @alex-mckenna, I'll implement the suggestions.

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
alex-mckenna: The "default" is current + 2 major, I don't think that's controversial here?

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
alex-mckenna: > This is actually wrong in another way

Then I'm not sure I understand why they're defined like `maybeIntFlag = maybeReadableFlag` and `maybeWordFlag = maybeReadableFlag`. Either you can just use `maybeReadableFlag` or these definitions are wrong, no?

clash-lib/src/Clash/Driver/Types.hs
alex-mckenna: > Don't we add dbg_ to prevent naming collisions?

No

> The underscore-to-no-underscore pattern is pretty well established.

Sure, it is more consistent to always use the same name scheme for lenses. It is a shame the default naming scheme is so intrusive though :cry:

martijnbastiaan: It is. We'll have somewhat sensible records usable in 2030 though :-).

clash-ghc/src-ghc/Clash/GHC/ClashFlag.hs
martijnbastiaan: Yeah, the point of my comment was to point out my own code is wrong, not your comment. Sorry for the confusion.

> Either you can just use maybeReadableFlag or these definitions are wrong, no?

We've established a (frankly stupid) pattern that int/word is passed `-flike-this=3` while strings are passed `-flike-that str`. `maybeReadableFlag` interprets a string, so it will use the latter pattern. However, if we then use it to parse ints/words we're in the "wrong". So, it's not that `maybeReadableFlag` is wrong, but its usage for ints/words is.

I've put a comment somewhere that I think we should be liberal in what type of flag passing we accept for both.

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
martijnbastiaan: Okay, I've set them all to 1.10 now. If this leads to any discussion I'm going to rip it from this PR.

PR #2247, opened 2022-06-17
Run `clash-testsuite` with 1 thread

martijnbastiaan:
On CI the testsuite spawns 64 threads to coordinate running tests. This
induces an incredible overhead, making Clash run an order of magnitude
slower than it should. Note that `-jN` continues to work: the testsuite
will still parallelize test running.

### DIFF
--- a/tests/clash-testsuite.cabal
+++ b/tests/clash-testsuite.cabal
@@ -118,7 +118,6 @@ library
 executable clash-testsuite
   import:              basic-config
   main-is:             Main.hs
-  ghc-options:         -threaded -with-rtsopts=-N
 
   build-tool-depends:
     clash-ghc:clash


christiaanb:
Is this only an issue on CI? or also on local dev machines? If only on CI, then perhaps make the change in https://github.com/clash-lang/clash-compiler/blob/master/.ci/cabal.project.local?

martijnbastiaan:
It's an issue on machines with a massive number of threads. The point is that `clash-testsuite` itself doesn't need many (any?) threads as it is completely I/O bound. To clarify further, if you run with `-j1024`, 1026 will be spawned right now: 2 for `clash-testsuite` itself, 1024 for all the subprocesses it spawns (`clash`, `iverilog`, ..). Previously 2048 would be spawned (provided you're running on a 1024-thread machine :-)).

christiaanb:
Alright, thanks for explaining!

martijnbastiaan:
@leonschoorl and I just discussed the magic number "2". I've set it to "1" now, which should also work. I can't test it right now because my machine is doing loads of other stuff, so I'll test it at home (with 24 threads). Leon approved this IRL, so I'll merge if it does.

leonschoorl (approved):

PR #2256, opened 2022-06-23
Add vivado support in CI

vmchale:
This adds support for Vivado in CI; it is tested by the [Xilinx FIFO branch](https://github.com/clash-lang/clash-compiler/pull/2182)


vmchale:
I think it is best to merge Vivado support and the Xilinx FIFO at the same time, as explained [here](https://github.com/clash-lang/clash-compiler/pull/2182#issuecomment-1164654003).

PR #2257, opened 2022-06-23
Add vivado support to `clash-testsuite`

vmchale:
This adds support for Vivado in CI.

It is tested downstream [here](https://github.com/clash-lang/clash-compiler/pull/2182).


tests/src/Test/Tasty/Vivado.hs
```diff
@@ -0,0 +1,53 @@
...
+          hdlSrc <- dir
+          tcl <- fromModuleName (HdlSource hdlSrc target) m top
+          hPutStr h tcl *>
+            hFlush h
+          runVivado hdlSrc ["-mode", "batch", "-source", fp]
+    | otherwise = pure (testPassed "Ignoring test due to --no-vivado")
+
+   where
+    vivado workDir args =
+      TestFailingProgram True "vivado" args NoGlob PrintNeither False (Just 0) (ExpectNotStdOut "expected: ") (Just workDir)
```
martijnbastiaan, line 48:
```suggestion
      TestFailingProgram True "vivado" args NoGlob PrintNeither False (Just 1) (ExpectNotStdOut "expected: ") (Just workDir)
```

I think I messed this up. This used to be `1` in your code @vmchale, but I changed it to `0` for debugging purposes and mindlessly squashed..

vmchale: Oh no! I think this is supposed to be 0, the `1` was in there for debugging purposes. `0` is exit success!

martijnbastiaan: agreed!

tests/src/Clash/Vivado.hs
```diff
@@ -0,0 +1,149 @@
...
+  -- | Qualified name
+  String ->
+  IO SimProj
+simProjFromClashEntities hdlDir qualName = do
+  (Just mHdl) <- readManifest (hdlDir </> qualName ++ "/clash-manifest.json")
+  let deps = T.unpack <$> transitiveDependencies mHdl
+  nextSimProj <- traverse (simProjFromClashEntities hdlDir) deps
+  pure $ mconcat nextSimProj <> asSimProj qualName mHdl
+
+fromModuleName ::
```
DigitalBrains1, line 61: I think this function has an unclear name and there's no documentation either. `mkTcl` feels appropriate for this and then rename the existing `mkTcl`? I dunno.

tests/src/Test/Tasty/Vivado.hs
DigitalBrains1, line 48: I think looking for the string "expected" is too brittle. I'd prefer if we enhance our `Expect` stuff and then make the regex `^@|^Error` match. Luckily, any errors will never occur in the first line of the code, so if it is easier we could match `\n@` and `\nError`.

Also, currently Vivado outputs critical warnings, and I think those should have made CI fail:
```
CRITICAL WARNING: [Vivado 12-1504] The IP is already part of the fileset 'sources_1'. Requested source '/home/peter/clash-test-1e8af8d974db0407/xilinx_tmp/.srcs/sources_1/ip/dcfifo/dcfifo.xci' will not be added.
CRITICAL WARNING: [filemgmt 20-1440] File '/home/peter/clash-test-1e8af8d974db0407/xilinx_tmp/.gen/sources_1/ip/dcfifo/synth/dcfifo.vhd' already exists in the project as a part of sub-design file '/home/peter/clash-test-1e8af8d974db0407/xilinx_tmp/.srcs/sources_1/ip/dcfifo/dcfifo.xci'. Explicitly adding the file outside the scope of the sub-design can lead to unintended behaviors and is not recommended.
```

 I looked into it and it turns out you can actually change the severity of most of Vivado's messages, so we can elevate all critical warnings to a level where Vivado exits with an error exit code:
```
set_msg_config -severity {CRITICAL WARNING} -new_severity ERROR
```
That should be added near the top of the Tcl CI script.

PS: Our `assert` function outputs a message prefixed with the word "Error: " in VHDL and with `@` in Verilog, that's the reason for the regex.

tests/clash-testsuite.cabal
```diff
@@ -12,6 +12,8 @@ maintainer:          Christiaan Baaij <christiaan.baaij@gmail.com>
 copyright:           Copyright © 2015 University of Twente
 category:            Testing
 build-type:          Simple
+data-files:
+  tcl/clash_namespace.tcl
```
DigitalBrains1, line 16: Could we make the name of the directory more generic? `global-files`? Otherwise we'll need another directory when we have a different reason to add a file.

tests/tcl/clash_namespace.tcl
```diff
@@ -0,0 +1,31 @@
...
+                if {$ipName in $seen} {
+                    continue
+                }
+                ${tclIface}::createIp
+                lappend seen $ipName
+            }
+        }
+        return $seen
+    }
+}
```
DigitalBrains1, line 31: This is not an implementation of the API I designed. It should be this (adapted to the current CI script, not exactly a match for the CI script Martijn and I wrote):
```suggestion
namespace eval clash {
    # Populate a namespace with a Clash-generated Tcl interface.
    # Namespace is clash::tclIface::$top::$baseName
    proc loadTclIface {top clashTclFile} {
        # Evaluate script code inside temporary throwaway namespace to
        # separate its definitions from ours and reduce the chance of
        # accidentally corrupting our code.
        namespace eval tmp "source $clashTclFile"
        set baseName [file rootname [file tail $clashTclFile]]
        tmp::createNamespace [namespace current]::tclIface::${top}::${baseName}
        namespace delete tmp
    }

    proc runClashScripts {} {
        if {![namespace exists tclIface]} {
            # There are no scripts
            return
        }

        # Identical names means identical IP, only one run needed even if it
        # occurs in multiple HDL directories.
        set seen [list]
        foreach topNs [namespace children tclIface] {
            foreach tclIface [namespace children $topNs] {
                set api [subst $${tclIface}::api]
                if {$api ne {1}} {
                    error "Error: $tclIface doesn't implement an API we\
                        support: api = \"$api\"."
                }
                set purpose [subst $${tclIface}::scriptPurpose]
                if {$purpose ne {createIp}} {
                    error "Error: $tclIface::scriptPurpose bogus value\
                        \"$purpose\"."
                }
                set ipName [subst $${tclIface}::ipName]
                if {$ipName in $seen} {
                    continue
                }
                ${tclIface}::createIp $ipName
                lappend seen $ipName
            }
        }
        return $seen
    }
}
````

(I think the diff is not more helpful because for arcane and unknowable reasons, the script in the repository has DOS line endings. This seems to trip up GitHub's diff functionality.)

**Review by DigitalBrains1 — CHANGES_REQUESTED** [review]
Wow, the three PR's together are a significant amount of work! Good job on putting all that together!

I do have some things I'd like to discuss in this PR.

I put this together while I'm really tired, so I hope my mistakes aren't too grave. I'd like to get the ball rolling on this, so I went ahead and did it now anyway.

File: tests/src/Clash/Vivado.hs [hunk 5]
```diff
@@ -0,0 +1,149 @@
...
+
+-- | Given something like @topEntity_dcfifo_240A2A9D3FF8D64B.tcl@, return the
+-- IP name, @dcfifo@
+parseTcl :: FilePath -> String
+parseTcl fp = takeWhile (/= '_') (drop 1 $ dropWhile (/= '_') fp)
+
+mSimProj :: FilePath -> T.Text -> FilePath -> SimProj -> SimProj
+mSimProj dir lib fp p@(SimProj vhd v ip)
+  | ".vhdl" `isSuffixOf` fp = SimProj ((lib, dir </> fp):vhd) v ip
+  | ".vhd" `isSuffixOf` fp = SimProj ((lib, dir </> fp):vhd) v ip
```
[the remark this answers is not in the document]
alex-mckenna: No we don't

tests/clash-testsuite.cabal
vmchale: I figured it'd make sense to put all `tcl` sources in one directory. But if you prefer to put all `data-files` in the same directory I suppose that could work :thinking:

tests/src/Clash/Vivado.hs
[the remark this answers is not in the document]
martijnbastiaan: We don't, good point.

tests/src/Clash/Vivado.hs
```diff
@@ -0,0 +1,149 @@
...
+  [i|
+source #{cabalDir}/tcl/clash_namespace.tcl
+
+# create clash namespace
+namespace eval clash {}
+
+set clash::hdl_dir #{hdlDir}
+
+# for cleanup
+file delete -force xilinx_tmp
```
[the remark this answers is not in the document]
vmchale: Yes.

tests/clash-testsuite.cabal
martijnbastiaan: I think Peter would like `data-files/tcl`

martijnbastiaan: in case we ever want something else than `tcl`.

martijnbastiaan: Peter approved IRL.

tests/tcl/clash_namespace.tcl
vmchale: What does checking the `$purpose` variable accomplish? I got rid of it because I thought it didn't do anything useful. 

Also this forces the API version to be `1`, which doesn't really enforce much (and IP cores are already already versioned!)

DigitalBrains1: The `tclIface::...::api` has nothing to do with Xilinx IP versions. It is the version of the Clash<->Tcl API that the generated script implements.

The idea is to lay the ground work for a future-proof Clash<->Tcl API. So when we need it for something else than Vivado's `create_ip` thing, we can bump the API, introduce a new script purpose and our scripts would check API and script purpose and invoke it in the correct manner. I wanted to codify `scriptPurpose` right now so this is established already when we later want to have more options.

And similarly, if we later need changes to how `tclIface::...::createIp` works, we bump the API as well. Our integration scripts should support all versions of the API so that when a user has created their own stuff that implements the API, they are not forced to upgrade their own stuff but our scripts gracefully handle the not-latest API they generate.

Properly checking values that currently have only one allowed option is simply defensive programming to hopefully give a meaningful error message if there is a mistake a user or we made, rather than blowing up spectacularly and unfathomably later.

vmchale: Oh I see, that makes sense.

tests/src/Test/Tasty/Clash.hs
```diff
@@ -129,6 +130,8 @@ data TestOptions =
     , verilate :: Verilate
     -- ^ Whether to run compatible tests through verilator as well as / in
     -- place of the simulator that would ordinarily be used.
+    , vivado :: Bool
+    -- ^ Whether to simulate via Vivado (such as when one depends on Xilinx IP)
```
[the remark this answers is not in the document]
vmchale: That makes sense.

tests/src/Clash/Vivado.hs
```diff
@@ -0,0 +1,149 @@
...
+
+launch_simulation -simset sim_1 -mode behavioral
+
+run -all
+
+quit
+|]
+ where
+  hdlDir = directory hdl
+  readSources = case buildTarget hdl of
```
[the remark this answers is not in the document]
vmchale: Oh yes I see.

tests/src/Test/Tasty/Clash.hs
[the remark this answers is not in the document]
DigitalBrains1: It suddenly randomly occurred to me it might be much less work if in addition to `hdlSim` (still as a `Bool`) we additionally define `hdlSimWith` with a list of tools. This saves changing the 121 lines in `Main.hs` that say `hdlSim=False`.

DigitalBrains1:
Should we perhaps set `-fclash-hdlsyn Vivado` when doing Vivado CI? It's not unlikely we'll create a test case that happens to need it.

vmchale:
Yes! You're right.

tests/src/Test/Tasty/Clash.hs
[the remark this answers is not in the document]
vmchale: I like `hdlSim` as a list because we can take `null hdlSim` to deduce the boolean we had before (and this enforces that everything is the same).

tests/src/Clash/Vivado.hs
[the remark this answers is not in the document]
vmchale: This is just how I organized the code, I don't think it's much better doing a `case` on the extension, but I don't think there's much advantage either way.

tests/src/Clash/Vivado.hs
vmchale: I've named it `tclFromManifest` which I think makes sense.

tests/src/Clash/Vivado.hs
```diff
@@ -0,0 +1,149 @@
...
+import Clash.Driver.Manifest (readManifest, Manifest (..))
+import Clash.Annotations.Primitive (HDL (..))
+
+import Data.String.Interpolate (i)
+import Data.List (isSuffixOf)
+
+import System.FilePath ((</>))
+import qualified Data.Text as T
+
+data HdlSource = HdlSource { directory :: FilePath, buildTarget :: HDL }
```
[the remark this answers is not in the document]
vmchale: I've renamed it to `hdlDir` there - hopefully that is a little better?

tests/src/Test/Tasty/Clash.hs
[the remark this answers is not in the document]
DigitalBrains1: Sure, I think it's fine.  `data HdlSim = DefHdlSim | HdlSim [Sim] | NoHdlSim` might be even more intuitive.

tests/src/Clash/Vivado.hs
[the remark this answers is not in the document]
DigitalBrains1: It looks rather convoluted to me.

First you `case` on the extension, to put it in two separate lists `vhdlFiles` and `verilogFiles`. Then you keep the `buildTarget` around just for these lines of code, meaning you need the datatype `HdlSource` to put it in, whereas otherwise you could have just kept `directory` from that record type. Then you `case` on the `buildTarget` to pick the right one of the two lists you built by casing on the extension. Why not just `case` on the extension once, here?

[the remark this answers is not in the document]
DigitalBrains1: Ah right, no, you do need to `case` on the extension to pick out the `.tcl` earlier, I overlooked that for a moment. But then you could just add all files in `vhdlFiles` with `read_vhdl` and add all files from `verilogFiles` with `read_verilog`, again eliminating the need to pass `buildTarget` around and `case` on it here. In practice, one of those will always be empty, unless we suddenly begin generating `.vhdl` and `.v` in the same project. I don't see a reason for it now, but I also don't see a reason to filter it out (which to me is the only purpose of this code, is there another?).

tests/src/Clash/Vivado.hs
```diff
@@ -0,0 +1,149 @@
...
+
+set_property top #{entity} [get_fileset sim_1]
+
+save_project_as sim -force
+
+launch_simulation -simset sim_1 -mode behavioral
+
+run -all
+
+quit
```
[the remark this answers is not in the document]
vmchale: I've emended this; it's now https://qbaylogic.slack.com/files/U77TC97H8/F03JZ50MEHY/untitled.tcl I think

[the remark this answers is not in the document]
vmchale: Ok, this is now based on your script [here](https://qbaylogic.slack.com/files/U77TC97H8/F03JZ50MEHY/untitled.tcl).

tests/src/Test/Tasty/Clash.hs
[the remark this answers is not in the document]
vmchale: Ok, I think I'll keep with a list of `Sim` for now, those cases can be handled with the current setup.

tests/src/Clash/Vivado.hs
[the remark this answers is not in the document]
vmchale: Ok, now I no longer split into VHDL files vs. Verilog files but rather assume based on the project type. I think this is fine since all the tests specify one HDL or the other.

vmchale:
I've disabled several tests: `Parameters`, `BlobVec`, `AndSpecificEnable`, `IntegralTB`. 

The last three fail in simulation, producing different results than the other simulators, which is concerning. `Paramters` fails with

```
ERROR: [XSIM 43-3225] Cannot find design unit topEntity.testBench in library work located at xsim.dir/work.
```

which doesn't seem interesting.

tests/src/Test/Tasty/Vivado/GenTcl.hs
```diff
@@ -0,0 +1,154 @@
...
+if {$myspace::ips ne {}} {
+  read_ip $myspace::ipFiles
+  generate_target {synthesis simulation} [get_ips $myspace::ips]
+}
+|]
+
+  ++ readSources
+
+  ++ [i|
+update_compile_order -fileset [current_fileset]
```
martijnbastiaan, line 134: Could you add a comment here stating that we need to do this due to a bug?

vmchale: Yes!

tests/src/Test/Tasty/Clash.hs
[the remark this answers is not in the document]
martijnbastiaan: I think `HdlSim []` and `NoHdlSim` would have the same meaning, which doesn't sit well to me. Doing the `null` check seems the way to go to me!

tests/src/Test/Tasty/Vivado/GenTcl.hs
```diff
@@ -0,0 +1,155 @@
...
+  hdlDir = directory hdl
+  readSources = case buildTarget hdl of
+    Verilog ->
+     concatMap (\v ->
+      [i|read_verilog ${myspace::hdl_dir}/#{v}\n|]) (snd <$> hdls)
+    VHDL ->
+     concatMap (\(lib, vhdl) ->
+        [i|read_vhdl -library #{lib} ${myspace::hdl_dir}/#{vhdl}\n|])
+      hdls
+    SystemVerilog -> error "SystemVerilog not implemented for Vivado tests"
```
martijnbastiaan, line 155:
```suggestion
    SystemVerilog ->
     concatMap (\v ->
      [i|read_verilog -sv -library #{lib} ${myspace::hdl_dir}/#{v}\n|]) (snd <$> hdls)
```

Should JustWork right?

martijnbastiaan (changes requested):
Overall LGTM, pretty happy to get Vivado in the testsuite! 

I'll have a quick look at the disabled test cases to see if it's just Vivado acting up, or if it warrants an issue report.

martijnbastiaan:
I've looked through the disabled tests, they fall into a few categories:

 * Vivado's `mod` misbehaves. Affected: `IntegralTB`. See https://github.com/clash-lang/clash-compiler/issues/2262.
 * Vivado stumbles upon `finish_and_return`. Affected: `I2Ctest`. We should write an issue for this, we probably shouldn't generate Icarus Verilog exclusive constructs.
 * `BuildSpecific` should be refactored such that it takes a fully qualified Haskell name. Test runners can use Manifest files from there to figure out what arguments to pass to the simulators. Affected: `T1786`, `T2117`, `FIR`. Probably a good idea to make an enhancement issue for this too. No need to fix that in this PR.
 * Vivado can't find `memory.list`. We should copy [what other simulators are doing](https://github.com/clash-lang/clash-compiler/blob/f74584fb26f224f1a7af7e71b250e5e0856f1068/tests/src/Test/Tasty/Iverilog.hs#L96-L99). Affected: `RomFile`, `BlockRamFile`. This should be fixed in this PR.
 * No clue what's happening. Affected: `BlobVec`, `AndSpecificEnable`. We should probably make an issue for both. No need to fix in this PR.
 * There's a weird interaction between the Vivado logic and `Parameters`. Vivado is instructed to simulate `topEntity.testBench`, even though this should be `testBench.testBench`. The latter does happen for other tests; I'm not sure what makes `Parameters` so special. In the interest of time, let's not fix it in this PR, but let's make an issue.
 * Vivado segfaults, which make up all the other failures. These are fine to disabled and don't need a separate issue report.

Here's a patch to get you started: [pointers.diff](https://github.com/clash-lang/clash-compiler/files/9033675/pointers.diff.txt). Let me know if you need help writing the issues :).

tests/Main.hs
```diff
@@ -409,22 +410,24 @@ runClashTest = defaultMain $ clashTestRoot
         , outputTest "MultiResult" def{hdlTargets=[VHDL]}
         , runTest "DSL" def
         , runTest "MultiResult" def
-        , runTest "T919" def{hdlSim=False}
+        , runTest "T919" def{hdlSim=[]}
         , runTest "T1524" def
         , runTest "T1786" def{
             hdlTargets=[VHDL]
           , buildTargets=BuildSpecific ["testEnableTB", "testBoolTB"]
+          , hdlSim=[GHDL, ModelSim] -- triggers error in vivado
```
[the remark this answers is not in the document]
vmchale: It displays a message `Internal Error`! Which I think means "error in Vivado" or the like.

tests/src/Test/Tasty/Vivado/GenTcl.hs
vmchale: SystemVerilog fails at the moment with        

```
ERROR: [Common 17-39] 'update_compile_order' failed due to earlier errors.

              while executing
          "update_compile_order -fileset [current_fileset]"
              (file "/tmp/clash-test-3421926f0b288949/sim.tcl" line 27)
```

I propose we fix the issue where files are not topologically sorted in the `clash-manifest.json` and then `update_compile_order` will not be needed (i.e. don't fix in this PR)

tests/src/Test/Tasty/Clash.hs
```diff
@@ -421,12 +447,14 @@ runTest1 modName opts@TestOptions{..} path target =
     , cgOutputDirectory=tmpDir
     }))
 
-  buildAndSimTests (buildTests, simTests) =
+  buildAndSimTests sim (buildTests, simTests) =
     case (isJust expectClashFail, hdlLoad, hdlSim) of
       (True, _, _) -> []
+      -- ugly, but for Vivado specifically we want to simulate but disable
+      -- building
+      (_, False, sims) | sim `elem` sims && sim == Vivado -> simTests
```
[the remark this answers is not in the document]
vmchale: Yes, sorry. I had this in place for the `DcFifo` case specifically but now that I look at it there has to be a better way (which I can think about in that PR).

tests/Main.hs
```diff
@@ -118,39 +118,40 @@ clashTestGroup testName testTrees =
...
-    , runTest "BlockRamTest" def{hdlSim=False}
+    , runTest "BlockRamTest" def{hdlSim=[]}
     , runTest "Calculator" def
-    , runTest "CHIP8" def{hdlSim=False}
-    , runTest "CochleaPlus" def{hdlSim=False}
+    , runTest "CHIP8" def{hdlSim=[]}
+    , runTest "CochleaPlus" def{hdlSim=[]}
     , let _opts = def { clashFlags=["-fclash-component-prefix", "test"]
                       , buildTargets=BuildSpecific ["test_testBench"]
+                      , hdlSim=[GHDL, ModelSim]
```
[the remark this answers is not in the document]
vmchale: I've added `hdlSimExcludeVivado` which I think covers this!

vmchale:
Thanks! I tried to implement the logic to copy `memory.list` but it is vexing me - even when I see the file in `/tmp/clash-test-f3a06fb30d580034/memory.list` and confirm the temporary directory is `/tmp/clash-test-f3a06fb30d580034` in Tcl, it still fails. I will look to fix it tomorrow.

Also `SystemVerilog` is not up and working as I said above.

tests/src/Test/Tasty/Vivado/GenTcl.hs
martijnbastiaan: Agreed

martijnbastiaan:
Ah the memory file stuff is more complex than I thought:

> The easy way to get memory files working with Vivado is to give them the .mem extension then add them to your project. You do this as you would for a design or simulation source using “Add Sources” then selecting “Files of type: Memory Initialization Files”. Vivado will automatically identify them as memory files and place them in the current working directory during simulation etc.

https://projectf.io/posts/initialize-memory-in-verilog/

For TCL I guess this means using `read_mem` (see [Vivado Design Suite Tcl Command Reference Guide](https://docs.xilinx.com/v/u/2019.2-English/ug835-vivado-tcl-commands)). In that case.. just disable them for now and make an issue. Let's discuss later with @DigitalBrains1 what to do.

Thanks for reporting the issues!

tests/src/Test/Tasty/Vivado/GenTcl.hs
DigitalBrains1: I don't think that's the actual error message though, I think it's the "earlier errors" that are the problem?

It might be related to libraries and such.

vmchale:
Great, sounds good!

tests/src/Test/Tasty/Clash.hs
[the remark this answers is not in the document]
DigitalBrains1: Haha yes I suppose it would! I was just going for easy readability.

DigitalBrains1:
> Could you squash the commits to a single one? The default GitHub squash feature produces some [pretty ugly commit messages](https://github.com/clash-lang/clash-compiler/commit/60101b06461e3996c8d0f0a7bff71654c9610719), so I'd like to prevent that this time.

You can edit the commit message that GitHub's squash feature produces to your heart's content! That text there is just the suggestion that GitHub gives for the message, it gives you a text box to change it.

.ci/gitlab/test.yml
```diff
@@ -83,14 +83,53 @@ prelude:doctests:
...
-    - cabal v2-run -- clash-testsuite -j$THREADS -p .SystemVerilog --hide-successes --no-modelsim
+    - cabal v2-run -- clash-testsuite -j$THREADS -p .SystemVerilog --hide-successes --no-modelsim --no-vivado
+
+# Vivado is quite slow, so we only run a subset of the tests on development branches
+# with it. The full testsuite gets run with Vivado every night on 'master'.
+suite:cores:
+  extends: .test-common-local
+  script:
+    - source /opt/tools/Xilinx/Vivado/2022.1/settings64.sh
+    - XILINX_LOCAL_USER_DATA=no cabal v2-run -- clash-testsuite -j$THREADS -p Cores --hide-successes --no-modelsim --no-ghdl --no-iverilog --no-verilator --no-symbiyosys
```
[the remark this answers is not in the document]
vmchale: Ah fair enough. Will do, one moment.

DigitalBrains1:
Here's the first part of my review, in several messages. I'm using commits rather than GitHub's code suggestions feature; this was more practical to me and you could just pick commits if you agree.

The first commit is [this](https://github.com/clash-lang/clash-compiler/commit/8673c29044167267b3145231bbd248313e65ec1f): _Quote generated strings, construct paths_

It fixes two issues. We might generate file or module names with Tcl special characters; any strings we generate should be quoted in Tcl.

Furthermore, the path to `clash_namespace.tcl` should not contain hardcoded path separators: I wonder what Windows would do with mixed path separators. Let's not find out.

DigitalBrains1:
The second commit is [this](https://github.com/clash-lang/clash-compiler/commit/1f493e1d0839ccfab35e6343c04ef8cb21450788). The `tests/data-files/tcl/clash_namespace.tcl` file was still a file with DOS line ends, so firstly I fixed that. That means the diff is pretty useless, here's the actual diff:
```patch
--- clash_namespace.tcl.old     2022-07-05 16:27:54.382560339 +0200
+++ clash_namespace.tcl 2022-07-05 16:19:23.599559036 +0200
@@ -6,6 +6,11 @@
         # separate its definitions from ours and reduce the chance of
         # accidentally corrupting our code.
         namespace eval tmp "source $clashTclFile"
+        if {[info procs tmp::createNamespace] eq {}} {
+            error "Error: $clashTclFile does not provide a procedure named\
+                \"createNamespace\". The Tcl script does not conform to the\
+                defined Clash<->Tcl API."
+        }
         set baseName [file rootname [file tail $clashTclFile]]
         tmp::createNamespace [namespace current]::tclIface::${top}::${baseName}
         namespace delete tmp
@@ -32,11 +37,21 @@
                     error "Error: ${tclIface}::scriptPurpose bogus value\
                         \"$purpose\"."
                 }
+                # In Tcl, you can call procedures with a partial name. So an
+                # invocation of "createIp" could call "createIpAlt" if
+                # "createIp" did not exist. Let's be strict here to prevent
+                # confusion: only accept the exact name "createIp".
+                set createIp ${tclIface}::createIp
+                if {[info procs $createIp] eq {}} {
+                    error "Error: $tclIface doesn't provide \"createIp\"\
+                        procedure."
+            }
+
                 set ipName [subst $${tclIface}::ipName]
                 if {$ipName in $seen} {
                     continue
                 }
-                ${tclIface}::createIp $ipName -dir ip
+                $createIp $ipName -dir ip
                 lappend seen $ipName
             }
         }
```

I improved the error handling in `loadTclIface`, hopefully giving a relatively user-friendly error message when the generated Tcl is wrong.

I also added more checking to `runClashScripts` because I was not aware (or had forgotten) that you can actually invoke Tcl procedures with a partial name as long as it is unique; this would make debugging very confusing if there is another procedure that happens to have a name that starts with `createIp`.

DigitalBrains1:
Then [this commit](https://github.com/clash-lang/clash-compiler/commit/8054139984080432ea98df6a3688dbb890ee232f); the previous two I consider essential, this one is a suggestion (though I really like it better :-).

We have two languages at our disposal here: Haskell and Tcl. Several things could be solved in either language and the end result would be semantically identical. But I thought you mentioned you'd like to do as much as possible in Haskell. I am definitely of the opinion you need to make a choice: do I use Haskell where possible and only Tcl where needed, or do I put most things in Tcl?

If you're doing as much as possible in Haskell, I‍ don't see a reason to use `file mkdir` in Tcl; just use `System.Directory.createDirectory` in Haskell. That's what this commit does.

DigitalBrains1:
[The next commit](https://github.com/clash-lang/clash-compiler/commit/4aadf2b03332502ce34ee8e45332ddcabe4d9976) is a continuation of our discussion [here](https://github.com/clash-lang/clash-compiler/pull/2257#discussion_r906024918). I'd already accepted your original version of that code, but had not mentioned that yet. But now that you've changed it to what it is now, I'd like to take it a bit further.

We get all files in a nice reverse-topological order (modulo bug we'll fix soon). Currently we don't generate any Tcl files that need any ordering at all, but in the future we might. It doesn't really make sense to me to throw away a perfectly good reverse-topo sort we might need later. So I changed the code to keep the order of all files, including the Tcl. And I quite like to annotate the type of the file in the data structure built; we can base decisions on that information.

This commit also fixes a misunderstanding that was not fixed yet. The first argument to the Tcl `clash::loadTclIface` is the _library_ where the generated Tcl script resides. Its only purpose so far is preventing name clashes between IPs with the same Tcl script name in different libraries; and those are filtered out later anyway as their `$ipName` will be the same as well. But future Tcl scripts might be properly bound to a library or occur multiple times. If you don't want to accept this commit, we will need to change the Tcl code to not have this argument at all because passing it anything else (`Test.Tasty.Vivado.GenTcl.parseTcl`) doesn't serve a purpose either.

Regarding `error "SystemVerilog not implemented for Vivado tests"`, we don't generate any SystemVerilog tests for Vivado; it seems like its main purpose was to have the `case` be exhaustive. If you somehow were to contrive to give the code in this commit any `.sv` files, they would be skipped just like all other files with extensions we don't handle (`.sdc` for example), albeit in a different place. So I removed the error. Keeping `Clash.Annotations.Primitive.HDL` around seems superfluous. If you really want to error on SystemVerilog, you can error when you encounter a `.sv` file.

Housekeeping: I moved `Test.Tasty.Vivado.GenTcl.asSimProj` up in the file because it makes the order a nice grouping. Its code is unchanged. Also, the Haddock docs for the first two arguments to `mkTcl` in the same file were accidentally on the second and third instead.

I still have more to review and comment, but the day is over. I'll continue tomorrow.

DigitalBrains1:
> (It's listing me as a coauthor because I cherry-picked.... ah well).

You can change it to whatever you want when you squash.

vmchale:
Looks good to me; keeping the ordering will hopefully pan out soon :crossed_fingers:

tests/data-files/tcl/clash_namespace.tcl
```diff
@@ -0,0 +1,60 @@
+namespace eval clash {
```
DigitalBrains1, line 1: Let's slap a licence on there so people can copy it and derive from it with peace of mind.

```suggestion
# Copyright   :  (C) 2021-2022, QBayLogic B.V.,
#                    2022     , Google Inc.
# License     :  BSD2 (see the file LICENSE)
# Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>

namespace eval clash {
```

tests/src/Test/Tasty/Vivado/GenTcl.hs
```diff
@@ -0,0 +1,151 @@
...
+
+import Clash.Driver.Manifest (Manifest (..))
+import Test.Tasty.Common (getManifests)
+
+import Data.Char (isLower)
+import Data.List (isSuffixOf)
+import Data.String.Interpolate (i)
+
+import System.FilePath ((</>))
+import qualified Data.Text as T
```
DigitalBrains1, line 19: Our style guide would like us to order them differently:
```suggestion
import Data.Char (isLower)
import Data.List (isSuffixOf)
import qualified Data.Text as T
import Data.String.Interpolate (i)
import System.FilePath ((</>))

import Clash.Driver.Manifest (Manifest (..))

import Paths_clash_testsuite (getDataDir)
import Test.Tasty.Common (getManifests)
```

vmchale: Oh you're right, sorry :grimacing:

DigitalBrains1:
And [this commit](https://github.com/clash-lang/clash-compiler/commit/691e67c7a327461a99b4b0025b69a474d9c5e275) changes the layout of the comments in `tests/Main.hs`. I had trouble making out which tests the comments applied to, so I tried to work out a more visually distinguishing formatting. Also, this one was rather hilarious:
```haskell
            -- TODO: When issue #2039 is fixed, it should be possible to drop
            -- compile-ultra.
          , runTest "BlobVec" def{clashFlags=["-fclash-compile-ultra"], hdlSim=hdlSimExcludeVivado}
            -- TODO: Vivado is disabled because it gives different results, see
            -- https://github.com/clash-lang/clash-compiler/issues/2268
```
This test is bracketed by TODOs :-D.

I made some minor changes as well. I added a TODO to `Cores.Xilinx.Floating` and removed a superfluous HTML anchor from the comment for `examples.FIR`.

vmchale:
Looks good to me!

I think you need to approve for it to be merged, but in any case thank you for your review + patience :)

DigitalBrains1:
I was just trying to get a PR fixing the topological sort out first; if we merge that, we can drop the `update_compile_order` right away! Is that okay?

DigitalBrains1:
Ah you accidentally picked a merge commit instead of a squash commit. Ah well.

tests/src/Test/Tasty/Clash.hs
```diff
@@ -137,7 +140,7 @@ allTargets = [VHDL, Verilog, SystemVerilog]
 instance Default TestOptions where
   def =
     TestOptions
-      { hdlSim=True
+      { hdlSim=[GHDL, ModelSim, Vivado]
```
DigitalBrains1, line 143: It is no longer running simulation in Verilator because of this default. It's still running the `make` step but not the `sim` step. Before this PR:
```
$ cabal run clash-testsuite -- -p DivZero -j8
. 
  tests
    shouldwork
      Basic
        DivZero
          VHDL
            clash (gen):                OK (1.60s)
            GHDL (import):              OK
            ghdl (make testBench):      OK (0.55s)
            ghdl (sim testBench):       OK
          Verilog
            clash (gen):                OK (1.69s)
            iverilog (make testBench):  OK
            iverilog (sim testBench):   OK
            verilator (make testBench): OK (0.95s)
            verilator (sim testBench):  OK
          SystemVerilog
            clash (gen):                OK (1.63s)
            modelsim (vlib):            OK (0.03s)
            modelsim (vlog):            OK (0.08s)
            modelsim (sim testBench):   OK (0.60s)
            verilator (make testBench): OK (1.09s)
            verilator (sim testBench):  OK

All 15 tests passed (2.74s)
```
after this PR:
```
$ cabal run clash-testsuite -- -p DivZero -j8
. 
  tests
    shouldwork
      Basic
        DivZero
          VHDL
            clash (gen):                OK (1.76s)
            GHDL (import):              OK (0.02s)
            ghdl (make testBench):      OK (0.72s)
            ghdl (sim testBench):       OK
            Vivado:                     OK (21.37s)
          Verilog
            clash (gen):                OK (1.73s)
            iverilog (make testBench):  OK
            iverilog (sim testBench):   OK
            verilator (make testBench): OK (1.25s)
            Vivado:                     OK (21.40s)
          SystemVerilog
            clash (gen):                OK (1.69s)
            modelsim (vlib):            OK (0.02s)
            modelsim (vlog):            OK (0.06s)
            modelsim (sim testBench):   OK (0.75s)
            verilator (make testBench): OK (1.22s)

All 15 tests passed (23.15s)
```

DigitalBrains1: I also think the
```haskell
-- | Specify whether a test will use a traditional simulator, verilator, or
-- both when executing. By default both are used, but this can be changed when
-- a test is only relevant with one type of tool.
data Verilate = SimAndVerilate | VerilateOnly | SimOnly
```
functionality could be folded into `hdlSim`.

tests/src/Test/Tasty/Clash.hs
```diff
@@ -449,26 +477,33 @@ runTest1 modName opts@TestOptions{..} path target =
...
         case verilate of
           VerilateOnly -> []
-          _ -> buildAndSimTests (verilogTests opts tmpDir)
+          _ -> buildAndSimTests ModelSim (verilogTests opts tmpDir)
 
       SystemVerilog ->
         case verilate of
           VerilateOnly -> []
-          _ -> buildAndSimTests (systemVerilogTests opts tmpDir)
+          _ -> buildAndSimTests ModelSim (systemVerilogTests opts tmpDir)
```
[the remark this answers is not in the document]
vmchale: I say another PR.

PR #2258, opened 2022-06-24
Don't terminate Verilog simulator on error in assertion functions

acairncross:
Simulators (at least ModelSim) terminate immediately on `$finish` making it impossible to set e.g. an exit code to flag the error to the shell. I also noticed that the SystemVerilog versions of these primitives do already use `$stop` instead.

## Still TODO:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,7 @@ jobs:
           - name: GHC 9.0.2, Multiple Hidden
             ghc: 9.0.2
             multiple_hidden: yes
+            workaround_ghc_mmap_crash: yes
 
           - name: GHC 9.2.5, Multiple Hidden
             ghc: 9.2.5
@@ -138,6 +139,7 @@ jobs:
       env:
         THREADS: 2
         CABAL_JOBS: 2
+        WORKAROUND_GHC_MMAP_CRASH: ${{ matrix.workaround_ghc_mmap_crash }}
         MULTIPLE_HIDDEN: ${{ matrix.multiple_hidden }}
         CI_COMMIT_BRANCH: ${{ github.base_ref }}
 
--- a/changelog/2023-05-17T13_42_59+02_00_use_stop_in_verilog
+++ b/changelog/2023-05-17T13_42_59+02_00_use_stop_in_verilog
@@ -1,0 +1,1 @@
++CHANGED: Use `$stop` in `assert` instead of `$finish`. This harmonizes VHDL and SystemVerilog simulations and will properly set an exit code. See [#2258](https://github.com/clash-lang/clash-compiler/pull/2258).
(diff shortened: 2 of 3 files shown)


martijnbastiaan:
@kloonbot run_ci aedbd33ced4a2f8ef96a00e359af2221250b3398

alex-mckenna:
Worth mentioning that this setting the exit code is something simulator-specific still. The only difference between `$stop` and `$finish` in the standard is that `$finish` exits the simulator process and `$stop` suspends the simulator and drops into an interactive prompt. The Verilog standard (at least the 2001 revision) gives you no mechanisms to stop with a particular exit code.

I don't think it would be an issue for CI (as CI should timeout if a test ever enters this prompt and fail anyway), but it might be worth leaving a note in the testsuite somewhere that things like failing assertions will now make tests run indefinitely _unless_ stopped with a timeout.

Another thing: we have [a primitive](https://github.com/clash-lang/clash-compiler/blob/df3d5d1484413f2366e64804dab760fdb57edf14/clash-lib/prims/commonverilog/Clash_Explicit_SimIO.primitives.yaml#L18) that use an `iverilog`-specific extension, `$finish_and_return` which is `$finish` where the argument is the exit code to return from the process. Should this also be changed to `$stop` for consistency? I think nobody has considered this since it's a more niche feature of Clash but this is already broken for all non-`iverilog` simulators and making it consistent in this PR might be a nice unexpected bonus

DigitalBrains1:
> Worth mentioning that this setting the exit code is something simulator-specific still. 

And if anyone knows how to do that in Vivado, I'm all ears! If you also know it for VHDL, I'm even more ears. I'm hoping for 200% ears here.

I'm selfishly declaring this on-topic enough to raise here. Can I do that? We'll see.

martijnbastiaan:
> I don't think it would be an issue for CI (as CI should timeout if a test ever enters this prompt and fail anyway), but it might be worth leaving a note in the testsuite somewhere that things like failing assertions will now make tests run indefinitely unless stopped with a timeout.

I think we'll just have to test it. You'd hope that the simulators would detect they're running in a  non-interactive environment.

martijnbastiaan (approved):
... in any case :heavy_check_mark:

DigitalBrains1:
Seeing Alex's comment, maybe the changelog message should be tweaked a bit regarding
> will properly set an exit code.

DigitalBrains1:
> I think we'll just have to test it. You'd hope that the simulators would detect they're running in a non-interactive environment.

I'd like to know what this test shows! Create a deliberately failing test based on this PR, and see what it does in the different simulators we run on CI.

martijnbastiaan:
Unfortunately, Icarus Verilog will just spin forever if we use `$stop` instead of `$finish`..

DigitalBrains1:
Perhaps we could tune it with `-fclash-hdlsyn` (we'd need to add `iverilog`).

martijnbastiaan:
I've moved this to an issue, given that this PR itself is not the solution and there is no obvious way to fix it.. https://github.com/clash-lang/clash-compiler/issues/2752.

PR #2259, opened 2022-06-28
Conditional assignments now declare their usage

vmchale:
`boolFromBit` etc. were broken by the recent work on implicit netlist usage, since there was no way that "internals" of a black box would be accounted for.

This is necessary for https://github.com/clash-lang/clash-compiler/pull/2182 and fixes Verilog synthesis. 

This may cause problems with uniqueness/identifiers since the `UsageMap` uses `Text` keys, and this does not address that.  

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Netlist/Util.hs
+++ b/clash-lib/src/Clash/Netlist/Util.hs
@@ -1148,7 +1148,7 @@ canUse _ _ = \case
 declareUse :: Usage -> Identifier -> NetlistMonad ()
 declareUse u i = usageMap %= Map.insertWith (<>) (Id.toText i) u
 
--- | As a sanity check, this will throw an exception if we run into a name
+-- | Like 'declareUse', but will throw an exception if we run into a name
 -- collision.
 declareUseOnce :: HasUsageMap s => Usage -> Identifier -> State.State s ()
 declareUseOnce u i = usageMap %= Map.alter go (Id.toText i)


clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -433,9 +438,13 @@ enableToBit bitName = \case
         [ (Just (BoolLit True), Literal Nothing (BitLit H))
         , (Nothing            , Literal Nothing (BitLit L))
         ]
+    declareUseOnce (Proc Blocking) uniqueBitName
     pure texp
   tExpr -> error $ "enableToBit: Got \"" <> show tExpr <> "\" expected Enable"
 
+declareUseOnce :: HasUsageMap s => Usage -> Identifier -> State s ()
+declareUseOnce u i = usageMap %= Map.insert (Id.toText i) u
```
martijnbastiaan, line 446: This should be moved next to `declareUse`. Could you document the differences?

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -456,6 +465,7 @@ boolFromBit boolName = \case
         [ (Just (BitLit H), Literal Nothing (BoolLit True))
         , (Nothing        , Literal Nothing (BoolLit False))
         ]
+    declareUseOnce (Proc Blocking) uniqueBoolName
```
martijnbastiaan, line 468: I don't think we ever build clocked processes with this DSL, so I would expect `Proc NonBlocking` here. Does it fail if you do?

clash-lib/src/Clash/Primitives/DSL.hs
martijnbastiaan: Also, with the "once" in the name I'd expect it to crash if there was already an identifier with the same name in the map.

vmchale: Oh fair enough!

clash-lib/src/Clash/Primitives/DSL.hs
vmchale: I think you're right, I looked at https://github.com/clash-lang/clash-compiler/blob/df3d5d1484413f2366e64804dab760fdb57edf14/clash-lib/src/Clash/Netlist/BlackBox.hs#L226 and I think `Proc NonBlocking` will do exactly what we want across VHDL and Verilog.

martijnbastiaan (approved):
Small change requested, but otherwise LGTM.

I do feel uncomfortable with the way we declare usage now. If it's this easy to forget, we should probably hide constructors that allow us to do so.

PR #2271, opened 2022-07-06
Partially fix topological sort of files

DigitalBrains1:
We don't track dependency information for include files and memory
files. Include files are generated from
`Clash.Netlist.BlackBox.Types.bbIncludes` or YAML/JSON `includes` keys,
and memory files are generated from `~TEMPLATE` holes in the primitive
template language. We should actually track these dependencies to
maintain a proper reverse topological sort order in our manifest files.
However, this commit makes the problem less severe. We used to list
these include and memory files last in the list, but it is extremely
likely that the HDL file that caused the include/memory file to be
generated depends on it. So by listing them first instead of last, we
fix the sort order for the common case that the include/memory files
_themselves_ do not depend on anything else.

To be precise, we sort the include files after the memory files because
it is probably difficult to impossible to construct a memory file that
depends on an include file, but the reverse is easily achieved.

Additionally, experimentation showed that the note stating transitive
dependencies in the manifest file are sorted topologically (rather than
_reverse_ topologically) is incorrect: dependencies are sorted like
files and components.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver.hs
+++ b/clash-lib/src/Clash/Driver.hs
@@ -479,9 +479,11 @@ generateHDL env design hdlState typeTrans peEval eval mainTopEntity startTime = 
       let
         components = map (snd . snd) (OMap.assocs netlist)
         filesAndDigests0 =
-             zip (map fst hdlDocs) hdlDocDigests
+          -- FIXME: We should track dependencies of `mfiles` and `dfiles` and
+          -- maintain the proper topological sort of all these.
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
For some reason this does not actually move include files earlier in the list. Further investigation is needed :-(.

martijnbastiaan:
Superseded by https://github.com/clash-lang/clash-compiler/pull/2347

PR #2274, opened 2022-07-08
Enable running the clash testsuite on windows

christiaanb:
Assumes the Cabal store is installed in `C:\cabal\store`, which is the `ghcup` default.

## Still TODO:

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -40,7 +40,10 @@ cabalClashBinDir :: IO String
 cabalClashBinDir = makeAbsolute rel_path
  where
   rel_path = printf templ platform ghcVersion3 (VERSION_clash_ghc :: String)
-  platform = "x86_64-linux" :: String -- XXX: Hardcoded
+  platform :: String -- XXX: Hardcoded
+  platform = case os of
+     "mingw32" -> arch <> "-windows"
+     _ -> arch <> "-" <> os
   templ = "dist-newstyle/build/%s/ghc-%s/clash-ghc-%s/x/clash/build/clash/" :: String
 
 -- | Set GHC_PACKAGE_PATH for local Cabal install. Currently hardcoded for Unix;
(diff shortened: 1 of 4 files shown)


martijnbastiaan:
Cool. I think there's two parts to this:

 * You made Verilator work under Windows
 * You made `clash-testsuite` work under Windows+Cabal

It's probably neatest to split this into two commits. I also have good news though :eyes:, `clash-testsuite` + IVerilog already worked on Windows when using Stack, so extending CI to test with Cabal is really easy: 

https://github.com/clash-lang/clash-compiler/blob/d5c021acf6b55b232289b37ea3dd28e9944e28d3/.github/workflows/ci.yml#L64-L65

Just add a `cabal run` there too.

christiaanb:
Sure, I can split it into multiple commits. 

I didn't actually meant to turn this into a proper PR, but a draft PR at first. Since we don't actually run the verilator tests since https://github.com/clash-lang/clash-compiler/pull/2257 was merged, I wanted to wait until that's fixed.

martijnbastiaan:
Sure!

martijnbastiaan:
@christiaanb Could you try adding  `cabal update`? If that doesn't work feel free to merge.

christiaanb:
But won’t I need to setup Cache as well? And how do I cache stuff on Windows, but not on Mac? Should I just setup a new matrix for Windows + Cabal?

christiaanb:
@DigitalBrains1 yeah, I’m waiting for that.

martijnbastiaan:
> But won’t I need to setup Cache as well? And how do I cache stuff on Windows, but not on Mac? Should I just setup a new matrix for Windows + Cabal?

Ugh good points. There's some good strategies for this, but I'll look into it later.

martijnbastiaan (approved):

tests/Main.hs
```diff
@@ -40,7 +40,10 @@ cabalClashBinDir :: IO String
 cabalClashBinDir = makeAbsolute rel_path
  where
   rel_path = printf templ platform ghcVersion3 (VERSION_clash_ghc :: String)
-  platform = "x86_64-linux" :: String -- XXX: Hardcoded
+  platform :: String -- XXX: Hardcoded
```
DigitalBrains1, line 43: Shouldn't the comment have been removed? Or do you still consider this hardcoding, even though softer? :-)

I could add it to a new incarnation of `random-stuff`.

christiaanb: I still consider it hardcoding, preferably we would get this string through some API call in the `Cabal` library.

PR #2276, opened 2022-07-08
Allow build on more recent OSX (M1 darwin 12.4)

vmchale:
Recent OSX development tools require you to use various framework flags, see [this blog post](https://gpanders.com/blog/exploring-mach-o-part-1/). 

## Still TODO:

  - [ ] ~Write a changelog entry (see changelog/README.md)~ not uploaded to Hackage

### DIFF
--- a/clash-cosim/src/cbits/Makefile
+++ b/clash-cosim/src/cbits/Makefile
@@ -7,7 +7,7 @@ CFLAGS = -Wall -O2
 
 UNAME = $(shell uname)
 ifeq ($(UNAME),Darwin)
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
Hmm, so I guess we're not building `clash-cosim` on CI, hence never having seen the error before?

vmchale:
I think it does build at least since clash-testsuite depends on it! But I guess OSX is older? Or maybe `stack` does something.

martijnbastiaan (approved):
Alright, well given that the older mac os versions don't seem to break on this (as per CI), I guess this is good :)

PR #2281, opened 2022-07-12
Use newer bash on Nix CI builds

christiaanb:
The 4.4 branch no longer exist on nixpkgs 22.05 or higher

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -68,7 +68,7 @@ stack-build:
   tags:
 
 nix-build:
-  image: nixos/nix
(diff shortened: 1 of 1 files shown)


martijnbastiaan (approved):
Given that we don't really care about the Nix image, we should probably lock it to a specific version. Could you change the `image: `  field to `nixos/nix:2.10.1`? Otherwise LGTM.

leonschoorl:
Can't we just ask nix for bash, without specifying a specific version?
I'd think there is no need for a specific version.

PR #2284, opened 2022-07-16
Use `sequentialTestGroup` to describe dependencies in `clash-testsuite`

martijnbastiaan:
See https://github.com/UnkindPartition/tasty/pull/343. This promises to do a few things:

* Reduce the startup time of `clash-testsuite` from ~half a minute to sub second figures
* Reduce complexity in `clash-testsuite` when specifying dependencies
* Make Tasty account for dependencies when using filters (`-p foo`) - this will make its command line suggestions upon a failing test more accurate

## Still TODO:

### DIFF
--- a/tests/src/Test/Tasty/Clash.hs
+++ b/tests/src/Test/Tasty/Clash.hs
@@ -12,7 +12,6 @@ import           Clash.Annotations.Primitive (HDL(..))
 import           Data.Coerce               (coerce)
 import           Data.Default              (Default, def)
 import qualified Data.List                 as List
-import           Data.List                 (intercalate)
 import           Data.Maybe                (isJust)
 import qualified Data.Text                 as T
 import qualified System.Directory          as Directory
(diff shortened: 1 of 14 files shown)


tests/src/Test/Tasty/Clash.hs
```diff
@@ -407,63 +336,39 @@ runTest1
...
-    -- ^ All the independent trees to run after "clash (gen)"
-    -> [TestTree]
-  sequenceIndependent gen = go
+    })
+
+  emptyGroup = testGroup "empty" []
+
+  buildAndSimTests sim (buildTests, simTests0)
+    | isJust expectClashFail = emptyGroup
+    | sim `notElem` hdlLoad  = emptyGroup
```
martijnbastiaan, line 362: There's a very slight change in behavior here: if a simulator isn't in `hdlLoad`, simulation also won't run. Is that fair @DigitalBrains1, or is there some subtlety I'm overlooking?

DigitalBrains1: Yeah, it's fair. It used to be (well for the past 10 days :-P ) that the Vivado tests wholly ignored the value of `hdlLoad`, but I still made sure `tests/Main.hs` correctly set `hdlLoad` to include Vivado or not. Since you also edited the Haddock for `vivadoTests`, all is well.

alex-mckenna:
I know this is still draft, but I :+1: so much

martijnbastiaan:
Haha tnx, I'm hoping the PR over at Tasty gets merged any time now..

alex-mckenna (approved):
Broadly speaking, this looks good. Let's open an issue now to replace tasty from GitHub with a version from (h/st)ackage when released

PR #2286, opened 2022-07-19
Make sure fromSLV doesn't cause out of bound errors in VHDL simulation

### LINKED ISSUE #2220: Generated fromSLV isn't always total for sum types
I'm still getting reports that the vhdl generated from working Clash code is erroring at runtime in GHDL. The "army of testers" (hah!) is in a different timezone so it takes time to interact to pin it down and I do not know yet if the bug is here or elsewhere, but there is one. 

GHDL is reporting "bound check failure" (which in earlier reports turned out to be a negative output from to_integer to somewhere not expecting it) in a generated library function, I think for Enum (or BitPack?):

> function fromSLV (slv : in std_logic_vector) return A_Index is
>  begin
>    return A_Index'val(to_integer(unsigned(slv)));   <-- HERE
>  end;

The A_Index target is a 'derive Enum'd data type consisting of just under 100 abstract values
> data A_Index  = A_J | ... | A_ILL

and it is represented in the vhdl by
> type A_Index is (A_J, ..., A_ILL);

I believe slv as the fromSLV input must be a 7-bit bitvector in principle. My question is if "you" (collectively) see any scope for say, slv having the top bit set, leading perhaps to a negative result as the input to A_Index'val, which might explain things?

It doesn't happen when clash runs its own code, from which the vhdl is generated.

I think this used to be done via an array in earlier clash versions?

There will be further experiments done this evening that may pin the problem more exactly.

Any comments meanwhile?

Regards

PTB

leonschoorl:
Fixes #2220

I've wrapped the extra checks in `translate_off/on` with the idea of not adding extra logic for it in hardware.
As the result shouldn't be used, because it'll throw an exception in Haskell.
But I'm not a 100% sure about this, as it (potentially) introduces differences between the resulting hardware and a VHDL simulation.

`toEnum @A_Index 3` will produce:
  * in Haskell: `Exception: toEnum{A_Index}: tag (3) is outside of enumeration's range (0,2)`
  * in VHDL simulation `A_0`
  * in hardware: ???

-----

### DIFF
--- a/changelog/2022-07-19T19_23_48+02_00_fix_toEnum_OOB_errors
+++ b/changelog/2022-07-19T19_23_48+02_00_fix_toEnum_OOB_errors
@@ -1,0 +1,1 @@
+FIXED: Fix out of bound errors in toEnum/fromSLV for sum types[#2220](https://github.com/clash-lang/clash-compiler/issues/2220)
--- a/clash-lib/src/Clash/Backend/VHDL.hs
+++ b/clash-lib/src/Clash/Backend/VHDL.hs
@@ -760,11 +760,25 @@ funDec (RenderEnums enums) _ t@(Sum _ _) | enums = Just
     "function" <+> "fromSLV" <+> parens ("slv" <+> colon <+> "in" <+> "std_logic_vector") <+> "return" <+> qualTyName t <+> "is" <> line <>
     "begin" <> line <>
     indent 2
-      ( "return" <+> qualTyName t <> "'val" <>
-        parens ("to_integer" <>
-          parens ("unsigned" <> parens "slv"))) <> semi <> line <>
+      (
+      translate_off (
+      "if unsigned(slv) <= " <> qualTyName t <> "'pos("<> qualTyName t <> "'high) then"
+      ) <> line <>
+        indent 2
+          ( "return" <+> qualTyName t <> "'val" <>
+            parens ("to_integer" <>
+              parens ("unsigned" <> parens "slv"))) <> semi <> line <>
+      translate_off (
+        "else" <> line <>
+        indent 2
+          ( "return" <+> qualTyName t <> "'val(0)") <> semi <> line <>
+        "end if" <> semi
+      )
+      ) <> line <>
     "end" <> semi
   )
+  where
+    translate_off body = "-- pragma translate_off" <> line <> body <> line <> "-- pragma translate_on"
 
 funDec _ syn t@(Vector _ elTy) = Just
   ( "function" <+> "toSLV" <+> parens ("value : " <+> qualTyName t) <+> "return std_logic_vector" <> semi <> line <>
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -610,6 +610,7 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "T2046C" def{hdlSim=[],clashFlags=["-Werror"],buildTargets=BuildSpecific["topEntity"]}
         , runTest "T2097" def{hdlSim=[]}
         , runTest "T2154" def{hdlTargets=[VHDL], hdlSim=[]}
+        , runTest "T2220_toEnumOOB" def{hdlTargets=[VHDL]}
         , runTest "T2272" def{hdlTargets=[VHDL], hdlSim=[]}
         ] <>
         if compiledWith == Cabal then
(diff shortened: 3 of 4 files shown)


DigitalBrains1:
When we looked at it, the code generated for `toEnum` already included a range check and a default value if outside of range. This also happens for the test case you added:

<details><summary>Generated VHDL</summary>

<br/>

`topEntity.vhdl`
```vhdl
-- Automatically generated VHDL-93
library IEEE;
use IEEE.STD_LOGIC_1164.ALL;
use IEEE.NUMERIC_STD.ALL;
use IEEE.MATH_REAL.ALL;
use std.textio.all;
use work.all;
use work.T2220_toEnumOOB_topEntity_types.all;

entity topEntity is
  port(x      : in std_logic_vector(1 downto 0);
       result : out T2220_toEnumOOB_topEntity_types.Maybe);
end;

architecture structural of topEntity is
  -- T2220_toEnumOOB.hs:7:1-9
  signal wild1             : signed(63 downto 0);
  signal \c$case_alt\      : T2220_toEnumOOB_topEntity_types.A_Index;
  signal \c$case_scrut\    : boolean;
  signal \c$case_alt_0\    : T2220_toEnumOOB_topEntity_types.A_Index;
  signal \c$case_scrut_0\  : boolean;
  signal \c$case_alt_1\    : T2220_toEnumOOB_topEntity_types.A_Index;
  signal \c$app_arg\       : signed(63 downto 0);
  signal \c$app_arg_0\     : signed(63 downto 0);
  -- T2220_toEnumOOB.hs:7:1-9
  signal \c$wild1_app_arg\ : signed(63 downto 0);

begin
  wild1 <= \c$wild1_app_arg\;

  \c$case_alt\ <= \c$case_alt_0\ when \c$case_scrut\ else
                  A_Index'val(0);

  \c$case_scrut\ <= tagToEnum(\c$app_arg_0\);

  \c$case_alt_0\ <= \c$case_alt_1\ when \c$case_scrut_0\ else
                    A_Index'val(0);

  \c$case_scrut_0\ <= tagToEnum(\c$app_arg\);

  \c$case_alt_1\ <= T2220_toEnumOOB_topEntity_types.fromSLV(std_logic_vector(resize(unsigned(std_logic_vector(wild1)),2)));

  \c$app_arg\ <= to_signed(1,64) when wild1 <= to_signed(2,64) else to_signed(0,64);

  \c$app_arg_0\ <= to_signed(1,64) when wild1 >= to_signed(0,64) else to_signed(0,64);

  \c$wild1_app_arg\ <= signed(std_logic_vector(resize(unsigned(x),64)));

  with (x) select
    result <= std_logic_vector'("0" & "--") when "11",
              std_logic_vector'("1" & (T2220_toEnumOOB_topEntity_types.toSLV(\c$case_alt\))) when others;


end;
```
</details>

Do you think it's even possible for the new code path to produce a value that is actually used or will it always be wrapped inside an equivalent range check in practice, meaning the GHDL bounds check failure is just because of eagerness?

The basic issue is of course VHDL enumeration's inability to express undefined values, but if it's in practice always wrapped then the generated code after this PR is functionally identical to before, and this PR makes simulation and synthesis no worse.

christiaanb:
I'm okay with the `translate off/on`, we only use the `fromSLV` in a couple of places:

1. When generating VHDL for vivado: converting the SLV value stored in the RAM/ROM array to the actual type
2. When generating VHDL for vivado: converting the SLV value in the Vec/RTree to the actual type when extracting an element
3. Converting `tagToEnum#`: this would throw an exception in Haskell when the Int doesn't actually map to the constructor
4. On the boundaries of Clash `Synthesis` regions

So for 1, 2 and 4 the only time you would pick the non-happy path through `fromSLV` is when you integrate a Clash design in a larger VHDL design, and feed it "illegal" values in from your VHDL. For 3 I think the hardware mapping an "illegal" value to some "unknown/undefined" value is fine. In none of the cases I would want an additional multiplexer in my path simply because we instructed Clash to render Haskell enums to VHDL enums.

DigitalBrains1:
Good analysis. I would think that case 3 would still lead to an undesired test-and-mux because `tagToEnum#` already creates these out-of-bounds checks itself. And it's operating on a signed number, so it tests both the MSB and the maximum value.

But maybe we could make mapping enumerations (that includes `Bool`) to `std_logic{,_vector}` the default later and then people only get the mux when they explicitly request VHDL enumerations, because the tests for out-of-bounds mux to an X value which the synthesis tool should be able to eliminate.

DigitalBrains1 (changes requested):

christiaanb:
Indeed, the Core for `toEnum` will have that bounds check, which is why `fromSLV` should not have an additional synthesizable bounds-check: we can't know for sure that synthesis tools will eliminate the second one.

Issue #1215, opened 2020-03-12
Include clock origin and phase in domain.


rowanG077:
In some cases a design requires clocks to be related somehow. For instance a low overhead clock crossing can be implemented if two clocks have the same origin, phase and the period of one clock is an integer multiple of the other. Clash currently doesn't provide a way to encode this in a domain meaning there is no way to specify two clocks are related in this way.

I would like to have two extra parameters in `DomainConfiguration`:

- `vOrigin :: String`: What the origin of the clock is. If not specified this would default to a unique string.
- `vPhase :: Natural`: Where Nat is between 0 and 359. If not specified defaults to 0.

Would you guys be oke with this? Am I missing some other parameter that could be useful?

leonschoorl:
Interesting idea
Do you have an idea of what this would look like in use?
For example what would be the type of this low overhead synchronizer?

rowanG077:
I haven't tested it in Hardware since I don't have an FPGA around at home. However I believe something like this should work. It was inspired by a [comment](https://groups.google.com/d/msg/clash-language/aZcM-45Li3o/R46kj1kYAAAJ) @christiaanb made in the google group.

```
{-# LANGUAGE UndecidableInstances #-}

module SimpleSynchronizer where

import Clash.Explicit.Prelude

import GHC.TypeLits          ( TypeError
                             , ErrorMessage(..)
                             )

import Data.Kind             ( Type )

type GearRatio (fastDom :: Domain) (slowDom :: Domain) = GearRatioCalc
                                                           (Div (DomainPeriod slowDom) (DomainPeriod fastDom))
                                                           ((DomainPeriod fastDom) <=? (DomainPeriod slowDom))
                                                           ((Mod (DomainPeriod slowDom) (DomainPeriod fastDom)) <=? 0)

type family GearRatioCalc (ratio :: Nat) (fastDomIsFaster :: Bool) (isMultiple :: Bool) :: Nat where
  GearRatioCalc r 'True 'True  = r
  GearRatioCalc _ 'False _ = TypeError ('Text "Invalid gearing:" ':$$:
                                        'Text "\tThe fast Clock domain does not have a " ':<>:
                                        'Text "smaller period then the slow Clock Domain.")
  GearRatioCalc _ 'True 'False = TypeError ('Text "Invalid gearing:" ':$$:
                                            'Text "\tThe fast Clock domain does not cleanly " ':<>:
                                            'Text "divide the slow Clock domain.")

[...]

PR #2295, opened 2022-08-09
Add dynamic clock periods

martijnbastiaan:
## Still TODO:


clash-lib/prims/verilog/Clash_Explicit_Testbench.primitives.yaml
```diff
@@ -17,7 +17,7 @@
       always @(~IF~ACTIVEEDGE[Rising][0]~THENposedge~ELSEnegedge~FI ~ARG[3]) begin
         if (~ARG[6] !== ~ARG[7]) begin
           $display("@%0tns: %s, expected: %b, actual: %b", $time, ~LIT[5], ~ARG[7], ~ARG[6]);
-          $finish;
+          $finish(0);
```
alex-mckenna, line 20: Are you sure you want to change this? If you're intending to set the exit code you can't in standard verilog, but some simulators (like `iverilog`) _may_ have extensions for this (e.g. `finish_and_return(0)` in `iverilog` is `exit(0)`).

What you've actually changed is the level of information printed when stopping simulation to be explicitly "print no information". The other options you can give according to the standard being

```
0: prints nothing
1: prints simulation time and location
2: prints simulation time, location and statistics about the used memory and CPU time
```

martijnbastiaan: I tried to resolve a rebase conflict (but messed up):

https://github.com/clash-lang/clash-compiler/commit/07073588403ff39291447a68c7b0bf8891e10f53

alex-mckenna: F

DigitalBrains1:
This doesn't contain the halfway option yet, where you have a static clock but specify explicitly the time of the first active edge (which enables you to adjust the phase of the clock without all the extra processing true dynamic clocks incur). Was this a conscious decision or oversight? I'm fine with the former :-).

DigitalBrains1 (commented):
> To prevent a combinatorial explosing of clock generation primitives

explosion

clash-lib/prims/verilog/Clash_Explicit_Testbench.primitives.yaml
martijnbastiaan: Fixed!

martijnbastiaan:
>  Was this a conscious decision

Somewhat. I wanted to get the absolute minimum in so we can use it in _the project that shall not be named_. However, I also think it's largely unrelated to this PR as I'd like to make that change on another level, as per @rowanG077's suggestion a while ago: https://github.com/clash-lang/clash-compiler/issues/1215.

DigitalBrains1:
Yeah, I mainly thought that as you were fiddling with periods anyway you might as well add phase. "On another level": that's also a valid criticism of this PR, all the information used to be in `DomainConfiguration`, now it's partly in `Clock` and partly in `DomainConfiguration`. Eventually we might want to move it all into `Clock` perhaps.

If we do implement something like #1215, we should think a bit about the unit we express it in. An integral amount of degrees might not be the best option. It seems Vivado only lets you specify it in degrees, and Quartus allows you to specify it in degrees, nanoseconds or picoseconds (neither force an integral amount, mind you). Of course, if Clash also generates the instantiation of the PLL, it doesn't matter much what unit they choose, but it's instructive to see what others have done with the same subject.

martijnbastiaan:
Hmmyeah, I think the proper way to do this is to change `_period` to `Maybe Nat`, instead of just `Nat`. However, this would also break the current API for a feature that, approximately, 0% of Clash users are interested in :-). So I think we should evaluate it this way - if it proves useful for a larger audience we can make it break the public API. I'm therefore hopeful that this PR gets accepted as is - its general approach anyway.

> Eventually we might want to move it all into Clock perhaps.

I think this gets you into trouble with `initBehavior` and `resetPolarity`. At some point we did encode it on a clock/reset level, but that turned out to be quite painful after we started adding properties.

> If we do implement something like https://github.com/clash-lang/clash-compiler/issues/1215, we should think a bit about the unit we express it in. An integral amount of degrees might not be the best option.

Why not as picoseconds†, like period?

† should be refactored into femtoseconds at some point

DigitalBrains1:
> So I think we should evaluate it this way - if it proves useful for a larger audience we can make it break the public API. I'm therefore hopeful that this PR gets accepted as is - its general approach anyway.

Sounds good.

> > Eventually we might want to move it all into Clock perhaps.
> 
> I think this gets you into trouble with `initBehavior` and `resetPolarity`. At some point we did encode it on a clock/reset level, but that turned out to be quite painful after we started adding properties.

I didn't define "it all" ;-). You could also make _just_ the period and phase information something encoded in `Clock`. Even the active edge definition seems more a property of a domain, since it affects a lot of generated HDL. It's more that a clock also needs to obey the active edge than that it is a property of the clock signal.

Anyway, we are getting a bit sidetracked thanks to my comments. I don't think it is going to affect this PR.

> Why not as picoseconds†, like period?

Sounds fine to me. I can't properly pin it down, but it feels more appropriate to me than an angular measure like degrees or radians. However, it is much more common to express phase in an angular measure, so I feel like I should consider that more appropriate? :-) The main difference is the scaling I suppose, degrees or radians scale with the frequency. With degrees, you are always dividing one period into 360 equally sized units. With picoseconds, you're dividing a 500 MHz clock into 2000 equally sized units, but a 50 MHz clock into 20,000 equally sized units. So in a sense you lose resolution as frequency increases. In another sense, you do not: it's still a picosecond.

Again, sidetracked.

changelog/2022-08-11T09_57_03+02_00_outputVerifier_takes_additional_clock
```diff
@@ -0,0 +1 @@
+CHANGE: `Clash.Explicit.Testbench.outputVerifier` now takes an additional clock as an argument: the clock used by the circuit under test. If your tests use the same domain for the test circuit and design under test, consider using `Clash.Explicit.Testbench.outputVerifier'`.
```
DigitalBrains1, line 1: Since `veryUnsafeSynchronizer` is part of the rather obscure but still _public_ API, could you add an entry for that as well?

tests/shouldwork/Signal/DynamicClocks.hs
```diff
@@ -0,0 +1,89 @@
+{-|
+Copyright  :  (C) 2022, Google Inc.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE NumericUnderscores #-}
```
DigitalBrains1, line 6: `NumericUnderscores` is actually enabled by default.

Also, this module generates a bunch of warnings when compiled with `-Wall`. My personal preference is if it doesn't do that :-). Do with it what you will.

martijnbastiaan: Ah yeah, we should force that on CI.

changelog/2022-08-11T09_57_03+02_00_outputVerifier_takes_additional_clock
martijnbastiaan: Good point and yes I can

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -446,11 +446,18 @@ unsafeSynchronizer
...
-    (snatToNum (clockPeriod @dom2))
-{-# INLINE unsafeSynchronizer #-}
+unsafeSynchronizer clk1 clk2 =
+  veryUnsafeSynchronizer (toEither clk1) (toEither clk2)
+ where
+  toEither :: forall dom. KnownDomain dom => Clock dom -> Either Int (Signal dom Int)
+  toEither (Clock _ maybePeriods)
+    | Just periods <- maybePeriods = Right (fromIntegral <$> periods)
+    | otherwise = Left (snatToNum (clockPeriod @dom))
+-- XXX: 'veryUnsafeSynchronizer' doesn't need a blackbox, but Clash doesn't currently
```
christiaanb, line 456:
```suggestion
-- XXX: 'unsafeSynchronizer' shouldn't need a blackbox, but Clash doesn't currently
```

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -472,26 +479,37 @@ unsafeSynchronizer _clk1 _clk2 =
...
+  -- If clock frequencies line up, we can return the signal unchanged
   same :: Signal dom1 a -> Signal dom2 a
   same (s :- ss) = s :- same ss
 
-  go :: Int -> Signal dom1 a -> Signal dom2 a
-  go relativeTime (a :- s)
-    | relativeTime <= 0 = a :- go (relativeTime + t2) (a :- s)
-    | otherwise = go (relativeTime - t1) s
+  goDynamic :: Int -> Signal dom1 Int -> Signal dom2 Int -> Signal dom1 a -> Signal dom2 a
+  -- goDynamic (traceShowId -> relativeTime) (t1 :- ts1) (t2 :- ts2) (a :- as)
```
christiaanb, line 504: The commented out code is left there intentionally?

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -194,7 +194,7 @@ module Clash.Explicit.Signal
   , fromEnable
   , enableGen
     -- * Clock
-  , Clock
+  , Clock(Clock)
```
christiaanb, line 197: Do we really need to expose the constructor? It's not synthesizable is it?

martijnbastiaan: Huh, no this was unintentional. Will fix.

changelog/2022-08-11T09_57_03+02_00_outputVerifier_takes_additional_clock
```diff
@@ -0,0 +1,2 @@
...
+CHANGE: `Clash.Explicit.Signal.veryUnsafeSynchronizer` now accepts either a static clock period or a dynamic one. If you don't use dynamic clocks, convert your calls to use `Left`.
```
DigitalBrains1, line 2:
```suggestion
CHANGE: `Clash.Explicit.Testbench.outputVerifier` now takes an additional clock as an argument: the clock used by the circuit under test. If your tests use the same domain for the test circuit and design under test, consider using `Clash.Explicit.Testbench.outputVerifier'`. [#2295](https://github.com/clash-lang/clash-compiler/pull/2295)
CHANGE: `Clash.Explicit.Signal.veryUnsafeSynchronizer` now accepts either a static clock period or a dynamic one. If you don't use dynamic clocks, convert your calls to use `Left`. [#2295](https://github.com/clash-lang/clash-compiler/pull/2295)
```

DigitalBrains1 (approved):
Official seal of approval

tests/shouldwork/Signal/DynamicClocks.hs
martijnbastiaan: Ah wait, I thought this was about `unittests`. Forcing this check on CI would be quite the task. I've fixed the warnings for `DynamicClocks.hs` only.

DigitalBrains1:
I removed the superfluous `NumericUnderscores`. Furthermore, I noticed that an earlier commit still exported the `Clock` constructor, but then a later commit removed it again. I also moved that in the right place. However, because I accidentally rebased on a newer `master`, the _Compare_ links are unhelpful. Here you can actually compare my changes:

https://github.com/clash-lang/clash-compiler/compare/7b28ffa..a4c309f5c2bee9bf89d55b1e77abc42be8672938

martijnbastiaan:
@mergify rebase

martijnbastiaan:
@mergify rebase

martijnbastiaan:
@mergify queue

martijnbastiaan:
@mergify rebase

martijnbastiaan:
@mergify rebase

PR #2296, opened 2022-08-11
DEC shouldn't care about the shape of proofs

### LINKED ISSUE #2289: Netlist generation fails on `Unsigned 1`
After some refactoring I hit a bug where Clash failed during Normalization. When trying reducing to reduce it to a smaller case I no-longer hit a Normalization but hit a netlist generation issue. So I probably will create another issue once this one is fixed.

Anyway I can reproduce the issue with GHC 8.10.7 and with GHC 9.0.2 with both Clash 1.6.3 and the current master.

When I run synthesis I get this error:

```
<no location info>: error:
    Clash error call:
    Clash.Netlist(1117): mkDcApplication undefined for: (Unsigned 1,Name {nameSort = User, nameOcc = "GHC.Tuple.(,)", nameUniq = 3963167672086036486, nameLoc = UnhelpfulSpan UnhelpfulNoLocationInfo},[Tick (NameMod PrefixName (LitTy (SymTy "$p2(%,%)"))) (Literal (NaturalLiteral 1)),Tick (NameMod PrefixName (LitTy (SymTy "$p2(%,%)"))) (Literal (NaturalLiteral 1))],[Just (Unsigned 64),Just (Unsigned 64)])
      error, called at src/Clash/Netlist.hs:1117:9 in clash-lib-1.7.0-inplace:Clash.Netlist
      mkDcApplication, called at src/Clash/Netlist.hs:842:16 in clash-lib-1.7.0-inplace:Clash.Netlist
      mkExpr, called at src/Clash/Netlist.hs:505:31 in clash-lib-1.7.0-inplace:Clash.Netlist
```

And the complete compilation output with DebugApplied enabled available here(GHC 9.0.2 + Clash master):  https://gist.github.com/rowanG077/3c330bd65a2c7f66907189eef96443c4

I have created a reproducer which contains 3 modules so to make it easy to debug it I have created a repo for it: https://github.com/rowanG077/clash-netlist-gen-bug

It's based on clash-starter so to run it just clone it and run `cabal run clash -- Example.TopEntity --verilog` It assumes clash sources are available one directory up.

christiaanb:
Whether we have `Refl : True ~ True` or `SomeAxiom : (1 <=? 2) ~ True` it doesn't matter, since when we normalize both sides we always end up with a proof of `True ~ True`.

Since DEC only fires for applications where all the type arguments are equal, we can deduce that all equality arguments witness the same equality, hence we don't have to care about the shape of the proof.

Fixes #2289

## Still TODO:

### DIFF
--- a/changelog/2022-08-11T22_15_00+02_00_fix2289
+++ b/changelog/2022-08-11T22_15_00+02_00_fix2289
@@ -1,0 +1,1 @@
+FIXED: Netlist generation fails for certain uses of GADTs [#2289](https://github.com/clash-lang/clash-compiler/issues/2289)
--- a/clash-lib/src/Clash/Normalize/Transformations/DEC.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/DEC.hs
@@ -1,6 +1,6 @@
 {-|
   Copyright  :  (C) 2015-2016, University of Twente,
-                    2021,      QBayLogic B.V.
+                    2021-2022, QBayLogic B.V.
                     2022,      LumiGuide Fietsdetectie B.V.
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
@@ -29,6 +29,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecursiveDo #-}
+{-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE TemplateHaskell #-}
 
 module Clash.Normalize.Transformations.DEC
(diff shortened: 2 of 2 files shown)


rowanG077:
Can confirm this fixes the bug. Thank you @christiaanb :+1:

martijnbastiaan:
Makes sense. Could you add the reproducer from https://github.com/rowanG077/clash-netlist-gen-bug to the testsuite?

christiaanb:
@martijnbastiaan yeah.... so I created a regression test https://github.com/clash-lang/clash-compiler/commit/1f8b057e7eb98d55969e1ea5bf0a1ddea4be5bf9 but as you can see, it doesn't fail on master: https://gitlab.com/clash-lang/clash-compiler/-/pipelines/611074184

I've tried both adding the modules to the `clash-testsuite.cabal` and not, and neither options fail... so yeah... I don't think I can add a regression test.

martijnbastiaan (approved):

PR #2270, opened 2022-07-05
Add Xilinx dual clock FIFO primitive to `clash-cores`

vmchale:
This adds Xilinx's DcFifo to clash-cores. The Haskell model and the IP are both tested.

Past discussion is [here](https://github.com/clash-lang/clash-compiler/pull/2182#issue-1222094757); it is bit cluttered. 

It's a draft PR because it depends on https://github.com/clash-lang/clash-compiler/pull/2257 
## Still TODO:

  - [ ] ~Write a changelog entry (see changelog/README.md)~ (clash-cores isn't on Hackage)
  - [x] Check copyright notices are up to date in edited files
  - [ ] Fix `full` behavior and add a test for it
  - [ ] Fix `overflow` behavior and add a test for it
  - [x] Remove unused / obsolete code
  - [x] Test reset+full behavior

PR #2305, opened 2022-08-22
Instantiate types in component declarations with their size

vmchale:
As Peter points out, we instantiate the DcFifo differently from the Xilinx documentation: https://github.com/clash-lang/clash-compiler/pull/2270#issuecomment-1221238224

With this, component declarations attach a size to types (which conforms to the example). 

## Still TODO:


DigitalBrains1:
Heh, I actually think _this time_ we don't need a Changelog, because the feature is not in a released branch yet! So from the point of view of a user following releases and reading the Changelog, the feature and the fix arrive at the same time.

DigitalBrains1 (approved):
If you remove the Changelog :-), I'm happy with this.

vmchale:
Done!

PR #2307, opened 2022-08-26
Add '+>>.' and '.<<+'

martijnbastiaan:
We've got `+>>` for `Vector`. It seems natural that its `BitVector` counterpart would exist. 

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Sized/BitVector.hs
+++ b/clash-prelude/src/Clash/Sized/BitVector.hs
@@ -1,5 +1,6 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente
+                  2022     , Google Inc.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 -}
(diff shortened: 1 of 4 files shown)


clash-prelude/src/Clash/Sized/BitVector.hs
```diff
@@ -25,9 +25,40 @@ module Clash.Sized.BitVector
...
+-- | Shift in a bit from the MSB side of a 'BitVector'. Equal to right shifting
+-- the 'BitVector' by one and replacing the MSB with the bit to be shifted in.
+--
+-- >>> 1 +>>. 0b1111_0000 :: BitVector 8
+-- 0b1111_1000
+-- >>> 0 +>>. 0b1111_0000 :: BitVector 8
+-- 0b0111_1000
+--
+(+>>.) :: forall n. KnownNat n => Bit -> BitVector n -> BitVector n
+b +>>. bv = replaceBit @_ @(Index n) maxBound b (shiftR bv 1)
```
rowanG077, line 52: I dislike using `shiftR`, `shiftL` because it clutters HDL with 64 bit wide numbers. If possible for free I'd rather see that avoided. An implementation would also be `pack . (b +>>)  . unpack` although that might be slower in simulation.

DigitalBrains1: We could define `shiftL1` and `shiftR1` being equal to 
```
(`shiftL` 1) / (`shiftR` 1)
```
 with much less bounds checking and stuff.

DigitalBrains1: And yes, sim speed for the route through `Vec n Bit` is poor.

martijnbastiaan: Are you worried about synthesis results @rowanG077, or just code clutter?

DigitalBrains1: I think you should also use `replaceBit#` directly because the path through `replaceBit` introduces an unnecessary `unsafeDupablePerfomIO` to catch `XException`s.

martijnbastiaan: In any case, I think we should make the black boxes for `shiftR#` and `shiftL#` (+ other operations having the same problem) special case on constants. I'll post an issue.

martijnbastiaan:
> I think you should also use `replaceBit#` directly because the path through `replaceBit` introduces an unnecessary `unsafeDupablePerfomIO` to catch `XException`s. [(from a comment)](https://github.com/clash-lang/clash-compiler/pull/2307#discussion_r955961902)

Done @DigitalBrains1

clash-prelude/src/Clash/Sized/BitVector.hs
rowanG077: @martijnbastiaan Not especially. I would be surprised if the extra wires wouldn't be removed. But more that we should at least try to make HDL "good". Unless there is a pretty good reason not to. (And a pretty good reason could be a lot of work or simulation performance.)

martijnbastiaan: The proposed way would be wildly inefficient in simulation yeah. But mostly I feel like this is a separate issue anyway: https://github.com/clash-lang/clash-compiler/issues/2308.

rowanG077 (approved):

clash-prelude/src/Clash/Sized/BitVector.hs
rowanG077: Since we have the issue to track it I'm fine with this PR as is.

Issue #2264, opened 2022-07-03
Refactor BuildSpecific to pass in the full qualified name


vmchale:
Right now, the test suite takes e.g. `PortProductsSum_testBench` as an argument to `BuildSpecific`. That makes it needlessly complicated to find the `clash-manifest.json` for a particular entity (here, `verilog/PortProductsSum.topEntity/clash-manifest.json`).

We should pass in the full qualified name in `BuildSpecific` so that we can adroitly simulate with Vivado.

DigitalBrains1:
This issue is caused by the following:

This code:
```haskell
module F where

g = ...
{-# ANN g (defSyn "h") #-}
```
generates:
```
verilog/F.g/h.v
```
and in our test suite, we invoke such a beast with:
```haskell
runTest "F" def{buildTargets=BuildSpecific ["h"]}
```

Note how it does not specify `g`, only `F` and `h`.

Now, first of all, this `h` is unique: Clash will avoid colliding with it and specifying multiple `defSyn "h"` is simply user error, so there will always only be one `h.v` (or `h.vhdl` / `h.sv`), and this will also be `top_component` in its `clash-manifest.json`.

However, the way I wrote the Tcl connector for Vivado (later also Quartus will be supported), you give it the name of the directory containing `clash-manifest.json` for the top entity (and only that). So you'd give it `F.g`, but our test suite does not pass the value `g` to the test.

Other HDL tools in our test suite simply glob over all directories, either picking HDL files or manifests from all of them, so they don't care about `g`. They _do_ need `h`.

PR #2319, opened 2022-08-31
GenTcl restructuring

martijnbastiaan:
This PR aims to do a few things:

 * Restructure to allow functions to be reused for synthesis, instead of just simulation. (Note that this is the reason for removing references to `Sim`.)
 * Add convenience functions for external integrations such as `clash-shake`
 * Clarify functions and their docs
 * Minor refactors to prevent some ??? when reading the code.

### DIFF
--- a/.ci/gitlab/test.yml
+++ b/.ci/gitlab/test.yml
@@ -101,7 +101,8 @@ suite:cores:
   extends: .test-common-local
   script:
     - source /opt/tools/Xilinx/Vivado/2022.1/settings64.sh
-    - cabal v2-run -- clash-testsuite -j$THREADS -p Cores --hide-successes --no-modelsim --no-ghdl --no-iverilog --no-verilator --no-symbiyosys
+    # TODO: Change 'Vector' to 'Cores' after merging #2270
+    - cabal v2-run -- clash-testsuite -j$THREADS -p Vector --hide-successes --no-modelsim --no-ghdl --no-iverilog --no-verilator --no-symbiyosys
   tags:
     - local
     - vivado-2022.1-standard
--- a/clash-lib/clash-lib.cabal
+++ b/clash-lib/clash-lib.cabal
@@ -215,6 +215,7 @@ Library
 
                       Clash.Driver
                       Clash.Driver.Manifest
+                      Clash.Driver.LocatedManifest
                       Clash.Driver.Types
 
                       Clash.Edalize.Edam
--- a/clash-lib/src/Clash/Driver/LocatedManifest.hs
+++ b/clash-lib/src/Clash/Driver/LocatedManifest.hs
@@ -1,0 +1,27 @@
+{-|
+Copyright  : (C) 2022, Google LLC
+License    : BSD2 (see the file LICENSE)
+Maintainer : QBayLogic B.V. <devops@qbaylogic.com>
+-}
+
+module Clash.Driver.LocatedManifest where
+
+import Clash.Driver.Manifest (Manifest)
+import System.FilePath ((</>), dropFileName)
+
+data LocatedManifest = LocatedManifest
+  { -- | Path pointing to the manifest file itself
+    lmPath :: FilePath
+
+    -- | Manifest file corresponding to the one at 'lmPath'
+  , lmManifest :: Manifest
+  }
+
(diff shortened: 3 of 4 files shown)


DigitalBrains1:
Ah, interesting to see development in this part!

1. I don't think common functionality should be in `clash-testsuite`, that is a package with a very narrow, specific purpose. And if you move it to `clash-lib`, I'd like to know how we offer this to our users? What machinery hooks into it? I had thought we would either integrate Vivado design generation into the starter project or make another project fulfilling this functionality. A Tcl script can do all the heavy lifting on its own (we have such a script), there is no need for any Haskell code per se; and this would have the benefit that it is easier to customise for users who are familiar with Vivado. Note that if you do it in Haskell, the user is _still_ writing Tcl inside a `.hs` file, so that doesn't gain them anything IMO. I had envisioned the code in `clash-testsuite` to be very narrow in purpose: to generate Tcl scripts to run Vivado inside `clash-testsuite`, that's it.
2. You don't need _any_ SDC or XDC files in simulation; we do behavioural simulation, constraints are meaningless there. IMO, SDC and XDC files should only be added in designs for synthesis.
3. [UG903 Using Constraints](https://docs.xilinx.com/r/en-US/ug903-vivado-using-constraints) explains that you need to put your `read_xdc` in the right place (before synthesis or between synthesis and implementation) and in the right order in your script and that you might need scoping parameters (for the DCFIFO, I will need scoping parameters). I had planned to write code this week to tie in to the Clash<->Tcl API to communicate all these parameters so you can do the correct `read_xdc` command in the correct place. We could merge this PR which handles the simplest cases, but it'll be updated soon to be more sophisticated.
4. You read all SDC files, but at a glance this seems to break all tests in the test suite? You correctly note that it will fail if a Clash design consists of multiple entities, but that is precisely the case with `testBench` and `topEntity`? I know I had it fail for me when I added `topEntity`'s SDC file.

martijnbastiaan:
Great, thanks for the comment!

1. Agreed. I don't think we should move it to `clash-lib` at all, it doesn't really have that much to do with Clash besides the manifest files. On the larger TCL script: my request would be to keep it composable. That is, users shouldn't have to _edit_ scripts, just compose them differently with custom TCL interspersed. On where to put it: I'm fine with keeping it in `clash-testsuite` until we've tried it in a project or two. Although I think we should definitely have a starter project that uses the TCL scripts, I think they should be accessible as a library - we shouldn't have to keep copy+pasting these scripts around.

2. Agreed, that should be removed from `mkBaseTcl`.

3. Okay, curious to hear your thoughts.

4. Yeah, I'll remove it.

vmchale:
Does this fix any of the failing tests that I commented out in order to get Vivado support in our test suite?

vmchale (approved):
Approve on the assumption that the `.xdc` files will not be read spuriously.

martijnbastiaan:
> Does this fix any of the failing tests that I commented out in order to get Vivado support in our test suite?

It doesn't try to fix any :). But this PR will probably be superseded by a future PR of @DigitalBrains1, so I'm leaving this pending for now.

martijnbastiaan:
I'll pop off the XDC/SDC commit. I'm going to assume the silence on all the other PRs is a silent approval @DigitalBrains1 ;-). Another PR should probably restructure the module to use https://github.com/clash-lang/clash-compiler/pull/2321.

DigitalBrains1:
Ah no, I had actually thought we had decided this was to be closed and superseded by #2321, so there was a bit of miscommunication. There is no need for `clash-testsuite` to be picking around in the `clash-manifest.json` files at all with #2321. All that's needed now is support for IP and we can replace a bunch of code in `clash-testsuite` by `source clashConnector.tcl` in the Tcl.

I think my work for this week is adding IP support (including new functionality to issue `read_xdc` with the correct paramters for `DcFifo`) to `clashConnector.tcl`. Then we can change `clash-testsuite`. And Vanessa, that doesn't help with #2264; frankly I don't really understand that issue but I'm planning on adding multiple top entities (that might be test benches) after #2270 is merged, although I might actually take a week off before that or something if you don't mind.

martijnbastiaan:
Obsolete thanks to new TCL scripts.

PR #2321, opened 2022-09-02
New: Bundled clashConnector.tcl

DigitalBrains1:
We now bundle a Tcl script that parses `clash-manifest.json` files and
issues the correct Vivado Tcl commands to read the sources in
Non-Project Mode.

The Clash<->Tcl API for creating Xilinx IP will be added soon; first,
its features need to be extended.

The script can be obtained from `Clash.DataFiles.tclConnector` in
`clash-lib` or by invoking the tool `static-files`. The name of the tool
is generic so we can add different output products to it later and still
have the name make sense.

## Still TODO:

The copyright entry in the Cabal file `clash-lib.cabal` wasn't updated for three years? It should even be bumped whenever anything in `clash-lib` is updated at all, right? Or at the very least when the file itself is edited?


clash-lib/tools/static-files.hs
```diff
@@ -0,0 +1,69 @@
...
+
+main :: IO ()
+main = do
+  args <- parseArgsOrExit patterns =<< getArgs
+  -- Since we got here, we know we got invoked with the sole mandatory option
+  -- @--tcl-connector@ and its mandatory argument
+  let force = args `isPresent` (longOption "force")
+      Just outFile = args `getArg` (longOption "tcl-connector")
+  createOkayOrDie outFile force
+  inFile <- getDataFileName $ "data-files" </> "tcl" </> "clashConnector.tcl"
```
martijnbastiaan, line 67: Could we expose these paths in `Clash.DataFiles` too? That way build systems (e.g. Shake) can just use that instead of copying it somewhere else.

clash-lib/tools/static-files.hs
```diff
@@ -0,0 +1,69 @@
...
+  copyFile inFile outFile
+  putStrLn $ "Tcl Connector written to " ++ outFile
```
martijnbastiaan, line 69: IMO just leave it out

changelog/2022-09-02T12_48_27+02_00_static_files_tcl
```diff
@@ -0,0 +1,8 @@
+ADDED: The executable `static-files` in `clash-lib` can be used to write a Tcl
+script to a file that allows Vivado to interact with the metadata generated by
+Clash. The Tcl proc `clash::readHdl` will issue `read_vhdl`/`read_verilog` calls
+that add all the generated HDL files, and `clash::readXdc` will issue `read_xdc`
+commands for constraints on the top entity.
+
+You can write a copy of the Tcl script to a file by invoking
+    cabal run clash-lib:static-files -- --tcl-connector clashConnector.tcl
```
martijnbastiaan, line 8:
```suggestion
    cabal run clash-lib:static-files -- --tcl-connector=clashConnector.tcl
```
?

clash-lib/tools/static-files.hs
```diff
@@ -0,0 +1,69 @@
...
+Obtain static files useful when working with Clash designs.
+
+Currently, only the Tcl connector is available.
+
+Usage:
+  static-files [--force] --tcl-connector=<outfile>
+
+Options:
+  -f, --force                  Overwrite existing files
+  --tcl-connector=<outfile>    Write a Tcl script to a file that can parse Clash
```
martijnbastiaan, line 36: Given that it's for Vivado specifically, I'd mention it in the name.

martijnbastiaan (commented):
Some name ideas: `clash-vivado-lib`? `clash-tcl`? `clash-tcl-vivado`? I don't really have an opinion :).

clash-lib/tools/static-files.hs
DigitalBrains1: Right, I knew I forgot something: I was going to suggest we expose them through the library as well. So now I know your opinion on that!

changelog/2022-09-02T12_48_27+02_00_static_files_tcl
DigitalBrains1: Apart from my personal preference to omit the `=`, the `=` also prevents bash from expanding a `~` leading the path, giving our users grief with errors like "Directory does not exist: ~". So, no, I'd rather keep it like this.

clash-lib/tools/static-files.hs
DigitalBrains1: I was planning to extend it to other tools some day, and it can already be used on Intel when you just do `clash::readMetadata` and then parse the `metadata` dict yourself. But if you still think it should mention Vivado specifically, I'll add it.

changelog/2022-09-02T12_48_27+02_00_static_files_tcl
martijnbastiaan: Well what I meant was: the functions docopt uses `--tcl-connector=....`. So can we actually use this form?

clash-lib/tools/static-files.hs
DigitalBrains1: I'll add a `--verbose` flag as discussed.

changelog/2022-09-02T12_48_27+02_00_static_files_tcl
DigitalBrains1: Yes, it is fine. I think you have to use the equals sign in the usage message, though.

PR #2326, opened 2022-09-07
Replace package `interpolate` by `string-interpolate`

DigitalBrains1:
Replace package `interpolate` by `string-interpolate`

Replace `neat-interpolate` in `clash-testsuite` by `string-interpolate`.

Cleaned up `InlinePrimitive`s: convert to YAML, use `__i` and indenting
to make code more readable.

Some doc is also added to `clash-lib:Clash.Util.Interpolate` to explain
its purpose, and the string identifying the module in some (probably
unreachable) error messages are fixed.

I was wondering about two things:
 - Why do we even use `InlinePrimitive` in `clash-prelude`? Why aren't they simply in `clash-lib/prims`?
 - Should we also replace
      ```haskell
      InlineYamlPrimitive [Verilog, SystemVerilog, VHDL]
      ```
      by
      ```haskell
      InlineYamlPrimitive [minBound..]
      ```
      ?

## Still TODO:


clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
```diff
@@ -99,40 +93,36 @@ veriBinaryPrim
...
-                          ,"format": "Haskell"
-                          ,"templateFunction": "#{show tclTFName}"
-                          }
-                        ]
-        }
-      }
-    ]
-    |]
+  InlineYamlPrimitive [Verilog, SystemVerilog] [__i|
+  BlackBox:
```
martijnbastiaan, line 97: I normally wouldn't comment on this, but given that this is in the diff anyway: the other blackboxes in this patch indent like this:

```suggestion
  InlineYamlPrimitive [Verilog, SystemVerilog] [__i|
    BlackBox:
```

which I like better.

martijnbastiaan (approved):
Looks good, nice cleanup

clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
DigitalBrains1: Right, if you look at its brother [here](https://github.com/clash-lang/clash-compiler/pull/2326/files/cfc24e2b27d972d12c572ea94efb665b050fceb5#diff-73ec9a585ecf605d2db31b5b6e420ccdfbea78007357c3f4a365529f66b13b03R30) you'll see that one was not intended, thanks for spotting it!
```haskell
vhdlBinaryPrim primName tclTFName funcName = InlineYamlPrimitive [VHDL] [__i|
  BlackBox:
```

DigitalBrains1:
> Why do we even use InlinePrimitive in clash-prelude? Why aren't they simply in clash-lib/prims?

As discussed: it's a stylistic thing; keeping Haskell and HDL definitions close together is actually kinda nice.

PR #2332, opened 2022-09-21
Tweak some Stack / Stack CI settings

DigitalBrains1:
- Pass `-Wall -Wcompat` to GHC when building with Stack (not just CI,
  also the `stack.yaml` in the root). These options only apply to the
  local packages, it will not change the GHC options for dependencies.
- Pass `--pedantic` to Stack in CI runs

    This switch adds `-Wall -Werror` to the options passed to GHC.
    Options specified in `stack.yaml` (like `-Wcompat`) are still
    passed; it is cumulative. Future versions of Stack might add
    behavior to `--pedantic`.
- Caching on GitHub Windows CI is broken in cache@v2:
  https://github.com/actions/cache/issues/666#issuecomment-1002945087
  cache@v3 exhibits the same issue, going back to cache@v1
- Drop unneeded `extra-deps`
- Use an LTS resolver for GitHub Stack CI with GHC 9.0.2, instead of a
  nightly
- While editing these files anyway, bumping all resolvers to the latest
  LTS

GitLab jobs `nix-build` and `debian-bindist` still do not have
-Wall -Werror. This could be addressed in a later PR.

## Still TODO:

### DIFF
--- a/.ci/stack-8.10.7.yaml
+++ b/.ci/stack-8.10.7.yaml
@@ -1,4 +1,7 @@
-resolver: lts-18.24
+resolver: lts-18.28
+
+ghc-options:
+  "$locals": -Wall -Wcompat
 
 packages:
 - clash-prelude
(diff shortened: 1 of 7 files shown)


leonschoorl (approved):
I've just updated lts-19.24 to lts-19.25, which was just released.

PR #2333, opened 2022-09-28
New: `Xilinx.Floating.fromU32`

DigitalBrains1:
New in `clash-cores`: convert `Unsigned 32` to `Float` with Xilinx IP.

## Still TODO:

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
@@ -1,14 +1,18 @@
 {-|
 Copyright  :  (C) 2021-2022, QBayLogic B.V.,
+                  2022     , Google Inc.,
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
 {-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Clash.Cores.Xilinx.Floating.Annotations
   ( veriBinaryPrim
   , vhdlBinaryPrim
+  , vhdlFromUPrim
+  , veriFromUPrim
   ) where
 
 import Prelude
@@ -17,6 +21,8 @@ import Data.String.Interpolate (__i)
 import Language.Haskell.TH.Syntax (Name)
 
 import Clash.Annotations.Primitive (Primitive(..), HDL(..))
+
+import Clash.Cores.Xilinx.Floating.BlackBoxes
 
 -- | The InlinePrimitive annotation for a binary function in VHDL.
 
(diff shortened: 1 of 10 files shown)


DigitalBrains1:
I'll remove draft status after I've looked at it when I'm not as tired. I'm afraid I'll do something dumb right now.

DigitalBrains1:
I don't intend to merge the final commit (_Temp hack: run Xilinx Floating test suite_). Rather, I'd like to fix the test suite soon to support multiple `buildTargets` (for all tools). But this is pretty useful to run the tests right now.

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
```diff
@@ -536,6 +522,30 @@ nanTest =
...
+fromUBasicSamples :: [(Unsigned 32, Float)]
+fromUBasicSamples = delayOutput0 (natToNum @F.FromU32DefDelay) $
+  map (\x -> (x, fromIntegral x))
+    [ 0
+    , 1
+    , maxBound
+    , 0b0010_1010_1010_1010_1010_1010_1010_1010
+    , 0b0101_0101_0101_0101_0101_0101_0101_0101
+    , 0b1010_1010_1010_1010_1010_1010_1010_1010
+    , 0b1101_0101_0101_0101_0101_0101_0101_0101
```
vmchale, line 534: The way `fromIntegral` works here is quite opaque to me! 

```
λ:> fromIntegral (0b0000_0000_1111_1111_1111_1111_1111_1111::Unsigned 32) :: Float
1.6777215e7
λ:> fromIntegral (0b0000_0000_1111_1111_1111_1111_1111_1111::Unsigned 32) :: Int
16777215
```

I wonder if there would be a clearer way to handle floats where we can show bits. Something like

```
u32f :: Word32 -> Float
u32f b = unsafePerformIO $ alloca $ \bytes -> poke (castPtr bytes) b *> peek bytes
```

(the function itself is stupid, but it makes clear which bits you are setting in the `Float` and how it all comes together)

vmchale (approved):

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
DigitalBrains1: I think I don't really understand what you mean or what it is that is vague to you, but I'll take a stab at an answer: are you looking for [`decodeFloat`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#v:decodeFloat)? For NaN, there's also [this](https://github.com/clash-lang/clash-compiler/blob/e0266278ae4f8f242559858e19c94d8f8a0e2be6/tests/shouldwork/Cores/Xilinx/Floating.hs#L29) in this very test, and decoding infinity is simply an application of [`isInfinite`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#v:isInfinite). That just leaves the sign bit. Also, of course, you can just `bitCoerce :: Float -> BitVector 32` if all you want is look at the bits.

I don't understand your `u32f`, I'm not familiar with peeking and poking outside of BASIC on the Commodore 64 ;-).

vmchale: Oh I see! Yes, `encodeFloat` or `bitCoerce` is much better than what I had. 

I guess `fromIntegral` works too, if you prefer.

DigitalBrains1: I'm not trying to encode a float, I'm trying to give it an `Unsigned 32` and the expected result of `fromU32` applied to that `Unsigned 32`. The Haskell implementation of `fromU32` is exactly `fromIntegral` followed by a dummy pipeline, and I'm trying to verify in this test whether interesting cases behave the same in Vivado. I'm letting Haskell do all the hard work of converting an integral number into a float, since I'm not trying to verify that Haskell does this correctly but rather that it does it in the same way as Vivado.

Are you confusing `fromIntegral :: (Integral a, Num b) => a -> b` with `bitCoerce`? `fromIntegral` just converts a 1 into a 1.0 and a 2 into a 2.0, but if you give it something which cannot be represented in floating point, it will get rounded.

vmchale: >fromIntegral just converts a 1 into a 1.0 and a 2 into a 2.0

Oh right, that makes a lot of sense, sorry for the diversion.

clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
```diff
@@ -126,3 +132,96 @@ veriBinaryPrim primName tclTFName funcName =
...
+      name: #{primName}
+      type: |-
+        #{primName}
+          :: ( KnownDomain dom            -- ARG[0]
+             , KnownNat d                 -- ARG[1]
+             , HasCallStack               -- ARG[2]
+             )
+          => Clock dom                    -- ARG[3]
+          -> Enable dom                   -- ARG[4]
+          -> DSignal dom n (Unsigned ..)  -- ARG[5]
```
martijnbastiaan, line 153: Instead of doing this, you could move this into the where clause:

```
let
  knownDomArg = 0
  knownNatArg = 1
  ...
in
  ...
```

And then splice these number in, e.g.

```
~ARG[#{inputArg}]
```

This would make the code much easier to change if we ever need to add/remove an argument, not to mention easier to read. Perhaps you can do something similar for `~SYM[3]`.

What do you think?

martijnbastiaan: Also, isn't the result missing from the type signature?

clash-cores/src/Clash/Cores/Xilinx/Floating/Explicit.hs
```diff
@@ -230,6 +234,43 @@ div = withFrozenCallStack $ divWith defConfig
...
+     , HasCallStack
+     )
+  => Clock dom
+  -> Enable dom
+  -> DSignal dom n (Unsigned 32)
+  -> DSignal dom (n + FromU32DefDelay) Float
+fromU32 = withFrozenCallStack fromU32With
+{-# INLINE fromU32 #-}
+
+-- | The default delay for conversion of @Unsigned 32@ to @Float@
```
martijnbastiaan, line 271: Could you comment whether this optimizes for anything? Clock speed I suppose?

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
martijnbastiaan, line 534: These are just picked at random I suppose?

tests/shouldwork/Cores/Xilinx/Floating/Annotations.hs
```diff
@@ -32,3 +33,26 @@ binEnTopAnn name =
           ]
     , t_output = PortName "result"
     }
+
+unTopAnn :: String -> TopEntity
+unTopAnn name =
```
martijnbastiaan, line 38: I suppose the `un` is for `unary`? This took me 30 seconds to figure out. Admittedly, I'm still somewhat covid-brained but.. :-). Could you spell it out?

martijnbastiaan (approved):
In general LGTM!

christiaanb (approved):

tests/shouldwork/Cores/Xilinx/Floating/Annotations.hs
DigitalBrains1: Do you also want me to spell out "binary" in "binTopAnn" then? I admit the name is pretty cryptic, I actually had a laugh about it, and I wouldn't put it in code exposed to our users. You picked the benign one, "unEnTopAnn" is almost like an exercise for your vocal cords.

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
DigitalBrains1: No... they are two bitpatterns (the classic 0x55 and 0xaa) but with special treatment for what would have been the sign bit if it were signed. I'm trying to suss out bugs in Vivado, thinking adverserially, "well it is _supposed_ to be unsigned, but is it really?".

martijnbastiaan: Ah ok, gotcha. It didn't have a comment which is why I figured they were more or less randomly picked.

tests/shouldwork/Cores/Xilinx/Floating/Annotations.hs
martijnbastiaan: For symmetry's sake let's :)

clash-cores/src/Clash/Cores/Xilinx/Floating/Explicit.hs
DigitalBrains1: Xilinx wrote a whole section on it:
> Latency and Rate Configuration
This parameter describes the number of cycles between an operand input and result output.
The latency of all operators can be set between 0 and a maximum value that is dependent
upon the parameters chosen. IP cores that are intended for many applications, such as the
Floating-Point Operator core, are designed to run as fast as the DSP primitives. To achieve
this, the fabric logic must be heavily pipelined, leading to relatively high latency. For
designs which run at a relatively low frequency, the latency can be reduced while timing can
still be met. For instance, if the fully pipelined design with latency=12 can meet timing at
400 MHz, then if the system clock in the user design is 70MHz, for example, then latency
can likely be reduced to six and timing can still be met. However, the relationship between
latency and achieveable clock speed is not linear. This is because the amount of logic
between register stages is roughly the same, so if one register is removed, the achieveable
clock frequency drops considerably because one register to register path now has almost
double the amount of logic as other register-to-register delays. The advantage of reducing
latency when clock speed allows is that the result appears sooner. There is no significant
gain in resources other than an inevitable reduction in registers for which there is an ample
supply. For the previous example, Xilinx recommends using latency=12 if the system clock
is > 0.5 * value in the resource and performance section on the Xilinx webpage for the IP
core. Other latency values of use are 6, 3, 2 and 1, though minimum may apply due to FIFOs
and feedback stages. Therefore, there is not much use in setting latency to 11 for this
reason.
>
> Note: The Accumulator operator has a minimum latency of 1.

The only things I wrote on it in this module's doc is:
> Most functions allow customization of the Xilinx IP. Valid combinations will need to be gleaned from, e.g., the Vivado wizard (IP Catalog -> Floating-point). 

> For each customizable operation, there also exists a function that uses the defaults. These functions use the settings from `defConfig` and the maximum delay for the Xilinx IP with that configuration. That delay is also defined as a type variable for delay annotation in circuits.

To answer your question briefly, yes, I think we can conclude from what Xilinx writes that this optimises for clock speed. Even though in this specific case, there is no DSP involved. Also, to answer more extensively, they appear to say you should only divide by integers for useful results, but their example is very simple: they pick 12, and useful latencies are clearly 6, 3, 2 and 1. But how to do that with 5? Is it 3, 2, 1 or 2, 1? Your guess is as good as mine. If you _really_ wanted to know, I suppose you could determine Fmax for 5, 3 and 2. I don't want to know that badly, though.

clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
DigitalBrains1: Oh, I like that idea!

And yes, the result is missing, this is just a cheat sheet for the correct numbers, and the result is `~RESULT`, no need for a cheat sheet. The code isn't correct Haskell anyway, I elided an unknown number with `..` (yeah, it's known here, it's 32, but I designed it to also work for `Unsigned 64` which I'll add later).

I would like to omit any unused variables from the assignments, though, so just `clockArg`, `enableArg` and `inputArg`. Otherwise you'll have useless code churn later when it were to change.

clash-cores/src/Clash/Cores/Xilinx/Floating/Explicit.hs
martijnbastiaan: Nah it's fine!

martijnbastiaan: Thanks for the comment!

clash-cores/src/Clash/Cores/Xilinx/Floating/Annotations.hs
DigitalBrains1: How do you like it now?

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
DigitalBrains1: I commented extensively.

PR #2336, opened 2022-10-12
Move `UniqMap` to a separate module

alex-mckenna:
This PR moves the `UniqMap` type to a new module, allowing the functions to be used as a qualified import instead of including the type in the name. In the process of refactoring some other benefits came about, including

  * the `UniqSet` type is removed, since it is morally a `UniqMap` that exposes a mostly similar API
  * the names of functions on `UniqMap` are named closer to the original functions in `containers`

## Still TODO:

  - [x] Document functions in `Clash.Data.UniqMap`


alex-mckenna:
I'm only including a copyright notice in the actual `Clash.Data.UniqMap` module; the other changes are mechanical fixes.

alex-mckenna:
Martijn is OOO, changing reviewer

alex-mckenna:
@mergify rebase

clash-lib/src/Clash/Data/UniqMap.hs
```diff
@@ -0,0 +1,234 @@
+{-
+Copyright   : (C) 2022 Alexander McKenna
```
martijnbastiaan, line 2: Most of the definitions in this file are copied from existing modules right? In that case, mention the original copyrights too.

martijnbastiaan (changes requested):
In general LGTM. Could you use the namespace you've created everywhere?

clash-lib/src/Clash/Core/Type.hs
```diff
@@ -104,7 +104,7 @@ import           Clash.Core.Name
 import {-# SOURCE #-} Clash.Core.Subst
 import           Clash.Core.TyCon
 import           Clash.Core.Var
-import           Clash.Unique
+import qualified Clash.Data.UniqMap as UniqMap (find, lookup)
```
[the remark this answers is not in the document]
alex-mckenna: Because they're the two functions that were used in this file already? `find` is a most easily-distinguishable renaming of the previous `lookupUniqMap'`, where the prime means _throw an error if this isn't in the map_

clash-lib/src/Clash/Data/UniqMap.hs
alex-mckenna: To be pedantic I have, because the original module had no copyright notice... I'll approximate one by looking at the history for the file

martijnbastiaan (approved):
Thanks for the refactor, looks much better!

PR #2339, opened 2022-10-22
CI: fix `nix-build` for newer kernels, clean up

DigitalBrains1:
We need to install `awk` before we can use it in `effective_cpus.sh`. `awk` is only used for newer kernels, like the one in Ubuntu 22.04. So the previous code succeeded only on older runner hosts.

The shebang line for `effective_cpus.sh` is changed to pick `bash` from the $PATH; it is pre-installed but not in /usr/bin.

An obsolete workaround for Circle CI and unneeded packages are removed.

It seems we only need to install `gawk` in the `before_script` (per above). `nix-build` will automatically fetch the following packages previously specified in the `before_script`:
findutils gnugrep gnused coreutils bash

These previously specified packages now seem unused: git ssh zstd.

I tested this PR against both `glanerbrug` (running a newer kernel) as well as `diepenheim` (running an older kernel). It all worked.

## Still TODO:

### DIFF
--- a/.ci/effective_cpus.sh
+++ b/.ci/effective_cpus.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # `nproc` doesn't account for limits set by cgroups/docker. This script tries
 # to determine the effective number of cpus we can use by inspecting the shares
 # it has been given.
@@ -14,11 +14,11 @@ elif [ -f /sys/fs/cgroup/cpu.max ]; then
   cfs_quota_us=$(cat /sys/fs/cgroup/cpu.max | awk '{ print $1 }')
   cfs_period_us=$(cat /sys/fs/cgroup/cpu.max | awk '{ print $2 }')
 else
-  echo "Could not determine number of effective CPUs"
+  echo "Could not determine number of effective CPUs" >&2
   exit 1
 fi
 
-if [[ ${cfs_quota_us} == -1 ]]; then
+if [[ ${cfs_quota_us} == @(-1|max) ]]; then
   # No limits set
   nproc
 else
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
For 1.6, I will combine PR #2337 and this one, because the automatic backport of 2337 (#2338) failed on the issue this PR fixes.

DigitalBrains1:
Ah I didn't dive deep enough with my git archeology. 

Commit 90fa58ba1 required `findutils`, `gnugrep` and `coreutils` (for `head` I presume) to run the symlinking logic, and its comment mentions that `git` and `openssh` are required for Circle CI.

Commit f33d2fe5f added a script which required `bash` and `gnused` (although probably `bash` could have been handled like I do in this PR).

Commit 31e92bb11 added `zstd` without explanation.

The code which clearly depended on `findutils`, `gnugrep`, `coreutils` and `gnused` is no longer there, making it likely they can be removed. Circle CI apparently wanted `git` and `openssh`, but we don't use Circle CI anymore, making it likely they can be removed.

This leaves `bash` and `zstd`. I don't know about `zstd`.

`bash` is odd in its own way: it is already there (you just need to get it from the `$PATH` instead of `/usr/bin`), perhaps it was not in an older version of the Docker image and that is why it is added? After all, if it was already there, then the line in the script
```bash
ln -s $(find /nix -type f -name bash | grep 'bash-[0-9]\+\.[0-9]\+' | head -n1) /bin
```
would have already worked fine without "installing" it again with `nix-env`. I think. I'm not that at home with nix to be sure.

martijnbastiaan (approved):
Thanks! The `zstd` should have been lodged in with another commit; at some point we switched to using `zstd` for archiving caches/artifacts for multithreaded compression/decompression.

DigitalBrains1:
> at some point we switched to using `zstd` for archiving caches/artifacts for multithreaded compression/decompression.

I thought that might have been the case, but we don't do any caching for `nix-build`, and it doesn't generate artifacts.

PR #2343, opened 2022-10-25
New: `Xilinx.Floating.fromS32`

martijnbastiaan:
New in `clash-cores`: convert `Signed 32` to `Float` with Xilinx IP.

---------------

I shamelessly copied Peter's homework and I'm not really sure about `fromSBasicSamples`. How did you obtain things like "smallest with rounding" @DigitalBrains1?

### DIFF
--- a/tests/shouldwork/Cores/Xilinx/Floating/TH.hs
+++ b/tests/shouldwork/Cores/Xilinx/Floating/TH.hs
@@ -9,7 +9,7 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
 module Floating.TH where
 
-import Clash.Prelude (natToNum, unpack, Unsigned)
+import Clash.Prelude (natToNum, unpack, Signed, Unsigned)
 
 import Prelude
 import Numeric.IEEE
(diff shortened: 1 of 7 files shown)


DigitalBrains1:
Looks fine to me!

> How did you obtain things like "smallest with rounding" @DigitalBrains1?

Just reasoning about it. You have 24 bits of mantissa (first bit implicitly one). Also, your input needs to have a one bit at the end or it will simply be moved into the exponent and your value is suddenly shorter. So 2\*\*24+1 will round to either 2\*\*24 or 2\*\*24+2 by necessity, and we want to exercise the rounding a bit to verify that GHC and Vivado do it the same as they should, since they both should follow IEEE-754. "Smallest with rounding" covers rounding of the binary pattern `0.10` to an integral number. The "more rounding tests" also checks how `1.10` and `0.01` round. `0.10` and `1.10` are the "ties". The standard IEEE-754 rounding is round to nearest, ties to even. I'm mostly interested in those ties as they are the subtle ones. By those rules, `0.10` should round to `0` and `1.10` to `10`.

For this one, I'd also add the negative versions of all those tests. Since float encodes with sign-magnitude, the magnitude of the negative versions needs to be the same.

Disclaimer: the above might contain glaring mistakes, this is so easy to trip up in, in my experience. I think I got it right.

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
```diff
@@ -555,6 +555,42 @@ fromUBasicSamples = delayOutput0 (natToNum @F.FromU32DefDelay) $
...
+    , 0b0010_1010_1010_1010_1010_1010_1010_1010
+    , 0b0101_0101_0101_0101_0101_0101_0101_0101
+    , 0b1010_1010_1010_1010_1010_1010_1010_1010
+    , 0b1101_0101_0101_0101_0101_0101_0101_0101
+
+    , -- Longest exactly representable
+      0b0000_0000_1111_1111_1111_1111_1111_1111
+
+    , -- Smallest with rounding
+      0b0000_0001_0000_0000_0000_0000_0000_0001
```
DigitalBrains1, line 581:
```suggestion
       0b0000_0001_0000_0000_0000_0000_0000_0001
      -0b0000_0001_0000_0000_0000_0000_0000_0001
```

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
```diff
@@ -555,6 +555,42 @@ fromUBasicSamples = delayOutput0 (natToNum @F.FromU32DefDelay) $
...
+
+      -- More rounding tests
+    , 0b0000_0001_0000_0000_0000_0000_0000_0011
+    , 0b0000_0010_0000_0000_0000_0000_0000_0001
```
DigitalBrains1, line 585:
```suggestion
    ,  0b0000_0001_0000_0000_0000_0000_0000_0011
    , -0b0000_0001_0000_0000_0000_0000_0000_0011
    ,  0b0000_0010_0000_0000_0000_0000_0000_0001
    , -0b0000_0010_0000_0000_0000_0000_0000_0001
```

DigitalBrains1 (commented):
Negative versions of the roundings.

Let's indent all values in the table with an additional space to allow for insertion of a `-` at the start. Note that there are negative numbers written in two's complement and negative numbers written with an unary minus; this represents the fact that in the first we are interested in the encoding of the `Signed 32` and in the latter we are interested in what happens with the magnitude of the number rather than the sign.

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
DigitalBrains1, line 581:
```suggestion
      -- Smallest with rounding
    ,  0b0000_0001_0000_0000_0000_0000_0000_0001
    , -0b0000_0001_0000_0000_0000_0000_0000_0001
```

tests/shouldwork/Cores/Xilinx/Floating/TH.hs
```diff
@@ -555,6 +555,42 @@ fromUBasicSamples = delayOutput0 (natToNum @F.FromU32DefDelay) $
     is = is0 ++ repeat (last is0)
     os = replicate d 0 ++ os0
 
+fromSBasicSamples :: [(Signed 32, Float)]
+fromSBasicSamples = delayOutput0 (natToNum @F.FromS32DefDelay) $
+  map (\x -> (x, fromIntegral x)) (specials ++ map (* (-1)) specials)
```
DigitalBrains1, line 560: Never mind! As you pointed out, `specials ++ map (* (-1)) specials` already covers it, and much better at that!

PR #2345, opened 2022-10-26
Test suite: give every target its own work dir

DigitalBrains1:
For each combination of individual targets in `BuildTargets` and individual tools in `Sim`, we create a new subdirectory and run the tool for that target in that directory. Clash generates HDL only once: the output is copied into every subdirectory.

This achieves several goals:
- Build targets run concurrently, achieving greater parallelism
- All tools now support multiple build targets
- Clean separation between tools

Before this PR, all tools ran from the same shared temporary directory, where they were exposed to each other's files. This did not cause any problems in practice, though.

It does mean that ModelSim's build stage and GHDL's import stage are no longer shared between multiple build targets, performing the work multiple times instead.

## Still TODO:

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -440,15 +440,11 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "T1786" def{
             hdlTargets=[VHDL]
           , buildTargets=BuildSpecific ["testEnableTB", "testBoolTB"]
-          , -- TODO: Enable multiple build targets for Vivado
-            hdlSim=hdlSim def \\ [Vivado] -- triggers error in vivado
           }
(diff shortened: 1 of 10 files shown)


DigitalBrains1:
- I removed the comment in `tests/Main.hs` that for the `Floating` test benches you need Vivado and it takes long to run. I added this comment back when we didn't have Vivado in the test suite / CI.
- I renamed the Tasty test names for SymbiYosys from `SymbiYosys` to `symbiyosys` in line with all the other tools which also use lowercase.
- I added an explanation in `tests/src/Test/Tasty/Vivado.hs` explaining why we need `XILINX_LOCAL_USER_DATA=no`. Originally this was a review comment I forgot to post for PR #2257.

martijnbastiaan (approved):

PR #2346, opened 2022-10-27
Consistently apply 'relativeTime' logic in DcFifo

martijnbastiaan:
This copies the 'relativeTime' logic from 'veryUnsafeSynchronizer'. Without a consistent way of dealing with this, a feedback loop can cause a '<\<loop\>>' output in Haskell simulations.

## Still TODO:

### DIFF
--- a/clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
+++ b/clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans -fno-warn-unused-top-binds #-}
@@ -7,6 +8,9 @@ module Test.Cores.Xilinx.DcFifo ( tests ) where
 
 import qualified Prelude as P
 import qualified Data.List as L
+
+import Control.DeepSeq (rnf)
+import Data.Bool (bool)
 import Data.Maybe (catMaybes)
 import Data.Proxy (Proxy (..))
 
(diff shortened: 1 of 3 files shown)


clash-cores/test/Test/Cores/Xilinx/DcFifo.hs
```diff
@@ -65,8 +76,84 @@ tests = testGroup "FIFO tests"
...
+  noWrites = pure Nothing
+  writeMaybe = bool <$> noWrites <*> writes <*> writeEnable
+
+  config = (defConfig @8){dcOverflow=True, dcUnderflow=True}
+  FifoOut{..} = dcFifo config clkW rstW clkR rstR writeMaybe readEnable
+  overflowSynced = dualFlipFlopSynchronizer clkW clkR rstR enaR False isOverflow
+  underflowSynced = dualFlipFlopSynchronizer clkR clkW rstW enaW False isUnderflow
+
+  readEnable =
+    delay clkR enaR True $
```
vmchale, line 150: I think you can remove the `delay` here and on line 154, makes the whole thing a little simpler!

vmchale (approved):

DigitalBrains1:
```diff
-  go st@(FifoState _ rt) rdClk wrClk@(tWr :- _) =
-    if rt < fromIntegral tWr
+  go st@(FifoState _ rt) rdClk wrClk =
+    if rt <= 0
```

Note what this accomplishes:
* Significantly, the old code skipped the first write clock edge. The first edge of all clocks in Clash coincide; you coud call this t = 0, but in HDL simulation it is actually t = `~LONGESTPERIOD`. It doesn't really matter what you define as the zero point in time, as long as all relations are kept intact.

    Because it skipped processing the first edge, the model was consistently one `Signal` constructor behind the rest of the circuit. It was stuck in the past.
* Second, but probably of no consequence, is that whenever read clock and write clock edges coincided, the old code would first handle the write clock edge and then the read clock edge, whereas the new code will first handle the read clock edge.

Issue #2266, opened 2022-07-03
Vivado test infrastructure fails on Parameters


vmchale:
```
ERROR: [XSIM 43-3225] Cannot find design unit topEntity.testBench in library work located at xsim.dir/work.
```

It seems to try to simulate `topEntity.testBench` instead of `testBench.testBench`, see https://github.com/clash-lang/clash-compiler/pull/2257#issuecomment-1172957230

vmchale:
Same thing happening here? :( https://gitlab.com/clash-lang/clash-compiler/-/jobs/2704503129

vmchale:
```
set_property top_lib {#{entity}} [current_fileset -sim]
```

fixes VHDL but breaks Verilog in the process.

martijnbastiaan:
Yeah that makes sense, Verilog doesn't have a library concept :).

DigitalBrains1:
* The `Parameters.hs` test is a bit messy. Most importantly, it omits something you should always include : `{-# NOINLINE topEntity #-}` and `{-# NOINLINE testBench #-}`.
    It also imports a lot of stuff it doesn't use, whereas I'd prefer my `-Wall` to be silent. And it defines primitives for all three HDL's but the test is only run for VHDL?
* Reordering the read files and removing `update_compile_order` fixes the test bench. So we're hitting some Vivado bug/shortcoming with `update_compile_order` here probably; no idea what's so different about it!
* What do you mean by _when the `testBench` bit is not monomorphic_? A working `testBench` is always monomorphic, so I'm missing some context here. There is the issue of #2196 relating to mandatory `INLINE` statements in test benches, but it doesn't apply to `Parameters.hs`.
* Indeed, `top_lib` will do the wrong thing for Verilog. I haven't really figured out SystemVerilog libraries and interworking between VHDL and Verilog yet in Vivado.

leonschoorl:
> it omits something you should always include : ...  {-# NOINLINE testBench #-}
Since `testBench` generally isn't used anywhere, I don't think this is needed or does anything at all.

DigitalBrains1:
Leon and I discussed this in real life.

First off, the `NOINLINE`s are not in any way related to the failures here; we discovered what causes it already.

Secondly, Leon and I agree that `topEntity` should get a `NOINLINE`, but we have different opinions about `testBench`.

[...]

PR #2347, opened 2022-10-30
Fix ordering issues in Manifest files (#2325, #2334)

### LINKED ISSUE #2325: Varying(!) sort order for transitive dependencies in manifest
Our source says:
https://github.com/clash-lang/clash-compiler/blob/f3e962be3588f2b85f07ab266ee21390c9b11fdc/clash-lib/src/Clash/Driver/Manifest.hs#L163-L170

At first I thought this TODO was outdated and we do a reverse topo sort now. Then, it turned out we do both (I here you gasping "but that is impossible!"... hold on).

The following source:
```haskell
module DepSort where

import Clash.Prelude

type T = Unsigned 8 -> Unsigned 8

f :: T
f = g . h
{-# NOINLINE f #-}
{-# ANN f (defSyn "f") #-}

g :: T
g = h . j
{-# NOINLINE g #-}
{-# ANN g (defSyn "g") #-}

h :: T
h = (+ 5)
{-# NOINLINE h #-}
{-# ANN h (defSyn "h") #-}

j :: T
j = (+ 10)
{-# NOINLINE j #-}
{-# ANN j (defSyn "j") #-}
```

gives a reverse topo sort:
```json
    "dependencies": {
        "transitive": [
            "DepSort.j",
            "DepSort.h",
            "DepSort.g"
        ]
    },
```

but splitting these functions into separate source files

<details><summary>like this</summary>

`DepSortF.hs`:
```haskell
module DepSortF where

import Clash.Prelude

import DepSortG
import DepSortH

f :: Unsigned 8 -> Unsigned 8
f = g . h
{-# NOINLINE f #-}
{-# ANN f (defSyn "f") #-}
```

`DepSortG.hs`:
```haskell
module DepSortG where

import Clash.Prelude

import DepSortH
import DepSortJ

g :: Unsigned 8 -> Unsigned 8
g = h . j
{-# NOINLINE g #-}
{-# ANN g (defSyn "g") #-}
```

`DepSortH.hs`:
```haskell
module DepSortH where

import Clash.Prelude

h :: Unsigned 8 -> Unsigned 8
h = (+ 5)
{-# NOINLINE h #-}
{-# ANN h (defSyn "h") #-}
```

`DepSortJ.hs`:
```haskell
module DepSortJ where

import Clash.Prelude

j :: Unsigned 8 -> Unsigned 8
j = (+ 10)
{-# NOINLINE j #-}
{-# ANN j (defSyn "j") #-}
```

</details>

gives us a topological sort!
```json
    "dependencies": {
        "transitive": [
            "DepSortG.g",
            "DepSortJ.j",
            "DepSortH.h"
        ]
    },
```

🤔

### LINKED ISSUE #2334: Manifest: wrong file order for include files
The following code:
```haskell
module FileOrder where

import Clash.Explicit.Prelude

topEntity ::
  Clock System ->
  Reset System ->
  Signal System Int ->
  Signal System Int ->
  Signal System Int ->
  Signal System Int
topEntity clk rst = assert clk rst "FileOrder"
{-# NOINLINE topEntity #-}
```
has the following manifest (VHDL):
```json
[...]
    "files": [
        {
            "name": "topEntity.sdc",
            "sha256": "13d4c1aafcc09a84d81d3b60d513cb901384a10e27e5a5a16572a600f15895f3"
        },
        {
            "name": "FileOrder_topEntity_types.vhdl",
            "sha256": "d02c6454a42e980254dbe81ffbdbaa1199419a1f9e1dcfca374692415139d3f1"
        },
        {
            "name": "topEntity.vhdl",
            "sha256": "e90c47000d5d18b95d9d11d375eb94693b67c49cf16a7fa493d541ff31d5bea7"
        },
        {
            "name": "topEntity_slv2string_2F58399B7F4E729C.vhdl",
            "sha256": "a94cf66fcd9f744ef94d77d5c4619960cbbacf140f8254b12cc2ee7e31bcf7b2"
        }
    ],
```

The file list is supposed to be sorted:
```haskell
  , fileNames :: [(FilePath, ByteString)]
    -- ^ Names and hashes of all the generated files for the @TopEntity@. Hashes
    -- are SHA256.
    --
    -- This list is reverse topologically sorted. I.e., a component might depend
    -- on any component listed before it, but not after it.
```
and we really want that as well. However, `topEntity.vhdl` depends on `topEntity_slv2string_....vhdl` which is listed after it.

I tried to fix this in https://github.com/clash-lang/clash-compiler/tree/topo-sort but it didn't work and I currently don't have the time to investigate. If you're working on this, feel free to (force) push to that branch and make it your own.

martijnbastiaan:
Fixes #2325

Fixes #2334

## Still TODO:

  - [x] Write tests
  - [x] Update Vivado TCL - it doesn't need to call `update_compile_order` anymore.

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -639,6 +639,8 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "T2220_toEnumOOB" def{hdlTargets=[VHDL]}
         , runTest "T2272" def{hdlTargets=[VHDL], hdlSim=[]}
         , outputTest "T2334" def{hdlTargets=[VHDL]}
+        , outputTest "T2325" def{hdlTargets=[VHDL]}
+        , outputTest "T2325f" def{hdlTargets=[VHDL]}
         ] <>
         if compiledWith == Cabal then
           -- This tests fails without environment files present, which are only
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -367,7 +367,7 @@ runClashTest = defaultMain $ clashTestRoot
           --
           -- tracked: https://github.com/clash-lang/clash-compiler/issues/2266
           let _opts = def { hdlTargets=[VHDL]
-                          , hdlSim=hdlSim def \\ [Vivado]
+                          , hdlSim=hdlSim def
                           }
           in runTest "Parameters" _opts
         , runTest "PopCount" def
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -638,6 +638,7 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "T2154" def{hdlTargets=[VHDL], hdlSim=[]}
         , runTest "T2220_toEnumOOB" def{hdlTargets=[VHDL]}
         , runTest "T2272" def{hdlTargets=[VHDL], hdlSim=[]}
+        , outputTest "T2334" def{hdlTargets=[VHDL]}
         ] <>
         if compiledWith == Cabal then
           -- This tests fails without environment files present, which are only
--- a/changelog/2022-10-30T16_44_13+01_00_fix_2334
+++ b/changelog/2022-10-30T16_44_13+01_00_fix_2334
@@ -1,0 +1,1 @@
+FIXED: Files in `clash-manifest.json` are now (correctly) listed in reverse topological order [#2334](https://github.com/clash-lang/clash-compiler/issues/2334)
--- a/changelog/2022-10-30T17_20_09+01_00_fix_2325
+++ b/changelog/2022-10-30T17_20_09+01_00_fix_2325
@@ -1,0 +1,1 @@
+Fixed: Dependencies in `clash-manifest.json` are now listed in reverse topological ordering [#2325](https://github.com/clash-lang/clash-compiler/issues/2325)
--- a/clash-lib/data-files/tcl/clashConnector.tcl
+++ b/clash-lib/data-files/tcl/clashConnector.tcl
@@ -38,7 +38,7 @@ namespace eval clash {
         variable topEntity
 
         CheckMetadataExists
-        set libs [lreverse [dict get $metadata $topEntity dependencies]]
+        set libs [dict get $metadata $topEntity dependencies]
         lappend libs $topEntity
         foreach lib $libs {
             foreach hdlFile [dict get $metadata $lib hdlFiles] {
--- a/clash-lib/src/Clash/DataFiles.hs
+++ b/clash-lib/src/Clash/DataFiles.hs
@@ -57,15 +57,6 @@ import Paths_clash_lib (getDataFileName)
   file by invoking
 
   > cabal run clash-lib:static-files -- --tcl-connector clashConnector.tcl
-
-  The Tcl Connector is currently affected by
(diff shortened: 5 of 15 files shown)


DigitalBrains1 (commented):
Thanks so much for fixing this!

I wonder if there is still some way to get files wrongly sorted. I had a whole manifesto in the commit msg for my attempt:
```markdown
We don't track dependency information for include files and memory
files. Include files are generated from
`Clash.Netlist.BlackBox.Types.bbIncludes` or YAML/JSON `includes` keys,
and memory files are generated from `~TEMPLATE` holes in the primitive
template language. We should actually track these dependencies to
maintain a proper reverse topological sort order in our manifest files.
However, this commit makes the problem less severe. We used to list
these include and memory files last in the list, but it is extremely
likely that the HDL file that caused the include/memory file to be
generated depends on it. So by listing them first instead of last, we
fix the sort order for the common case that the include/memory files
_themselves_ do not depend on anything else.

To be precise, we sort the include files after the memory files because
it is probably difficult to impossible to construct a memory file that
depends on an include file, but the reverse is easily achieved.
```

What could happen, I think, is that a primitive can reference another HDL file (because it refers to something that is generated in a separate HDL file). And if that primitive generates an include file that also references that HDL file, you'd want the include file to be in between the two HDL files in the dependency sort.

So maybe we still need a little warning somewhere that not every conceivable dependency is sorted, because that would require us to group HDL files and include files as they come to life rather than first listing all include files.

I had this FIXME in `clash-lib/src/Clash/Driver.hs` line 483:
```haskell
-- FIXME: We should track dependencies of `mfiles` and `dfiles` and
-- maintain the proper topological sort of all these.
```

DigitalBrains1:
Now that `update_compile_order` is gone, you could try to enable all the tests that currently says "Vivado segfaults" to see if that is still the case. Because a different compile order might actually avoid whatever dumb thing Vivado is trying to do when it segfaults.

Ah, also: you can re-enable the "Parameters" test (`tests/Main.hs` line 365 and further), because that one was plagued by `update_compile_order` being crappy. This PR closes issue #2266 !

martijnbastiaan:
> Ah, also: you can re-enable the "Parameters" test (tests/Main.hs line 365 and further), because that one was plagued by update_compile_order being crappy. This PR closes issue https://github.com/clash-lang/clash-compiler/issues/2266 !

Done, that works now. The others still segfault unfortunately.

martijnbastiaan:
Also, manifesto added to the relevant commit.

christiaanb (approved):

tests/Main.hs
```diff
@@ -367,7 +367,7 @@ runClashTest = defaultMain $ clashTestRoot
           --
```
DigitalBrains1, line 367: This comment also needs to go

DigitalBrains1: Wow I thought CI would take a while, it was done in under 32 minutes?

martijnbastiaan: ah flip, you're right.

martijnbastiaan: > 12 jobs for
[fix-2325-2334](https://gitlab.com/clash-lang/clash-compiler/-/commits/fix-2325-2334)
in 18 minutes and 50 seconds

GitLab :-). GHA usually takes longer, but it was feeling generous today!

PR #2353, opened 2022-11-07
 Use femtoseconds for dynamic clocks 

### LINKED ISSUE #2328: Clash should store clock periods in `fs` instead of `ps`
We currently store the clock period in picoseconds. The limitations of `ps` hasn't been a problem so far, but with the introduction of [dynamic clocks](https://github.com/clash-lang/clash-compiler/pull/2295) users might wish to represent tiny frequency differences which are currently unrepresentable:

```
clashi> hzToPeriod 200.001e6
4999
clashi> hzToPeriod 200.002e6
4999
```

To fix this, we should should store clock periods in femtoseconds (the minimal time unit in Verilog/VHDL) instead. To prevent silent breakage, I propose the following newtype:

```haskell
newtype Femtosecond = Femtosecond Natural
```

Users of `hzToPeriod` shouldn't notice any difference, as long as we change the type signature accordingly from:

https://github.com/clash-lang/clash-compiler/blob/804067a6e52263329f286d742293723010b9cc9f/clash-prelude/src/Clash/Signal/Internal.hs#L1542-L1542 

to

```haskell
hzToPeriod :: HasCallStack => Ratio Natural -> Femtosecond
```

martijnbastiaan:
Over at https://github.com/bittide/bittide-hardware/ we're trying to
model systems that can change the frequency of their clocks by tiny
steps to achieve synchronization. We currently store clock periods in
picoseconds, making these tiny changes unrepresentable:

    >>> hzToPeriod 200.001e6
    4999
    >>> hzToPeriod 200.002e6
    4999

This patch switches dynamic clocks over to femtoseconds, which does make
it representable:

    >>> hzToFs 200.001e6
    Femtoseconds 4999975
    >>> hzToFs 200.002e6
    Femtoseconds 4999950

Note that this patch does not touch any public API, so it should be
fully backwards compatible.

Although this closes #2328, a new issue should be opened indicating we
should (probably) harmonize dynamic and static clock frequency
representation.

## Still TODO:

  - [x] PR cover letter
  - [x] Write a changelog entry for added flag
  - [x] Remove outdated warning
  - [x] Add warning saying users should use `-fclash-timescale-precision 1fs`
  - [x] Add readthedocs flag documentation
  - [x] Add `ClashOpts` new field documentation
  - [x] Module documentation for `Clash.Backend.Verilog.Time`


docs/developing-hardware/flags.rst
```diff
@@ -275,3 +275,10 @@ Clash Compiler Flags
   When using one of ``--vhdl``, ``--verilog``, or ``--systemverilog``, this
   flag refers to synthesis target. For example, running Clash with
   ``clash My.Module -main-is top --vhdl`` would synthesize ``My.Module.top``.
+
+-fclash-timescale-precision
+  Sets the second part of Verilog's ``timescale 100fs/100fs``. E.g., setting this
+  flag to ``1fs`` would make Clash generate Verilog files with ``timescale 100fs/1fs``
+  as its header.
+
+  **Default:** ``100ps``
```
christiaanb, line 284:
```suggestion
  **Default:** ``100fs``
```

clash-prelude/tests/Clash/Tests/Signal.hs
```diff
@@ -216,8 +217,14 @@ case_dynamicStaticEq = do
 
     clk11 = clockGen @H11
     clk77 = clockGen @H77
-    dclk11 = dynamicClockGen @H11 (pure (hzToPeriod 11))
-    dclk77 = dynamicClockGen @H77 (pure (hzToPeriod 77))
+
+    -- We construct periods in a roundabout way (i.e., using 'hzToPeriod' instead
+    -- of using 'hzToPs'), to prevent rounding errors.
```
christiaanb, line 222:
```suggestion
    -- of using 'hzToFs'), to prevent rounding errors.
```

christiaanb, line 222:
```suggestion
    -- of using 'hzToPs'), to prevent rounding errors between periods of the static clocks and the initial periods of the dynamic clocks.
```

clash-prelude/tests/Clash/Tests/Signal.hs
```diff
@@ -240,8 +247,8 @@ case_dynamicHasEffect = do
 
     clk11 = clockGen @H11
     clk77 = clockGen @H77
-    dclk11 = dynamicClockGen @H11 (pure (hzToPeriod 11))
-    dclk77lying = dynamicClockGen @H77 (pure (hzToPeriod 78))
+    dclk11 = dynamicClockGen @H11 (pure (hzToFs 11))
```
christiaanb, line 250: Why can we use `hzToFs` here, and not do the roundabout way of doing `1000 * hzToPeriod 11` like we did earlier? Perhaps worth a comment why this is different.

christiaanb (approved):

clash-prelude/tests/Clash/Tests/Signal.hs
martijnbastiaan: Good catch, we can't!

clash-prelude/tests/Clash/Tests/Signal.hs
martijnbastiaan: Not only the initial periods, but broader point taken!

clash-prelude/default.nix
```diff
@@ -3,8 +3,7 @@
 with nixpkgs.pkgs;
 with gitignore;
 
-# We disable tests as doctests don't play nice with nix
-# (issue with ghc plugins)
+# We disable tests as doctest-parallel don't play nice with nix
```
DigitalBrains1, line 6: Arguably `doctest-parallel` is plural (that was the point, right), but still:
```suggestion
# We disable tests as doctest-parallel doesn't play nice with nix
```

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -929,6 +935,21 @@ tbClockGen done = Clock (done `seq` SSymbol) Nothing
 {-# NOINLINE tbClockGen #-}
 {-# ANN tbClockGen hasBlackBox #-}
 
+-- | Femtoseconds expressed as an 'Int64'. Is a newtype to prevent accidental
+-- mixups with picoseconds - the unit used in 'DomainConfiguration'.
+--
+newtype Femtoseconds = Femtoseconds Int64
+  -- No 'Integral' instance to prevent accidental picoseconds / femtoseconds mixup
+  deriving (Show, Eq, Generic, NFDataX, NFData, Lift)
```
DigitalBrains1, line 943: Maybe also a `TryDomain` instance?

martijnbastiaan: Yeah good call

DigitalBrains1 (approved):

PR #2357, opened 2022-11-10
Tcl connector: add verbosity, dry-run

DigitalBrains1:
`clash::verbosity` sets the logging level and the boolean `clash::dryRun` enables printing of actions instead of performing them. Log level 1, the default, is as verbose as the script was before. Log lines are now prefixed by `Clash [n]:` where _n_ is the log level of the message that follows.

Documentation will be added later.

Alternatively, if you prefer, we can extend this PR with the doc before merging.

## Still TODO:


martijnbastiaan (approved):
Very nice!

Just a comment with something along the lines of:


```
Set verbosity level. Accepted levels: 1 - 4. Example:

    set clash::verbosity 3
```

..would suffice.

PR #2359, opened 2022-11-18
Add GHC 9.4 to compatibility notes

martijnbastiaan:


DigitalBrains1:
The fact that the ǂ (alveolar click) and the ‡ (double dagger) are so similar makes me think you should use different footnote marks. According to [Wikipedia](https://en.wikipedia.org/wiki/Note_(typography)#Numbering_and_symbols),  *, †, ‡, §, ‖, ¶, in that order, are fairly common in English. Alternatively, we could simply do <sup>1</sup>, <sup>2</sup>, <sup>3</sup>.

PR #2364, opened 2022-11-21
Add Xilinx specific 'dualFlipFlopSynchronizer'

martijnbastiaan:
Uses ASYNC_REG to instruct Vivado to pack register tightly and ignore timing checks.

--------

Even though it's a draft, I wouldn't mind a look @DigitalBrains1 

## Still TODO:

  - [ ] VHDL implementation (is this possible?)
  - [ ] More sensible test
  - [ ] Documentation

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Synchronizer.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Synchronizer.hs
@@ -1,3 +1,11 @@
+{-|
+  Copyright   :  (C) 2022 Google Inc
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+  Xilinx specific synchronization primitives.
+-}
+
 {-# LANGUAGE QuasiQuotes #-}
 
 module Clash.Cores.Xilinx.Synchronizer where
@@ -5,7 +13,7 @@ module Clash.Cores.Xilinx.Synchronizer where
 import Clash.Annotations.Primitive
 import Clash.Explicit.Prelude
 
-import Data.String.Interpolate      (__i)
+import Data.String.Interpolate (__i)
 
 -- | See https://docs.xilinx.com/r/en-US/ug901-vivado-synthesis/ASYNC_REG
 dualFlipFlopSynchronizer ::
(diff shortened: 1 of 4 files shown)


rowanG077:
I think to make this synchronize safe for multi-bit constructs there needs to be a third register in the source domain with a `max_delay` timing constraint on the data whose outputs feeds into first register in the synchronizer chain. Without it the separate wires going into the first synchronizing register could have wildly different delays leading to issues.

It's hard to make this generic. Because often you want the register in the source domain to be part of the logic you write. e.g. the async FIFO contains state registers where the grey codes for the pointers are stored. These feed directly into the synchronizer.

So unfortunately I offer no answers here on how to make this a nice API but just food for thought :).

martijnbastiaan:
True. I did limit the input to one bit here, maybe we should do that for Prelude's version too. For word sync I guess people could use the dual clock FIFO, this seems to be what Xilinx recommends anyway.

DigitalBrains1:
Multi-bit sync with flip flop synchronizers is unsolvable in the general case; it's why FIFO synchronizers use Gray code with, for Vivado, `set_max_delay` as well as `set_bus_skew`. With those guarantees, for each individual bit, you get either the value from the source domain from clock tick `n-1` or from clock tick `n`. And since the Gray code plus the `+1` behaviour of the counter guarantees that only a single bit changes on every source domain clock tick, you get either the previous, or the new value of the Gray-encoded counter. If you don't use Gray code, your multi-bit data will be garbled as each individual bit might belong to the value it was at tick `n-1` or at tick `n`.

And with Gray code but without the constraints, you might get samples from source domain clock tick `n-k` for `k` larger than 1, which would again garble the counter.

So I don't think we need to care about multi-bit constructs here. Perhaps the doc could carry a warning, but it can probably be very generic that clock crossing requires a lot of care and attention, and that we only apply `ASYNC_REG` here.

DigitalBrains1 (commented):
I feel this might be in the grey area of where `clash-prelude` ends and `clash-cores` starts. We have _some_ vendor-specific stuff in Prelude, and this is pretty basic.

Why did you omit the reset? Or is this still to be done?

DigitalBrains1:
Damn, it threw away my review. Must have pressed the oh-shit-button.

clash-cores/src/Clash/Cores/Xilinx/Synchronizer.hs
```diff
@@ -0,0 +1,74 @@
...
+  . unsafeSynchronizer clk1 clk2
+{-# NOINLINE dualFlipFlopSynchronizer #-}
+{-# ANN dualFlipFlopSynchronizer hasBlackBox #-}
+{-# ANN dualFlipFlopSynchronizer (InlineYamlPrimitive [Verilog, SystemVerilog] [__i|
+ BlackBox:
+   kind: Declaration
+   name: Clash.Cores.Xilinx.Synchronizer.dualFlipFlopSynchronizer
+   template: |-
+     // begin dualFlipFlopSynchronizer
+     (* ASYNC_REG = "TRUE" *) reg ~GENSYM[sync_a][0], ~GENSYM[sync_b][1];
```
DigitalBrains1, line 41: I wonder if we couldn't just alter the `clash-prelude` `dualFlipFlopSynchronizer` to have a black box that says:
```
~IF~VIVADO~THEN(* ASYNC_REG = "TRUE" *) ~FIreg ~GENSYM[sync_a][0], ~GENSYM[sync_b][1];
```

Then you'd just have to specify `-fclash-hdlsyn vivado` which might already be needed for some workaround anyway.

DigitalBrains1:
The `BitSize a ~ 1` constraint doesn't prevent multi-bit sync issues. If you sync two bits separately, but then write some logic that depends on the state of both, you still get really odd results. If for instance in the source domain it is known that for two bits, the combination `11` will never occur, if you dual FF sync those two bits to another domain, suddenly you have a new state where that combination _will_ occur. And your logic would go haywire even though it looked like an unreachable state.

clash-cores/src/Clash/Cores/Xilinx/Synchronizer.hs
rowanG077: In fact it's would be really a maintainer nightmare to go vendor specific here. We would need to move the `asyncFIFOSynchronizer` to Xilinx as well using the Xilinx specific `dualFlipFlopSynchronizer`. If possible I would greatly prefer to add this to the original `asyncFIFOSynchronizer`

DigitalBrains1: I think `asyncFifoSynchronizer` might be better off not using another prim for its FF synchroniser anyway. Since it needs to specify additional constraints, it might as well use a private FF chain with those constraints, once we decide how we add those constraints in the first place.

martijnbastiaan: We could just dump it in there anyway. Vendors treat unrecognized attributes as comments anyway AFAIK.

martijnbastiaan:
Okay, so after discussing it offline: we'd like to make this part of the existing `dualFlipFlopSynchronizer`. However, this thing is a complex beast (with reset lines, enables, etc.) so writing a black box for it isn't easy. For now we'll just publish this as part of Bittide instead.

PR #2369, opened 2022-12-02
fix spi core, extra clock edge in sample trailing mode

lmbollen:
The current implementation of the SPI core has a bug where, `SPIMode1` or `SPIMode3`, the clock would feature one too many edgeswhile SS is active, causing the idle polarity to switch after each communication. This could lead to unexpected behavior by the connected peripheral.

This problem is fixed by the change in `spiGen`.

The second problem that arose is that after fixing spiGen, the circuit would not sample the input data at the right time, causing part of the sent bits to become undefined.

This is fixed by the changes in `spiCommon`.

## Still TODO:

### DIFF
--- a/clash-cores/src/Clash/Cores/SPI.hs
+++ b/clash-cores/src/Clash/Cores/SPI.hs
@@ -1,5 +1,6 @@
 {-|
   Copyright   :  (C) 2019, Foamspace corp
+                     2022, Google LLC
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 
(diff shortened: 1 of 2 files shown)


martijnbastiaan (changes requested):
In general LGTM, but I think it could improve VCS wise. [This article](https://dev.to/wordssaysalot/art-of-writing-a-good-commit-message-56o7) covers it. More specifically:

* Start your commit message with a capital letter
* Describe what you did, e.g. ```Correct number of edges generated in `SPIMode{1,3}` ```.
* Move your PR cover letter to the commit body. 

> Before this patch, the SPI core would generate one too many edges causing [..]. After this patch, `spiGen` no longer waits for a superfluous edge either.

martijnbastiaan (approved):

PR #2370, opened 2022-12-06
Tcl connector: Handle plain constraint files

DigitalBrains1:
When the Tcl connector finds a file named `*.sdc` or `*.xdc`, and it is not handled by an accompanying Clash<->Tcl interface file, it will issue a `read_xdc` for that file. Exception: for every top component (as declared in `clash-manifest.json`), it will only load the `.sdc` file with the same file name as the top component name for the actual top entity in the design, not for subentities. Clash emits `create_clock` statements in those files, and we should only load it for the actual top entity.

Note that because all `.sdc` and `.xdc` files are loaded with an unadorned `read_xdc <file>`, they should use globally unique identifiers to match only the intended signals, entities and so on. That way, they will not unintendedly match things outside the correct entities as the statements are not scoped to any entity explicitly.

Furthermore, Clash<->Tcl interface files are now named ~~`*.tclash`~~ `*.clash.tcl`.

It's possible that you would want to pass Vivado a constraint file with the extension `.tcl`. At least in Project Mode, and maybe in Non-Project Mode, this signifies an "unmanaged" constraint file: Vivado will allow GUI edits to files named 
`.sdc` and `.xdc`. By picking our own extension for Clash<->Tcl interface files, we make it possible to name constraint
files `.tcl`. Without this change, the Tcl connector script would try to parse such a file as a Clash<->Tcl interface file and error out.

Finally, the first commit improves the coding style of the connector script.

## Still TODO:

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/DcFifo/Internal/BlackBoxes.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/DcFifo/Internal/BlackBoxes.hs
@@ -65,7 +65,7 @@ dcFifoBBF _isD _primName args _resTys
   bbMeta = emptyBlackBoxMeta
     { N.bbKind = N.TDecl
     , N.bbIncludes =
-      [ ( ("dcfifo", "tcl")
+      [ ( ("dcfifo", "clash.tcl")
         , BBFunction (show 'dcFifoTclTF) 0 (dcFifoTclTF dcConfig))
       ]
     -- TODO: Make this blackbox return multiple results, instead of a tuple. See:
--- a/clash-cores/src/Clash/Cores/Xilinx/Floating.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Floating.hs
@@ -8,7 +8,7 @@ Support for the [Xilinx Floating-Point LogiCORE IP v7.1](https://www.xilinx.com/
 
 The functions in this module make it possible to use the Xilinx IP in Clash.
 Compilation will instantiate the Xilinx IP. Clash will output a TCL script
-named @floating_point@/.../@.tcl@ which needs to be executed in the Vivado
+named @floating_point@/.../@.clash.tcl@ which needs to be executed in the Vivado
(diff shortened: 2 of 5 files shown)


martijnbastiaan (approved):
I think I'm going to refer people here if they want to see properly formatted, semantically separated commits :+1:.

martijnbastiaan:
I was slightly leaning towards `.clash.tcl` anyway, so it sounds good to me. My reasoning for the `.primitives.yaml` was that:

* It signals that they're primitives (doh)
* Editors will do the correct thing and apply YAML syntax highlighting

Similar reasoning would go for `.clash.tcl`. (I already said this in the office, but writing here for prosperity.)

DigitalBrains1:
Right, for posterity broadly my reasons for initially picking `.tclash`:
* To me, multiple extensions are each a unit; often they indicate layers. I can use `gunzip` to turn a `.tar.gz` into a `.tar` and I can decrypt with GnuPG to turn an `.ods.gpg` into an `.ods`. The Apache http server can also use multiple extensions to [encode content variants](https://httpd.apache.org/docs/2.4/content-negotiation.html#naming) in server driven content negotiation. There, the order doesn't matter, `.nl.iso8859-1.html` is equivalent to `.html.iso8859-1.nl`. It's no longer a matter of layers, rather they're aspects, but they are still separate units. `.clash.tcl` is one unit.
* You might want to configure your computer to offer custom actions in your file manager or use custom syntax highlighting etc. in your editor. I suspect some tools will only allow to match the final extension. You lose the "lazy" option of automatic highlighting and gain the power user option of custom highlighting and stuff like autocompletion. I might well be too lazy to configure more than an equivalence to `.tcl`, but I _could_.<sup>1</sup>

<sup>1</sup> I just checked. Both my editor vim and my file manager Thunar are happy to match on `.clash.tcl` as well. I never use the file manager to browse Clash output, but to me it's about the broader principle of the structure of file names in general, not just how I use Clash specifically.

PR #2374, opened 2022-12-15
Make `resetGlitchFilter` less resource intensive

martijnbastiaan:
This PR turned out to be slightly more involved than I anticipated: the current iteration doesn't simulate without initial values.

## Still TODO:

### DIFF
--- a/.ci/gitlab/test.yml
+++ b/.ci/gitlab/test.yml
@@ -61,27 +61,27 @@ build-clash-dev:
 cores:unittests:
   extends: .test-common
   script:
-    - cabal v2-run clash-cores:unittests
+    - cabal v2-run -- clash-cores:unittests --hide-successes
 
 cosim:unittests:
   extends: .test-common
   script:
-    - cabal v2-run clash-cosim:tests
+    - cabal v2-run -- clash-cosim:tests
 
 prelude:unittests:
   extends: .test-common
   script:
-    - cabal v2-run clash-prelude:unittests
+    - cabal v2-run -- clash-prelude:unittests --hide-successes
 
 lib:doctests:
   extends: .test-common
   script:
-    - cabal v2-run clash-lib:doctests
+    - cabal v2-run -- clash-lib:doctests -j${THREADS}
 
 lib:unittests:
   extends: .test-common
   script:
-    - cabal v2-run clash-lib:unittests
+    - cabal v2-run -- clash-lib:unittests --hide-successes
 
 prelude:doctests:
   extends: .test-common
--- a/changelog/2022-12-15T11_04_34+01_00_use_counter_in_resetGlitchFilter
+++ b/changelog/2022-12-15T11_04_34+01_00_use_counter_in_resetGlitchFilter
@@ -1,0 +1,3 @@
+CHANGED: `resetGlitchFilter` now uses a counter instead of shift register, allowing glitch filtering over much larger periods.
+CHANGED: `resetGlitchFilter` now filters glitches symmetrically, only deasserting the reset after the incoming reset has stabilized. For more information, read [#2374](https://github.com/clash-lang/clash-compiler/pull/2374).
+CHANGED: `resetGlitchFilter` does not support domains with unknown initial values anymore. Its previous behavior could lead to unstable circuits. Domains not supporting initial values should consider using power-on-resets and glitch filters specifically designed for this environment.
--- a/clash-prelude/src/Clash/Explicit/Reset.hs
+++ b/clash-prelude/src/Clash/Explicit/Reset.hs
@@ -1,5 +1,6 @@
 {-|
 Copyright  :  (C) 2020-2021, QBayLogic B.V.
+                  2022     , Google LLC
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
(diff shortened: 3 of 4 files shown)


changelog/2022-12-15T11_04_34+01_00_use_counter_in_resetGlitchFilter
```diff
@@ -0,0 +1 @@
+CHANGED: `resetGlitchFilter` now uses a counter instead of shift register, allowing much larger periods to be filtered. Its implementation is now symmetrical too: it will only deassert resets after seeing the incoming reset deasserted for a configurable amount of time.
```
lmbollen, line 1: >  configurable amount of time.

to
>  configurable number of cycles.

?

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -176,12 +173,14 @@ resetSynchronizer clk rst = rstOut
                          $ unsafeFromReset rst
 {-# NOINLINE resetSynchronizer #-} -- Give reset synchronizer its own HDL file
 
-data GlitchFilterState = Idle | InReset
+data GlitchFilterState = Asserted | Deasserted
```
lmbollen, line 176: Add haddock comment referencing 'resetGlitchFilter'.

lmbollen (approved):

DigitalBrains1:
Why did you make it symmetric?

> the current iteration doesn't simulate without initial values.

Yeah, as discussed, I think we shouldn't try to fix that in this particular function; it is just only for platforms with initial values. Shall we make the function error out¹ both in simulation and HDL generation when its domain has `InitBehavior` `Undefined`?

¹ We could enforce this at the type level but all these constraints carry over to encompassing functions and it just gets annoying that the functions in your design accumulate all these constraints. That's why in similar situations we decided we'd rather error out (as long as we make sure we do that both in simulation and HDL generation). Though I'll admit that this particular function is often only found very near the top, so its constraints usually don't carry far :-).

.ci/gitlab/test.yml
```diff
@@ -71,7 +71,7 @@ cosim:unittests:
 prelude:unittests:
   extends: .test-common
   script:
-    - cabal v2-run clash-prelude:unittests
```
DigitalBrains1, line 74: In this line, you add an option terminator (`--`), but not in the others. Shouldn't they all need it? Could you please make it consistent regardless?

DigitalBrains1 (changes requested):
I already asked why you made it symmetric. I'm not sure it's the right thing to do, but if it is, could you document why you made it symmetric in the commit msg and I think also the Changelog?

martijnbastiaan:
It _seemed_ like the right thing to do. Do we want a glitch in the reset to immediately yoink it to the deasserted state?

.ci/gitlab/test.yml
martijnbastiaan: They should be added to the others. In fact, it causes CI failures now.

DigitalBrains1:
Really? I thought it was more customary without the deassertion filter. Like, you don't want a reset to happen when the reset pulse is an unintended glitch. That's why you filter stuff that you deem too short.

> Do we want a glitch in the reset to immediately yoink it to the deasserted state?

Why not? We've responded to the reset.  Why would you keep the reset asserted longer? If you need long reset pulses, we have `holdReset`. A reset glitch filter prevents _triggering_ a reset on a glitch.

Maybe it helps to consider the source of the glitches? Like, a button only glitches when you press it down or when you release it, not when you keep it pressed down or it is in the up position. In this case, all you need to filter is the _repetition_. Once it starts glitching, you know there's going to be a real change. But we want the glitch filter for more things than just button debouncing, it's just an example of glitches (in fact, the previous implementation with the shift register might be wholly unsuitable for button debouncing, waayy too long periods of bounce).

martijnbastiaan:
I don't really mind either way. If you want me to change it I'll do it.

> Why not? We've responded to the reset. Why would you keep the reset asserted longer?

It makes the implementation (every so slightly) more complex AFAICT if you need to account for asymmetry. But again, I don't really mind either way. Neither the increased hardware complexity nor the increased wait time (in context of pressing physical buttons) seem like real issues to me.

DigitalBrains1:
If this were the _addition_ of a glitch filter I would never have balked at the symmetric implementation. However, you changed an existing implementation to have different qualities, that's why my ears pricked up :-). I think the difference in the amount of LUTs is too small to consider relevant. By changing the deassertion, users might need to change their test suite for a design: a test runs a few cycles longer¹, hard-coded samples or hard-coded cycle numbers will need to be adapted. If a change is an improvement, that is perfectly fine. I'm not convinced here, kinda on the fence, but I'll let you make the call.

Are you changing the implementation to debounce a button? (By the way, the button is definitely not the only signal where a glitch actually always introduces a change in level.) We should probably also just have a button debouncing function, although there are so many possible implementations X-D. I think Gergő has one in his book? I've written one once some day that debounces a bunch of buttons at the same time. All buttons needed to stop bouncing before the bounce was considered settled, so there were cases where it was insufficient (really strange key rollover properties), but it did only require one counter for a group of buttons. Also, as soon as it saw a change in level it immediately reported that as the new state, but after that "the latch went opaque" until the signal settled.

¹ I'm reminded of one of our own test cases that had a hard-coded duration but the test was slightly changed without changing the duration and it no longer ran the last few samples, reducing the test coverage.

DigitalBrains1 (approved):

martijnbastiaan:
The original documentation said:

> It will then stay asserted for as long as the given reset was asserted consecutively.

Sooo.. the old implementation was symmetrical? I'm confused.

martijnbastiaan:
Right, it would keep the reset asserted for at least _n_ cycles guaranteed, but if a deasserted reset happened during that period (or after) it will immediately deassert. That's a bit harder to replicate, so I'm going to opt to make it fully symmetrical (and update the docs). 

Thanks for the reviews!

DigitalBrains1:
Right, that's a rather odd implementation. I forgot about that `msb` check during this discussion.

I think after the reset was asserted for _n_ ¹ cycles consecutively, and then after _n+k_ cycles a single unasserted sample occurs, it will be asserted for cycle numbers _n_ through _2n+k_ give or take a sample (so for a duration of _n+k_, in other words, a length _n_ is guaranteed).

Properly symmetrical sounds fine, and actually, many tests will respond the same, I think (since it will deassert _n_ cycles after the reset is released, if the reset isn't actually glitchy in the test). That's awesome, if I had realised that earlier I would have been more positive! You can see it in the doctest; only the glitched deassertion changes.

¹ Configurable value for `resetGlitchFilter`.

martijnbastiaan:
Good point, that's great.

Issue #2279, opened 2022-07-10
Improved Jargon in Haddock


alex-mckenna:
Browsing through the docs for `clash-prelude` before I had a thought: it would be nice to pick some jargon to disambiguate the following:

  * code which is translatable to HDL, but not synthesizable HDL
  * code which is not translatable to HDL

To avoid bikeshedding too much on the words used I'll propose here _(non-)synthesizable_ for talking about vendor tooling being able to synthesize generated HDL, and _(non-)translatable_ for talking about Clash being able to generate HDL in the first place. This has some benefits:

  * places where we warn about non-synthesizable code currently may refer to non-translatable code and this would tidy up the documentation for these functions
  * by introducing the concept of _translatable Clash_ in documentation we can make it clearer to users the subset of Haskell types which Clash will generate HDL for (i.e. `[]` is non-translatable because functions on it have no way of setting a bound for recursion unlike functions on `Vec`)

The obvious downside is explaining such a general term is a difficult thing to find a place for in API documentation. For other resources, like a hypothetical example-based Clash book it would be easy to introduce into the flow of the documentation.

CC @DigitalBrains1, who often has good ideas about documentation

PR #2377, opened 2022-12-16
Add `Clash.Magic.simulation`

martijnbastiaan:
Use cases:

 * Use a fast path in simulation
 * Deal with `hasX` and friends

## Still TODO:


leonschoorl:
I think the documentation could make it more clear that this is only true in Haskell simulation, and false for simulation done on the HDL.

martijnbastiaan:
What do you think of the current doc?

alex-mckenna:
This is as deeply magical as it is useful and I love it. I wonder if we could make a type-level version of this work, i.e.

```haskell
type Foo = If Simulation ClashType HaskellType
```

martijnbastiaan:
> This is as deeply magical as it is useful and I love it.

I love the implementation complexity vs usefulness ratio haha. 

Not sure about the type-level version! I'm inclined to say _probably_ not.

DigitalBrains1 (changes requested):
LGTM in general! One phrasing thingy.

DigitalBrains1:
> > This is as deeply magical as it is useful and I love it.
> 
> I love the implementation complexity vs usefulness ratio haha.

Yeah, I absolutely adore it too!

changelog/2022-12-16T12_15_46+01_00_add_Clash_Magic_simulation
```diff
@@ -0,0 +1 @@
+ADDED: `Clash.Magic.simulation`, a way to differentiate between simulation and HDL translation.
```
DigitalBrains1, line 1: (My phrasing thingy. I forgot to check it into the review)

I just checked all occurrences of the word _HDL_ in `clash-prelude`. As I‍ suspected, we usually talk about generating HDL / generated HDL. (Also, I encountered "synthesized to HDL" which I object to.)

So I'd like to suggest to change "HDL translation / translating to HDL" to "HDL generation / generating HDL".

martijnbastiaan: RIP https://github.com/clash-lang/clash-compiler/issues/2279. But yeah, agreed.

DigitalBrains1: I'm fine with HDL being _generated_ for _translatable_ functions! So no need to bury #2279 yet. Translatability is a static property, generation a process. They can co-exist.

We'd just need to weed out that "synthesized to HDL" because I'd like to reserve the word synthesis and its companion word forms for what #2279 calls synthesisable code.

clash-prelude/src/Clash/Magic.hs
```diff
@@ -30,6 +30,9 @@ module Clash.Magic
   -- ** Functions to control Clash's (de)duplication mechanisms
   , deDup
   , noDeDup
+
+  -- ** Utilities to differentiate between simulation and normalization
```
DigitalBrains1, line 34: Soft suggestion: normalization -> generating HDL. No need to confuse people new to the jargon with yet another.

clash-prelude/src/Clash/Magic.hs
```diff
@@ -234,3 +237,8 @@ noDeDup
   :: forall a . a -> a
 noDeDup = id
 {-# NOINLINE noDeDup #-}
+
+-- | 'True' in Haskell simulation. Replaced by 'False' when translating to HDL.
```
DigitalBrains1, line 241:
```suggestion
-- | 'True' in Haskell simulation. Replaced by 'False' when generating HDL.
```

DigitalBrains1 (changes requested):

christiaanb (approved):

PR #2378, opened 2022-12-17
Don't quote bit literals with `-fclash-force-undefined`

### LINKED ISSUE #2360: -fclash-force-undefined produces invalid values for BitVector with undefined bits.
Title says it all I think.

Reproducer:
```haskell
module Example.Project where

import Clash.Prelude

topEntity ::(BitVector 8)
topEntity = deepErrorX "broken"

```
compiled with `cabal run -- clash --verilog Example.Project -fclash-force-undefined=0`, produces:
```verilog
** GENERATED BY CLASH 1.6.4. DO NOT MODIFY.
*/
`timescale 100fs/100fs
module topEntity
    ( // No inputs

      // Outputs
      output wire [7:0] result
    );


  assign result = 8'b'0''0''0''0''0''0''0''0'; // invalid value


endmodule
```

HDL was generated with 1.6.4, but the problem was encountered with 1.7.0.
Problem also exists for systemverilog and vhdl.

alex-mckenna:
When rendering bit literals, we use the `char` function from the prettyprinter in most cases. For undefined bits we were using the `int` function, and wrapping in single quotes. This is not the same: the `char` function prints a single Haskell `Char`, but does _not_ wrap that character in quotes.

Simply removing the quote marks solves the issue.

Fixes #2360.

## Still TODO:

  - [ ] ~Check copyright notices are up to date in edited files~

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -635,6 +635,7 @@ runClashTest = defaultMain $ clashTestRoot
         , outputTest "T2325f" def{hdlTargets=[VHDL]}
         , runTest "T2342A" def{hdlSim=[]}
         , runTest "T2342B" def{hdlSim=[]}
+        , runTest "T2360" def{hdlSim=[],clashFlags=["-fclash-force-undefined=0"]}
         ] <>
         if compiledWith == Cabal then
           -- This tests fails without environment files present, which are only
--- a/changelog/2022-12-17T15_33_33+01_00_render_force_undefined_bits_correctly.md
+++ b/changelog/2022-12-17T15_33_33+01_00_render_force_undefined_bits_correctly.md
@@ -1,0 +1,2 @@
+FIXED: Clash now renders undefined bits set via `-fclash-force-undefined`
+correctly [#2360](https://github.com/clash-lang/clash-compiler/issues/2360).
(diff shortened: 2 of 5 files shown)


martijnbastiaan (approved):

tests/Main.hs
```diff
@@ -635,6 +635,7 @@ runClashTest = defaultMain $ clashTestRoot
         , outputTest "T2325f" def{hdlTargets=[VHDL]}
         , runTest "T2342A" def{hdlSim=[]}
         , runTest "T2342B" def{hdlSim=[]}
+        , runTest "T2360" def{hdlSim=[]}
```
DigitalBrains1, line 638: Shouldn't this pass `-fclash-force-undefined=0` to properly test the issue?

DigitalBrains1 (changes requested):

tests/Main.hs
DigitalBrains1: And it also doesn't actually run it in any HDL tool? I think the problem only crops up on actual running?

alex-mckenna: Shh, _you saw nothing_

DigitalBrains1 (approved):

Issue #2005, opened 2021-11-19
Improve documentation for `XException`


alex-mckenna:
Currently, the documentation of the [`XException`](https://hackage.haskell.org/package/clash-prelude-1.4.6/docs/Clash-XException.html) module is not very friendly to beginners. Ideally (in my opinion) the following should be included:

  * the difference between the Haskell `error` / `undefined` and the Clash `errorX` / `undefined`
  * the relationship between `undefined` and `errorX` in Clash (i.e. one fails with a custom message)
  * the difference between `errorX` and `deeperrorX`
  * examples for all of the above

Additionally, there are some problems with the documentation that does exist:

  * the documentation for `XException` itself says "An exception representing an "uninitialized" value.". Especially for VHDL users, there is a difference between uninitialised (`U`), don't care (`-`) and undefined (`X` or `W`) in HDL. I think we should explain how Clash's `XException` corresponds to different forms of undefined / uninitialised in IEEE 1164 and the 4-value logic of IEEE 1364 / IEEE 1800. I think relating back to undefined values in HDLs we support will help users build intuition
  * some functions, e.g. `isX` make reference to weak head normal form, however this is explained nowhere in the Clash documentation as far as I know. I think this module should make an effort to explain terms like this before using them
  * an ASCII representation of bottom (`_|_`) is used, however this symbol may not be known by hardware designers who are not Haskell programmers (or even more novice Haskell programmers)

CC @DigitalBrains1, who started the discussion on slack that led to me making this issue

DigitalBrains1:
I think it would also be good to mention how `Bit` and [`BitVector`](https://hackage.haskell.org/package/clash-prelude-1.5.0/candidate/docs/Clash-Sized-BitVector.html) are proper three-valued types and how `XException` interacts with [`BitPack`](https://hackage.haskell.org/package/clash-prelude-1.5.0/candidate/docs/Clash-Class-BitPack.html). Actually, we don't even really explain that `Bit` and `BitVector` are three-valued in the [`BitVector` module documentation](https://hackage.haskell.org/package/clash-prelude-1.5.0/candidate/docs/Clash-Sized-BitVector.html).

How `errorX` interacts with `BitPack`:
```haskell
>>> pack (errorX "Where did I go?" :: Unsigned 8)
0b...._....
>>> unpack $ pack (errorX "Where do I go?" :: Unsigned 8) :: Unsigned 8
[...]

PR #2380, opened 2022-12-18
Add documentation to `Clash.XException`

martijnbastiaan:
First stab at fixing #2005. This is very much a work in progress, but feel free to critique / help out.

Pinging @alex-mckenna @DigitalBrains1 

## Still TODO:

  - [ ] A lot

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,12 +1,16 @@
 {
   "cSpell.words": [
-    "NOINLINE",
-    "WHNF",
+    "dflipflop",
     "fmap",
     "frontends",
     "inlines",
+    "NOINLINE",
     "scrutinee",
-    "subterms"
+    "subterms",
+    "Twente",
+    "Verilog",
+    "VHDL",
+    "WHNF"
   ],
   "files.exclude": {
     "**/*.dyn_hi": true,
--- a/clash-prelude/src/Clash/XException.hs
+++ b/clash-prelude/src/Clash/XException.hs
@@ -6,14 +6,58 @@ Copyright  :  (C) 2016,      University of Twente,
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
-'XException': An exception for uninitialized values
+This module houses utilities to deal with /undefined/ values in Clash. An undefined
+value is translated to @x@ in VHDL and Verilog. At its core, this is handled by
+a special 'Exception' called 'XException'.
 
->>> show (errorX "undefined" :: Integer, 4 :: Int)
-"(*** Exception: X: undefined
-CallStack (from HasCallStack):
+Like any 'Exception' in Haskell, any term can take on this value:
+
+>>> error "help" :: Int
+*** Exception: help
 ...
->>> showX (errorX "undefined" :: Integer, 4 :: Int)
-"(undefined,4)"
+>>> errorX "help" :: Int
+*** Exception: X: help
+...
+
+Within the Haskell ecosystem, and throughout this module, you will find these
+exception values being referred to as /bottom/. This is sometimes written as a
+symbol: @_|_@.
+
+All 'Exception's, including 'XException', are \"infectious\". Any value depending
+on an 'Exception' will itself be replaced by it:
+
+>>> let x = error "help" :: Int
+>>> x + 1
+*** Exception: help
+
+In typical non-IO Haskell, exceptions are generally used to halt execution in case
+of /the impossible/. E.g., situations where GHC can't proof a branch won't be
+evaluated, but it nevertheless won't:
+
+@
+if | a < 0     = "less than"
+   | a == 0    = "equal"
+   | a > 0     = "greater than"
+   | otherwise = error "Can't happen"
+@
+
+In Clash they are used far more pervasively, as hardware architectures often
+deal with /undefinedness/. For example, the initial value of a 'dflipflop' is
+/undefined/:
+
+>>> showX (sampleN @System 3 (dflipflop (pure 5)))
+"[undefined,5,5]"
+
+Note that we had to use a special version of 'show' called 'showX' here, as the
+former treats any bottom as an immediate cause to end program execution:
+
+>>> show (sampleN @System 10 (dflipflop (pure 5)))
+"[*** Exception: X: First value of dflipflop undefined
+
+In this we find the heart of "Clash.XException": a way to continue simulating
+when facing /undefinedness/. The two most important classes are 'NFDataX' and
+'ShowX'. You are encouraged to read their documentation first before diving into
+other definitions.
 -}
 
 {-# LANGUAGE CPP #-}
@@ -476,7 +520,10 @@ rwhnfX = (`seqX` ())
 {-# INLINE rwhnfX #-}
 
 -- | Class that houses functions dealing with /undefined/ values in Clash. See
--- 'deepErrorX' and 'rnfX'.
+-- this module's documentation for a general introduction.
+--
+-- TODO
+--
 class NFDataX a where
   -- | Create a value where all the elements have an 'errorX',
   -- but the spine is defined.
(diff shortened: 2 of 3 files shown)


clash-prelude/src/Clash/XException.hs
```diff
@@ -6,14 +6,58 @@ Copyright  :  (C) 2016,      University of Twente,
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
-'XException': An exception for uninitialized values
+This module houses utilities to deal with /undefined/ values in Clash. An undefined
+value is translated to @x@ in VHDL and Verilog. At its core, this is handled by
```
alex-mckenna, line 10: Is this true? I vaguely remember some clash undefined values becoming "don't care" in VHDL

clash-prelude/src/Clash/XException.hs
```diff
@@ -6,14 +6,58 @@ Copyright  :  (C) 2016,      University of Twente,
...
+>>> error "help" :: Int
+*** Exception: help
+...
+>>> errorX "help" :: Int
+*** Exception: X: help
 ...
->>> showX (errorX "undefined" :: Integer, 4 :: Int)
-"(undefined,4)"
+
+Within the Haskell ecosystem, and throughout this module, you will find these
```
alex-mckenna, line 22: I wonder if it's a good idea to mention bottom at all. If I was unfamiliar with the concept I would now be confused about what it means. Attempting to look up it's definition may lead to PLT type definitions which only add more confusion

clash-prelude/src/Clash/XException.hs
```diff
@@ -6,14 +6,58 @@ Copyright  :  (C) 2016,      University of Twente,
...
+
+@
+if | a < 0     = "less than"
+   | a == 0    = "equal"
+   | a > 0     = "greater than"
+   | otherwise = error "Can't happen"
+@
+
+In Clash they are used far more pervasively, as hardware architectures often
+deal with /undefinedness/. For example, the initial value of a 'dflipflop' is
```
alex-mckenna, line 45: Let's not use undefinedness. It's not really a proper word, and it reads clunkily IMO. I can't think of a good single word to take it's place

.vscode/settings.json
```diff
@@ -1,12 +1,16 @@
 {
   "cSpell.words": [
-    "NOINLINE",
-    "WHNF",
+    "dflipflop",
```
alex-mckenna, line 3: Or in Dutch, _de flipflop_

alex-mckenna (commented):
Some initial thoughts, by no means is this a proper review

.vscode/settings.json
DigitalBrains1: I think this thing is properly spelled "D flip-flop", not as a single word.

I don't have time to review further now.

DigitalBrains1: Looking at the rest of the words, I think it's spelled front-end, with a dash, like the flip-flop.

martijnbastiaan: We have a function called `dflipflop`, so it comes up often :sweat:

DigitalBrains1: You can't have VSCode ignore all words that are typeset in a fixed width font? Just an idea, I dunno. Seems pretty dumb that it is spell checking code.

clash-prelude/src/Clash/XException.hs
martijnbastiaan: Yeah if we don't we'll have to change existing documentation. But that's fine.

.vscode/settings.json
martijnbastiaan: `dflipflop` is mentioned in the documentation.

clash-prelude/src/Clash/XException.hs
martijnbastiaan: "Unknown values"? "Undefined values"?

.vscode/settings.json
DigitalBrains1: So are the names of other functions. I'm saying this should be spell checked:
> This is a variant of dflipflop

but this should be left alone:
> This is a variant of `dflipflop`

because the fixed width font indicates this, in all likeliness, is code.

But I don't use VSCode and I don't use spell checking, so I'm just saying it sounds dumb that it's trying to spell check something that by its formatting is very likely to not be prose.

alex-mckenna: I appreciate the fact the most replies are on my dumb joke comment 😅

martijnbastiaan: :cry:

martijnbastiaan: In any case I'll delete it from the additions. People will have to suffer through the red scribbles.

clash-prelude/src/Clash/XException.hs
martijnbastiaan: Ooh yes, you're right! It does translate to `-`.

alex-mckenna: In other places it does become X though I think. This is maybe okay, we should probably note that a Clash X is not necessarily the same as X in VHDL or Verilog (since I think outside of BiSignal, we would just treat a Verilog Z as also being X)

This might be fiddly to get right, so I guess keep at it until @DigitalBrains1 doesn't object (since I'm not so rigourous when it comes to doing things properly)

.vscode/settings.json
DigitalBrains1: Maybe we should correct the spelling of frontend (should be: [front-end](https://www.oxfordlearnersdictionaries.com/definition/english/front-end_2) (adjective), [front end](https://www.oxfordlearnersdictionaries.com/definition/english/front-end_1) (noun)) and remove that misspelled word from the word list as well. In a separate PR. Unless someone can show a dictionary that spells it as such.

And both `NOINLINE` and `fmap` in the word list are of the same class as `dflipfop`: they denote code rather than prose, and if I would use them in documentation I'm writing, I would format them as code with a fixed-width font.

> We have a function called dflipflop, so it comes up often 😓

```
$ git grep dflipflop
clash-cores/src/Clash/Cores/LatticeSemi/ICE40/IO.hs:dflipflopE
clash-cores/src/Clash/Cores/LatticeSemi/ICE40/IO.hs:dflipflopE = delay (deepErrorX "dflipflopE: undefined initial value")
clash-cores/src/Clash/Cores/LatticeSemi/ICE40/IO.hs:    mux latch_dIn_0 (dflipflopE pkgPinInRead) pkgPinInRead
clash-prelude/src/Clash/Explicit/Reset.hs:-- For synchronous domains two sequential dflipflops are used:
clash-prelude/src/Clash/Explicit/Signal.hs:  , dflipflop
clash-prelude/src/Clash/Explicit/Signal.hs:dflipflop
clash-prelude/src/Clash/Explicit/Signal.hs:dflipflop = \clk i ->
clash-prelude/src/Clash/Explicit/Signal.hs:    (deepErrorX "First value of dflipflop undefined")
clash-prelude/src/Clash/Explicit/Signal.hs:{-# INLINE dflipflop #-}
clash-prelude/src/Clash/Signal.hs:  , dflipflop
clash-prelude/src/Clash/Signal.hs:dflipflop
clash-prelude/src/Clash/Signal.hs:dflipflop =
clash-prelude/src/Clash/Signal.hs:  E.dflipflop (fromLabel @(HiddenClockName dom))
clash-prelude/src/Clash/Signal.hs:{-# INLINE dflipflop #-}
```

I see one occurrence (`Reset.hs`) and that can trivially be replaced by _D flip-flop_. This PR adds one further occurrence as yet.

I'm personally not even requesting you remove them from this word list even though I do think they probably shouldn't be on there. It feels like a slippery slope: exactly how often does a word need to occur before we add it? At the most extreme, do we add all functions in `clash-prelude` as well as `base`? (Of course we don't)

Perhaps you can just mentally reconcile the red scribble with an indication that you need to format something as a fixed-width font? :-D

martijnbastiaan:
I don't think I'll have the time to pick this up any time soon.

Issue #2384, opened 2022-12-25
Clash Sampling (v1.6.3) and Synthesized Hardware disagree on redundant bit replacement


IamfromSpace:
Hello again!  Still greatly enjoying my adventures with Clash :)  I've recently stumbled across an issue that may simply be user error.

I have a simple function that takes in an ASCII BitVector and returns the index of lower and uppercase letters (a = 0, b = 1, A = 26, ..., Z = 51), dropping two unneeded bits.

When defined like this, it works just fine in the simulator, but in actual hardware (Alchitry Cu with ice40 HX8K via the IceStorm toolchain) only 0 has defined behavior.

```haskell
parse :: BitVector 8 -> Unsigned 6
parse x = fromIntegral (if x >= 97 then x - 97 else x - 65 + 26)
```

If I instead define it like this, it works correctly in both contexts:

```haskell
parse :: BitVector 8 -> Unsigned 6
  let y = fromIntegral x :: Unsigned 8
  in truncateB (if y >= 97 then y - 97 else y - 65 + 26)
```

I'm not totally sure what the expected result should be.  I do see that this note on the BitVector module should have caught my attention:

> as it is not safe to coerce between different size BitVector. To change the size, use the functions in the [Resize](https://hackage.haskell.org/package/clash-prelude-1.6.3/docs/Clash-Class-Resize.html#t:Resize) class.

Should `fromIntegral` be a safe way to resize?  If not, should (/could) the simulator also return undefined?

rowanG077:
Could you post your generated HDL here with a minimal test case? If I synthesize `parse` using this `topEntity` with clash 1.6.3:

```Haskell
topEntity :: Signal System (BitVector 8) -> Signal System (Unsigned 6)
topEntity a = fmap parse a
```

I get this verilog:

```
/* AUTOMATICALLY GENERATED VERILOG-2001 SOURCE CODE.
** GENERATED BY CLASH 1.6.3. DO NOT MODIFY.
*/
`timescale 100fs/100fs
module topEntity
    ( // Inputs
      input [7:0] a


      // Outputs
    , output wire [5:0] result
    );
  wire signed [63:0] c$app_arg;
  wire signed [63:0] c$case_alt;
  wire signed [63:0] c$case_alt_0;
  wire [7:0] c$bv;
  wire [7:0] c$bv_0;

  assign c$app_arg = (a >= 8'b01100001) ? c$case_alt : c$case_alt_0;

[...]

PR #2385, opened 2022-12-30
Note about fromInteger/toInteger in Concrete(ish) Fixed Types

IamfromSpace:
This came out of a discussion in #2384 about `fromIntegral` and how it was generally better to steer away from it (though conversion turned out not to bet the issue there).  This seemed like the places to call out the potential risks associated while very briefly touching on the reasoning.  This was my best attempt at concisely capturing the consideration and steering users towards the right places.  More than happy to make edits or close if folks feel this is unnecessary.  Cheers!

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Sized/Internal/BitVector.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/BitVector.hs
@@ -212,10 +212,15 @@ type role BitVector nominal
 
 -- * Type definitions
 
--- | A vector of bits.
+-- | A vector of bits
 --
 -- * Bit indices are descending
 -- * 'Num' instance performs /unsigned/ arithmetic.
+--
+-- __NB__: The usual Haskell method of converting an integral numeric type to
+-- another, 'fromIntegral', is not well suited for Clash as it will go through
+-- 'Integer' which is arbitrarily bounded in HDL. Instead use
+-- 'Clash.Class.BitPack.bitCoerce' and the 'Resize' class.
 --
 -- BitVector has the <https://downloads.haskell.org/ghc/latest/docs/html/users_guide/exts/roles.html type role>
 --
@@ -223,8 +228,8 @@ type role BitVector nominal
 -- type role BitVector nominal
 -- ...
 --
--- as it is not safe to coerce between different size BitVector. To change the
--- size, use the functions in the 'Clash.Class.Resize.Resize' class.
+-- as it is not safe to coerce between different sizes of BitVector. To change
+-- the size, use the functions in the 'Resize' class.
 data BitVector (n :: Nat) =
     -- | The constructor, 'BV', and  the field, 'unsafeToNatural', are not
     -- synthesizable.
@@ -237,7 +242,12 @@ data BitVector (n :: Nat) =
 
 -- * Bit
 
--- | Bit
+-- | A single bit
+--
+-- __NB__: The usual Haskell method of converting an integral numeric type to
+-- another, 'fromIntegral', is not well suited for Clash as it will go through
+-- 'Integer' which is arbitrarily bounded in HDL. Instead use
+-- 'Clash.Class.BitPack.bitCoerce' and the 'Resize' class.
 data Bit =
   -- | The constructor, 'Bit', and  the field, 'unsafeToInteger#', are not
   -- synthesizable.
@@ -665,6 +675,9 @@ maxBound# = let m = 1 `shiftL` natToNum @n in BV 0 (m-1)
 {-# NOINLINE maxBound# #-}
 {-# ANN maxBound# hasBlackBox #-}
 
+-- | __NB__: 'fromInteger'/'fromIntegral' can cause unexpected truncation, as
+-- 'Integer' is arbitrarily bounded during synthesis.  Prefer
+-- 'Clash.Class.BitPack.bitCoerce' and the 'Resize' class.
 instance KnownNat n => Num (BitVector n) where
   (+)         = (+#)
   (-)         = (-#)
@@ -780,6 +793,9 @@ times# bv1 bv2 = undefErrorP "mul" bv1 bv2
 instance KnownNat n => Real (BitVector n) where
   toRational = toRational . toInteger#
 
+-- | __NB__: 'toInteger'/'fromIntegral' can cause unexpected truncation, as
+-- 'Integer' is arbitrarily bounded during synthesis.  Prefer
+-- 'Clash.Class.BitPack.bitCoerce' and the 'Resize' class.
 instance KnownNat n => Integral (BitVector n) where
   quot        = quot#
   rem         = rem#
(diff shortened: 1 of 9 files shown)


clash-prelude/src/Clash/Sized/Internal/BitVector.hs
```diff
@@ -500,6 +500,9 @@ maxBound# :: forall n . KnownNat n => BitVector n
 maxBound# = let m = 1 `shiftL` fromInteger (natVal (Proxy @n))
             in  BV 0 (m-1)
 
+-- | __NB__: `fromInteger` can cause unexpected truncation, as `Integer` is
+-- arbitrarily bounded during synthesis.  Prefer
+-- `Clash.Class.BitPack.bitCoerce` and the `Resize` class.
```
martijnbastiaan, line 505:
```suggestion
-- | __NB__: 'fromInteger' can cause unexpected truncation, as 'Integer' is
-- arbitrarily bounded during synthesis.  Prefer
-- 'Clash.Class.BitPack.bitCoerce' and the 'Resize' class.
```

martijnbastiaan: If you use this Haddock can link it in the documentation.

martijnbastiaan (changes requested):
Seems sensible to add this comment!

DigitalBrains1:
In #2384, IamfromSpace [wrote](https://github.com/clash-lang/clash-compiler/issues/2384#issuecomment-1364847840):
> I might take a stab at a docs PR to call this out in sized types that support Integral, unless there's some reason not to!

Thank you! We definitely need improved documentation, and contributions are very welcome!

Your notes are perfect for people well familiar with Haskell, knowing where `fromInteger` and `toInteger` are defined and knowing that `fromIntegral` is the combination of the two. I think also mentioning `fromIntegral` both where 
 you mention `fromInteger` as well as where you mention `toInteger` will also give a nudge to people less experienced in Haskell.

Also, we should mention this near the top of the Haddock of all our custom `Num` instance types, so even people who don't know how to resize or convert at all can learn. I'll deepen your stab with my own later. Bloody `Num` instances! ;-P

martijnbastiaan:
@DigitalBrains1 I propose merging this PR (with Haddock fixes); it's a perfectly good improvement over the status-quo after all. Let's not make perfect the enemy of good :).

DigitalBrains1:
I've extended the doc a bit and also fixed our own examples, it turns out we used `fromIntegral` in examples! 🥴

@IamfromSpace, I have my suggested version at [clash-compiler/docs/note-integer-dangers-of-num-and-integral-instances](https://github.com/clash-lang/clash-compiler/tree/docs/note-integer-dangers-of-num-and-integral-instances)(, do you agree with these changes?

Something went wrong when picking a base commit for this PR, as its parent is d326554ff . I rebased to current master without issues.

@Martijn, do you also agree with these changes? :-)

martijnbastiaan:
> @martijnbastiaan, do you also agree with these changes? :-)

Jup looks good!

clash-prelude/src/Clash/Sized/Internal/BitVector.hs
IamfromSpace: Whoops!  Been writing too much markdown :laughing:

IamfromSpace:
Glad this change is helpful!  The docs have been quite helpful, but if I see opportunities I'll try and present more changes.

@DigitalBrains1 Your changes also make sense to pair with these to me, I've brought them in from your branch.

> Something went wrong when picking a base commit for this PR, as its parent is https://github.com/clash-lang/clash-compiler/commit/d326554ff8941e189924edba605dcb00b430e031

You'll have to forgive my neurosis when it comes to all things git, haha, the original ancestor was intentional, as it allows the change to also be brought into any 1.0+ version with a simple merge, should that be desired :).  Being easily able to add this to 1.6.x seemed worthwhile, whether or not it's leveraged.  I've just incorporated everything so we get both your changes and my original ancestor.  Final diff ends the same!

I didn't feel like this needed a changelog or copyright update (the latter isn't strictly necessary), but please let me know if I should incorporate them and I'll take care of both.

martijnbastiaan:
@IamfromSpace Our flow always forks from `master`. We then apply a label `backport-*` which gets picked up by Mergify which subsequently backports it. I've added the label for this PR too.

Could you apply some `git rebase origin/master -i` foo and remove the merge commit from here?

IamfromSpace:
Ah, gotcha.  I'm personally not a fan of these kinds of git workflows.  I see the appeal, but I think they ultimately make things more complicated.  That said, that's as much of a fuss as I'll make about it :laughing: I'm more than happy to conform to local processes.  I've simply reset my branch to @DigitalBrains1 proposed one, since that has everything taken care of.  For future PRs I'll have them fast-forward ready.

I think this makes it good to go, let me know if there's anything else needed :)

DigitalBrains1:
I do wonder what happens with CI if you base it on a commit from three years ago... wouldn't it run the CI setup from 3 years ago?

I force-pushed to 2d70a9a11e54319d9e53a6bfa5e84dd791f6ab78 as that's based on the latest tip (we want CI to actually test what is going to be committed, not something that looks like it merges without conflicts 😆).

clash-prelude/src/Clash/Sized/Internal/BitVector.hs
DigitalBrains1: I know the feeling! It is so difficult switching between the two.

DigitalBrains1:
@kloonbot run_ci f2d9dcd588f45687c4cc5afb07494785698419b7

IamfromSpace:
:+1: :+1:

> I do wonder what happens with CI if you base it on a commit from three years ago... wouldn't it run the CI setup from 3 years ago?
>
> I force-pushed to https://github.com/clash-lang/clash-compiler/commit/2d70a9a11e54319d9e53a6bfa5e84dd791f6ab78 as that's based on the latest tip (we want CI to actually test what is going to be committed, not something that looks like it merges without conflicts laughing).

There are certainly challenges/trade-offs of an "(reasonably) earliest ancestor approach."  Automation has to merge to the target to test the final deal.  And like you said, you might get a CI flow that's too old (however, CI flows are great cases for earliest-ancestor changes) or the PR needs to include a merge commit manually, so that the head includes the most up to date CI.  Again, no real debate from me, just since you asked :)

Thank you for taking care of these updates!

martijnbastiaan:
Hmm, I've seen the current CI failure before. It seems to happen sporadically. I'll look into it.

martijnbastiaan:
@mergify rebase

DigitalBrains1:
@mergify rebase

DigitalBrains1:
@kloonbot run_ci  d149589

martijnbastiaan (approved):

PR #2391, opened 2023-01-03
Add fallback value to concurrency.group

martijnbastiaan:


DigitalBrains1:
This is now giving failures on 1.6 as well, so I'm marking it for backport.

PR #2392, opened 2023-01-03
Github CI: upgrade to actions/cache@v3 and fix caching on Windows

leonschoorl:
With this fix in place we can upgrade to actions/cache@v3 and get rid of the deprecation warnings caused by the cache action, while keeping the cache working on Windows.

And I'll also throw in a fix for the flaky build-clash-dev job over on gitlab CI

### DIFF
--- a/.ci/build_clash_dev.sh
+++ b/.ci/build_clash_dev.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 set -xeou pipefail
 
+# Make sure all our deps are build, and listed in the .ghc.environment file
+cabal --write-ghc-environment-files=always v2-build --only-dependencies clash-lib
+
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
So the `clash-build-dev` thing had two possible paths:
* It works as intended, nothing is compiled. It is picked up from the artifact generated by the `build` job.
* For some reason it decides to recompile everything or at least a lot. If I'm not mistaken, in this case again two things could happen. It would compile succesfully and finish succesfully, or it would fail with the CPP message.

Perhaps it depends on the amount of things that it would not pick up in the problematic case whether it would have `prettyprinter` or not.

DigitalBrains1 (approved):
LGTM!

PR #2399, opened 2023-01-06
New: `clashCompileError`

DigitalBrains1:
Based upon the version proposed by @basile-henry in issue #2020:
https://github.com/clash-lang/clash-compiler/issues/2020#issuecomment-1032467299

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Primitives/Prelude.hs
+++ b/clash-lib/src/Clash/Primitives/Prelude.hs
@@ -1,0 +1,30 @@
+{-|
+  Copyright   :  (C) 2022     , Myrtle.ai,
+                     2023     , QBayLogic B.V.,
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+  Blackbox functions for primitives in one of the @Prelude@ modules.
+-}
+
+{-# LANGUAGE TemplateHaskellQuotes #-}
+
+module Clash.Primitives.Prelude
+  ( clashCompileErrorBBF
+  ) where
+
+import Data.Either (lefts)
+import GHC.Stack (HasCallStack)
+
+import Clash.Core.TermLiteral (termToDataError)
+import Clash.Netlist.BlackBox.Types (BlackBoxFunction)
+import Clash.Netlist.Types ()
+
+clashCompileErrorBBF :: HasCallStack => BlackBoxFunction
+clashCompileErrorBBF _isD _primName args _ty
+  | [ _hasCallstack
+    , either error id . termToDataError -> msg
+    ] <- lefts args
+  = pure $ Left $ "clashCompileError: " <> msg
+  | otherwise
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Primitives/Prelude.hs
```diff
@@ -0,0 +1,44 @@
...
+import Data.Text.Prettyprint.Doc.Extra (Doc)
+
+clashCompileErrorBBF :: BlackBoxFunction
+clashCompileErrorBBF _isD _primName _args _ty =
+  pure (Right (emptyBlackBoxMeta, bb))
+ where
+  [ _hasCallStack
+    , msg
+    , _ ] = [0..]
+  bb = BBFunction "Clash.Prelude.ExtraBB.clashCompileErrorBBF" 0
```
martijnbastiaan, line 31:
```suggestion
  bb = BBFunction (show 'clashCompileErrorBBF) 0
```

martijnbastiaan (approved):
LGTM (assuming copyright / CI fixes). As far as the `clash-dev` failure goes: it seems to happen sporadically and I can't reproduce it locally. Given that it is not anything user-facing I'd be fine with disabling the check for 9.2.

clash-lib/src/Clash/Primitives/Prelude.hs
DigitalBrains1: Huh. Weird that my eyes really never caught on the string! Thanks for spotting it!

DigitalBrains1:
GHC 9.2 complaining about incomplete pattern matches (which actually pointed out a real bug in my code) made me realise it should probably do better on error reporting in the case a user passes some string which cannot be reduced to a literal by Clash. So I did some improvements there. I also trimmed the Changelog entry a bit; I can be a bit of a blabbermouth.

DigitalBrains1:
Martijn, could you review these changes:
https://github.com/clash-lang/clash-compiler/compare/5fc0f35..9605ab2a43ef392cac0060ebf44ac0e82b606f23

clash-lib/src/Clash/Primitives/Prelude.hs
```diff
@@ -0,0 +1,56 @@
...
+clashCompileErrorTemplate ::
+  HasCallStack =>
+  Backend s =>
+  BlackBoxContext ->
+  State s Doc
+clashCompileErrorTemplate bbCtx
+  | [ _hasCallStack
+    , (msgE, _, _)
+    ] <- bbInputs bbCtx
+  = case exprToString msgE of
```
martijnbastiaan, line 48: You could reuse `TermLiteral` in `clashCompileErrorBBF`; this would save you from having to write the error message yourself.

DigitalBrains1:
Ah, actually using `TermLiteral` changes things a lot, I think it's an improvement. Do you agree?

DigitalBrains1:
Ah, it can be even shorter; that does produce a useless call stack as part of the error but the code is so nice and simple this way.

martijnbastiaan:
Definitely an improvement!

PR #2403, opened 2023-01-15
CI: fix shared runners, use matrix

DigitalBrains1:
PR #2402 incorrectly changed `.ci/gitlab/test.yml`, not moving back jobs to shared runners as intended.

Gitlab CI now supports `parallel:matrix` trigger jobs, so use that.

## Still TODO:

### DIFF
--- a/.ci/gitlab/common.yml
+++ b/.ci/gitlab/common.yml
@@ -1,6 +1,6 @@
 .common:
   image: ghcr.io/clash-lang/clash-ci-$GHC_VERSION:2022-11-14
-  timeout: 2 hours
+  timeout: 10 minutes
   stage: build
   variables:
     # Note that we copy+paste the image name into CACHE_FALLBACK_KEY. If we don't,
@@ -26,5 +26,15 @@
     - .ci/setup.sh
   after_script:
     - tar -cf - $(ls -d /root/.cabal /root/.stack $(pwd)/.ci/bindist/linux/debian/*/build || true) | zstd -T${THREADS} -3 > cache.tar.zst
+
+# We run tests on local machines if:
+#
+#   * A job may take more than 3 minutes to complete on public runners.
+#   * A job needs specific capabilities public runners cannot provide, e.g.
+#     more than 4GB memory of memory, proprietary synthesis tooling.
+#
+.common-local:
+  extends: .common
+  timeout: 2 hours
   tags:
     - local
--- a/.ci/gitlab/test.yml
+++ b/.ci/gitlab/test.yml
@@ -5,7 +5,29 @@ stages:
   - build
   - test
 
-.test-common:
+.test-nocache:
+  extends: .common
+  needs: ["build"]
+  stage: test
+  cache:
+    paths: []
+  before_script:
+    - unset SNAPCRAFT_LOGIN_FILE
+    - unset HACKAGE_PASSWORD
+    - export THREADS=$(./.ci/effective_cpus.sh)
+    - export CABAL_JOBS=$(./.ci/effective_cpus.sh)
+    - export clash_lib_datadir=$(pwd)/clash-lib/
+    - export clash_cosim_datadir=$(pwd)/clash-cosim/
+    - export
+    - tar -xf dist.tar.zst -C /
+
+    # Not all package in cache get packed into dist.tar.zst, so we need to
+    # regenerate the package database
+    - ghc-pkg recache --package-db=$HOME/.cabal/store/ghc-$GHC_VERSION/package.db
+  after_script:
+
+.test-cache-local:
+  extends: .common-local
   needs: ["build"]
   stage: test
   before_script:
@@ -13,6 +35,8 @@ stages:
     - unset HACKAGE_PASSWORD
     - export THREADS=$(./.ci/effective_cpus.sh)
     - export CABAL_JOBS=$(./.ci/effective_cpus.sh)
+    - export clash_lib_datadir=$(pwd)/clash-lib/
+    - export clash_cosim_datadir=$(pwd)/clash-cosim/
     - export
     - tar -xf cache.tar.zst -C / || true
     - tar -xf dist.tar.zst -C /
@@ -23,16 +47,6 @@ stages:
     - ghc-pkg recache --package-db=$HOME/.cabal/store/ghc-$GHC_VERSION/package.db
   after_script:
     - tar -cf - /root/.cabal/packages | zstd -T${THREADS} -3 > cache.tar.zst
-
-.test-common-shared:
-  extends:
-    - .common
-    - .test-common
-
-.test-common-local:
-  extends:
-    - .common-local
-    - .test-common
 
 # 'build' publishes its build files as an artifact. These build files are reused
 # by the tests below.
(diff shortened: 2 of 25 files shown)


DigitalBrains1:
PR #2402 added this comment to the `clash-build-dev` job:
> Normally, this job is small. But it is flaky on GHC 9.2; it sometimes fails and we don't know yet why. When it fails, it recompiles things it should have picked up from the 'build' issue and then it is a larger job, so we keep it on local runners for now.

It turns out it's not just that job; that job fails when it happens, but the other jobs also sometimes fail to pick up the build. In the CI for this commit, _all_ the jobs for the `tests: [9.2.5]` pipeline that ran on shared runners did a full recompile, burning shared runner CI minutes like mad. Conversely, all the jobs in that pipeline running on local runners, _including `clash-build-dev`_, picked up the `build` artifact and didn't compile stuff anew.

We should consider what to do about this for now.

.ci/gitlab/test.yml
```diff
@@ -29,6 +29,8 @@ stages:
 
 .test-common:
   extends: .test-common-local
+  # Run on shared runners
+  tags:
```
martijnbastiaan, line 33: I've probably written something like this myself before, but I had to double take on this before getting how this worked. Could you add to the comment that this overrides / empties the tags set in `.test-common-local`, hence the local runners not picking it up anymore (because they need `local`)?

martijnbastiaan:
> Gitlab CI now supports parallel:matrix trigger jobs, so use that.

Awww yiss!

I think we can also remove this hack now that [this](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/102564) has been merged:

https://github.com/clash-lang/clash-compiler/blob/05ff87f75e6e159f196e36c4bf6f1fdccc017ccb/.gitlab-ci.yml#L141-L168

Only if you feel like doing that, otherwise we can merge this PR IMO.

martijnbastiaan:
It looks like our caching/artifact strategy is broken:

https://gitlab.com/clash-lang/clash-compiler/-/jobs/3605382618

Each job get its own cache id now:

> lib:unittests-ci-shared-matrix-ghcr.io

Having said that, cache shouldn't matter - the artifacts should contain everything that matters. Perhaps we should directly invoke the binaries instead of using `cabal run` to bypass cache invalidation?

.ci/gitlab/test.yml
DigitalBrains1: It is literally what was there before we moved off shared runners, and the most confusing thing to me is:
- We define `.test-common-local` which adds the tag `local`
- We derive `.test-common` from `.test-common-local` removing the tag

It makes much more sense to me to do the reverse, but I was minimizing the diff :-)

DigitalBrains1:
It is not really clear to me what the replacement for the `set_github_status.sh` thing should be. Feel free to add it to this PR, though!

On the caching; note that for `prelude:unittests`[https://gitlab.com/clash-lang/clash-compiler/-/jobs/3605382615#L23] it picked up the cache from the previous CI run but still recompiled.

.ci/gitlab/test.yml
martijnbastiaan: yeah let's go with code clarity over minimal diffs :-)

DigitalBrains1:
Ugh I tried to have `build` provide the cache for the test jobs, but they don't always pick it up. Don't you just love CI.

DigitalBrains1:
I also figured out why we had:
```
.common:

.test-common-local:
  extends: .common

.test-common
  extends: .test-common-local
```

This is because `.common` sets the `local` tag. `.test-common-local` keeps it, and `.test-common` clears it again. I suppose that makes sense; with a few comments added at least.

DigitalBrains1:
> Having said that, cache shouldn't matter - the artifacts should contain everything that matters. Perhaps we should directly invoke the binaries instead of using cabal run to bypass cache invalidation?

Given all the issues with these caches, perhaps that is the better option. I'd rather really solve the issue but maybe this compromise is good enough and more robust.

DigitalBrains1:
A bunch of things where changed to get everything running smoothly.

* Refactor the shared YAML config blocks
* Shared runners now have a low timeout (10 minutes)
* There was confusion about what should set the `local` tag to run it on a local runner. This was set in `.common`, and then set again in many jobs. The exceptions seem to be `hackage-release-candidate` and `hackage-release`; these got the `local` tag through inheritance of `.common`, but I feel these jobs can be on shared runners. @martijnbastiaan , do you agree?
* There is some issue where running unit tests and doctests with `cabal v2-run` sometimes triggers a recompile of everything instead of picking up everything from the `dist.tar.zst`. We don't want this on shared runners. `cabal list-bin` is not supported on the Cabals we use for GHC 8.x, so instead we now hardcode the path to those binaries and run the binaries directly. This will at least never cause a recompilation.
* As we invoke the binaries directly, we can dispense with the cache and `.ci/setup.sh` for doctests and unit tests.
* A warning on GHC 9.2's clash-prelude doctest was fixed.
* Some debugging output was removed.

.ci/gitlab/test.yml
```diff
@@ -45,40 +63,38 @@ build:
...
-   - local
 
 # Tests run on shared runners:
 
 cores:unittests:
-  extends: .test-common
+  extends: .test-nocache
   script:
-    - cabal v2-run -- clash-cores:unittests --hide-successes
+    - ./dist-newstyle/build/*/*/clash-cores-*/t/unittests/build/unittests/unittests --hide-successes
```
martijnbastiaan, line 72:
```suggestion
    - ./.ci/run_test.sh clash-cores unittests --hide-successes
```

Moving all the `*` stuff to a separate script seems a little bit cleaner to me.

.gitlab-ci.yml
```diff
@@ -25,44 +25,20 @@ stages:
...
-  variables:
-    GHC_VERSION: 9.0.2
-
-tests-9.2:
-  extends: .common-trigger
-  variables:
-    GHC_VERSION: 9.2.5
+  parallel:
+    matrix:
+      - GHC_VERSION: [8.8.4, 8.10.7, 9.0.2, 9.2.5]
```
martijnbastiaan, line 36: This is so much cleaner! Very nice.

.ci/gitlab/common.yml
```diff
@@ -26,5 +26,9 @@
     - .ci/setup.sh
   after_script:
     - tar -cf - $(ls -d /root/.cabal /root/.stack $(pwd)/.ci/bindist/linux/debian/*/build || true) | zstd -T${THREADS} -3 > cache.tar.zst
+
+.common-local:
```
martijnbastiaan, line 30: We should probably have some rule of thumb on when to run jobs on public runners / local runners. What about:

```suggestion
# We run tests on local machines if:
# 
#   * A job may take more than 10 minutes to complete on public runners
#   * A job needs specific capabilities public runners cannot provide, e.g.
#     more than 4GB memory of memory, proprietary synthesis tooling.
#
.common-local:
```

martijnbastiaan (approved):
In general looks good to me! Thanks for the rabbit hole diving :-).

.ci/gitlab/test.yml
DigitalBrains1: That will need a bit more than that as there are three patterns:
1. `./dist-newstyle/build/*/*/clash-prelude-*/t/unittests/build/unittests/unittests`
2.`./dist-newstyle/build/*/*/clash-cosim-*/build/test/test`
3. `/dist-newstyle/build/*/*/clash-testsuite-*/x/clash-testsuite/build/clash-testsuite/clash-testsuite`

What's the difference between 1 and 2, you reckon? Whether there are multiple `test-suite` stanzas in the `.cabal` file or not? As in, `clash-cosim` has only one, maybe that's why it has a different path to the executable. We could also keep that one separate and make a script for 1 and 3 only.

.ci/gitlab/common.yml
DigitalBrains1: 10 minutes is the timeout. I'd say that if a job is expected to take more than, say, 3 minutes, maybe it's better to do a local one. `prelude:doctests` is already at that limit, btw.

Would we really gain anything by running jobs taking more than that on public runners?

.ci/gitlab/test.yml
martijnbastiaan: `clash-cosim` has a custom setup, perhaps that's it?

> The fact that I couldn't readily tell what the structure was, was the reason to forego the separate script solution initially. 

Sure, that makes sense.

> Do you want to make this nicer

Nah it's fine. I'd probably have done something like `find dist-newstyle/ -executable -type f -name $bin_name | grep $lib_name`, but then again, that doesn't feel particularly nice either.

.ci/gitlab/common.yml
martijnbastiaan: Sure, 3, 5, 7 - all works for me :-). As long as there's some guidance.

> Would we really gain anything by running jobs taking more than that on public runners?

My original motivation was moving anything not in the critical time path of a pipeline to public runners. This was before any limits on GitLab (which I'm happy to see implemented, anything to block those damn crypto miners..).

PR #2404, opened 2023-01-16
Export Wrapping constructor

### LINKED ISSUE #2292: Question about the performance of the Fixed type
I'm implementing an FM radio receiver in clash [here](https://github.com/adamwalker/fpga-fm-radio). The [fixed point](https://hackage.haskell.org/package/clash-prelude-1.6.3/docs/Clash-Sized-Fixed.html) data types have been very useful for keeping track of the decimal point and make DSP super nice in Clash.

However, the performance of the generated code is not so nice. As stated in the docs, the Num operations saturate on overflow. This adds extra logic and prevents Vivado from using the very efficient DSP blocks. My filters implemented using the Signed type take 2 levels of logic, but using the Fixed type they blow out to 10+.

Why do the fixed types saturate on overflow? IMO, they should just wrap and be a zero-cost type-level way of tracking where the decimal point is. When doing DSP stuff you usually just size things so overflow wont happen, or you use accumulators that are big enough and only perform the saturation at the end. 

Would it be possible to change the fixed types to not wrap so that the generated logic is more efficient? Alternatively, I can implement my own type, but I think clash users would benefit from a more efficient Fixed type.

adamwalker:
Fixes  #2292 

Is a changelog entry needed?

### DIFF
--- a/changelog/2023-02-14T13_04_58+11_00_export_Wrapping_constructor
+++ b/changelog/2023-02-14T13_04_58+11_00_export_Wrapping_constructor
@@ -1,0 +1,1 @@
+CHANGED: Export the constructor for the `Wrapping` type in the `Clash.Num.Wrapping` module. See [#2292](https://github.com/clash-lang/clash-compiler/issues/2292)
--- a/clash-prelude/src/Clash/Num/Wrapping.hs
+++ b/clash-prelude/src/Clash/Num/Wrapping.hs
@@ -8,7 +8,7 @@ Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Clash.Num.Wrapping
-  ( Wrapping(fromWrapping)
+  ( Wrapping(..)
   , toWrapping
   ) where
 


martijnbastiaan (approved):
LGTM, but please add a changelog entry :). See `changelog/README.md` for instructions. It's probably best to refer to https://github.com/clash-lang/clash-compiler/issues/2292 too.

martijnbastiaan:
@adamwalker Do you still want this?

adamwalker:
Yes. I've added a changelog.

alex-mckenna:
@kloonbot run_ci 38dc731

martijnbastiaan:
@kloonbot run_ci 38dc73158cb5a13e5b9464a56a4794af365523fe

(Is it dead?)

DigitalBrains1:
Kloonbot errors with

> [...] (refusing to allow a GitHub App to create or update workflow `.github/workflows/ci.yml` without `workflows` permission)

so I think the problem is that `master` has updated that file and this needs to be rebased so it doesn't "change" the file.

@mergify rebase

DigitalBrains1:
@mergify rebase

DigitalBrains1:
@kloonbot run_ci cf1a379

PR #2405, opened 2023-01-20
GitLab CI: Use manual cache bust token

DigitalBrains1:
Since we need a PR to update the cache bust token in CACHE_FALLBACK_KEY anyway, we might as well use a manual token instead.

## Still TODO:


DigitalBrains1:
We suspect the nightly CI failures on 1.6 might be due to a corrupted cache. This PR, when backported, will bust the cache, and hopefully it will pass again.

DigitalBrains1:
@mergify rebase

leonschoorl (approved):

PR #2411, opened 2023-01-23
Test suite: multiple tests in one file

DigitalBrains1:
PR #2345 made it possible to use multiple build targets in all tools.
This PR leverages that functionality, combining build targets that were
previously split over multiple top entity files.

Not counting Vivado, 90% of `clash-testsuite`'s execution time is spent
on Clash startup costs. These costs are shared between the multiple
build targets, making the test suite a bit quicker.

Additionally, an unintended test bench structure in some tests was fixed.

The `RotateC` tests were written in a way that causes a functionally
identical copy of the `topEntity` to be included in the `testBench`,
instead of the usual case where the `testBench` uses a `topEntity` in
its own entity directory.

I can only think of one functional difference: such an included version
of the code will likely elide many enable lines as they are driven by
`enableGen`. With a separate entity directory, this transformation is
omitted because the `topEntity` has actual enable lines. Clearly if the
intent is to test the `~ISACTIVEENABLE` mechanism and similar
mechanisms, this is problematic, but this was not the case in these test
benches. However, it is clearly generally intended that the top entity
lives in its own entity dir, and the code surreptitiously does something
else than a cursory reading or general expectations suggest. In the
future, as some feature is added to Clash, it might even have more
consequences: the deviating structure might not be accounted for due to
it being unintended.

## Still TODO:


martijnbastiaan:
An unintended (happy) consequence of this is that this should make the testsuite a bit quicker too. Ignoring Vivado, 90% of `clash-testsuite`'s execution time is spent on Clash startup costs.

DigitalBrains1:
That wasn't really unintended! :-D

I've incorporated your comment into the commit messages and the PR cover letter (as well as adjusting the message for the second commit a bit).

martijnbastiaan (approved):

PR #2418, opened 2023-02-07
GitLab CI: Increase snap publish timeout

DigitalBrains1:
PR #2403 decreased the timeout of jobs on shared runners to 10 minutes, and PR #2415 moved the snap publish job to local runners because it is a big job. PR #2415 should have also set the timeout back to 2 hours, but I missed it.

## Still TODO:


leonschoorl (approved):
LGTM

PR #2427, opened 2023-02-22
Fix Xilinx clock wizard and add Tcl IP generation

hiddemoll:
This PR fixes the Xilinx clock generation primitives and adds support for clock primitive instantiation on Xilinx without manually going through the Vivado Clocking Wizard. The updated black boxes generate Tcl code to instantiate an MMCM through the Clocking Wizard. The clock primitives are limited to 1 output clock, and either a single-ended or a differential input clock. The generated Tcl script conforms with the Clash<->Tcl API.
​
Fixed in this PR:
- Broken Haskell simulation
The Haskell simulation of the Xilinx clock primitives was incorrect due to a polarity mismatch in the lock output of Xilinx `clockWizard` and `clockWizardDifferential`: Haskell simulation was the inverse of HDL simulation and hardware. In Haskell simulation, the primitives had an `Enable` as their output, which was _asserted_ when the PLL lock was _deasserted_. This corresponds to an _asserted_ reset input to the primitive in the simulation model, meaning the `Enable` output was asserted iff the PLL reset input was asserted. HDL simulation and hardware meanwhile had an active high lock signal output (which is usually used as an active low reset signal to the circuit). Finally, the simulation model did not account for different frequencies of the input and output domains, instead treating individual input samples as individual output samples without resampling.
This has been fixed by changing the type to `Signal Bool` and asserting it when the PLL has locked.
- Wrong port names of blackbox primitives
The blackbox primitives for `clockWizard` and `clockWizardDifferential` had capitalized port names. Also, the `clockWizardDifferential` primitive had 2 identical port names for the differential input clock signals.
- Remove `Asynchronous` constraint from Xilinx `clockWizard` and `clockWizardDifferential`
Originally intended to signal that these functions react synchronously to the incoming reset and that the outgoing lock signal is an asynchronous signal. Since synchronous reset signals are a subset of asynchronous reset signals this constraint on the input is vacuous. The constraint on the lock output does not convey this information at all and is wrong. Note that it is still necessary to synchronize the lock output in your design.
​
## Known problems:
- The `locked` signal is asynchronous. Therefore, this function is unsafe.
- The Clocking Wizard uses input and output frequencies to determine the multiplier and divider for the MMCM, but Clash uses periods instead. The conversion from period (picoseconds, Natural) to frequency (MHz, Double) is not exact. This is already not exact when using `hzToPeriod`, e.g. `hzToPeriod 300e6` translates to a period of `3333` ps, which becomes `300.03000300030004` in Tcl.
​
## Still TODO:
​


clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
```diff
@@ -0,0 +1,126 @@
...
+clockWizardTclTemplate bbCtx = pure bbText
+ where
+  ((_,stripVoid -> kdIn,_):(_,stripVoid -> kdOut,_):_:(nm,_,_):_) = bbInputs bbCtx
+  Just compName = exprToString nm
+  KnownDomain _ clkInPeriod _ _ _ _ = kdIn
+  KnownDomain _ clkOutPeriod _ _ _ _ = kdOut
+  clkInFreq :: Double
+  clkInFreq  = (1.0 / (fromInteger clkInPeriod  * 1.0e-12)) / 1e6
+  clkOutFreq :: Double
+  clkOutFreq = (1.0 / (fromInteger clkOutPeriod * 1.0e-12)) / 1e6
```
martijnbastiaan, line 50: You can use [periodToHz](https://hackage.haskell.org/package/clash-prelude-1.6.4/docs/Clash-Explicit-Signal.html#v:periodToHz) here. You'll have to convert it to a `Double` if needed though.

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
```diff
@@ -0,0 +1,126 @@
...
+clockWizardDifferentialTclTemplate bbCtx = pure bbText
+ where
+  ((_,stripVoid -> kdIn,_):(_,stripVoid -> kdOut,_):_:(nm,_,_):_) = bbInputs bbCtx
+  Just compName = exprToString nm
+  KnownDomain _ clkInPeriod _ _ _ _ = kdIn
+  KnownDomain _ clkOutPeriod _ _ _ _ = kdOut
+  clkInFreq :: Double
+  clkInFreq  = (1.0 / (fromInteger clkInPeriod  * 1.0e-12)) / 1e6
+  clkOutFreq :: Double
+  clkOutFreq = (1.0 / (fromInteger clkOutPeriod * 1.0e-12)) / 1e6
```
martijnbastiaan, line 101: Same, use `periodToHz`.

clash-prelude/src/Clash/Xilinx/ClockGen.hs
```diff
@@ -12,9 +12,21 @@ PLL and other clock-related components for Xilinx FPGAs
...
+-- import qualified Clash.Primitives.Xilinx.ClockGen as P
+
+
+-- -- | List of known TemplateFunctions to prevent Hint from firing. This
+-- -- improves Clash startup times.
+-- knownTemplateFunctions :: HashMap String TemplateFunction
+-- knownTemplateFunctions =
+--   HashMap.fromList $ map (first show) $
+--     [ ('P.clockWizardDifferentialTclTF, P.clockWizardDifferentialTclTF)
+--     ]
```
martijnbastiaan, line 28: Remove please

clash-lib/prims/commonverilog/Clash_Xilinx_ClockGen.primitives.yaml
```diff
@@ -4,39 +4,51 @@
     type: |-
       clockWizard
         :: ( KnownDomain domIn confIn       -- ARG[0]
-           , KnownDomain domOut confOut )   -- ARG[1]
-        => SSymbol name                    -- ARG[2]
-        -> Clock  pllIn                    -- ARG[3]
-        -> Reset pllIn                     -- ARG[4]
-        -> (Clock pllOut, Enable pllOut)
+           , KnownDomain domOut confOut     -- ARG[1]
+           , KnownDomain domPllLock confPllLock ) -- ARG[2]
```
martijnbastiaan, line 8: `KnownDomain` takes one argument; could you fix that?

martijnbastiaan (changes requested):
Nice changes, this definitely makes the primitive more usable. 

Two remarks on the output `Signal domPllLock Bool`:

 * The domain `domPllLock` isn't bound by an argument, so _any_ domain can be instantiated here. This is incredibly unsafe.
 * We never produce asynchronous signals in Clash, we shouldn't break this pattern. 

So I propose:

* Renaming the current functions to `unsafeClockWizard[Differential]`.
* Rename `domPllLock` to `domOut` again. It should be documented that this an asynchronous signal and the user should think twice about using it. 
* Create a new function `clockWizard[Differential]` which calls `unsafeClockWizard[Differential]` and synchronizes the signal using a `dualFlipFlopSynchronizer`.

DigitalBrains1:
> Two remarks on the output `Signal domPllLock Bool`:
> 
> * The domain `domPllLock` isn't bound by an argument, so _any_ domain can be instantiated here. This is incredibly unsafe.
> * We never produce asynchronous signals in Clash, we shouldn't break this pattern.

I explicitly requested Hidde to write it like this. This is how we do it and I think have always done it in `Clash.Clocks`, which provides the constraints and implementation for the Intel PLL `alteraPll`. It is actually a better way to represent reality. The PLL lock signal _is_ asynchronous and any domain is a lie we have to maintain because Clash doesn't know about asynchronous signals. Same, by the way, for the reset input `Reset domIn` in these primitives; luckily all these primitives are fine with purely asynchronous reset inputs.

The reason I have requested it is precisely because I'm working on a class to make these PLL's more user friendly. Currently we always have to write

```haskell
(clk100, pllStable) = clockWizard (SSymbol @"clkWizard50to100") clk50 rstIn
rstSync = resetSynchronizer clk $ unsafeFromLowPolarity pllStable
```

which means we're forcing the user to remember this and teaching them to use functions that begin with the word `unsafe`. I'd like to remedy both by changing the invocation to

```haskell
(clk100, rstSync) = clocksSynchronizedReset (clockWizard (SSymbol @"clkWizard50to100")) clk50 rstIn
```

but this implementation absolutely requires that `pllStable` is a `Signal domIn Bool` to get unsurprising simulation behaviour in Clash (for HDL, it doesn't matter, since the domain is a lie anyway). So I asked Hidde to make the domain as unrestricted as it has always been for `alteraPll`, so I can restrict it to `domIn` in `clocksSynchronizedReset`.

By the way, for multiple clocks, `clocksSynchronizedReset` works as follows:

instead of

```haskell
(clk100, clk150, pllLocked :: Signal DomIn Bool) = alteraPll (SSymbol @"alterapllmulti") clk50 rst50
rst100 = resetSynchronizer clk100 $ unsafeFromLowPolarity $ unsafeSynchronizer clk50 clk100 pllStable
rst150 = resetSynchronizer clk150 $ unsafeFromLowPolarity $ unsafeSynchronizer clk50 clk150 pllStable
```

you write

```haskell
(clk100, rst100, clk150, rst150) = clocksSynchronizedReset (alteraPll (SSymbol @"alterapllmulti")) clk50 rst50
```

You might actually need to do something more clever with multiple domains (i.e., a reset manager), but in this simple case, you can use the simple implementation.

DigitalBrains1:
By the way, along with my `clocksSynchronizedReset` will come improved documentation in both `Clash.Xilinx.ClockGen` as well as `Clash.Intel.ClockGen`. The scope of _this_ PR was agreed to be as small as viable: fixing the Xilinx clock wizard so it could be used, and that is that. As far as I'm concerned, further improvements can come in future PR's.

(Nitpick: you wrote _synchronizes the signal using a `dualFlipFlopSynchronizer`_, but you probably meant `resetSynchronizer`, a DFF sync is not what we want to see here).

clash-lib/prims/commonverilog/Clash_Xilinx_ClockGen.primitives.yaml
DigitalBrains1: Please also look at the layout: the comments should be vertically aligned. This might "automatically" happen by fixing `KnownDomain` though ;-).

martijnbastiaan:
Agreed on keeping the scope of this PR limited.

> which means we're forcing the user to remember this and teaching them to use functions that begin with the word unsafe. I'd like to remedy both by changing the invocation to
>
> ```haskell
> (clk100, rstSync) = clocksSynchronizedReset (clockWizard (SSymbol @"clkWizard50to100")) clk50 rstIn
> ```

I get what you're saying, but this just hides the fact that `clockWizard` is really unsafe to use. Anything that _doesn't_ have `unsafe` prepended to its name should be assumed to be safe w.r.t. domain crossing / synchronicity. This is a precedent we set in `clash-prelude` (though subsequently break in the clock wizard primitives). So my very strong preference would still be to rename the current primitive yielding asynchronous signals to `unsafe*` and introducing a non-unsafe version that does the synchronization for you.

Got to run, will elaborate on later.

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
```diff
@@ -0,0 +1,126 @@
...
+    = True
+  valid _ = False
+
+clockWizardTclTemplate
+  :: Backend s
+  => BlackBoxContext
+  -> State s Doc
+clockWizardTclTemplate bbCtx = pure bbText
+ where
+  ((_,stripVoid -> kdIn,_):(_,stripVoid -> kdOut,_):_:(nm,_,_):_) = bbInputs bbCtx
```
DigitalBrains1, line 43: Unfortunately this style generates warnings on GHC 9.2 (non-exhaustive pattern match). Let's do it like this:

```haskell
clockWizardTemplate bbCtx
  | ((_,stripVoid -> kdIn,_):(_,stripVoid -> kdOut,_):_:(nm,_,_):_) <- bbInputs bbCtx
  , Just compName <- exprToString nm
  , KnownDomain _ clkInPeriod _ _ _ _ <- kdIn
  , KnownDomain _ clkOutPeriod _ _ _ _ <- kdOut
  = ...
  | otherwise
  = error ("clockWizardTclTemplate: bad bbContext: " <> show bbCtx)
```

christiaanb:
We could make it safe by adding `Not (domOut == pllLock), Not (domIn == pllLock)` as a constraint. That way `domOut` and `pllLock` cannot unify when `clockWizzard` is invocated. Meaning you must assign a domain to `pllLock` that is apart from `domOut` and `domIn`; thus forcing you to synchronize if you want to use the signal in either of those domains.

This would make it safe as it prevents accidental synchronization.

It would make it inconvenient to use though, as you have to create a `KnownDomain` instance for the concrete `pllLock` domain.

clash-prelude/src/Clash/Xilinx/ClockGen.hs
```diff
@@ -95,11 +106,8 @@ clockWizardDifferential
...
   -> Reset domIn
   -- ^ Reset for the PLL
-  -> (Clock domOut, Enable domOut)
+  -> (Clock domOut, Signal domPllLock Bool)
   -- ^ (Stable PLL clock, PLL lock)
-clockWizardDifferential !_name (Clock _ Nothing) (Clock _ Nothing) rst =
-  (Clock SSymbol Nothing, unsafeCoerce (toEnable (unsafeToHighPolarity rst)))
-clockWizardDifferential !_name _ _ _ =
-  error "clockWizardDifferential: no support for dynamic clocks"
+clockWizardDifferential !_ !_ = clocks
```
DigitalBrains1, line 111: Hah, I'm asking you to do A, you do A, and now I'm asking you to do B. Because I later realised this solution has a disadvantage: if people get creative with different clocks connected to the two clock inputs here, they will only encounter the `clocks: dynamic clocks unsupported` error message when they do it wrong on the positive phase. The negative phase is no longer checked. So could you change it to:

```suggestion
clockWizardDifferential !_ clk@(Clock _ Nothing) (Clock _ Nothing) =
  clocks clk
clockWizardDifferential !_ _ _ =
  error "clockWizardDifferential: dynamic clocks not supported"
```

(the rephrasing of the error message is intentional, I'm trying to go one step towards a single phrasing in `clash-prelude`).

DigitalBrains1 (changes requested):
Ah, nice work! I just have a few little things.

DigitalBrains1:
> Meaning you must assign a domain to pllLock that is apart from domOut and domIn; thus forcing you to synchronize if you want to use the signal in either of those domains.

This has surprising simulation results in Clash simulation, though. In HDL `unsafeSynchronizer` is just a wire, but in Clash **it resamples**. The simulation model of the lock output in Clash is that it is deasserted for as long as the `Reset domIn` is asserted. So it always starts as a signal in `domIn`. Suppose `domIn` has a period of 10 ns and `domOut` has a period of 2 ns. A 20 ns reset pulse on `Reset domIn` will keep `domOut` with a deasserted lock signal for 10 ticks, a period of 20 ns. Now suppose we introduce a domain for the lock signal that has a period of 100 ns. Suddenly a 20 ns pulse on `Reset domIn` will have the lock signal in `domOut`, with a period of 2 ns, asserted for 50 ticks which is 100 ns. This just feels weird, to introduce a domain that resamples like that purely to avoid people accidentally treating it as synchronous.

DigitalBrains1:
> The updated Blackboxes generates Tcl to instantiate a PLL or MMC through the Clocking Wizard.

This, by the way, is inaccurate. It always generates an MMCM. The automatic mode which picks the one the wizard likes best is only supported on UltraScale and UltraScale+, so we stuck to the default MMCM so it also works on at least 7 series (I don't think anything older, but not sure).

DigitalBrains1:
> So my very strong preference would still be to rename the current primitive yielding asynchronous signals to unsafe* and introducing a non-unsafe version that does the synchronization for you.

I fully support this (in a new PR).

DigitalBrains1:
Since we support Vivado in CI, we could (in a future PR) add a simple test that verifies we can instantiate and simulate a `clockWizard`. I'm talking about a dummy simulation, not actually verifying the output of the primitive, just that Vivado does not throw any errors while doing it. We could just have a counter on the output clock that will count towards test completion (`done = cnt .==. pure maxBound`). Shall we make a TODO issue for that?

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
DigitalBrains1: I suggest we leave it as is for this PR and in a future PR, define a function in `Clash.Signal.Internal` and expose it through the usual channels:

```haskell
periodToHz' :: (Integral a, Fractional b) => a -> b
periodToHz' = fromRational . toRational . periodToHz . fromIntegral
```

Or actually I'd like to propose something more radical and name it `periodToHz0` and also, for Clash 1.8, rename all functions ending in a prime symbol to functions with a `0` suffix, and then provide the prime version as perpetual synonyms for that. And never introduce any more functions ending in a prime symbol in `clash-prelude`. I'm not a big fan of prime symbols, you might have noticed.

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
```diff
@@ -0,0 +1,126 @@
...
+import qualified Data.String.Interpolate  as I
+import Data.Text.Prettyprint.Doc.Extra    (Doc)
+
+
+clockWizardTclTF :: TemplateFunction
+clockWizardTclTF =
+  TemplateFunction used valid clockWizardTclTemplate
+ where
+  used = [0..4]
+  valid bbCtx
```
DigitalBrains1, line 30: I missed this bit in my review. We've discussed the `used` in the office. Furthermore, this `valid` is incomplete, as it doesn't check `kdIn` and `kdOut` are actually `KnownDomain` constraints. But as far as I'm concerned, you make `valid` a `const True` and let the `TclTemplate` handle printing the error. In fact, I prefer the error message that produces :-).

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
hiddemoll: @DigitalBrains1 What do you think of generalizing the existing function (`periodToHz`) and adding a hint to the changelog that users can use type applications (`periodToHz @Natural @(Ratio Natural)`) to get the old behavior back - in case they get type errors?

DigitalBrains1: Yes, that's fine. People will get an extra trip through `Integer` but this is safe and the function doesn't need to perform blindingly fast at all.

DigitalBrains1:
I've completed a basic functionality test bench; it is in the branch [peter/xilinx-clockwiz](https://github.com/clash-lang/clash-compiler/tree/peter/xilinx-clockwiz). It adds `seClkToDiffClk` to our test bench functions, which still needs a bit more documentation. The idea is you put that commit first, followed by the `Add Xilinx clock wizard support through Tcl` commit. You can either include it here in this PR, or we create a separate PR for it if you want this merged quickly without more work.

Now that I see the commit summary, I realise I did not review the commit message. I'd like it to be more explicit. After all, you fix the API, you fix everything else about it (broken Haskell simulation, wrong port names, was there more?), and you add Tcl instantiation support. But the only thing you mention is that last bit :-D. Could you rephrase it to something like "Fix Xilinx clock wizard, add Tcl IP generation"? And also change the title of this PR accordingly?

clash-prelude/src/Clash/Clocks/Deriving.hs
```diff
@@ -78,6 +78,16 @@ derive' n = do
       let c = varT $ mkName ("c" ++ show n') in
       [t| Clock $c |]
 
+    -- | Generate @!_ = knownDomain dom@ with fresh @dom@ variable
```
DigitalBrains1, line 81: Ahh, now I get it. It's the `unsafeCoerce` that is wreaking havoc; and `Clash.Intel.ClockGen.altpll` prevents the `KnownDomain`s from being eliminated itself, which is why that one does work. But the only reason for the `unsafeCoerce` was to avoid `BangPatterns` which is enabled by default anyways. Plus it is no longer needed because we pattern match to detect dynamic clocks. And to top it off you're introducing bang patterns yourself here. So let's just get rid of the `unsafeCoerce` instead:

```diff
diff --git a/clash-prelude/src/Clash/Clocks/Deriving.hs b/clash-prelude/src/Clash/Clocks/Deriving.hs
index 8553e8f69..dfe62c369 100644
--- a/clash-prelude/src/Clash/Clocks/Deriving.hs
+++ b/clash-prelude/src/Clash/Clocks/Deriving.hs
@@ -13,12 +13,12 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 module Clash.Clocks.Deriving (deriveClocksInstances) where
 
 import Control.Monad               (foldM)
+import Clash.Promoted.Symbol       (SSymbol(..))
 import Clash.Explicit.Signal       (unsafeSynchronizer)
 import Clash.Signal.Internal
 import Language.Haskell.TH.Compat
 import Language.Haskell.TH.Syntax
 import Language.Haskell.TH.Lib
-import Unsafe.Coerce               (unsafeCoerce)
 
 conPatternNoTypes :: Name -> [Pat] -> Pat
 #if MIN_VERSION_template_haskell(2,18,0)
@@ -46,24 +46,22 @@ derive' n = do
 #endif
 
   -- Function definition of 'clocks'
-  let clk = mkName "clk"
   let rst = mkName "rst"
 
   -- Implementation of 'clocks'
+  clkImpl <- [| Clock SSymbol Nothing |]
   lockImpl <- [| unsafeSynchronizer clockGen clockGen
                    (unsafeToLowPolarity $(varE rst)) |]
   let
     noInline  = PragmaD $ InlineP (mkName "clocks") NoInline FunLike AllPhases
-    clkImpls  = replicate n (clkImpl clk)
+    clkImpls  = replicate n clkImpl
     instTuple = mkTupE $ clkImpls ++ [lockImpl]
     funcBody  = NormalB instTuple
     errMsg    = "clocks: dynamic clocks unsupported"
     errBody   = NormalB ((VarE 'error) `AppE` (LitE (StringL errMsg)))
     instFunc  = FunD (mkName "clocks")
       [ Clause
-          [ AsP
-              clk
-              (conPatternNoTypes 'Clock [WildP, conPatternNoTypes 'Nothing []])
+          [ (conPatternNoTypes 'Clock [WildP, conPatternNoTypes 'Nothing []])
           , VarP rst]
           funcBody
           []
@@ -91,9 +89,6 @@ derive' n = do
       let p = varT $ mkName "pllLock" in
       [t| KnownDomain $p |]
 
-
-    clkImpl clk = AppE (VarE 'unsafeCoerce) (VarE clk)
-
 -- Derive instances for up to and including to /n/ clocks
 deriveClocksInstances :: Int -> Q [Dec]
 deriveClocksInstances n = mapM derive' [1..n]
```
(this diff is against the original code of this file for clarity).

Your solution boils down to:

```haskell
instance Clocks (Clock c1, Clock c2, Signal pllLock Bool) where

  type ClocksCxt (Clock c1, Clock c2, Signal pllLock Bool) =
    (KnownDomain c1, KnownDomain c2, KnownDomain pllLock)

  clocks clk@(Clock _ Nothing) rst =
    ( unsafeCoerce clk
    , unsafeCoerce clk
    , unsafeSynchronizer clockGen clockGen (unsafeToLowPolarity rst))
   where
    !_ = knownDomain @c1
    !_ = knownDomain @c2
  clocks _ _ = error "clocks: dynamic clocks unsupported"
  {-# NOINLINE clocks #-}
```

while mine is
```haskell
instance Clocks (Clock c1, Clock c2, Signal pllLock Bool) where

  type ClocksCxt (Clock c1, Clock c2, Signal pllLock Bool) =
    (KnownDomain c1, KnownDomain c2, KnownDomain pllLock)

  clocks (Clock _ Nothing) rst =
    ( Clock SSymbol Nothing
    , Clock SSymbol Nothing
    , unsafeSynchronizer clockGen clockGen (unsafeToLowPolarity rst))
  clocks _ _ = error "clocks: dynamic clocks unsupported"
  {-# NOINLINE clocks #-}
```

DigitalBrains1 (changes requested):
Could you reorder the commits? If you put _Generalize `periodToHz`_ first, you don't need _Change frequency calculation to use `periodToHz`_.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -1617,11 +1619,19 @@ hzToFs freq = Femtoseconds (floor ((1.0 / freq) / 1.0e-15))
...
 -- 200000000 % 1
 --
 -- __NB__: This function is /not/ synthesizable
-periodToHz :: Natural -> Ratio Natural
-periodToHz period = 1.0 / (1.0e-12 * fromIntegral period)
+periodToHz :: (Integral a, Fractional b) => a -> b
+periodToHz = fromRational . toRational . periodToHzx
+ where
+  periodToHzx :: Integral a => a -> Ratio Natural
+  periodToHzx period = 1.0 / (1.0e-12 * fromIntegral period)
```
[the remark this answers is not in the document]
DigitalBrains1: Ugh, there's no call stack on the exception emitted when you pass in a negative argument:
```
*** Exception: arithmetic underflow
```

Let's handle that.
```suggestion
periodToHz :: HasCallStack => (Integral a, Fractional b) => a -> b
periodToHz period
  | period > 0
  = fromRational $ 1.0 / (fromIntegral period * 1e-12)
  | otherwise
  = withFrozenCallStack $ error "Negative or zero period"
```

You'll need to do this at the top of the file:
```diff
-import GHC.Stack                  (HasCallStack)
+import GHC.Stack                  (HasCallStack, withFrozenCallStack)
```

and your changelog entry needs to mention the changed exception for 0:
```
CHANGED: `periodToHz` now throws an `ErrorCall` with call stack when called with the argument 0 (zero), instead of a `ratioZeroDenomException :: ArithException`.
```

changelog/2023-03-01T10_02_55+01_00_generalized_periodToHz
```diff
@@ -0,0 +1 @@
+Changed: Generalized `periodToHz`. Use type applications (`periodToHz @Natural @(Ratio Natural)`) to get the old behaviour back, in case type errors arise.
```
DigitalBrains1, line 1:
```suggestion
CHANGED: Generalized `periodToHz`. Use type applications (`periodToHz @Natural @(Ratio Natural)`) to get the old behavior back, in case type errors arise.
```

DigitalBrains1 (changes requested):

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
```diff
@@ -0,0 +1,139 @@
...
+
+import Clash.Backend (Backend)
+import Clash.Netlist.BlackBox.Util (exprToString)
+import Clash.Netlist.Types
+import Clash.Netlist.Util (stripVoid)
+import Clash.Signal (periodToHz)
+
+import Control.Monad.State (State)
+import qualified Data.String.Interpolate as I
+import Data.Text.Prettyprint.Doc.Extra (Doc)
```
DigitalBrains1, line 22: https://clash-lang.readthedocs.io/en/latest/hacking-on-clash/style.html#imports

```suggestion
import Control.Monad.State (State)
import qualified Data.String.Interpolate as I

import Clash.Signal (periodToHz)

import Clash.Backend (Backend)
import Clash.Netlist.BlackBox.Util (exprToString)
import Clash.Netlist.Types
import Clash.Netlist.Util (stripVoid)
import Data.Text.Prettyprint.Doc.Extra (Doc)
```
(`Clash.Signal` is a _related third party import_).

I am aware that much of our code is not setting a good example here. Do as I say, not as I do ;-).

DigitalBrains1 (changes requested):

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
DigitalBrains1, line 22: https://clash-lang.readthedocs.io/en/latest/hacking-on-clash/style.html#imports :

```suggestion
import Clash.Backend (Backend)
import Clash.Netlist.BlackBox.Util (exprToString)
import Clash.Netlist.Types
import Clash.Netlist.Util (stripVoid)
import Clash.Signal (periodToHz)

import Control.Monad.State (State)
import qualified Data.String.Interpolate as I
import Data.Text.Prettyprint.Doc.Extra (Doc)
```
(`Clash.Signal` is a _related third party import_)

I am aware that our code is not really an example of proper application of these guidelines. Do as I say, not as I do ;-).

clash-prelude/src/Clash/Signal/Internal.hs
DigitalBrains1, line 1634: (This is a repost from a contribution that somehow disappeared. _Very_ frustrating, I'm glad Christiaan still had an e-mail notification with it.)

Ugh, there's no call stack on the exception emitted when you pass in a negative argument:
```
*** Exception: arithmetic underflow
```
Let's handle that.
```suggestion
periodToHz :: HasCallStack => (Integral a, Fractional b) => a -> b
periodToHz period
  | period > 0
  = fromRational $ 1.0 / (fromIntegral period * 1e-12)
  | otherwise
  = withFrozenCallStack $ error "Negative or zero period"
```

You'll need to do this at the top of the file:

```diff
-import GHC.Stack                  (HasCallStack)
+import GHC.Stack                  (HasCallStack, withFrozenCallStack)
```
and your changelog entry needs to mention the changed exception for 0:
```
CHANGED: `periodToHz` now throws an `ErrorCall` with call stack when called with the argument 0 (zero), instead of a `ratioZeroDenomException :: ArithException`.
```

clash-lib/src/Clash/Primitives/Xilinx/ClockGen.hs
```diff
@@ -0,0 +1,139 @@
...
+clockWizardTclTF :: TemplateFunction
+clockWizardTclTF =
+  TemplateFunction used valid clockWizardTclTemplate
+ where
+  knownDomIn = 0
+  knownDomOut = 1
+  -- knownDomPllLock = 2
+  name = 3
+  -- clk = 4
+  -- rst = 5
```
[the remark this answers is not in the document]
martijnbastiaan: Let's do this in another PR. The current state of this PR is already a massive improvement over the status quo, and we want to use it even if there are some style issues in the implementation.

martijnbastiaan (approved):
@DigitalBrains1's suggestions look good. I'm fine with merging after applying them.

hiddemoll:
I have already applied all Peters changes, with the exception of two:
1. I have not yet applied the infinite-list for the blackbox parameters
2. The periodToHz changes. Peter moved these to their own PR, which got merged over the weekend

I think I have only one more thing to do for this PR, which is removing my periodToHz changes, rebase those from Peter and then remove those commits without breaking other things.

DigitalBrains1 (approved):

DigitalBrains1 (changes requested):
Whoops, hit "Approve" too soon.

DigitalBrains1:
> Two remarks on the output `Signal domPllLock Bool`:

After much internal discussion we decided to leave this part alone in this PR, i.e., leave it at `domOut`, and tackle this in a future PR where we rename these primitives to `unsafe`... and introduce safe versions.

changelog/2023-03-01T12_18_28+01_00_xilinx_clocking_wizard
```diff
@@ -0,0 +1,2 @@
+Changed: Remove `Asynchronous` constraint. Originally intended to signal that `clockWizard` and `clockWizardDifferential` handle asynchronous resets. Since synchronous resets are a subset of synchronous resets this constraint is already met.
```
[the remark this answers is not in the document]
DigitalBrains1: Please note that your current version is not identical to my suggestion. If you don't want to use the GitHub interface for this (I dislike it because it causes load on CI with commits that get squashed afterwards), you could press _Edit_ on my message and simply copy-paste the bit that's between the _suggestion_ tags.

changelog/2023-03-01T12_18_28+01_00_xilinx_clocking_wizard
```diff
@@ -0,0 +1,3 @@
+CHANGED: Remove `Asynchronous` constraint from Xilinx `clockWizard` and `clockWizardDifferential`. Originally intended to signal that these functions react asynchronously to the incoming reset and that the outgoing lock signal is an asynchronous signal. Since synchronous resets are a subset of synchronous resets this constraint on the input is vacuous. The constraint on the lock output does not convey this information at all and is wrong. Note that it is still necessary to synchronize the lock output in your design.
+FIXED: Fix Xilinx ClockGen primitives. The port names of the Xilinx ClockGen primitives are now lowercase. The type of the lock output of Xilinx `clockWizard` and `clockWizardDifferential` is now `Signal Bool` instead of `Enable`, which was not the correct type here as a circuit should be kept in reset while the clock is stabilizing. This also fixes a polarity mismatch between hardware (and HDL simulation) and Haskell simulation. Also loosens the domain on the outgoing `Signal Bool`.
```
DigitalBrains1, line 2:
```suggestion
FIXED: Fix Xilinx ClockGen primitives. The port names of the Xilinx ClockGen primitives are now lowercase. The type of the lock output of Xilinx `clockWizard` and `clockWizardDifferential` is now `Signal Bool` instead of `Enable`, which was not the correct type here as a circuit should be kept in reset while the clock is stabilizing. This also fixes a polarity mismatch between hardware (and HDL simulation) and Haskell simulation. The lock signal is now also correctly resampled to the output domain.
```

DigitalBrains1 (changes requested):

DigitalBrains1 (approved):

PR #2428, opened 2023-02-25
Use prettyprinter-interp

DigitalBrains1:
When we splice a `string-interpolate` quasi quoter into a `prettyprinter` `Doc`, it is more efficient to use the `prettyprinter-interp` quasi quoter wrappers.

## Still TODO:


martijnbastiaan:
Perhaps unrelated, but does this package allow us to use `#` unescaped?  I encountered a few:

```
    • Data.String.Interpolate.__i: unescaped # symbol without interpolation brackets
```
..the other day and I thought it was quite obnoxious. We certainly didn't have to do this before.

DigitalBrains1:
No, `string-interpolate` requires you escape your pounds. While that may be a bit of an annoyance, I'm actually in favour of such rigour. If a character is special, you _should_ need to escape it to use literally, and if a character is not special, you shouldn't be allowed to escape it. There have been so many examples of strings that were accidentally right and then suddenly wrong later. Most famously when they added the `\u` escape for unicode code points and suddenly everybody who had escaped a `u` (which before was a literal `u`, there was no need to escape but it was allowed) suddenly had a broken string. Admittedly the reverse of your wish, though :-).

martijnbastiaan (approved):

PR #2429, opened 2023-02-27
Make `clash-ffi` a haskell library

kleinreact:
This PR reorganizes `clash-ffi` to be used like a normal Haskell library as part of an external project. 

To this end, `clash-ffi` now gets built as a standard Haskell library, which then is imported by the external project setting up the specific simulator actions. The external project is built as a `foreign-library`, which then can be loaded into the simulator. This approach introduces more flexibility in the sense that every specific simulation application results in its own VPI-loadable shared library. This is the way of usage, as it  is intended by standardized interfaces, like VPI, VHPI or FLI. 

A guideline for setting up a project using `clash-ffi` has been added to the `README.md`.

The PR also fixes a minor bug in the calculation of time values before they are sent to the simulator via the FFI.

### DIFF
--- a/clash-ffi/README.md
+++ b/clash-ffi/README.md
@@ -9,16 +9,51 @@ a standard interface (e.g. VPI / VHPI / FLI). Currently only VPI is supported.
 
 All interaction with a simulator follows the same general process:
 
-  * the Haskell code performing the FFI is compiled as a shared library. Any
-    libraries specified as `foreign-library` in `clash-ffi.cabal` are copied
-    into a `lib/` directory in the project on successful build. Haskell FFI
-    code *must* export the entry point
+  * The Haskell code performing the FFI is compiled as a foreign
+    shared library. To this end, your project's cabal must be extended
+    by a `foreign-library` section, e.g.,
 
-    ```c
-    void clash_ffi_main(void);
+    ```cabal
+    foreign-library <my-clash-ffi-lib-name>
+      default-language: Haskell2010
+      includes:         vpi_user.h
+      include-dirs:     </path/to/vpi_user.h>
+      build-depends:    clash-ffi,
+                        ...
+      type:             native-shared
+      lib-version-info: 0:1:0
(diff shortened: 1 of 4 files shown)


martijnbastiaan:
A quick drive-by comment: `cocotb` builds one `.so` file  (upon installation) and dynamically loads test benches / device under test drivers. We could replicate this behavior using [hint](https://hackage.haskell.org/package/hint). Still, the current proposal is an improvement over the status-quo. (And whether we want to go the `hint` route is up for debate.)

martijnbastiaan (approved):
Very nice. Could you add the example to CI? Just GitLab CI should be good. Let me know if you need help.

martijnbastiaan:
Thanks! Feel free to (enable auto-)merge.

clash-ffi/example/run-iverilog.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+# This is just a minimalistic script for demonstrating the process of
+# running the clash-ffi example using the Icarus Verilog VVP runtime
+# engine. The script is not designed to work in any possible system
+# environment and may not work immediatly for you. It is intended to
+# serve as an easy starter instead. Adapt it too you needs if it's not
+# working out-of-the-box for you.
```
DigitalBrains1, line 8:
```suggestion
# environment and may not work immediately for you. It is intended to
# serve as an easy starter instead. Adapt it to your needs if it's not
# working out-of-the-box for you.
```

clash-ffi/example/run-iverilog.sh
```diff
@@ -0,0 +1,36 @@
...
+
+###############################
+
+# adapt these variables, if the tools are not in your PATH already
```
DigitalBrains1, line 12: I feel the transitive use of _adapt_ is only possible with a phrase starting with _to_. I'm not 100% sure, but I prefer _change_. Actually I don't like change, but I mean, in this sentence.

The comma can go.

DigitalBrains1 (commented):
I was just glancing through this and notice some natural language thingies.

PR #2432, opened 2023-02-28
Workaround Haddock/CPP #if issues

### LINKED ISSUE #2278: Malformed haddock in `clash-prelude`
The API documentation for [`head` and `tail`](https://hackage.haskell.org/package/clash-prelude-1.6.3/docs/Clash-Sized-Vector.html#v:head) for the current stable version of Clash is slightly malformed, resulting in the documentation being rendered badly. I see:

Screenshot from 2022-07-10 17-53-58

Looking at the source, the functions causing problems attempt to use the preprocessor to change what is displayed:

```haskell
{- | Extract the first element of a vector

>>> head (1:>2:>3:>Nil)
1

#if __GLASGOW_HASKELL__ >= 900
>>> head Nil
<BLANKLINE>
<interactive>:...
    • Couldn't match type ‘1’ with ‘0’
      Expected: Vec (0 + 1) a
        Actual: Vec 0 a
    • In the first argument of ‘head’, namely ‘Nil’
      In the expression: head Nil
      In an equation for ‘it’: it = head Nil

#else
>>> head Nil
<BLANKLINE>
<interactive>:...
    • Couldn't match type ‘1’ with ‘0’
      Expected type: Vec (0 + 1) a
        Actual type: Vec 0 a
    • In the first argument of ‘head’, namely ‘Nil’
      In the expression: head Nil
      In an equation for ‘it’: it = head Nil

#endif
-}
```

leonschoorl:
This eliminates the `# [line number]` artifacts we sometimes got in the  generated documentation in place of excluded `#if` blocks.

Fixes #2278

## Still TODO:

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -30,6 +30,8 @@ package clash-prelude
   ghc-options: +RTS -qn4 -A128M -RTS -j4
   -- workaround for plugins not loading in Haddock with GHC-8.6
   haddock-options: --optghc=-fdefer-type-errors
+  -- Workaround for Haddock/CPP #if issues https://github.com/haskell/haddock/issues/1382
+  haddock-options: --optghc="-optP -P"
   -- Don't pollute docs with 1024 SNat literals
   haddock-options: --optghc=-DHADDOCK_ONLY
 


martijnbastiaan:
What do you mean with:

> But then line numbers in error messages don't match the user sourcecode anymore.

What error messages are we talking about?

leonschoorl:
Like this:
```haskell
{-# LANGUAGE CPP #-}

#if 0

Some block

that is removed
by
CPP.
#endif

x :: Int
x = False  -- < this is line 13
```

Normally ghc reports an error with nice location information:
```
$ ghc testcpp.hs 
[1 of 1] Compiling Main             ( testcpp.hs, testcpp.o )

testcpp.hs:13:5: error:
    • Couldn't match expected type ‘Int’ with actual type ‘Bool’
    • In the expression: False
      In an equation for ‘x’: x = False
   |
13 | x = False  -- < this is line 13
   |     ^^^^^
```

But when we add these workaround `-optP -P` arguments to ghc everywhere, we get errors like this:
```
$ ghc testcpp.hs -optP -P
[1 of 1] Compiling Main             ( testcpp.hs, testcpp.o )

/tmp/ghc673736_0/ghc_1.hscpp:228:5: error:
    • Couldn't match expected type ‘Int’ with actual type ‘Bool’
    • In the expression: False
      In an equation for ‘x’: x = False
    |
228 | x = False  -- < this is line 13
    |     ^^^^^
```

That's why I only add them to ghc run by haddock.

martijnbastiaan (approved):

PR #2435, opened 2023-03-06
New: `indexTuple`

DigitalBrains1:
This function can be used to bind variables to numbers indexing successive elements in a list-like structure. This is really useful for writing primitives like:

```haskell
  let primName = 'Clash.Cores.Xilinx.Floating.Explicit.fromU32With
      ( _knownDomainArg :: Int
       , _knownNatArg
       , _hasCallStackArg
       , clockArg
       , enableArg
       , inputArg) = indexTuple
  in InlineYamlPrimitive [VHDL] [__i|
    BlackBox:
      name: #{primName}
      template: |-
        [...]
        ~IF~ISACTIVEENABLE[#{enableArg}]~THEN      aclken : in std_logic;
```

Before GHC 9.2, we could simply do
```haskell
a : b : c : d : _ = [0 :: Int ..]
```
but since GHC 9.2 this gives a warning:
```
Indices.hs:5:3: warning: [-Wincomplete-uni-patterns]
    Pattern match(es) are non-exhaustive
    In a pattern binding:
        Patterns of type ‘[Int]’ not matched:
            []
            [_]
            [_, _]
            [_, _, _]
  |
5 |   a : b : c : d : _ = [0 :: Int ..]
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
and we like our code to be warning-free.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
+++ b/clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
@@ -1,0 +1,62 @@
+{-|
+Copyright  :  (C) 2023,      QBayLogic B.V.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+
+{-# LANGUAGE TemplateHaskell #-}
+
+module Clash.Class.IndexTuple.Deriving
+  (IndexTuple(..), deriveIndexTupleInstances) where
+
+import Control.Monad.Extra (concatMapM)
+import Language.Haskell.TH
+
+class IndexTuple a where
+  -- | This function can be used to bind variables to numbers indexing
+  -- successive elements in a list-like structure. If the type of the indices
+  -- can be derived from context, it is enough to provide a tuple with the
+  -- variables like:
+  --
+  -- @
+  -- (a, b) = indexTuple
+  -- (a, b, c, d, e, f) = indexTuple
+  -- @
+  --
+  -- and so on. @a@ will be 0, @b@ 1, @c@ 2, etcetera. All variables will have
+  -- the same type, and specifying the type of any one of the variables is
+  -- enough. The type needs to be an instance of `Num`.
+  --
+  -- If a type needs to be specified here, you can use
+  --
+  -- @
+  -- (a :: Int, b, c) = indexTuple
+  -- @
+  --
+  -- or
+  --
+  -- @
+  -- a :: Int
+  -- (a, b, c) = indexTuple
+  -- @
+  --
+  -- depending on preference.
+  indexTuple :: a
+
+-- Derive instance for /n/-tuple
+deriveIndexTupleInstance :: Int -> DecsQ
+deriveIndexTupleInstance n = do
+  let e (i :: Int) = varT $ mkName $ "e" ++ show i
+      e0  = e 0
+      eqT i = [t| $(e i) ~ $e0 |]
+      instCxt0 = appT (tupleT n) [t| Num $e0 |]
+      instCxt1 = foldl appT instCxt0 $ map eqT [1 .. n - 1]
+      instType  = foldl appT (tupleT n) $ map e [0 .. n - 1]
+      funRHS = tupE $ map (litE . integerL) [0 .. toInteger n - 1]
+  [d| instance $instCxt1 => IndexTuple $instType where
+        indexTuple = $funRHS
+   |]
+
+-- Derive instances for up to and including to /n/-tuples
+deriveIndexTupleInstances :: Int -> DecsQ
+deriveIndexTupleInstances n = concatMapM deriveIndexTupleInstance [1..n]
--- a/changelog/2023-03-06T15_37_28+01_00_indexTuple
+++ b/changelog/2023-03-06T15_37_28+01_00_indexTuple
@@ -1,0 +1,2 @@
+NEW: `indexTuple`. This function can be used to bind variables to numbers
+indexing successive elements in a list-like structure.
--- a/clash-prelude/clash-prelude.cabal
+++ b/clash-prelude/clash-prelude.cabal
@@ -298,6 +298,8 @@ Library
 
   other-modules:
                       Clash.Class.AutoReg.Instances
+                      Clash.Class.IndexTuple
+                      Clash.Class.IndexTuple.Deriving
                       Clash.CPP
                       Clash.Signal.Bundle.Internal
                       Language.Haskell.TH.Compat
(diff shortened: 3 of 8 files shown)


clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
```diff
@@ -0,0 +1,62 @@
...
+-- Derive instance for /n/-tuple
+deriveIndexTupleInstance :: Int -> DecsQ
+deriveIndexTupleInstance n = do
+  let e (i :: Int) = varT $ mkName $ "e" ++ show i
+      e0  = e 0
+      eqT i = [t| $(e i) ~ $e0 |]
+      instCxt0 = appT (tupleT n) [t| Num $e0 |]
+      instCxt1 = foldl appT instCxt0 $ map eqT [1 .. n - 1]
+      instType  = foldl appT (tupleT n) $ map e [0 .. n - 1]
+      funRHS = tupE $ map (litE . integerL) [0 .. toInteger n - 1]
```
martijnbastiaan, line 55:
```suggestion
  let 
    e (i :: Int) = varT $ mkName $ "e" ++ show i
    e0  = e 0
    eqT i = [t| $(e i) ~ $e0 |]
    instCxt0 = appT (tupleT n) [t| Num $e0 |]
    instCxt1 = foldl appT instCxt0 $ map eqT [1 .. n - 1]
    instType  = foldl appT (tupleT n) $ map e [0 .. n - 1]
    funRHS = tupE $ map (litE . integerL) [0 .. toInteger n - 1]
```

I'm going to be _that guy_ and say we indent blocks with 2 spaces

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
```diff
@@ -0,0 +1,62 @@
...
+  --
+  -- or
+  --
+  -- @
+  -- a :: Int
+  -- (a, b, c) = indexTuple
+  -- @
+  --
+  -- depending on preference.
+  indexTuple :: a
```
martijnbastiaan, line 44: This function _could_ take a starting index. I doubt it's useful though, and if we want to introduce it that would be a good moment to try [Retrie](https://hackage.haskell.org/package/retrie-1.2.1.1/docs/Retrie.html). I thought I'd mention it anyway.

martijnbastiaan (approved):
How about converting existing inline blackboxes to use this? That way readers referred here by the CHANGELOG would have some diffs to look at.

Either way, approve from me.

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
```diff
@@ -0,0 +1,62 @@
...
+  --
+  -- @
+  -- (a :: Int, b, c) = indexTuple
+  -- @
+  --
+  -- or
+  --
+  -- @
+  -- a :: Int
+  -- (a, b, c) = indexTuple
```
martijnbastiaan, line 40: Also I'm a sucker for doctests :eyes:, any reason not to use that here?

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
DigitalBrains1: We _do_? My, you live and learn :-)

I don't think I ever indented a let-block like that, but it's fine by me!

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
DigitalBrains1: Syntax. I didn't like
```
>>> :{
... multiline
:}
```
much. But I can change it if you want.

christiaanb:
What does this add over:

```
a : b : c : d : _ = [0..]
```

DigitalBrains1:
> What does this add over:
> 
> ```
> a : b : c : d : _ = [0..]
> ```

```
Indices.hs:5:3: warning: [-Wincomplete-uni-patterns]
    Pattern match(es) are non-exhaustive
    In a pattern binding:
        Patterns of type ‘[Int]’ not matched:
            []
            [_]
            [_, _]
            [_, _, _]
  |
5 |   a : b : c : d : _ = [0 :: Int ..]
  |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

martijnbastiaan:
You'll have to disable the new uni-pattern warning in GHC >=9.2 for that.

christiaanb:
Meh, a case where it is actually fine to ignore the warning…

DigitalBrains1:
Definitely, `Int` is one of them
```haskell
>>> let high = fromInteger @Int $ 2^63 - 5
>>> [high..]
[9223372036854775803,9223372036854775804,9223372036854775805,9223372036854775806,9223372036854775807]
```

christiaanb:
Since we’re actually in the clash prelude, we can do:

```
import Clash.Signal.Internal

enumFromS :: Enum a => a -> Signal dom a
enumFromS a = a :- enumFromS (succ a)
```
Which is the `Signal` counterpart to `enumFrom` which is used in the designating of `[x..]` as you probably know

And then we can do:


```
a :- b :- c :- d :- _ = fromEnumS 0
```

and we won’t get any pattern match warnings.

DigitalBrains1:
That only helps ourselves and not our users, since we don't give API guarantees for `Internal` modules. Or is `Clash.Signal.Internal` different? Also, we might one day want to decouple `clash-cores` from the rest, and then we can't use it either if the API is unstable :-).

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
leonschoorl: If we wanted something like this that takes a starting index, I'd like to call that something else, maybe `indexTupleFrom`.
And keep this one starting from 0.

martijnbastiaan:
Nice trick @christiaanb. It feels a bit wrong to use `Signal` that way IMO, but sure.

DigitalBrains1:
Let's first discuss this internally, this PR is getting a bit noisy :-). I replied on Slack about the Signal trick.

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
DigitalBrains1: It seems the name will change anyway! `indexTuple` does not get good vibes.

clash-prelude/src/Clash/Class/IndexTuple/Deriving.hs
```diff
@@ -0,0 +1,62 @@
...
+  -- @
+  --
+  -- depending on preference.
```
leonschoorl, line 43: From this description alone it looks like a slightly different copy of  `indicesI`.
Which it is, but not without reason.
Maybe add something along the lines of:
> This is similar to [`indicesI`](https://hackage.haskell.org/package/clash-prelude-1.6.4/docs/Clash-Sized-Vector.html#v:indicesI) but can be without incurring non-exhaustive pattern warnings with `-Wincomplete-uni-patterns` enabled.
>Note that `-Wincomplete-uni-patterns` is implied by `-Wall` starting from GHC x.y.
> It's intended for use in defining inline primitives, see this example.

Where `this example` is a link to an example of using it in an inline primitive, either in `Clash.Annotations.Primitive` or possible in `Clash.Tutorial`.

leonschoorl:
Superseded by using `infinite-list` like in #2441

PR #2436, opened 2023-03-09
Generalize period<->frequency conversion, use

DigitalBrains1:
`periodToHz`, `hzToPeriod`, `fsToHz` and `hzToFs` can be generalized so
they can be more easily used in cases where we don't want `Natural` and
`Ratio Natural` as the types.

However, the common use of `vPeriod=hzToPeriod 33e6` means we want the
argument to `hzToPeriod` to be monomorphic. Making it polymorphic means
it defaults to the inferior `Double` and warns about this defaulting. So
we merely generalize the return types of these functions.

These functions used to throw exceptions without call stacks when called
with zero. By changing that to `ErrorCall` we can get a stack trace,
solving an immense frustration in debugging a Clash design.

Currently, the frequency calculation in for the Intel PLLs depends on
the unit of the period of a `KnownDomain` (picoseconds). If this unit is
to be changed internally in the future (to femtoseconds), this
calculation would produce the wrong frequency. By using `periodToHz`,
the unit of the period no longer matters and can be changed internally.
This use case is what prompted this PR.

Harmonized and slightly improved some documentation.

## Still TODO:

(part of this commit message pilfered from a message written by Hidde)

### DIFF
--- a/changelog/2023-03-09T17_04_39+01_00_generalize_freqtools
+++ b/changelog/2023-03-09T17_04_39+01_00_generalize_freqtools
@@ -1,0 +1,6 @@
+CHANGED: Generalized the return types of `periodToHz` and `hzToPeriod`. Use a
+type application (`periodToHz @(Ratio Natural)`, `hzToPeriod @Natural`) to get
+the old behavior back, in case type errors arise.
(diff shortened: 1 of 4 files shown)


martijnbastiaan (approved):
Clear improvement, minimal impact. LGTM!

PR #2439, opened 2023-03-16
Drop snap package

DigitalBrains1:
The Clash snap package has not been a recommended way to use Clash for quite some time, and it is a hassle to support.

In order to build a snap package, we build .deb packages for Clash with Ubuntu 20.04 LTS. But the interaction between the Debian build system and GHC is problematic, requiring significant effort to support and to upgrade to a more recent Ubuntu release.

Additionally, snap packages have their own issues on distributions other than Ubuntu. Given that we no longer recommend people use our snap package and given the effort required to keep supporting it, we have decided to drop the snap package.

## Still TODO:


martijnbastiaan (approved):
Sad, but understandable.

PR #2443, opened 2023-03-24
GitLab CI: hack to deal with GHC heisenbug

DigitalBrains1:
Every now and then, GHC will exit with the error

```
out: mmap 131072 bytes at (nil): Cannot allocate memory
out: Try specifying an address with +RTS -xm<addr> -RTS
out: internal error: m32_allocator_init: Failed to map
    (GHC version 9.0.2 for x86_64_unknown_linux)
    Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

(when the binary is named `out`). For some reason this problem has become more pronounced for us. Since we invoke GHC/Clash an awful amount of times in some of our CI tests, the chances of hitting it in one of those invocations are really high. This commit wraps the relevant binaries in a Bash script that will retry for a total of three times when this error message is observed. It seems reasonable to think that the odds of it failing three times in a row are very small.

We think this problem occurs on every combination of GHC version and Linux kernel version, but we are seeing it on (almost?) exclusively on GHC 9.0.2.

## Still TODO:


martijnbastiaan (approved):
I'm not sure we did @christiaanb, so we could try it once and see if that fixes things (I kinda doubt it though). Otherwise LGTM.

DigitalBrains1:
Interestingly, compiling a binary with `-rtsopts` and then invoking the binary with `+RTS -xm20000000 -RTS` does seem to fix the problem. At least, I don't see it anymore where the problem was very pronounced before. Note that since it is actually `clash` that might fail, we'd need to compile Clash itself with `-rtsopts`. But that might be the better fix.

It is interesting how the [doc](https://downloads.haskell.org/ghc/latest/docs/users_guide/runtime_control.html#rts-flag--xm%20%E2%9F%A8address%E2%9F%A9) seems to suggest that this problem is not about compiled Haskell binaries at all:
> ❗Warning
> [...] Do not use unless GHCi fails with a message like [...]

GHCi has nothing to do with the problem we see here whatsoever.

DigitalBrains1:
And indeed instead of compiling with `-rtsopts` and then passing RTS options at runtime, it is also simply possible to compile with `--with-rtsopts=-xm20000000`, as you would expect to be the case.

DigitalBrains1:
Superseded by #2444

PR #2444, opened 2023-03-29
Add flag to workaround GHC heisenbug on CI

DigitalBrains1:
In very specific tests in GitLab CI we are affected by GHC bug #19421. We can work around the issue by passing `-with-rtsopts=-xm20000000` when compiling an affected binary. This is a stopgap measure until the real bug is fixed.

We have seen the bug:
- In `clash-testsuite` in `clashLibTest`s
- In `ffi:example` in the `clash` binary itself
- In `prelude:doctests`, probably in the `doctests` binary itself, although this is not certain.

and then only in GHC 9.0.2, although the bug should be in other versions of GHC as well.

This workaround was applied only to GHC 9.0.2 on CI and only to those cases that were observed to go wrong, although as a consequence now the `clash` binary is always built with the RTS option.

I tried to limit applying this RTS option as much as possible because of [the documentation](https://downloads.haskell.org/~ghc/9.0.2/docs/html/users_guide/runtime_control.html#rts-flag--xm%20%E2%9F%A8address%E2%9F%A9):
> ❗Warning
> This option is for working around memory allocation problems only. Do not use unless GHCi fails with a message like [...]

Our problem has nothing to do whatsoever with GHCi, but even if we ignore that, it still says "do not use unless". It would be possible to restrict the building of the `clash` binary with the option to _just_ `ffi:example`, by giving it its own special `cabal.project.local`, but I did not take it that far in this PR.

The `ffi:example` test should no longer be affected by the GHC bug, but it is a fairly useless test in its current state as it does not fail when there are issues, and the tested code is actually faulty. It is disabled to be fixed later.

## Still TODO:

  - ~~Check copyright notices are up to date in edited files~~ This is a temporary CI workaround. The sooner it is removed, the better. No need for copyright messages if the code is soon to be removed again.

### DIFF
--- a/.ci/cabal.project.local
+++ b/.ci/cabal.project.local
@@ -10,7 +10,7 @@ package *
 
 package clash-prelude
   ghc-options: -Werror
-  flags: +doctests +multiple-hidden
+  flags: +doctests +multiple-hidden -workaround-ghc-mmap-crash
   tests: True
   benchmarks: True
 
@@ -24,6 +24,7 @@ package clash-lib
 
 package clash-ghc
   ghc-options: -Werror
+  flags: -workaround-ghc-mmap-crash
 
 package clash-cosim
   ghc-options: -Werror
@@ -47,7 +48,7 @@ package clash-lib-hedgehog
 package clash-testsuite
   ghc-options: -Werror
   -- enable cosim
-  flags: +cosim +multiple-hidden
+  flags: +cosim +multiple-hidden -workaround-ghc-mmap-crash
 
 package clash-benchmark
   ghc-options: -Werror
--- a/.ci/gitlab/test.yml
+++ b/.ci/gitlab/test.yml
@@ -138,12 +138,13 @@ ffi:interface-tests:
   script:
     - ./dist-newstyle/build/*/*/clash-ffi-*/x/ffi-interface-tests/build/ffi-interface-tests/ffi-interface-tests --smallcheck-max-count 2000
 
-ffi:example:
-  extends: .test-cache-local
-  script:
-    - cabal build clash # ensure clash has been built (avoids some legacy cabal issue)
-    - cd clash-ffi/example && ./run-iverilog.sh
-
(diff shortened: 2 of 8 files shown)


.ci/cabal.project.local
```diff
@@ -10,7 +10,7 @@ package *
 
 package clash-prelude
   ghc-options: -Werror
-  flags: +doctests +multiple-hidden
+  flags: +doctests +multiple-hidden -workaround-heisenbug
```
martijnbastiaan, line 13: Don't you want `+workaround-heisenbug`?

DigitalBrains1: No, that would set it for all GHC versions. `.ci/setup.sh` turns it on just for GHC 9.0.2.

leonschoorl: I like that `workaround-ghc19421` is more specific then `workaround-heisenbug`.
But I still hate it, the name is meaningless to my mind.
And also confusingly if one were to read the [GHC-9.0.2 release notes](https://downloads.haskell.org/ghc/9.0.2/docs/html/users_guide/9.0.2-notes.html), it lists 19421 as fixed.

I propose something like `workaround-ghc-mmap-crash`

DigitalBrains1:
If people like the new version, feel free to merge and get this show on the road again!

martijnbastiaan (approved):

.ci/cabal.project.local
martijnbastiaan: Pretty strong word for such a minor thing 😅.

Suggestion seems fine to me. Approved the PR too.

DigitalBrains1:
> In prelude:doctests, probably in the doctests binary itself, although this is not certain.

I tried to determine this but the bug is being a cute little heisenbug. I can't even produce locally at all, it mainly happens on CI. I tried passing `--verbose` to the doctest to see if it printed anything before dying, but with `--verbose`, _the bug does not appear_. So I gave up. I could determine its location, but not its speed (how quickly it happens). It was Heisenberg through and through in that respect!

PR #2445, opened 2023-03-29
clash-ffi: fix example

christiaanb:


DigitalBrains1:
Could you include the improvements in [peter/fix_clash-ffi_example](https://github.com/clash-lang/clash-compiler/tree/peter/fix_clash-ffi_example)? It can become a single commit in my opinion.

But maybe it's better to wait first until PR #2444 is merged because CI is totally broken now.

kleinreact (approved):
LGTM. I also added some fixes regarding simulation termination, which are required for running the example with Questa / ModelSim and a script for doing exactly that.

Furthermore, I cherry picked @DigitalBrains1's fixes. Feel free to squash them if you like.

DigitalBrains1:
You accidentally picked up an old version of my branch. I've force-pushed this branch which fixes my YAML mistake. The GitLab CI documentation examples are pretty liberal with special characters in plain YAML scalars; I just took it too far :-).

DigitalBrains1:
Hey, that's really weird... In the `after-script`:
```yaml
  after_script:
    - tar -cf - /root/.cabal | zstd -T${THREADS} -3 > cache.tar.zst
    - rm -r "$MASKBINS"
```
the variable `THREADS` is still defined but `MASKBINS` no longer, as can be seen in the job log. I don't know why one is and the other is not, it makes no sense to me. Unless `THREADS` is coming from somewhere else, but that doesn't seem very likely (does it?).

The purpose of the `rm` was to clean up after ourselves... we've already seen that these jobs hardly start in a clean environment, so I was worried the runner would accumulate temporary dirs on all runs. It might do that now because the directory is not deleted afterwards. But I don't understand why that variable is no longer available.

Initially, I didn't even `export MASKBINS`, because they were clearly not running in a subshell (`export` would be useless), so they must be running in a shared shell, and children didn't need the variable, only that shell running the `script`s. Then I noticed that it was no longer set in `after_script` and I `export`ed it anyway. But that didn't even fix it.

kleinreact:
> Hey, that's really weired ...

It actually makes sense to me. If the commands get executed **after** the script then the scope of the script got already closed and your exports are gone. I don't see `THREADS` still being there: `zstd` does not complain about an empty `-T` option.

DigitalBrains1:
But `script` does get the vars set in `before_script` so why not `after_script`?

Anyway, I thought of a solution. If you use
```bash
export MASKBINS="$(mktemp -d -p "$(readlink -f .)" maskbins.XXXXXXXXXX)"
```
in the `before_script`, the directory gets created in the current directory and hopefully the checkout stage will remove it then like it does with other leftover files, so the `rm` in the `after_script` can be removed.

martijnbastiaan:
> The purpose of the rm was to clean up after ourselves... we've already seen that these jobs hardly start in a clean environment, so I was worried the runner would accumulate temporary dirs on all runs. It might do that now because the directory is not deleted afterwards. But I don't understand why that variable is no longer available.

GitLab Runner _should_ clean the environment before starting a job. I've never seen this causing problems either, do you have?

DigitalBrains1:
Really? Every job log starts with
```
Removing .ghc.environment.x86_64-linux-8.6.5
Removing cabal.project.local
Removing cache.tar.zst
Removing clash-cosim/src/cbits/cosim_clash.o
Removing clash-cosim/src/cbits/cosim_vpi.o
Removing clash-cosim/src/cbits/cosim_vpi.sl
Removing clash-cosim/src/cbits/cosim_vpi.vpi
Removing clash-cosim/src/cbits/libcosim_clash.so
Removing clash-cosim/src/prims/verilog/Clash_CoSim_CoSimInstances.primitives.yaml
Removing dist-newstyle/
Removing dist.tar.zst
```

DigitalBrains1:
Yean and if it doesn't clean `/tmp` like it cleans `/builds/clash-lang/clash-compiler`, and how would it know _what_ to clean from `/tmp`, it will accumulate.

martijnbastiaan:
`/tmp` is not a bound volume AFAIK, so it will start clean with every docker instance. In any case, this is something for the folks at GitLab to worry about. If we have any evidence that GitLab Runner messes up here, then it's fine to apply workarounds (and report a bug..).

PR #2451, opened 2023-04-14
Expose non-`XException` bottoms in `hasX`

### LINKED ISSUE #2450: `hasX` might hide exceptions other than `XException`
If the value evaluated with `hasX` contains multiple exceptions, only the first one encountered will determine the outcome of `hasX`.

```haskell
>>> let a = errorX "XException" :> error "ErrorCall" :> Nil
>>> let b = error "ErrorCall" :> errorX "XException" :> Nil
>>> hasX a
Left "X: XException\nCallStack (from HasCallStack):\n  errorX, called at <interactive>:1:9 in interactive:Ghci2"
>>> hasX b
*** Exception: ErrorCall
  error, called at <interactive>:2:9 in interactive:Ghci3
```

In my opinion, both should have thrown an `ErrorCall`. The documentation for `hasX` describes it as _fully evaluate a value_. It is currently not evaluating anything beyond the `XException`.

Currently, `hasX` reduces its argument to normal form using `NFData`, and catches `XException`s. This checks only one of the exceptions the argument might throw. Instead, we probably need to first check for exceptions other than `XException` and only after that check for all exceptions (which boils down to just the `XException`s of course):

```haskell
hasX :: (NFDataX a, NFData a) => a -> Either String a
hasX a =
  unsafeDupablePerformIO
    (catch
      (evaluate (rnfX a `seq` rnf a) >> return (Right a))
      (\(XException msg) -> return (Left msg)))
```

martijnbastiaan:
Fixes #2450

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/XException.hs
+++ b/clash-prelude/src/Clash/XException.hs
@@ -2,7 +2,7 @@
 Copyright  :  (C) 2016,      University of Twente,
                   2017,      QBayLogic, Google Inc.
                   2017-2019, Myrtle Software Ltd,
-                  2021-2022, QBayLogic B.V.
+                  2021-2023, QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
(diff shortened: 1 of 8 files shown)


clash-prelude/src/Clash/XException.hs
```diff
@@ -307,12 +306,12 @@ maybeIsX = maybeX isX
...
 -- If a data structure contains multiple 'XException's, the "first" message is
--- picked according to the implementation of 'rnf'.
-hasX :: NFData a => a -> Either String a
+-- picked according to the implementation of 'rnfX'.
+hasX :: NFDataX a => a -> Either String a
 hasX a =
   unsafeDupablePerformIO
     (catch
-      (evaluate (rnf a) >> return (Right a))
+      (evaluate (rnfX a) >> return (Right a))
```
christiaanb, line 314: This works for things like: `hasX (3, errorX “QQ”)`? I’d be surprised, since rnfX would return the tuple, not throw the XException.

martijnbastiaan: Right... it should be `rnf (forceX ..)` I suppose. These would cause two traversals unfortunately.

christiaanb: I think you need to do:
```
catch (evaluate (rnf a) >> return (Right a))
 (\(XException msg) -> evaluate (rnfX a) >> return (Left msg))
```
So you need both the NFData and NFDataX constraint.

martijnbastiaan: Alright, that makes sense. I'll add a TODO stating we could make `hasX` a type class method to prevent multiple traversals (even in the worst case).

DigitalBrains1 (approved):
LGTM!

DigitalBrains1:
For bonus points, you could extend the Haddock examples for `maybeHasX` and `hasX` with an example with both `XException` and `_|_` like we now test in `clash-prelude:unittests`.

Nice realisation that you only need `rnfX` when `rnf` actually found an `XException`, btw.

PR #2452, opened 2023-04-18
Add `XPM_CDC_GRAY`

martijnbastiaan:
## Still TODO:

  - [x] Figure out how to use XPM in combination with `launch_simulation`
    - [x] `set_property -name xsim.elaborate.xelab.more_options -value {-L xpm} -objects 
[current_fileset -sim]`
    - [x] fix weird `vl_logic` errors (https://support.xilinx.com/s/question/0D54U00006jQGuESAW/how-can-i-use-xpmcdcgray-using-launchsimulation)
  - [X] ~~Implement realistic Haskell model~~
    - [X] ~~Implement gray encoder/decoder~~
  - [x] Figure out weird start up behavior
    - [x] [Due to Xilinx's GSR](https://support.xilinx.com/s/question/0D52E00007Eu4OESAZ/gsr-usage?language=en_US) 
  - [x] Add `HasCallStack` to `xpmCdcGray#`
  - [x] Make initial values configurable. Only do auto-detection in non-`With` version.

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Gray.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Gray.hs
@@ -1,0 +1,99 @@
+{-|
+  Copyright   :  (C) 2023, Google LLC
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Clash.Cores.Xilinx.Xpm.Cdc.Gray
+  ( xpmCdcGray
+  , XpmCdcGrayConfig(..)
+  , xpmCdcGrayWith
+  ) where
+
+import GHC.Stack (HasCallStack)
+
+import Clash.Explicit.Prelude
+
+import Clash.Cores.Xilinx.Xpm.Cdc.Gray.Internal (xpmCdcGray#)
+
+-- | Synchronizes an 'Unsigned' from the source clock domain to the destination
+-- clock domain using Gray code. It instantiates Xilinx's @XPM_CDC_GRAY@, so no
+-- additional constraint definitions are needed. In order to synchronize data
+-- from the source domain to the destination domain, consecutive inputs need to
+-- be unchanged, successors (+1), or predecessors (-1). Overflows are okay. That is,
+-- 'maxBound' can be followed by zero, and vice versa. If this invariant is
+-- violated, the component might not transfer data correctly until the invariant
+-- is upheld again and the source clock has sampled the input once, followed by
+-- /4/ samples in the destination domain.
+--
+-- If you need all data to be transferred from the source domain to the destination
+-- domain, make sure the destination domain samples the input domain at least
+-- twice. Alternatively, use a FIFO.
+--
+-- Read more in [PG382](https://docs.xilinx.com/r/en-US/pg382-xpm-cdc-generator/XPM_CDC_GRAY).
+--
+-- __N.B.__: The simulation model does not detect invariant violations.
+--
+-- __N.B.__: In order to simulate initial values, both the source and destination
+--           domain need to support them. If the source and destination domain
+--           disagree on this property, use of this function will fail to
+--           simulate and translate to HDL.
+--
+xpmCdcGray ::
+  forall n src dst.
+  ( 2 <= n, n <= 32
+  , KnownNat n
+  , KnownDomain src
+  , KnownDomain dst
+  , HasCallStack
+  ) =>
+  Clock src ->
+  Clock dst ->
+  Signal src (Unsigned n) ->
+  Signal dst (Unsigned n)
+xpmCdcGray = xpmCdcGrayWith XpmCdcGrayConfig{..}
+ where
+  stages = d4
+  initialValues =
+    case (initBehavior @src, initBehavior @dst) of
+      (SDefined, SDefined) -> True
+      (SUnknown, SUnknown) -> False
+      _ -> clashCompileError $ "xpmCdcGray: domains need to agree on initial value "
+                            <> "behavior. To set initial value usage explicitly, "
+                            <> "consider using 'xpmCdcGrayWith'."
+{-# INLINE xpmCdcGray #-}
+
+-- | Configuration for 'xpmCdcGrayWith'
+data XpmCdcGrayConfig stages = XpmCdcGrayConfig
+  { -- | Number of synchronization stages. I.e., number of registers in the
+    -- destination domain. Note that there is always a register in the source
+    -- domain.
+    stages :: SNat stages
+
+    -- | Initialize registers used within the primitive to /0/. Note that
+    -- 'xpmCdcGray' will set this to 'True' if both domains support initial
+    -- values, to 'False' if neither domain does, and will otherwise emit an
+    -- error.
+  , initialValues :: Bool
+  }
+
+-- | Like 'xpmCdcGray', but with a configurable number of stages. Also see
+-- 'XpmCdcGrayConfig'.
+xpmCdcGrayWith ::
+  forall stages n src dst.
+  ( 2 <= n, n <= 32
+  , 2 <= stages, stages <= 10
+  , KnownNat n
+  , KnownDomain src
+  , KnownDomain dst
+  , HasCallStack
+  ) =>
+  XpmCdcGrayConfig stages ->
+  Clock src ->
+  Clock dst ->
+  Signal src (Unsigned n) ->
+  Signal dst (Unsigned n)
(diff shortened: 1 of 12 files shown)


tests/shouldwork/Cores/Xilinx/XpmCdcGrayTypes.hs
```diff
@@ -0,0 +1,73 @@
...
+  actual = xpmCdcGrayWith @stages @width (XpmCdcGrayConfig SNat) clkA clkB counter
+
+  done =
+    outputVerifierWith
+      (\clk rst -> assertBitVector clk rst "outputVerifier Port A")
+      clkB clkB noRstB
+      (expected)
+      (pack <$> actual)
+
+  noRstA = noRst @a
```
cuddlefishie, line 48: This isn't used in this function anywhere, is this some left over dead code?

tests/shouldwork/Cores/Xilinx/XpmCdcGrayTypes.hs
```diff
@@ -0,0 +1,73 @@
...
+ where
+  counter = delay clkA enableGen 0 (counter + 1)
+
+  actual = xpmCdcGrayWith @stages @width (XpmCdcGrayConfig SNat) clkA clkB counter
+
+  done =
+    outputVerifierWith
+      (\clk rst -> assertBitVector clk rst "outputVerifier Port A")
+      clkB clkB noRstB
+      (expected)
```
cuddlefishie, line 45: unneeded parenthesis

tests/shouldwork/Cores/Xilinx/XpmCdcGray.hs
```diff
@@ -0,0 +1,41 @@
...
+  done =             Types.tb @D3 @D5 @16     @10     @100     Proxy Proxy SNat expected
+  expected = $(Types.expected @D3 @D5 @16     @10     @100     Proxy Proxy SNat SNat SNat)
+{-# ANN tb2 (TestBench 'topEntity) #-}
+
+tb3 = done
+ where
+  --                          src dst  width  stages  samples
+  done =             Types.tb @D3 @D5 @16     @2      @100     Proxy Proxy SNat expected
+  expected = $(Types.expected @D3 @D5 @16     @2      @100     Proxy Proxy SNat SNat SNat)
+{-# ANN tb3 (TestBench 'topEntity) #-}
```
cuddlefishie, line 41: Is there a reason why `XpmCdcGrayTypes` also defines a `D11` but it's not tested here? I would expect there to be a test that uses that domain or that it's removed if these tests cover all interesting behaviour already

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Gray.hs
```diff
@@ -0,0 +1,89 @@
...
+
+-- | Synchronizes an 'Unsigned' from the source clock domain to the destination
+-- clock domain using Gray code. It instantiates Xilinx's @XPM_CDC_GRAY@, so no
+-- additional constraint definitions are needed. In order to synchronize data
+-- from the source domain to the destination domain, consecutive inputs need to
+-- be stable, successors (+1), or predecessors (-1). Overflows are okay. That is,
+-- 'maxBound' can be followed by zero, and vice versa. If these invariants are
+-- violated, the component might not transfer data correctly until the source clock
+-- has sampled the input once, followed by /4/ samples in the destination
+-- domain.
```
DigitalBrains1, line 30: If there are 4 synchronizer stages, wouldn't this be _5_ samples? The first sample of the first synchronizer stage after an invariant violation might be corrupted, but, when the invariant is upheld once again, the second sample will be clean. Followed by three more synchronizer stages before it is observable.

Also, let's make sure the user understands that data will be corrupted until they uphold the invariants again, because a naive reading of this documentation might lead to the conclusion that as long as you wait _n_ samples for certain _n_, data will be clean.

```suggestion
-- 'maxBound' can be followed by zero, and vice versa. If these invariants are
-- violated, the component might not transfer data correctly until the invariant
-- is upheld again and the source clock has sampled the input once, followed by
-- /4/ samples in the destination domain.
```

Note that the extra sample is now part of the text :-).

DigitalBrains1 (commented):
The code is not a reflection of the effort :-). It took some work to straighten this all out, it's almost like CDC logic is never easy... Kudos on hunting down all the little snags!

I have some minor thingies that aren't very crucial, but also a major point regarding documentation; at least, if I am correct in my analysis.

tests/shouldwork/Cores/Xilinx/XpmCdcGrayTypes.hs
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE BangPatterns #-}
+
+module XpmCdcGrayTypes where
```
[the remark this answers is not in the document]
DigitalBrains1: I see @cuddlefishie also pointed out the `dropN`, I did not read that review before finally submitting mine.

tests/shouldwork/Cores/Xilinx/XpmCdcGrayTypes.hs
martijnbastiaan: You're right, same story for `dropN`.

tests/shouldwork/Cores/Xilinx/XpmCdcGray.hs
martijnbastiaan: Hmm, well I guess we'd want to test domains being exactly twice at fast and _more_ than twice as fast, so I'll add D10 too.

martijnbastiaan:
> The code is not a reflection of the effort :-). 

Ain't that the truth.

@DigitalBrains1 @cuddlefishie I've applied your suggestions.

DigitalBrains1 (approved):
LGTM!

martijnbastiaan:
Great! I think there's been enough scrutiny offline, so I'll merge this tomorrow if there's no further feedback. And then I'm going to pump out all the other CDC primitives like there's no tomorrow.

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Gray.hs
```diff
@@ -0,0 +1,103 @@
+{-|
+  Copyright   :  (C) 2023, Google LLC
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE RecordWildCards #-}
+
+-- See [Note: eta port names for tdpbram]
+{-# OPTIONS_GHC -fno-do-lambda-eta-expansion #-}
```
christiaanb, line 10: Why do you do this in combination with the NOINLINE over at
https://github.com/clash-lang/clash-compiler/pull/2452/files#diff-e9f756a2ad405040ed110e5eca448f27ac331270f762b039437671580e660149R103?

Why not just tag `xpmCdcGrayWith` with an `INLINE` instead?

The reason I'm asking is the reason that we do https://github.com/clash-lang/clash-compiler/blob/95ee5b9421991abfea0783e9295e09c2383ad87f/clash-prelude/src/Clash/Explicit/BlockRam.hs#L1217 is because of https://github.com/clash-lang/clash-compiler/blob/95ee5b9421991abfea0783e9295e09c2383ad87f/clash-prelude/src/Clash/Explicit/BlockRam.hs#L1223

Is that the case for this XPM_CDC_GRAY as well? That Vivado complains when XPM_CDC_GRAY does not live in its own architecture?

DigitalBrains1: If that is the case, I feel you should also link `[Note: true dual-port blockRAM separate architecture]` because the `NOINLINE` on `xpmCdcGrayWith` is a bit surprising and should be documented.

martijnbastiaan: This is _just_ for pretty names in the HDL. I guess it's not that relevant given that `XPM_CDC_GRAY` has pretty names to begin with. Vivado does not complain about anything if we remove the flag + NOINLINE. I assumed the NOINLINE was necessary for Clash to pick up on the arg / port names.

Does this answer your question? Would you like it removed instead given the macro already has pretty port names?

christiaanb: Yes, I’d like the ghc option to be removed and have `xpmCdcGrayWith` marked `INLINE` instead.

DigitalBrains1: It should probably also be eta-reduced then.

martijnbastiaan: Done

PR #2454, opened 2023-04-21
Improve speed of Xilinx tdpbram simulation

DigitalBrains1:
This PR should make writing data in `clash-cores:Clash.Cores.Xilinx.BlockRam.tdpbram` a lot quicker as it reduces the number of bit operations on the underlying `Natural`s that make up a `BitVector` very significantly for all write operations.

A version of `updateRam` that was never merged was this code:
### Version 1
```haskell
goAdjust oldDat =
  let
    -- .&. and .|. will take care of any undefined bits in 'byteEnable'
    bitEnable = byteMaskToBitMask byteEna
    datBv = pack dat .&. bitEnable
    oldDatBv = pack oldDat .&. complement bitEnable
    newDatBv = datBv .|. oldDatBv
  in
    unpack newDatBv
```
The problem with this code occurs when the byte enable has an undefined bit. Bits that are zero in both the old memory value and the data input will remain zero, but not so for one bits. If the memory contained `0b0000_1111`, the data to write was also `0b0000_1111`, and the byte enable was undefined, the memory would then be updated to be `0b0000_....`, with only the zero bits retained. This could be very confusing to users looking at the data, and even more so if they aren't looking at `BitVector`s but say a sum type where some constructors are retained while others are erased, to name an example.

So the code that was eventually merged is:
### Version 2
```haskell
goAdjust oldDat =
  let
    datBv0 = pack dat
    oldDatBv0 = pack oldDat
    bitEnable = byteMaskToBitMask byteEna
    -- Skip any bits that are the same in old and new: by handling this
    -- explicitly, we avoid these bits becoming undefined unnecessarily.
    -- .&. and .|. will take care of any undefined bits in 'byteEna'.
    bitEnableMasked = bitEnable .&. (oldDatBv0 `xor` datBv0)
    datBv1 = datBv0 .&. bitEnableMasked
    oldDatBv1 = oldDatBv0 .&. complement bitEnableMasked
    newDatBv = datBv1 .|. oldDatBv1
  in
    unpack newDatBv
```

It retains both 0-bits and 1-bits, giving the most precise result possible given the inputs.

This does come at a performance penalty though. Let's analyse how many operations on `Natural` are needed to implement a bit operation on `BitVector`:

|  | `(.&.) @(BitVector _)` | `(.\|.) @(BitVector _)` | `xor @(BitVector _)` | `complement @(BitVector _)` |
| --- | --- | --- | --- | --- |
| `(.&.) @Natural` | 5 | 4 | 1 | 1 |
| `(.\|.) @Natural` | 2 | 3 | 1 | |
| `xor @Natural` <sup>1</sup>| 1 | 1 | 2 | 2 |
| `complementMod` startup| 1 | 1 | 1 | 1

<sup>1</sup> Includes `complementMod` costs

The cost of [`complementMod`](https://hackage.haskell.org/package/clash-prelude-1.6.4/docs/Clash-Sized-Internal-Mod.html#v:complementMod) is twofold: a shared startup cost per `BitVector` operation and the cost of `xor @Natural` for each use.

This PR rewrites the code to dive into the `BitVector` and reduce the number of operations needed on `Natural`, while keeping the behaviour.

If we now look at the cost of the code above in `Natural` operations and this PR, it's:

| | Version 1 | Version 2 | This PR
| --- | --- | --- | --- |
| `(.&.) @Natural` | 15 | 21 | 6
| `(.\|.) @Natural` | 7 | 10 | 5
| `xor @Natural` | 7 | 10 | 3
| `complementMod` startup| 4 | 6 | 1

These costs are fixed irrespective of whether bits are undefined or not, so it speeds up all forms of writing to memory.
## Still TODO:

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/BlockRam/Internal.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/BlockRam/Internal.hs
@@ -20,6 +20,7 @@ import Clash.Explicit.Prelude hiding (enable)
 import Clash.Annotations.Primitive (Primitive(InlineYamlPrimitive), hasBlackBox)
 import Clash.Explicit.BlockRam.Model (TdpbramModelConfig(..), tdpbramModel)
 import Clash.Sized.Internal.BitVector (BitVector(BV), undefined#)
+import Clash.Sized.Internal.Mod (complementMod)
 import Clash.XException.MaybeX (MaybeX(..))
 
 import Data.String.Interpolate (__i)
@@ -78,6 +79,7 @@ isActiveWriteEnable = fmap (/= 0)
 
 -- | Update a true dual-port block RAM with byte enables
 updateRam ::
+  forall nBytes a .
   ( KnownNat nBytes
   , BitSize a ~ (8 * nBytes)
   , BitPack a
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
After a comment by Martijn, I realised I completely forgot to include the test I had in mind to actually test all possible bit patterns fed to `updateRam`. A small adjustment to the existing unit test is sufficient to get full coverage, including ill-defined `BitVector`s (in `BV mask val`, `mask` and `val` don't normally have any 1 bits at the same positions, but it is good to know such `BitVector`s still lead to correct results).

martijnbastiaan:
This looks good to me, I'll await the test code before given official "LGTM" :+1:

DigitalBrains1:
I redefined the old and new values in the `accesBv` unit tests to cover all possible bit combinations. (There are 16 of them and they already used a `BitVector 16`. Pretty practical!) I also added two tests where individual byte enables were zero: while the existing tests have overlap for the purpose of testing bit combinations, they actually omit that case. Testing with byte enables all-zero is superfluous as this follows a different code path entirely.

PR #2458, opened 2023-04-25
Add `XPM_CDC_SINGLE` to `clash-cores`

martijnbastiaan:
Question for reviewers:

 * ~~Should the primitive synchronize `(BitPack a, BitSize a ~ 1) => a` instead of `Bit`? Pro: no pack/unpack overhead when syncing a `Bool`. Con: more complicated type signature.~~ I've kinda answered by question during the implementation of `XPM_CDC_HANDSHAKE`: packing/unpacking is very annoying.

## Still TODO:


DigitalBrains1 (commented):
LGTM! I just have a few small points.

DigitalBrains1:
I think it would be really good if you added this bit to `xpmCdcGray` as well:

```haskell
-- __N.B.__: In order to simulate initial values, both the source and destination
--           domain need to support them. If the source and destination domain
--           disagree on this property, use of this function will fail to
--           simulate and translate to HDL. You can explicitly set it using
--           'xpmCdcGrayWith'.
```

It's currently missing that final sentence, which could confuse users. They might even think they cannot use this XPM at all when the domains don't agree on that setting.

tests/Main.hs
```diff
@@ -503,6 +503,12 @@ runClashTest = defaultMain $ clashTestRoot
               , hdlSim=[Vivado]
               , buildTargets=BuildSpecific ["tb" <> show n | n <- [(1::Int)..7]]
               }
+          , runTest "XpmCdcSingle" $ def
+              { hdlTargets=[VHDL, Verilog]
+              , hdlLoad=[]
+              , hdlSim=[Vivado]
+              , buildTargets=BuildSpecific ["tb" <> show n | n <- [(1::Int)..7]]
```
DigitalBrains1, line 510: You don't wanna run poor `tb0`?

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Single.hs
```diff
@@ -0,0 +1,89 @@
...
+  , BitSize a ~ 1
+  ) =>
+  Clock src ->
+  Clock dst ->
+  Signal src a ->
+  Signal dst a
+xpmCdcSingle = xpmCdcSingleWith XpmCdcSingleConfig{..}
+ where
+  registerInput = True
+  stages = d4
```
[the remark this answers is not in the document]
martijnbastiaan: Yes, but also! We should probably have a `defXpmCdcSingleConfig`: this makes it easier to use non-standard values without having to define them all.

martijnbastiaan:
Thanks for the comments, I agree with them all. I'll also have a look at import ordering.

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Single.hs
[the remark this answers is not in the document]
martijnbastiaan: So... `defXpmCdcSingleConfig :: (KnownDomain src, KnownDomain dst) => Clock src -> Clock dst -> XpmCdcSingleConfig 4`. Not exactly the simplicity I was hoping to get..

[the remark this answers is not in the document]
martijnbastiaan: I'm going to omit it for now, we can always add it later without API breakage.

martijnbastiaan:
I've also added module documentation to `Clash.Cores.Xilinx.Xpm` and `Clash.Cores.Xilinx.Xpm.Cdc`, and hid all internal modules from `clash-cores`. I think the Haddock looks quite a bit nicer now!

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc.hs
```diff
@@ -2,10 +2,23 @@
...
+  A collection of Clock Domain Crossing (CDC) focused Xilinx Parameterized
+  Macros (XPMs). These macros are supported by all modern Xilinx FPGA platforms,
+  including UltraScale and UltraScale+. Use of these macros is preferred over
+  manual clock domain crossings, as they include any constraints needed to make
+  them work reliably. They will also automatically suppress warnings typically
+  generated by manual CDC, leading to fewer false negatives in logs.
+
+  __N.B.__: Care should always be taken when doing CDC, even when using vendor
+            vetted primitives. Always read the documentation and make sure the
+            primitives's prerequisites and invariants are upheld.
```
DigitalBrains1, line 15:
```suggestion
            primitives' prerequisites and invariants are upheld.
```

DigitalBrains1 (approved):
If you fix this small grammar mistake you have my blessings!

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Single.hs
```diff
@@ -0,0 +1,92 @@
...
+  , BitPack a
+  , BitSize a ~ 1
+  ) =>
+  XpmCdcSingleConfig stages ->
+  Clock src ->
+  Clock dst ->
+  Signal src a ->
+  Signal dst a
+xpmCdcSingleWith XpmCdcSingleConfig{..} clkSrc clkDst input =
+  xpmCdcSingle# registerInput initialValues stages clkSrc clkDst input
```
DigitalBrains1, line 91: Wait a min... now that you're no longer doing the `no-do-lambda-eta-expansion` dance, this can just be eta reduced, right? For one thing, it might get inlined more often in user code.

martijnbastiaan, line 91:
```suggestion
xpmCdcSingleWith XpmCdcSingleConfig{..} =
  xpmCdcSingle# registerInput initialValues stages
```

tests/shouldwork/Cores/Xilinx/XpmCdcSingleTypes.hs
```diff
@@ -0,0 +1,85 @@
...
+  -- | Expected data
+  Vec n (BitVector 1) ->
+  Signal b Bool
+tb Proxy Proxy initVals regInput SNat expectedDat = done
+ where
+  counter = delay clkA enableGen 0 (counter + 1)
+
+  actual =
+    xpmCdcSingleWith
+      @stages @(BitVector 1)
```
martijnbastiaan, line 43:
```suggestion
      @stages @(Unsigned 1)
```

PR #2459, opened 2023-04-26
Add `XPM_CDC_ARRAY_SINGLE` to `clash-cores`

martijnbastiaan:
Depends on https://github.com/clash-lang/clash-compiler/pull/2458

## Still TODO:

  - [X] ~~Write a changelog entry (see changelog/README.md)~~
  - [X] Check copyright notices are up to date in edited files

PR #2461, opened 2023-04-29
Add `XPM_CDC_HANDSHAKE` to `clash-cores`

martijnbastiaan:
Depends on https://github.com/clash-lang/clash-compiler/pull/2459
## Still TODO:

  - [x] ~~Clean up var names in Haskell model~~ They're inspired by Xilinx's. I can't think of much better names..
  - [X] ~~Randomized / non-happy path tests?~~ Upon reviewer insistence :-). The implementation of this primitive was reverse engineered from the schematic, so I consider the chance that there are any difference to be pretty small.

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake.hs
@@ -1,0 +1,144 @@
+{-|
+  Copyright   :  (C) 2023, Google LLC
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE RecordWildCards #-}
+
+module Clash.Cores.Xilinx.Xpm.Cdc.Handshake
+  ( xpmCdcHandshake
+  , XpmCdcHandshakeConfig(..)
+  , xpmCdcHandshakeWith
+  ) where
+
+import Clash.Explicit.Prelude
+
+import GHC.Stack (HasCallStack)
+
+import Clash.Cores.Xilinx.Xpm.Cdc.Handshake.Internal (xpmCdcHandshake#)
+
+-- | Synchronizes data from the source clock domain to the destination. For this
+-- to function correctly, a full handshake must be completed before another data
+-- transfer is initiated. The handshake is considered completed when both sides
+-- have acknowledged the transfer and the handshake signals have been reset.
+--
+-- By default, it uses four synchronization stages in both source and
+-- destination domains, and auto-detects whether to use initial values for the
+-- synchronization registers. Use 'xpmCdcHandshakeWith' to change these
+-- settings. For more information see [PG382](https://docs.xilinx.com/r/en-US/pg382-xpm-cdc-generator/XPM_CDC_HANDSHAKE).
+--
+-- __N.B.__: In order to simulate initial values, both the source and destination
+--           domain need to support them. If the source and destination domain
+--           disagree on this property, use of this function will fail to
+--           simulate and translate to HDL. You can explicitly set it using
+--           'xpmCdcHandshakeWith'.
+xpmCdcHandshake ::
+  forall a src dst.
+  ( 1 <= BitSize a, BitSize a <= 1024
+  , KnownDomain src
+  , KnownDomain dst
+  , BitPack a
+  , NFDataX a
+  , HasCallStack
+  ) =>
+  Clock src ->
+  Clock dst ->
+  -- | Word to synchronize to destination domain. This value should not change
+  -- when @src_send@ is asserted.
+  "src_in" ::: Signal src a ->
+
+  -- | Assertion of this signal allows the @src_in@ bus to be synchronized to the
+  -- destination clock domain. This signal should only be asserted when @src_rcv@
+  -- is deasserted, indicating that the previous data transfer is complete. This
+  -- signal should only be deasserted once @src_rcv@ is asserted, acknowledging
+  -- that the @src_in@ has been received by the destination logic.
+  "src_send" ::: Signal src Bool ->
+
+  -- | Asserting this signal indicates that data on @dest_out@ has been captured
+  -- by the destination logic. This signal should be deasserted once @dest_req@ is
+  -- deasserted, completing the handshake on the destination clock domain and
+  -- indicating that the destination logic is ready for a new data transfer.
+  "dst_ack" ::: Signal dst Bool ->
+
+  -- | @dest_req@ indicates that @dest_out@ contains valid data. It can be
+  -- acknowledges by asserting @dst_ack@. @src_rcv@ indicates that the destination
+  -- domain has acknowledged a data transfer.
+  ( "dest_out" ::: Signal dst a
+  , "dest_req" ::: Signal dst Bool
+  , "src_rcv"  ::: Signal src Bool
+  )
+xpmCdcHandshake = xpmCdcHandshakeWith XpmCdcHandshakeConfig{..}
+ where
+  srcStages = d4
+  dstStages = d4
+  initialValues =
+    case (initBehavior @src, initBehavior @dst) of
+      (SDefined, SDefined) -> True
+      (SUnknown, SUnknown) -> False
+      _ -> clashCompileError $ "xpmCdcHandshake: domains need to agree on initial value "
+                            <> "behavior. To set initial value usage explicitly, "
+                            <> "consider using 'xpmCdcHandshakeWith'."
+{-# INLINE xpmCdcHandshake #-}
+
+-- | Configuration for 'xpmCdcHandshakeWith'
+data XpmCdcHandshakeConfig srcStages dstStages = XpmCdcHandshakeConfig
+  { -- | Number of synchronization stages in the source domain
+    srcStages :: SNat srcStages
+
+    -- | Number of synchronization stages in the destination domain
+  , dstStages :: SNat dstStages
+
+    -- | Initialize registers used within the primitive to /0/. Note that
+    -- 'xpmCdcHandshake' will set this to 'True' if both domains support initial
+    -- values, to 'False' if neither domain does, and will otherwise emit an
+    -- error.
+    --
+    -- This value is ignored in Clash simulation on domains configured to not
+    -- support initial values.
+  , initialValues :: Bool
+  }
+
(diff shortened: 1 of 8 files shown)


clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake/Internal.hs
```diff
@@ -0,0 +1,229 @@
...
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE ViewPatterns #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+{-# OPTIONS_GHC -fno-do-lambda-eta-expansion #-}
```
martijnbastiaan, line 12: Remove

DigitalBrains1 (commented):
These review suggestions were still pending from yesterday. I'll continue the review next week, but you can have these two morsels now :-P

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake.hs
```diff
@@ -0,0 +1,137 @@
...
+  ) where
+
+import GHC.Stack (HasCallStack)
+
+import Clash.Explicit.Prelude
+
+import Clash.Cores.Xilinx.Xpm.Cdc.Handshake.Internal (xpmCdcHandshake#)
+
+-- | Synchronizes an array of bits from the source clock domain to the
+-- destination.
```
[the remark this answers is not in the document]
DigitalBrains1: No, actually, make that

```suggestion
-- | Synchronizes data from the source clock domain to the destination. For this
-- to function correctly, a full handshake&#x2014;an acknowledgment that the
-- data transfer was received and a resetting of the handshake
-- signals&#x2014;must be completed before another data transfer is initiated.
-- By default, it uses four synchronization stages in both source and
-- destination domains, and auto-detects whether to use initial values for the
-- synchronization registers. Use 'xpmCdcArraySingleWith' to change these
-- settings.For more information see [PG382](https://docs.xilinx.com/r/en-US/pg382-xpm-cdc-generator/XPM_CDC_HANDSHAKE)
```

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake.hs
```diff
@@ -0,0 +1,136 @@
...
+  , KnownDomain dst
+  , BitPack a
+  , NFDataX a
+  , HasCallStack
+  ) =>
+  XpmCdcHandshakeConfig srcStages dstStages ->
+  Clock src ->
+  Clock dst ->
+
+  -- | Word to synchronize to destination domain
```
DigitalBrains1, line 111:
```suggestion
  -- | Word to synchronize to destination domain. This value should not change
  -- when @src_send@ is asserted.
```

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake/Internal.hs
```diff
@@ -0,0 +1,227 @@
...
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE ViewPatterns #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+module Clash.Cores.Xilinx.Xpm.Cdc.Handshake.Internal where
```
DigitalBrains1, line 12: Could you please reorder the imports?

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake/Internal.hs
```diff
@@ -0,0 +1,227 @@
...
+  ( 2 <= srcStages, srcStages <= 10
+  , 2 <= dstStages, dstStages <= 10
+  , 1 <= BitSize a, BitSize a <= 1024
+  , KnownDomain src
+  , KnownDomain dst
+  , BitPack a
+  , NFDataX a
+  , HasCallStack
+  ) =>
+  -- | Initial values supported
```
DigitalBrains1, line 187:
```suggestion
  -- | Initial value usage
```

tests/Main.hs
```diff
@@ -510,6 +510,12 @@ runClashTest = defaultMain $ clashTestRoot
               , hdlSim=[Vivado]
               , buildTargets=BuildSpecific ["tb" <> show n | n <- [(1::Int)..7]]
               }
+          , runTest "XpmCdcHandshake" $ def
+              { hdlTargets=[VHDL, Verilog]
+              , hdlLoad=[]
+              , hdlSim=[Vivado]
+              , buildTargets=BuildSpecific ["tb" <> show n | n <- [(1::Int)..7]]
```
DigitalBrains1, line 517:
```suggestion
              , hdlLoad=[Vivado]
              , hdlSim=[Vivado]
              , buildTargets=BuildSpecific ["tb" <> show n | n <- [(0::Int)..7]]
```

tests/shouldwork/Cores/Xilinx/XpmCdcHandshakeTypes.hs
```diff
@@ -0,0 +1,128 @@
...
+  -- | Expected data (Maybe + data)
+  Vec n (BitVector 9) ->
+  Signal b Bool
+tb Proxy Proxy opts expectedDat = done
+ where
+  actual = top @(Unsigned 8) opts clkA clkB
+
+  done =
+    outputVerifierWith
+      (\clk rst -> assertBitVector clk rst "outputVerifier Port A")
```
DigitalBrains1, line 102: Since there are no undefined samples, how about using regular `assert`? Actually, regular `outputVerifer` would do as there is only a single `assert` in the test bench, so it's not important what its label is. I kinda forgot about that factoid in previous reviews of these XPM's, it seems rather laborious to use `outputVerifierWith` if you don't have multiple `assert`s anyway.

DigitalBrains1 (changes requested):
Chugging along on the XPM reviews...

tests/shouldwork/Cores/Xilinx/XpmCdcHandshakeTypes.hs
DigitalBrains1: The tangential advantage of `assert` over `assertBitVector` is its readability.

tests/Main.hs
martijnbastiaan: This one actually starts at 1! But I'll renumber the tests. Don't know what got to me when I started at 1, phah.

DigitalBrains1: 🤣 You got me there! This was to see if I was still paying attention, right?

tests/shouldwork/Cores/Xilinx/XpmCdcHandshakeTypes.hs
DigitalBrains1: Ah. But this only fails in HDL simulation, by definition, and there the readability of a `Maybe (Unsigned 8)` is not going to be great no matter which `assert`.

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake/Internal.hs
```diff
@@ -0,0 +1,228 @@
...
+  srcRcv = xpmCdcSingleWith (defOpts dstStages) clkDst clkSrc dstAck
+
+  srcSendFf = delay clkSrc enableGen (initVal False) srcSend
+  srcHsDataFf = delay clkSrc (toEnable (not <$> srcSendFf)) (initVal (unpack 0)) srcIn
+  dstOutEna = toEnable (srcSendFfSynced .&&. fmap not dstReq)
+
+  dstOut =
+    delay
+      clkDst dstOutEna (initVal (unpack 0))
+      (unsafeSynchronizer clkSrc clkDst srcHsDataFf)
```
DigitalBrains1, line 221: Why is this safe? It seems like if the timing skew between some bit out of `unsafeSynchronizer` and `srcSendFf` is large enough (several clock ticks), the data might not be stable. `dstOutEna` might respond to `srcSend` while the data is still propagating.

It seems to me the `unsafeSynchronizer` needs a timing constraint that puts the propagation delay below those several clock ticks. Are you sure you spotted all the timing constraints in the XPM model?

tests/shouldwork/Cores/Xilinx/XpmCdcHandshakeTypes.hs
DigitalBrains1: As discussed, this now uses regular ol' `outputVerifier'`.

clash-cores/src/Clash/Cores/Xilinx/Xpm/Cdc/Handshake.hs
```diff
@@ -0,0 +1,145 @@
...
+-- | Synchronizes data from the source clock domain to the destination. For this
+-- to function correctly, a full handshake must be completed before another data
+-- transfer is initiated. A handshake is considered completed if an
+-- acknowledgment that the data transfer was received and the handshake signals
+-- have been reset.
+--
+-- By default, it uses four synchronization stages in both source and
+-- destination domains, and auto-detects whether to use initial values for the
+-- synchronization registers. Use 'xpmCdcArraySingleWith' to change these
+-- settings. For more information see [PG382](https://docs.xilinx.com/r/en-US/pg382-xpm-cdc-generator/XPM_CDC_HANDSHAKE).
```
DigitalBrains1, line 29:
```suggestion
-- | Synchronizes data from the source clock domain to the destination. For this
-- to function correctly, a full handshake must be completed before another data
-- transfer is initiated. The handshake is considered completed when both sides
-- have acknowledged the transfer and the handshake signals have been reset.
--
-- By default, it uses four synchronization stages in both source and
-- destination domains, and auto-detects whether to use initial values for the
-- synchronization registers. Use 'xpmCdcHandshakeWith' to change these
-- settings. For more information see [PG382](https://docs.xilinx.com/r/en-US/pg382-xpm-cdc-generator/XPM_CDC_HANDSHAKE).
```

DigitalBrains1 (approved):

PR #2464, opened 2023-05-02
Add SimOnly data type

christiaanb:
A container for data you only want to have around during simulation and is ignored during synthesis. Useful for carrying around things such as:

  * A map of simulation/vcd traces
  * Co-simulation state or meta-data
  * etc.


martijnbastiaan (approved):
I've been in need of this multiple times, LGTM.

PR #2466, opened 2023-05-11
Fix error messages for `Unsigned`'s `pred` implementations

kleinreact:
Just found this by accident.

### DIFF
--- a/clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
@@ -308,10 +308,9 @@ instance KnownNat n => Enum (Unsigned n) where
 
   pred n
     | n == minBound =
-        error $ "'pred' was called on (" <> show @(Unsigned n) maxBound <> " :: "
-             <> "Unsigned " <> show (natToNatural @n) <> ") and caused an "
-             <> "underflow. Use 'satPred' and specify a SaturationMode if you "
-             <> "need other behavior."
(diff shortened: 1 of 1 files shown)


clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
```diff
@@ -308,7 +308,7 @@ instance KnownNat n => Enum (Unsigned n) where
 
   pred n
     | n == minBound =
-        error $ "'pred' was called on (" <> show @(Unsigned n) maxBound <> " :: "
+        error $ "'pred' was called on (" <> show @(Unsigned n) minBound <> " :: "
```
christiaanb, line 311: I mean... we might as well do:
```suggestion
        error $ "'pred' was called on (0 :: "
```

clash-prelude/src/Clash/Sized/Internal/Unsigned.hs
```diff
@@ -308,7 +308,7 @@ instance KnownNat n => Enum (Unsigned n) where
...
              <> "Unsigned " <> show (natToNatural @n) <> ") and caused an "
              <> "underflow. Use 'satPred' and specify a SaturationMode if you "
```
christiaanb, line 313: While we're changing the error message anyway, how about:
```suggestion
             <> "overflow. Use 'satPred' and specify a SaturationMode if you "
```
In my book, an underflow is something that happens to representations of fractional numbers, not whole numbers.

martijnbastiaan (approved):

PR #2468, opened 2023-05-17
Mention `hwSeqX` requirement

kleinreact:
We just discussed that this may be good to mention.


martijnbastiaan:
I think we should also mention it the functions documentation - this is what you would read as an HLS user.

kleinreact:
I moved the hint to the function's documentation then, to avoid redundencies.

martijnbastiaan (approved):

PR #2491, opened 2023-06-08
Various small fixes - part 12

DigitalBrains1:
## Still TODO:

  - ~~Check copyright notices are up to date in edited files~~ These are so small, let's just skip this step.

### DIFF
--- a/clash-lib/src/Clash/Backend/SystemVerilog.hs
+++ b/clash-lib/src/Clash/Backend/SystemVerilog.hs
@@ -1126,7 +1126,7 @@ expr_ _ (Identifier id_ (Just (Indexed ((Unsigned w),_,_))))  = do
 expr_ _ (Identifier _ (Just (Indexed ((BitVector _),_,0)))) = do
   iw <- Ap $ use intWidth
   traceIf True ($(curLoc) ++ "WARNING: synthesizing bitvector mask to dontcare") $
-    verilogTypeErrValue (Signed iw)
+    verilogTypeErrValue (Unsigned iw)
 
 -- See [Note] bitvector projection
 expr_ _ (Identifier id_ (Just (Indexed ((BitVector w),_,1)))) = do
--- a/clash-lib/src/Clash/Primitives/DSL.hs
+++ b/clash-lib/src/Clash/Primitives/DSL.hs
@@ -840,7 +840,7 @@ compInBlock compName inPorts0 outPorts0 =
 
 -- | Convert a 'LitHDL' to a 'TExpr'
 --
--- __N.B.__: Clash 1.6 changed 'instDecl'\'s type signature. Where it would
+-- __N.B.__: Clash 1.8 changed 'instDecl'\'s type signature. Where it would
 --           previously accept 'LitHDL' in its generics/parameters argument, it
 --           now accepts a 'TExpr'. This function is mostly there to ease this
 --           transition.
--- a/clash-cores/clash-cores.cabal
+++ b/clash-cores/clash-cores.cabal
@@ -53,15 +53,12 @@ common basic-config
     MagicHash
     NoImplicitPrelude
     PostfixOperators
+    NoStarIsType
     ScopedTypeVariables
     TemplateHaskell
     TypeApplications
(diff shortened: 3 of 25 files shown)


clash-lib/src/Clash/Backend/SystemVerilog.hs
```diff
@@ -1126,7 +1126,7 @@ expr_ _ (Identifier id_ (Just (Indexed ((Unsigned w),_,_))))  = do
 expr_ _ (Identifier _ (Just (Indexed ((BitVector _),_,0)))) = do
   iw <- Ap $ use intWidth
   traceIf True ($(curLoc) ++ "WARNING: synthesizing bitvector mask to dontcare") $
-    verilogTypeErrValue (Signed iw)
+    verilogTypeErrValue (Unsigned iw)
```
martijnbastiaan, line 1129: The commit message reads:

> BitVector masks were changed to use Natural or unsigned, but these
codepaths weren't changed.

I'm not sure what this means. What PR introduced this?

DigitalBrains1: 505604a33

martijnbastiaan: The _or unsigned_ is really unclear to me. Should it have read

> ... to use `Natural` for `Unsigned` and `BitVector`

?

DigitalBrains1: I don't think our data type `Unsigned` has anything to do with it though? I did fix a few typo's in this commit message, so it might be another one. I also thought it was curiously worded, but I interpreted "Natural or unsigned" as "Natural, which is unsigned rather than signed". I'll shoot Leon a message what he meant.

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -840,7 +840,7 @@ compInBlock compName inPorts0 outPorts0 =
 
 -- | Convert a 'LitHDL' to a 'TExpr'
 --
--- __N.B.__: Clash 1.6 changed 'instDecl'\'s type signature. Where it would
+-- __N.B.__: Clash 1.8 changed 'instDecl'\'s type signature. Where it would
 --           previously accept 'LitHDL' in its generics/parameters argument, it
 --           now accepts a 'TExpr'. This function is mostly there to ease this
 --           transition.
```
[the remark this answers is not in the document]
leonschoorl: Example of what that looks like rendered:
https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Bits.html#v:bitSizeMaybe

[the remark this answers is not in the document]
DigitalBrains1: Hmmm... I'm not incredibly impressed with `@since`. This bit sounds nice:
> `@since` annotation can be used to convey information about when the function was introduced or when it has changed in the way significant to the user.

Right, so two purposes: when it was introduced, or when things were significantly changed. But then it continues:

> While multiple such annotations are not an error, only the one to appear in the comment last will be used.

Well, for when a function was introduced, that's not a problem. I don't think we particularly need that functionality, though. But what about when things were significantly changed? You might end up trying to write:

```
@since 1.8 @foo@ has feature A.

@since 1.12 @foo@ additionally has feature B.
```

_But you can't write that._ By the time you're writing the bit about feature B, you'll have to convert those paragraphs to regular paragraphs or something similar. And if you don't realise it and don't check, you'll end up just with the latter paragraph because it **silently** drops the other one.¹

I'm inclined to think `@since` is mostly useful for indicating when a function was introduced, and not for the other stated purpose.

¹ Might this explain the curious phrasing "changed in _the_ way significant to the user"? It can only document _one_ way, so that will be _the_ way indeed, it's unique. 😜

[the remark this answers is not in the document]
martijnbastiaan: Sounds like we should solely use it for 

>  annotation can be used to convey information about when the function was introduced

then

martijnbastiaan (approved):

Issue #1772, opened 2021-04-19
`bLit` should handle `0x`, `0b`, and `0o` literals


martijnbastiaan:
With `BinaryLiterals` and friends we can write:

```haskell
0xFF
0o345
0b0012
```

but [`bLit`](https://hackage.haskell.org/package/clash-prelude-1.4.1/docs/Clash-Sized-BitVector.html#v:bLit) only supports binary literals. Ideally it would support all of them.

basile-henry:
What about have similar `hexLit`/`hLit` and `octLit`/`oLit` functions instead?

martijnbastiaan:
Good idea. I like the more explicit prefixes too!

leonschoorl:
Slightly related, I've been wondering if it would be more convenient if `bLit` took its length information from the provided string. 

Right now `bLit` needs to get the length information from its context, and it will zero extend or truncate the provided value as needed:
```
clashi> $$(bLit "10")

<interactive>:9:4: error:
    • No instance for (KnownNat n0) arising from a use of ‘bLit’
    • In the expression: bLit "10"
      In the Template Haskell splice $$(bLit "10")
      In the expression: $$(bLit "10")

clashi> $$(bLit "10") :: BitVector 2
10
clashi> $$(bLit "10") :: BitVector 4
0010
clashi> $$(bLit "10") :: BitVector 1
0
```

Instead we make the resulting type depend on the amount of bits provide:
```
clashi> :set +t
clashi> -- ^ this make it print the result type

clashi> $(bLit "10")
10
it :: BitVector 2

clashi> $(bLit "10") :: BitVector 4
  error:
    • Couldn't match type ‘2’ with ‘4’
      Expected: BitVector 4
        Actual: BitVector 2

clashi> $$(bLit "0010")
0010
it :: BitVector 4
```

basile-henry:
I like inferring the type from the String length very much! :smile: It's more concise and probably what a user would expect. We can easily recover the old behavior by wrapping with an explicit `resize` if needed.

martijnbastiaan:
We should absolutely do this! I was very surprised it didn't do this while writing a tutorial the other day.

leonschoorl:
We even have both.
But in that case I'd like to follow `Clash.Sized.Vector`'s pattern of having  `foo` and `fooI`.
Where `fooI` is `foo` except it takes its size information implicitly from context.

christiaanb:
[...]

PR #2492, opened 2023-06-10
Generate correct SDC for Vivado 

NadiaYvette:
Xilinx' Vivado tcl tool rejects some identifiers generated by clash because they contain '$', so this avoids generating such identifiers.


martijnbastiaan:
This only adds a test case, perhaps you forgot a commit?

alex-mckenna:
Actual fix is still WIP, I encouraged a push for visibility

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -401,6 +401,9 @@ data ClashOpts = ClashOpts
   , opt_timescalePrecision :: Period
   -- ^ Timescale precision set in Verilog files. E.g., setting this would sets
   -- the second part of @`timescale 100fs/100fs@.
+  , opt_sanitizeNames :: Bool
+  -- ^ Sanitize ID's for particular tools that may reject
+  -- otherwise valid names.
```
alex-mckenna, line 406: Let's be clearer about this: we expect to sanitise some names anyway since this happens when you don't use `unsafeFromCoreId`. What this flag does is force names to be sanitised in places where they wouldn't normally be.

clash-lib/src/Clash/Netlist/Id.hs
```diff
@@ -242,5 +242,8 @@ fromCoreId = withIdentifierSetM fromCoreId#
 -- spliced at verbatim in the HDL. As opposed to 'fromCoreId', the resulting
 -- Identifier might be generated at a later point as it is NOT added to an
 -- IdentifierSet.
-unsafeFromCoreId :: HasCallStack => Id -> Identifier
-unsafeFromCoreId = unsafeMake . nameOcc . varName
+unsafeFromCoreId :: HasCallStack => SanitizeNames -> Id -> Identifier
+unsafeFromCoreId san = unsafeMake . ckId . nameOcc . varName where
+  ckId s = case san of
+    Sanitize -> Data.Text.map (\c -> if c == '$' then '_' else c) s
+    NoSanitize -> s
```
alex-mckenna, line 249: `ckId` isn't a particularly meaningful name. A better name might be `maybeSanitize`

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -192,6 +192,10 @@ data IdentifierSet
       -- ^ Identifier store
     } deriving (Generic, NFData, Show)
 
+-- This is an option flag to clean up names that may be valid identifiers
+-- but that particular tools e.g. Vivado may reject (tcl with '$' in ids).
+data SanitizeNames = Sanitize | NoSanitize deriving (Eq, Bounded, Read, Show)
```
alex-mckenna, line 197: We don't need all these instances, In fact, I don't think we need any of these instances

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -192,6 +192,10 @@ data IdentifierSet
       -- ^ Identifier store
     } deriving (Generic, NFData, Show)
 
+-- This is an option flag to clean up names that may be valid identifiers
+-- but that particular tools e.g. Vivado may reject (tcl with '$' in ids).
```
alex-mckenna, line 196: This needs to start with `-- |` to be a haddock comment.

clash-lib/src/Clash/Netlist.hs
```diff
@@ -322,6 +323,7 @@ mkNetDecl (id_,tm) = preserveVarEnv $ do
 
   where
     multiDecls pInfo args0 = do
+      san <- asks _sanitizeNames
```
alex-mckenna, line 326: We should be consistent and use `Lens.view sanitizeNames` instead of `asks _sanitizeNames`. I don't particularly like lenses, but this is the style we tend to use here so let's stick with it

tests/Main.hs
```diff
@@ -717,6 +717,7 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "T2342A" def{hdlSim=[]}
         , runTest "T2342B" def{hdlSim=[]}
         , runTest "T2360" def{hdlSim=[],clashFlags=["-fclash-force-undefined=0"]}
+        , outputTest "T2242" def{hdlTargets=[Verilog], clashFlags=["-fclash-sanitise-netlist-ids","-fclash-hdlsyn","vivado"]}
```
alex-mckenna, line 720: The solution we picked didn't end up being vivado-specific, we can remove `-fclash-hdlsyn vivado` from the options

alex-mckenna: We can probably also test all `hdlTargets`, i.e. remove `hdlTargets` from here and just use the default. Since our fix isn't Verilog-specific either we would expect VHDL to also work this way. You would need to add `mainVHDL` and `mainSystemVerilog` to the test

alex-mckenna (commented):
Some initial thoughts, but overall this stayed a pretty small change like I hoped

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
```diff
@@ -95,6 +95,7 @@ flagsClash r = [
   , defFlag "fclash-edalize"                     $ NoArg (liftEwM (setEdalize r))
   , defFlag "fclash-no-render-enums"             $ NoArg (liftEwM (setNoRenderEnums r))
   , defFlag "fclash-timescale-precision"         $ SepArg (setTimescalePrecision r)
+  , defFlag "fclash-sanitise-netlist-ids"        $ NoArg (liftEwM (setSanitizeNames r))
```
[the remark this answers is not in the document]
NadiaYvette: Implemented!

clash-lib/src/Clash/Driver/Types.hs
NadiaYvette: Also done.

clash-lib/src/Clash/Netlist/Id.hs
NadiaYvette: Also done!

clash-lib/src/Clash/Netlist.hs
NadiaYvette: Implemented!

tests/Main.hs
NadiaYvette: A raft of cleanups of the issues you pointed out awaits further review, whenever you might have time. This one in particular

clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
```diff
@@ -95,6 +95,7 @@ flagsClash r = [
   , defFlag "fclash-edalize"                     $ NoArg (liftEwM (setEdalize r))
   , defFlag "fclash-no-render-enums"             $ NoArg (liftEwM (setNoRenderEnums r))
   , defFlag "fclash-timescale-precision"         $ SepArg (setTimescalePrecision r)
+  , defFlag "fclash-sanitize-netlist-ids"        $ NoArg (liftEwM (setSanitizeNames r))
```
alex-mckenna, line 98: New flags should be documented in [this file](https://github.com/clash-lang/clash-compiler/blob/master/docs/developing-hardware/flags.rst). I would make sure we document

  - that some parts of vendor tooling may not accept all valid basic identifiers
  - this flag may result in user-specified names being mangled
  - this flag is a bit of a big hammer and shouldn't be used unless needed

tests/shouldwork/Issues/T2242.hs
```diff
@@ -0,0 +1,16 @@
...
+  Signal System (Unsigned 8) ->
+  Signal System (Unsigned 8)
+topEntity = dflipflop
+{-# NOINLINE topEntity #-}
+
+mainVerilog :: IO ()
+mainVerilog = do
+  [topDir] <- getArgs
+  content <- readFile $ topDir </> show 'topEntity </> "topEntity.v"
+  if '$' `elem` content then error "Error: output contains $ symbol" else pure ()
```
alex-mckenna, line 16: You still need to include `mainVHDL` and `mainSystemVerilog` here, since those functions will be needed to test against other HDLs in the test suite

alex-mckenna (commented):
Another two comments. I think after this we can request a review from someone else to check that our idea / implementation is sensible.

tests/shouldwork/Issues/T2242.hs
NadiaYvette: mainVHDL and mainSystemVerilog added!

NadiaYvette:
Let me know if there are other good issues to work on.

alex-mckenna:
@NadiaYvette I think https://github.com/clash-lang/clash-compiler/issues/1772 is also a pretty good first issue, this one will take you into `clash-prelude` (the standard library)

martijnbastiaan:
Thanks @NadiaYvette for the effort, and thanks @alex-mckenna for the guidance. Super cool to see this coming out of ZuriHac. I've cut my review into two parts:

## In general
I don't think this is the morally correct way of fixing this issue. What Vivado is complaining about is the name in bold (`-name {c$arg}`), not the part in italics (`get_ports {c$arg}`):

> create_clock -name {**c$arg**} -period 5.000 -waveform {0.000 2.500} [get_ports {_c$arg_}]

This means that our HDL could remain as it is, we just need to think of sanitized names when rendering the TCL. E.g., we could emit:

> create_clock -name {c_arg} -period 5.000 -waveform {0.000 2.500} [get_ports {c$arg}]

I think it'd therefore be neater if we'd go this way. The TCL would always work, flag or no flag.

## This approach
Having said the previous - I still think it's worthwhile to have a flag like this. In the RealWorld(tm) we'll have to deal with broken tooling all the time, and Clash should be able to generate non-surprising, very simple identifier names. I do think we should change the flag name to `-fclash-simple-basic-identifiers`. "Sanitizing" doesn't really make sense to me, shouldn't Clash always generate "sane" identifier names? :wink: 

The complexity of this PR can be much reduced if you add the flag here:

https://github.com/clash-lang/clash-compiler/blob/925995c78eb3914e2396ffe1867b018bf85ea1a1/clash-lib/src/Clash/Netlist/Id.hs#L98-L108

The flag should then be used in `make#`:

https://github.com/clash-lang/clash-compiler/blob/925995c78eb3914e2396ffe1867b018bf85ea1a1/clash-lib/src/Clash/Netlist/Id/Internal.hs#L118-L131

There is precedence for a very similar flag, which enables/disables extended identifiers entirely, see `-fclash-no-escaped-identifiers` and the first argument to `makeSet`. Implementing it this way means users of `Netlist.Id` don't have to change.

As for the worries mentioned in the issue associated with this PR:

> Changing the ID generation is indeed a little hairy here. The identifiers which cause problems ultimately come from unsafeFromCoreId, meaning we don't do the sanitising that we get from makeBasic etc. 

This isn't true, all core ids are "sanitized" to valid HDL identifiers in `mkUniqueNormalized` (also see `fromCoreId`). Given that this isn't exactly type directed I'm sure there's bugs.. but that's not for this PR.

Again, thanks for the effort, let us know if you have any questions / need help!

martijnbastiaan (changes requested):
Thanks @NadiaYvette for the effort, and thanks @alex-mckenna for the guidance. Super cool to see this coming out of ZuriHac. I've cut my review into two parts:

## In general
I don't think this is the morally correct way of fixing this issue. What Vivado is complaining about is the name in bold (`-name {c$arg}`), not the part in italics (`get_ports {c$arg}`):

> create_clock -name {**c$arg**} -period 5.000 -waveform {0.000 2.500} [get_ports {_c$arg_}]

This means that our HDL could remain as it is, we just need to think of sanitized names when rendering the TCL. E.g., we could emit:

> create_clock -name {c_arg} -period 5.000 -waveform {0.000 2.500} [get_ports {c$arg}]

I think it'd therefore be neater if we'd go this way. The TCL would always work, flag or no flag.

## This approach
Having said the previous - I still think it's worthwhile to have a flag like this. In the RealWorld(tm) we'll have to deal with broken tooling all the time, and Clash should be able to generate non-surprising, very simple identifier names. I do think we should change the flag name to `-fclash-simple-basic-identifiers`. "Sanitizing" doesn't really make sense to me, shouldn't Clash always generate "sane" identifier names? :wink: 

The complexity of this PR can be much reduced if you add the flag here:

https://github.com/clash-lang/clash-compiler/blob/925995c78eb3914e2396ffe1867b018bf85ea1a1/clash-lib/src/Clash/Netlist/Id.hs#L98-L108

The flag should then be used in `make#`:

https://github.com/clash-lang/clash-compiler/blob/925995c78eb3914e2396ffe1867b018bf85ea1a1/clash-lib/src/Clash/Netlist/Id/Internal.hs#L118-L131

There is precedence for a very similar flag, which enables/disables extended identifiers entirely, see `-fclash-no-escaped-identifiers` and the first argument to `makeSet`. Implementing it this way means users of `Netlist.Id` don't have to change.

As for the worries mentioned in the issue associated with this PR:

> Changing the ID generation is indeed a little hairy here. The identifiers which cause problems ultimately come from unsafeFromCoreId, meaning we don't do the sanitising that we get from makeBasic etc. 

This isn't true, all core ids are "sanitized" to valid HDL identifiers in `mkUniqueNormalized` (also see `fromCoreId`). Given that this isn't exactly type directed I'm sure there's bugs.. but that's not for this PR.

Again, thanks for the effort, let us know if you have any questions / need help!

alex-mckenna:
> I think it'd therefore be neater if we'd go this way. The TCL would always work, flag or no flag.

Right, that makes sense. I had read the issue as it not being fine with the identifier in any position, not just as the argument to `-name`.

> I still think it's worthwhile to have a flag like this.

Yeah, perhaps a flag which limits basic identifiers to just `[A-Za-z0-9_]+` (obviously while still following HDL-specific rules since you can make invalid identifiers already with just that)

> all core ids are "sanitized" to valid HDL identifiers

I still don't particularly trust our name generation, but agreed that's not for this issue.

martijnbastiaan:
> I still don't particularly trust our name generation

I'll say this: if `Clash.Netlist.Id` ends up generating out-of-spec names I'll hand-deliver $favorite_drink to your doorstep<sup>terms and conditions apply</sup>.

Issue #2497, opened 2023-06-14
VIO blackbox should optionally error if it cannot faithfully reproduce user set probe names


martijnbastiaan:
Over at [bittide-hardware](https://github.com/bittide/bittide-hardware/) we use VIOs to probe on-chip tests. For this setup it is crucial that the probe names given in the HDL correspond to the probe names inferred by Vivado. However, the VIO blackboxes currently do not guarantee this:

https://github.com/clash-lang/clash-compiler/blob/b80511bbdd0e6ab03b221c50c46c850b9a2b3d59/clash-cores/src/Clash/Cores/Xilinx/VIO/Internal/BlackBoxes.hs#L169

Ideally, the black box would (optionally) error if `assign` did not yield the name the user asked for.

PR #2499, opened 2023-06-14
Error on unreliable name generation in VIO blackbox

martijnbastiaan:
Fixes https://github.com/clash-lang/clash-compiler/issues/2497

## Still TODO:

  - [x] Cleanup
  - [ ] Optional error
    - [x]  ~~I'll introduce another PR that puts the VIO in its own HDL file which should fix 99% of naming collisions.~~ Impeded: https://github.com/clash-lang/clash-compiler/issues/2500.

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/VIO/Internal/BlackBoxes.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/VIO/Internal/BlackBoxes.hs
@@ -31,7 +31,7 @@ import qualified Data.List.Infinite as Infinite
 import qualified Data.Text as T (pack, concat)
 
 import Control.Arrow (first)
-import Control.Monad (when, forM, zipWithM)
+import Control.Monad (when, zipWithM)
 import Control.Monad.State (State)
 import Control.Exception (assert)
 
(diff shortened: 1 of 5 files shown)


clash-cores/src/Clash/Cores/Xilinx/VIO/Internal/BlackBoxes.hs
```diff
@@ -168,17 +187,17 @@ vioProbeBBTF bbCtx
                 _         -> pure <$> DSL.assign (head inNames) singleInput
             _ -> zipWithM DSL.assign inNames inPs
 
-        outProbes <-
-          forM (zip outNames outTys) $ uncurry DSL.declare
-        outProbesBV <-
-          forM (zip outputNames outProbes) $ uncurry (DSL.toBvWithAttrs attrs)
+        outProbes <- zipWithM DSL.declare outNames outTys
+        outProbesBV0 <- zipWithM (DSL.toBvWithAttrs attrs) userOutputNames outProbes
+        inProbesBV0 <- zipWithM (DSL.toBvWithAttrs attrs) userInputNames inProbes
```
kleinreact, line 192: This introduces a lot of uncessary names. Why not
```haskell
  ...
  inProbesBV <- zipWithM toNameCheckedBv userInputNames inProbes
  ...
where
  toNameCheckedBv x =
    -- The HDL attribute 'KEEP' is added to the signals connected to the
    -- probe ports so they are not optimized away by the synthesis tool.    
    fmap (checkNameCollision x) . DSL.toBvWithAttrs [StringAttr' "KEEP" "true"] x
```
which also avoids the introduction of `attrs`.

tests/Main.hs
```diff
@@ -193,7 +193,19 @@ runClashTest = defaultMain $ clashTestRoot
       , clashTestGroup "Cores"
         [ clashTestGroup "Xilinx"
           [ clashTestGroup "VIO"
-            [ runTest "OutputBusWidthExceeded" def{
+            [ runTest "DuplicateOutputNames" def{
+                hdlTargets=[VHDL]
```
kleinreact, line 197: Any particular reason you just test `VHDL`?

kleinreact (approved):
LGTM. Code still can be cleaned up a bit (see my comments).

clash-cores/src/Clash/Cores/Xilinx/VIO/Internal/BlackBoxes.hs
martijnbastiaan: Good idea!

tests/Main.hs
martijnbastiaan: Yes! The Cores testsuite is getting quite slow because of Vivado. This particular test doesn't care about which HDL it does - name generation / collision detection is the same across all HDLs.

PR #2509, opened 2023-06-17
Combine function names when specializing

### LINKED ISSUE #2508: Clash  doesn’t pick the name of the function being specialized
Christiaan described this as:

> It doesn’t pick the name of the function being specialized when it has a NOINLINE, while the argument being specialized on does not

over at https://github.com/clash-lang/clash-compiler/issues/2502#issuecomment-1592916523. I'm actually unsure why we would _ever_ pick the argument's name instead of the function's name.

martijnbastiaan:
Given a global binder `accum` and an application `f accum`, Clash
now calls the new, specialized binder `f_accum` instead of just
`accum`, provided that both are marked `NOINLINE`/`OPAQUE`. This
more accurately reflects the body of the function and will result
in more sensible file names. For example, previously Clash
would generate a separate file `accum.{v,vhdl}` that contained the
inlined bodies of both `f` and `accum`. After this patch, it will
generate `f_accum.{v,vhdl}`.

Tabulated the new behavior looks like:

| OPAQUE        | Old name for `f g` | New name for `f g` |
|---------------|--------------------|--------------------|
| `f`           | `g`                | `f`                |
| `g`           | `g`                | `g`                |
| `f` and `g`   | `g`                | `f_g`              |
| !`f` and !`g` | `g`                | `f_g`              |

Fixes #2508
## Still TODO:

  - [X] Update comments

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017-2018, Google Inc.,
-                    2021-2023, QBayLogic B.V.
+                    2021-2024, QBayLogic B.V.
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -45,6 +45,7 @@ import qualified Data.Set.Ordered as OSet
 import qualified Data.Set.Ordered.Extra as OSet
 import qualified Data.Text as Text
 import qualified Data.Text.Extra as Text
+import GHC.BasicTypes.Extra (isNoInline)
 import GHC.Stack (HasCallStack)
 
 #if MIN_VERSION_ghc(9,0,0)
@@ -331,10 +332,10 @@ cache. Consider these applications which differ only by ticks:
 
 If one of these had been specialized, the other two would hit that term in the
 specialization cache, saving Clash from having to re-do work which is in effect
-the same. To preserve this behaviour, we use 'stripAllTicks' on the keys for
+the same. To preserve this behavior, we use 'stripAllTicks' on the keys for
 the specialization cache.
 
-TODO While this preserves the old behaviour, the old behaviour is likely not
+TODO While this preserves the old behavior, the old behavior is likely not
 quite what we want. Using a value from the specialization cache may change the
 ticks present, which can affect naming / debugging information in generated HDL.
 We may also not want to look at ticks, as then the specialization cache will
@@ -448,31 +449,42 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
                                        args
               -- Determine name the resulting specialized function, and the
               -- form of the specialized-on argument
-              (fId,inl',specArg') <- case specArg of
+              (newName, inl', specArg') <- case specArg of
                 Left a@(collectArgsTicks -> (Var g,gArgs,_gTicks)) -> if isPolyFun tcm a
                     then do
                       -- In case we are specializing on an argument that is a
-                      -- global function then we use that function's name as the
-                      -- name of the specialized higher-order function.
+                      -- global function then we use that function's name as part
+                      -- of the name of the specialized higher-order function.
                       -- Additionally, we will return the body of the global
                       -- function, instead of a variable reference to the
                       -- global function.
                       --
-                      -- This will turn things like @mealy g k@ into a new
-                      -- binding @g'@ where both the body of @mealy@ and @g@
+                      -- This will turn things like @mealy g@ into a new
+                      -- binding where both the body of @mealy@ and @g@
                       -- are inlined, meaning the state-transition-function
                       -- and the memory element will be in a single function.
+                      --
+                      -- The name of the new function depends on NOINLINE / OPAQUE
+                      -- annotations.
+                      --
+                      --    OPAQUE    | Name for @f g@
+                      --   -----------|---------------
+                      --    f and g   | f_g
+                      --    f         | f
+                      --    g         | g
+                      --    !f and !g | f_g
+                      --
                       gTmM <- fmap (UniqMap.lookup g) $ Lens.use bindings
                       return
-                        ( g
+                        ( specializeName (inl, varName f) (bindingSpec <$> gTmM, varName g)
                         , preferNoInline inl (maybe noUserInline bindingSpec gTmM)
                         , maybe specArg (Left . (`mkApps` gArgs) . bindingTerm) gTmM
                         )
-                    else return (f,inl,specArg)
-                _ -> return (f,inl,specArg)
+                    else return (varName f, inl, specArg)
+                _ -> return (varName f, inl, specArg)
               -- Create specialized functions
               let newBody = mkAbstraction (mkApps bodyTm (argVars ++ [specArg'])) (boundArgs ++ specBndrs)
-              newf <- mkFunction (varName fId) sp inl' newBody
+              newf <- mkFunction newName sp inl' newBody
               -- Remember specialization
               (extra.specialisationHistory) %= UniqMap.insertWith (+) f 1
               (extra.specialisationCache)  %= Map.insert (f,argLen,specAbs) newf
@@ -487,6 +499,19 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
 #else
     noUserInline = NoUserInline
 #endif
+
+    specializeName :: (InlineSpec, Name Term) -> (Maybe InlineSpec, Name Term) -> Name Term
+    specializeName (spec0, n0) (spec1, n1)
+      | (ann0 && ann1) || (not ann0 && not ann1)
+      = n0{nameOcc=nameOcc n0 <> "_" <> stripMods (nameOcc n1)}
+      | ann0 = n0
+      | otherwise = n1
+     where
+      ann0 = isNoInline spec0
+      ann1 = fmap isNoInline spec1 == Just True
+
+    stripMods :: Text.Text -> Text.Text
+    stripMods = Text.takeWhileEnd (/= '.')
 
     collectBndrsMinusApps :: Term -> [Name a]
     collectBndrsMinusApps = reverse . go []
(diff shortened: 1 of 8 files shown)


tests/shouldwork/Issues/T2508.hs
```diff
@@ -0,0 +1,46 @@
...
+{-# OPAQUE topEntity #-}
+
+mainVHDL :: IO ()
+mainVHDL = do
+  [topDir] <- getArgs
+
+  -- 'bbWrapper' should get its own file, so we expect two: one for 'topEntity',
+  -- one for 'bbWrapper'.
+  --
+  -- XXX: Naming doesn't make sense. 'topEntity1' should be called 'bbWrapper'.
```
martijnbastiaan, line 36: copy+paste error, should be remove before merging

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -486,6 +487,17 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
     noUserInline = NoUserInline
 #endif
 
+    specializeName :: Name Term -> Name Term -> Name Term
+    specializeName n0 n1 =
+      n1{nameOcc=stripMods (nameOcc n0) <> "_" <> stripMods (nameOcc n1)}
```
martijnbastiaan, line 492: This throws away provenance info of the first name. While this is not a regression over the status-quo, it's probably still good to add a comment. Perhaps we want to store a tree of provenance info in `Name`s in the future.

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -486,6 +487,17 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
     noUserInline = NoUserInline
 #endif
 
+    specializeName :: Name Term -> Name Term -> Name Term
```
christiaanb, line 490: If only one of the binders has a NOINLINE/OPAGUE, we should only pick that name, and not combine it with the name that didn't have a NOINLINE/OPAGUE.

Only function names with a NOINLINE/OPAGUE should be preserved.

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -486,6 +487,17 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
...
+    specializeName n0 n1 =
+      n1{nameOcc=stripMods (nameOcc n0) <> "_" <> stripMods (nameOcc n1)}
+
+    stripMods :: Text.Text -> Text.Text
+    stripMods t = fromMaybe t (lastMaybe (Text.splitOn "." t))
```
christiaanb, line 495:
```suggestion
    stripMods = Text.takeWhileEnd (/= '.')
```

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
christiaanb: I think we should remember the provonance of at least one of the names. To make things simpler, that would be the provenance of the function being specialized.

christiaanb (dismissed):

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
martijnbastiaan: Would you agree that in the future it could be desirable to remember both names? (Not in this PR though.)

martijnbastiaan:
@christiaanb Would you mind reviewing the updates I pushed? See the updated cover letter for the new behavior. Hopefully you'll agree that the test cases I had to update due to name changes now make more sense. I do think the current behavior is still sub-optimal, would you entertain:

| OPAQUE        | Old name for `f g` | New name for `f g` | IMO desired name for  `f g` |
|---------------|--------------------|--------------------|-----------------------------|
| `f`           | `g`                | `f`                | `f`                         |
| `g`           | `g`                | `g`                | `g` (unenthusiastically, but understandable)    |
| `f` and `g`   | `g`                | `f_g`              | `f_g`                       |
| !`f` and !`g` | `g`                | `g`                | `f_g` or `f`                |

The problems I have with preferring the _argument_'s name is that it feel conceptually wrong to me. For two reasons:

1. I feel that it is wrong to pick `b` in `f a b`. I.e., prefer one argument over another.
2. I feel that the resulting binder `f a` _mainly_ has `f`'s form, with `a` inside it. E.g., for `mealy accum`, the main structure is still a mealy machine, not `accum`.

This is why I'd propose calling `f g`, `f_g` if neither binder has a `NOINLINE` pragma.

martijnbastiaan:
Either way, I think the current PR is an improvement over the status-quo, so I'd be happy merging it as-is.

christiaanb:
Now that we have `PrefixName` ticks, I'd propose the following:

| OPAQUE | Old name for `f g` | New name for `f g` |
|-|-|-|
| `f`  | `g` | `f`, and add a `PrefixName g` tick to the specialised argument |
| `g` | `g` | `g`, and add a `PrefixName f` tick to the application of the specialised function |
| `f` and `g`   | `g`  | `f_g` |
| !`f` and !`g` | `g`  | `f_g` |

That way we can still keep track of the "lost" names.

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -488,6 +500,18 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
...
+    specializeName (spec0, n0) (Just spec1, n1)
+      | isNoInline spec0
+      , isNoInline spec1
+      = n0{nameOcc=nameOcc n0 <> "_" <> stripMods (nameOcc n1)}
+      | isNoInline spec1
+      = n1
+    specializeName (_, n0) _ = n0
+
+    stripMods :: Text.Text -> Text.Text
+    stripMods = Text.takeWhileEnd (/= '.')
```
christiaanb, line 513: This doesn't work when e.g. have `Data.Bits..&.`

christiaanb: Perhaps we should just add another issue, and leave this as is for now. We might want to refactor names so module name are stored separately.

leonschoorl:
Why does this depend on OPAQUE/NOINLINE (directly)?

I think it makes sense to always use the form `f_g`, if both `f` and `g` come from source code and aren't made up by clash.
It makes sense to for the specialization of `map myFunction` to be called `map_myFunction` if the `myFunction` name is available to the transformation. Even without me having to annotate myFunction.

Detection of should be possible via the [`NameSort`](https://hackage.haskell.org/package/clash-lib-1.8.1/docs/Clash-Core-Name.html#t:NameSort) inside of [`Name`](https://hackage.haskell.org/package/clash-lib-1.8.1/docs/Clash-Core-Name.html#t:Name) I assume.

Then OPAQUE/NOINLINE could be used to have the indirect effect of guaranteeing that such a name is preserved by GHC.

christiaanb:
The only name that is preserved are the ones with a NOINLINE/OPAQUE annotation; that's the "rule". Why would H/O functions be the exception to this? Why are H/O functions prone to name-mangling?

martijnbastiaan:
What I think @leonschoorl is saying: if there _is_ a name available (whether the function was annotated with NOINLINE or not), why not use it?

christiaanb:
If you have a `NOINLINE` on `functionWeCareAbout`, and not on `configuration`, then I do not think that `functionWeCareAbout configuration` should result in a module called `functionWeCareAbout_configuration`.

If you do care that modules corresponding to h/o-function take on the name of their function argument, you can add a `NOINLINE` to the top-level binder that you'll use in the argument position.

clash-lib/src/Clash/Normalize/Transformations/Specialize.hs
```diff
@@ -488,6 +500,18 @@ specialize' (TransformContext is0 _) e (Var f, args, ticks) specArgIn = do
...
 #endif
 
+    specializeName :: (InlineSpec, Name Term) -> (Maybe InlineSpec, Name Term) -> Name Term
+    specializeName (spec0, n0) (Just spec1, n1)
+      | isNoInline spec0
+      , isNoInline spec1
+      = n0{nameOcc=nameOcc n0 <> "_" <> stripMods (nameOcc n1)}
+      | isNoInline spec1
+      = n1
+    specializeName (_, n0) _ = n0
```
christiaanb, line 510: This implement:

| OPAQUE | Name for `f g` |
|-|-|
| f and g | `f_g` |
| otherwise | g |

And thus does not implement what's in the comments at https://github.com/clash-lang/clash-compiler/pull/2509/files#diff-38424dec2448cd00ad74cbd954d63cd04bc9d70ae89eaf7eecec8087669d2a33R470-R475

christiaanb: In addition, I thought we all agreed that we would like to have:

| OPAQUE | Name for `f g` |
| - | - |
| f and g | `f_g` |
| f | `f` |
| g | `g` | 
| otherwise | `f_g` |

martijnbastiaan: We've got OPAQUE f and g:

```
      | isNoInline spec0
      , isNoInline spec1
      = n0{nameOcc=nameOcc n0 <> "_" <> stripMods (nameOcc n1)}
```

resulting in `f_g`. We've got just OPAQUE `g`, which does:

```
      | isNoInline spec1
      = n1
```

I.e., just `g`. Then for everything else we pick `f`:

```
    specializeName (_, n0) _ = n0
```


This wasn't my intention though, I wanted to write:

```
      | isNoInline spec0
      = n0
    specializeName (_, n1) _ = n1
```

..to match the comment.

martijnbastiaan: > In addition, I thought we all agreed that we would like to have:

Yes, though this code predates that discussion :-)

martijnbastiaan:
I've updated the cover letter's table and the implementation to reflect what we discussed in the office.

martijnbastiaan:
> I've updated the cover letter's table and the implementation to reflect what we discussed in the office.

Happy with this PR now @christiaanb? :)

christiaanb (approved):
LGTM. Don't forget to raise the issue w.r.t. `stripMods`'s implementation being "bad".

PR #2511, opened 2023-06-17
Warn against non-opaque primitive functions on GHC >= 9.4

### LINKED ISSUE #2510: Clash should warn against just using `NOINLINE` for primitives on GHC >= 9.4
The following:

```haskell
import Data.String.Interpolate (__i)
import Clash.Explicit.Prelude
import Clash.Annotations.Primitive

class X a where
  x :: a

instance X (Signal System Int) where
  x = pure 3

instance X a => X (Signal System Int -> a) where
  x !_i = x

bb :: X a => a
bb = x
{-# NOINLINE bb #-}
{-# ANN bb hasBlackBox #-}
{-# ANN bb (
  let bbName = show 'bb
  in InlineYamlPrimitive [minBound..] [__i|
    BlackBox:
      name: "#{bbName}"
      kind: Expression
      template: "this does not end up in HDL with NOINLINE~ARG[1]"
|]) #-}

bbWrapper :: X a => a
bbWrapper = bb
{-# NOINLINE bbWrapper #-}

topEntity :: Signal System Int -> Signal System Int
topEntity i = bbWrapper i
```

..does not end up using the blackbox of `bb`, but instead uses the core unfolding. Clash point this out and encourage use of `OPAQUE`.

martijnbastiaan:
In order for Clash to work properly, some names need to remain untouched by GHC's compilation process. Examples of these names include top entities annotated with a `Synthesize` pragma, or functions having an associated black box implementation. For the longest time, we have tried to instruct GHC to do so by marking functions as "no inline":

```haskell
myFunction :: Unsigned 8 -> Unsigned 8
myFunction = (+1)
{-# NOINLINE myFunction #-}
```

While this does work in the majority of cases, GHC will still touch the name of this function in some, making Clash fail to recognize it. Hence, GHC 9.4 introduce a new pragma `OPAQUE`:

```haskell
myFunction :: Unsigned 8 -> Unsigned 8
myFunction = (+1)
{-# OPAQUE myFunction #-}
```

This PR introduces a change making Clash warn against usage of `NOINLINE` on GHCs 9.4 and newer. To remain compatible with older GHCs, Clash's code base itself defines a CPP macro that defines `CLASH_OPAQUE` as either `NOINLINE` or `OPAQUE` depending on the version of GHC the code base is compiled with. This means that you'll find the following pattern across it:

```haskell
myFunction :: Unsigned 8 -> Unsigned 8
myFunction = (+1)
{-# CLASH_OPAQUE myFunction #-} -- Expands to NOINLINE < 9.4
                                -- Expands to OPAQUE >= 9.4
```

This macro is local to Clash itself, and is not reusable within other projects. If your project suffers from similar cross-compatibility problems, consider using a similar strategy.

-----------------------------

Warn against non-opaque primitive functions on GHC >= 9.4.

Fixes #2510

## Still TODO:

  - [X] ~~Write a changelog entry (see changelog/README.md)~~ 9.4 / 9.6 support is not in a Hackage release yet
  - [x] Go through all primitives and mark them `OPAQUE` with CPP :cry:. Note that we also want to mark top entities as `OPAQUE` as GHC can touch them too in some circumstances.
  - [x] Add example from issue #2510 to CI
  - [x] Extend PR cover letter in such a way that it is useful for people finding the link now scattered throughout the code base.

### DIFF
--- a/benchmark/tests/ManyEntitiesEqual.hs
+++ b/benchmark/tests/ManyEntitiesEqual.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE RankNTypes #-}
 
 module ManyEntitiesEqual where
@@ -34,42 +35,50 @@ type Top n =
     => Signal System (Unsigned n)
     -> Signal System (Unsigned n)
 
-{-# NOINLINE top0 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top0 #-}
 {-# ANN top0 (defSyn "top_0") #-}
 top0 :: Top 64
 top0 = entity (SNat @64)
 
-{-# NOINLINE top1 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top1 #-}
 {-# ANN top1 (defSyn "top_1") #-}
 top1 :: Top 64
 top1 = entity (SNat @64)
 
-{-# NOINLINE top2 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top2 #-}
 {-# ANN top2 (defSyn "top_2") #-}
 top2 :: Top 64
 top2 = entity (SNat @64)
 
-{-# NOINLINE top3 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top3 #-}
 {-# ANN top3 (defSyn "top_3") #-}
 top3 :: Top 64
 top3 = entity (SNat @64)
 
-{-# NOINLINE top4 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top4 #-}
 {-# ANN top4 (defSyn "top_4") #-}
 top4 :: Top 64
 top4 = entity (SNat @64)
 
-{-# NOINLINE top5 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top5 #-}
 {-# ANN top5 (defSyn "top_5") #-}
 top5 :: Top 64
 top5 = entity (SNat @64)
 
-{-# NOINLINE top6 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top6 #-}
 {-# ANN top6 (defSyn "top_6") #-}
 top6 :: Top 64
 top6 = entity (SNat @64)
 
-{-# NOINLINE top7 #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE top7 #-}
 {-# ANN top7 (defSyn "top_7") #-}
 top7 :: Top 64
 top7 = entity (SNat @64)
--- a/clash-cores/src/Clash/Cores/Xilinx/BlockRam.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/BlockRam.hs
@@ -6,6 +6,8 @@
   Xilinx block RAM primitives
 -}
 
+{-# LANGUAGE CPP #-}
+
 -- See [Note: eta port names for tdpbram]
 {-# OPTIONS_GHC -fno-do-lambda-eta-expansion #-}
 
@@ -76,4 +78,5 @@ tdpbram clkA enA addrA byteEnaA datA clkB enB addrB byteEnaB datB =
       clashCompileError "tdpbram: domain A needs a rising active edge"
     (_, SFalling) ->
       clashCompileError "tdpbram: domain B needs a rising active edge"
-{-# NOINLINE tdpbram #-}
+-- See: https://github.com/clash-lang/clash-compiler/pull/2511
+{-# CLASH_OPAQUE tdpbram #-}
--- a/clash-prelude/src/Clash/Magic.hs
+++ b/clash-prelude/src/Clash/Magic.hs
@@ -16,6 +16,8 @@ Refer to "Clash.Annotations.TopEntity" for controlling naming of entities
 (VHDL) / modules ((System)Verilog) and their ports.
 -}
 
+{-# LANGUAGE CPP #-}
+
 module Clash.Magic
   (
   -- ** Functions to control names of identifiers in HDL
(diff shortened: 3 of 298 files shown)


benchmark/tests/ManyEntitiesEqual.hs
```diff
@@ -34,42 +34,74 @@ type Top n =
     => Signal System (Unsigned n)
     -> Signal System (Unsigned n)
 
+#if __GLASGOW_HASKELL__ >= 904
```
rowanG077, line 37: Why not make an intermediate macro.

```C
#if __GLASGOW_HASKELL__ >= 904
#define CLASH_NOINLINE OPAQUE
#else
#define CLASH_NOINLINE NOINLINE
#endif
```

martijnbastiaan: That does make it even more work, I'd have to add `#include blaat.h` everywhere and I can't do that with a regex. But.. it would be quite a bit neater so I'll try.

alex-mckenna: Minor bikeshedding: I think calling it `CLASH_NOINLINE` will be seen by some users and confuse them, i.e. _What is the different between `NOINLINE` and `CLASH_NOINLINE`?_ I think `CLASH_PRIMITIVE` would be a better name --- primitives are the only things which should be so aggressively left alone AFAICT

leonschoorl:
Now that we're properly marking all primitives as OPAQUE, we could try re-enabling the worker/wrapper transform on clash-prelude when compiling with GHC >= 9.4.

benchmark/tests/ManyEntitiesEqual.hs
martijnbastiaan: Primitives and things with synthesize annotations.

tests/src/Test/Tasty/Clash.hs
```diff
@@ -241,6 +241,10 @@ instance IsTest ClashGenTest where
       , "-odir", oDir
       , "-hidir", oDir
       , "-fclash-debug", "DebugSilent"
+
+      -- PR #2511 introduced a lot of CPP in shouldwork/*. To prevent very tedious
+      -- work, we enable CPP by default.
```
leonschoorl, line 246: How ironic that this comment about tedious CPP work now generates a CPP error

benchmark/tests/ManyEntitiesEqual.hs
rowanG077: Malicious mode: 

```
  if impl(ghc >= 9.4)
    CPP-Options: -DNOINLINE=OPAQUE
  else
    CPP-Options: -DNOINLINE=NOINLINE
```

martijnbastiaan: I _love_ it

tests/src/Test/Tasty/Clash.hs
martijnbastiaan: That's hilarious haha

benchmark/tests/ManyEntitiesEqual.hs
martijnbastiaan: I've picked a slightly different name than suggested here: `CLASH_OPAQUE`. I've also added a comment referring people to this PR to prevent confusion. The other suggestions:

* `CLASH_NOINLINE`: Uses "old" terminology, I'd prefer to use the new one.
* `CLASH_PRIMITIVE`: It's not only primitives, also top entities. Basically anything that has some annotation attached to it Clash should later pick up on.
* `NOINLINE`: Love it but no :D 

I'll unresolve the conversation so it is easier to spot for reviewers.

martijnbastiaan:
This PR has been scripted for 99%. Here are the steps:

1. In VSCode search for `^\{-# NOINLINE (.*) #-\}`, replace with `-- See: https://github.com/clash-lang/clash-compiler/pull/2511\n{-# CLASH_OPAQUE $1 #-}` in `clash-prelude,clash-cores,benchmark,tests/shouldwork,tests/shouldfail,examples`.
2. Create a file `need_cpp` using `git status --short | grep -E '^ M ' | grep -E '.hs$' | sed 's/^ M //g' > need_cpp`
3. Add `CPP` pragmas everywhere by running [🍝 `python3 add_cpp.py` 🍝](https://gist.github.com/martijnbastiaan/fbe6221ead4afbe877db217c51abb185).

clash-cores/src/Clash/Cores/Xilinx/BlockRam.hs
```diff
@@ -7,6 +7,8 @@
 -}
 
 -- See [Note: eta port names for tdpbram]
+{-# LANGUAGE CPP #-}
+
 {-# OPTIONS_GHC -fno-do-lambda-eta-expansion #-}
```
DigitalBrains1, line 12: This one is a bit unfortunate...

martijnbastiaan: Ah.. I'll change it before merging!

benchmark/tests/ManyEntitiesEqual.hs
rowanG077: I'm happy with that! `CLASH_INLINE` was just something I thought at in the second I posted the comment.

clash-prelude/src/Clash/Magic.hs
```diff
@@ -16,6 +16,9 @@ Refer to "Clash.Annotations.TopEntity" for controlling naming of entities
 (VHDL) / modules ((System)Verilog) and their ports.
 -}
 
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP #-}
```
christiaanb, line 20: CPP appears double

christiaanb (approved):
There's two more places where we need to replace NOINLINE by OPAQUE for GHC 9.4+:

* https://github.com/clash-lang/clash-compiler/blob/48f7f52feab5aa1619b6945d6f1dbbb9a7ec445d/clash-cosim/src/Clash/CoSim/CodeGeneration.hs#L82-L83
* https://github.com/clash-lang/clash-compiler/blob/48f7f52feab5aa1619b6945d6f1dbbb9a7ec445d/clash-ghc/src-ghc/Clash/GHC/GenerateBindings.hs#L214

Otherwise LGTM

clash-cores/src/Clash/Cores/Xilinx/BlockRam.hs
DigitalBrains1: I've tested the code base for all combinations of a single-line Haskell comment immediately preceding the `LANGUAGE CPP` line and this is the only culprit. The other comments pertain to the whole file, and are thus in their right place.

I had to brush up on my `awk` skills (you could say I had become a bit awkward at it?), but this is the script and its output:
<details>

```
clash-compiler$ find * -type f -print0 | xargs -0 gawk '/^--/{found=1; comment=$0; next} found && /^{-# LANGUAGE CPP #-}/{ print FILENAME; print comment ; print ; print "" } { found=0 }'
clash-cores/src/Clash/Cores/Xilinx/BlockRam.hs
-- See [Note: eta port names for tdpbram]
{-# LANGUAGE CPP #-}

tests/shouldwork/Basic/Replace.hs
-- See: https://github.com/clash-lang/clash-compiler/issues/365
{-# LANGUAGE CPP #-}

tests/shouldwork/Vector/FirOddSize.hs
-- see https://github.com/clash-lang/clash-compiler/issues/383
{-# LANGUAGE CPP #-}

tests/shouldwork/Vector/IndexInt2.hs
-- which generates special verilog when the index is constant.
{-# LANGUAGE CPP #-}

tests/shouldwork/Numbers/DivideByZero.hs
-- Test that the compiler doesn't crash on any DivideByZero exceptions
{-# LANGUAGE CPP #-}
```
</details>

PR #2513, opened 2023-06-17
Update compatibility matrix in README

### LINKED ISSUE #2494: Clarification on supported GHC versions
Which versions exactly are supported at the moment? The table for GHC compatibility in the README is outdated and there are also some related questions about CI.

Current table:

> |      | Linux | Windows | macOS |
> |------|-------|---------|-------|
> | 8.6  | ✔️     | ✔️      | ✔️     |
> | 8.8  | ✔️     | ❌      | ✔️     |
> | 8.10 | ✔️     | ✔️      | ❌     |
> | 9.0  | ✔️     | ✔️²     | ✔️     |
> | 9.2¹ | ❌     | ❌      | ❌     |
> | 9.4³ | ❌     | ❌      | ❌     |
>
>
> ¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change will be reverted in 9.4.
>
> ² GHC 9.0.2 on Windows fails to compile `clash-cores`. We therefore don't run the Clash testsuite on CI for this combination.
>
> ³ A critical dependency of Clash, `hint`, has not yet been ported to GHC 9.4.

### GHC 9.2

Now seems to be supported, as it's tested in CI for all three platforms. So ❌ is outdated. However, the footnote around the regression for error messages still applies, so maybe it should not be ✔️, either?

### GHC 9.4

Seems to be supported, so probably should be ✔️. Has the issue with `hint` been resolved? If so, footnote 3 can be removed.

Surprisingly, CI only runs the Windows and macOS tests (added in https://github.com/clash-lang/clash-compiler/pull/2367/commits/4fa47b41f0319e34f21f95ec315f9917789dcc3d). Is there any reason for not running the Linux tests or has it just been forgotten?

### GHC 9.6

~~Not run in CI yet, but if I remember correctly, it at least works for @alex-mckenna on Linux (via the Nix flake). Does this already warrant a ✔️ on Linux? Can we add tests for 9.6 to CI?~~
I just saw that GHC 9.6 support has been added, but [the GitHub Action depends on a Stackage nightly for GHC 9.6](https://github.com/clash-lang/clash-compiler/pull/2467/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7), which still doesn't exist. Not sure how the GitLab CI is run, but seems like a ✔️ for Linux is justified? 

Thanks!

martijnbastiaan:
Fixes #2494

## Still TODO:

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,20 +63,21 @@ git checkout 1.2
 Note that release branches might contain non-released patches.
 
 ## GHC compatibility
-|      | Linux | Windows | macOS |
-|------|-------|---------|-------|
-| 8.6  | ✔️     | ✔️      | ✔️     |
-| 8.8  | ✔️     | ❌      | ✔️     |
-| 8.10 | ✔️     | ✔️      | ❌     |
-| 9.0  | ✔️     | ✔️²     | ✔️     |
-| 9.2¹ | ❌     | ❌      | ❌     |
-| 9.4³ | ❌     | ❌      | ❌     |
+|      | Linux | Windows | macOS | Clash (released) | Clash (development version)
+|------|-------|---------|-------|------------------|--------------------------
+| 8.6  | ✔️     | ✔️       | ✔️     | 1.0 - 1.6        | ✔️
+| 8.8  | ✔️     | ❌      | ✔️     | 1.0 - 1.6        | ✔️
+| 8.10 | ✔️     | ✔️       | ❌    | 1.2 - 1.6        | ✔️
+| 9.0  | ✔️     | ✔️²      | ✔️     | 1.4 - 1.6        | ✔️
+| 9.2  | ⚠️¹    | ⚠️¹      | ⚠️¹    |                  | ⚠️¹️
+| 9.4  | ✔️     | ✔️       | ✔️     |                  | ✔️
+| 9.6  | ✔️     | ❔³     | ❔³   |                  | ✔️
 
-¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change will be reverted in 9.4.
+¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change was reverted in 9.4.
 
 ² GHC 9.0.2 on Windows fails to compile `clash-cores`. We therefore don't run the Clash testsuite on CI for this combination.
 
-³ A critical dependency of Clash, `hint`, has not yet been ported to GHC 9.4.
+³ We use Stack to test on Windows/macOS. There is no resolver yet that uses GHC 9.6, so this remains untested.
 
 ## Cabal
 To use Cabal you need both Cabal and GHC installed on your system. We recommend using [ghcup](https://www.haskell.org/ghcup/). For more information, see [https://www.haskell.org/downloads/](https://www.haskell.org/downloads/).


alex-mckenna (approved):
Looks good, but since `hint` now has a release in hackage that supports GHC 9.6, maybe we should get the hint PR in, cut a release and then we can remove the "(unreleased)" from here?

Whatever order you want to do things is fine though

martijnbastiaan:
Yeah that seems sensible. We're not going to backport the GHC upgrades though, so we'll need to make a major release. Which will take some time..

DigitalBrains1:
Did you mean
* Clash 1.6
* Clash 1.8 (unreleased)

or is the 1.4 deliberate (after all, 1.4 did support those GHC's, but it also supported GHC 8.4) and only the 1.6 a mistake?

DigitalBrains1 (changes requested):
Per comment above

martijnbastiaan:
> Did you mean

Yes!

martijnbastiaan:
I've also indicated that we don't test 9.6 for Windows/macOS yet. @DigitalBrains1 what do you think?

DigitalBrains1:
We don't actually support 1.4 anymore, right? If we go down that road, we could also list 1.2 and 1.0. I think it's better to just list supported versions.

Alternatively, you could differentiate between "released" or "stable" or whatnot, and "unreleased" or "`master` only"...

It's definitely a good idea to indicate when we actually don't test stuff like with 9.6 and Windows and macOS.

martijnbastiaan:
Like this?

README.md
```diff
@@ -63,20 +63,21 @@ git checkout 1.2
...
+|------|-------|---------|-------|------------------|--------------------------
+| 8.6  | ✔️     | ✔️       | ✔️     | 1.6              | ✔️
+| 8.8  | ✔️     | ❌      | ✔️     | 1.6              | ✔️
+| 8.10 | ✔️     | ✔️       | ❌    | 1.6              | ✔️
+| 9.0  | ✔️     | ✔️²      | ✔️     | 1.6              | ✔️
+| 9.2  | ⚠️¹    | ⚠️¹      | ⚠️¹    |                  | ⚠️¹️
+| 9.4  | ✔️     | ✔️       | ✔️     |                  | ✔️
+| 9.6  | ✔️     | ❔³     | ❔³   |                  | ✔️
 
 ¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change will be reverted in 9.4.
```
DigitalBrains1, line 76: How about this? (I don't have a clue about the correct number of spaces in the table. I know it renders the same regardless, but I try to keep it aligned on a fixed width font. However, the overly wide Unicode chars mean that my editor is no longer fixed width...)

```suggestion
|      | Linux | Windows | macOS | Clash (released) | Clash (development version)
|------|-------|---------|-------|------------------|--------------------------
| 8.6  | ✔️     | ✔️       | ✔️     | ✔️              | ✔️
| 8.8  | ✔️     | ❌      | ✔️     | ✔️              | ✔️
| 8.10 | ✔️     | ✔️       | ❌    | ✔️              | ✔️
| 9.0  | ✔️     | ✔️²      | ✔️     | ✔️              | ✔️
| 9.2  | ⚠️¹    | ⚠️¹      | ⚠️¹    |                  | ⚠️¹️
| 9.4  | ✔️     | ✔️       | ✔️     |                  | ✔️
| 9.6  | ✔️     | ❔³     | ❔³   |                  | ✔️

¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change was reverted in 9.4.
```

DigitalBrains1: Right, I could get it aligned by temporarily replacing all wide Unicode chars by some normal placeholder, then align the table, and then put the chars back. Shall I do this or don't you care?

martijnbastiaan: I'm strongly in favor of mentioning specific version numbers. It should be easy to see an upgrade path if you're trying to port some old codebase over. Also, you might be reading an out-of-date README (obviously that's not ideal, but mentioning version numbers would at least make it less misleading). I would also be in favor of adding a column `Clash (EOL)`, but we don't have to do that now. Basically I often look up [this table](https://www.snoyman.com/base/) even though almost all versions mentioned there are EOL.

On the space alignment: go ahead! I'm also happy with the current table, so feel free to fix & merge.

DigitalBrains1: So why are you limiting it to Clash 1.6? Older versions support these GHC's....

README.md
```diff
@@ -63,20 +63,21 @@ git checkout 1.2
...
-| 8.10 | ✔️     | ✔️      | ❌     |
-| 9.0  | ✔️     | ✔️²     | ✔️     |
-| 9.2¹ | ❌     | ❌      | ❌     |
-| 9.4³ | ❌     | ❌      | ❌     |
+|      | Linux | Windows | macOS | Clash (released) | Clash (development version)
+|------|-------|---------|-------|------------------|--------------------------
+| 8.6  | ✔️     | ✔️       | ✔️     | 1.6              | ✔️
+| 8.8  | ✔️     | ❌      | ✔️     | 1.6              | ✔️
+| 8.10 | ✔️     | ✔️       | ❌    | 1.6              | ✔️
+| 9.0  | ✔️     | ✔️²      | ✔️     | 1.6              | ✔️
```
DigitalBrains1, line 71:
```suggestion
| 8.6  | ✔️     | ✔️       | ✔️     | 1.0              | ✔️
| 8.8  | ✔️     | ❌      | ✔️     | 1.0              | ✔️
| 8.10 | ✔️     | ✔️       | ❌    | 1.2              | ✔️
| 9.0  | ✔️     | ✔️²      | ✔️     | 1.4              | ✔️
```

README.md
DigitalBrains1, line 76:
```suggestion
¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change was reverted in 9.4.
```

martijnbastiaan: Haven't you learned over the years that the answer when talking to me is usually laziness? :wink:

README.md
DigitalBrains1: I've updated the suggestion to include the range as we discussed. Is this the version you'd like to see?

martijnbastiaan: Yes looking good

DigitalBrains1 (approved):

PR #2529, opened 2023-07-05
New: `DiffClock`

### LINKED ISSUE #2518: Incorrect handling of diffferential clock pairs
Currently we use two separate `Clock dom` arguments to represent differential clock inputs (see [clockWizardDifferential](https://hackage.haskell.org/package/clash-prelude-1.6.4/docs/Clash-Xilinx-ClockGen.html#v:clockWizardDifferential))
However, this leads to Clash generating two `create_clock` constraints for this clock signal, which is incorrect.

It will at least mess up timing for Vivado.

We should  either find a way to detect this situation and only generate a single constraint on the `P`channel of the pair, or make a separate datatype (maybe `DiffClock dom`) that represents a differential clock pair.

DigitalBrains1:
A new type is introduced to signify a differential clock signal that is passed to the design on two ports in antiphase. This is used by Xilinx `clockWizardDifferential` (note that Intel solves this case differently, so if we were to add an Intel primitive it would not need this type).

With this new type, we prevent the creation of more `create_clock` statements in the global SDC file than is correct: only the positive phase of a differential pair should have a `create_clock` statement.

Fixes #2518

## Still TODO:


leonschoorl (approved):
LGTM
:clock6: - :clock1230:

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -16,6 +16,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE QuasiQuotes #-}
```
christiaanb, line 19: Why? There are no other changes in this file that require it.

clash-lib/src/Clash/Core/Util.hs
```diff
@@ -618,9 +619,9 @@ shouldSplit0 tcm (TyConApp tcNm tyArgs)
     && nameOcc a2Nm == "Clash.Signal.Internal.KnownDomain"
   isHidden _ _ = False
 
-  -- Currently we're only interested in splitting of Clock, Reset, and Enable
```
DigitalBrains1, line 621: I removed this comment as it appeared stale to me: the list includes file handles as well. I don't think we need to update the comment, which things are split off by `splitTy` is pretty apparent from the code.

tests/shouldwork/Xilinx/ClockWizard.hs
```diff
@@ -19,15 +19,15 @@ createDomain vXilinxSystem{vName="DomOut", vPeriod=hzToPeriod 300_000_000}
 
 topEntity ::
   Clock DomIn ->
-  Clock DomIn ->
+  DiffClock DomIn ->
```
martijnbastiaan, line 22: I don't understand this change, how can a _single_ clock change to a `DiffClock`?

martijnbastiaan (commented):
In general LGTM. Two things:

* I'm not very fond of the name `seClockToDiffClock`. Why not just `clockToDiffClock`? If that's not part of a release yet I'd prefer to change it. But then again, not really for this PR I suppose.
* Next Monday I'm at our ASIC client again, I'll investigate how they handle diff clocks (if applicable). I'll change my review to :heavy_check_mark: if my insight there doesn't change my view on this PR.

tests/shouldwork/Xilinx/ClockWizard.hs
DigitalBrains1: This test isn't really suitable to be synthesised, I'm pretty cavalier with the domains and `unsafeSynchronizer`. The first input was the non-inverted clock and the second input the inverted clock, and the non-inverted clock was fed to a clock wizard with a single-ended input as well. This test tests that both `clockWizard` and `clockWizardDifferential` assert their locked output and make actual progress. So with the new setup, we need a single-ended clock and a differential clock, where previously we could just feed the non-inverted signal to both.

DigitalBrains1:
> I'm not very fond of the name `seClockToDiffClock`. Why not just `clockToDiffClock`? If that's not part of a release yet I'd prefer to change it. But then again, not really for this PR I suppose.

Now that we have two types `Clock` and `DiffClock` I can see how that makes sense. Before, it did not make sense. It would be saying you're turning a clock into a clock. You'd have needed to say what you're turning into what for it not to be nonsensical.

But now that the name can also refer to the type names instead of the concepts represented, I'm fine with changing it to that.

martijnbastiaan (approved):

PR #2537, opened 2023-07-12
Fix argument parsing in `clashCompileError`

DigitalBrains1:
Even though it has type `String -> a` and thus only a single argument, when `a` is itself a function type, we have to account for additional arguments in `BlackBoxFunction`.

## Still TODO:


martijnbastiaan (approved):
Good point

PR #2538, opened 2023-07-13
Add `Clash.Cores.Xilinx.Ila`

martijnbastiaan:
Adds support for [Xilinx Integrated Logic Analyzer v6.2](https://docs.xilinx.com/v/u/en-US/pg172-ila).

I've purposely left out validation of configuration parameters in the black box: they're fairly difficult to get right, it happens after normalization anyway, and Vivado will already do this.

## Still TODO:

  - [x] Clean up
  - [x] Write documentation
  - [x] Write (HDL generation) tests

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Ila.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ila.hs
@@ -1,0 +1,142 @@
+{-|
+Copyright  :  (C) 2023, Google Inc,
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Support for [Xilinx Integrated Logic Analyzer v6.2](https://docs.xilinx.com/v/u/en-US/pg172-ila).
+An Integrated Logic Analyzer (ILA) is a feature provided by Xilinx in its design
+tools, notably Vivado, that allows designers to debug their FPGA logic in
+real-time. It stores the signals it samples in a ring buffer, allowing users to
+see values before and after a trigger point.
+
+It is necessary to read the product guide linked above in order to effectively
+use the IP. Clash simulation is not applicable for this IP.
+
+When using the generated ILAs make sure you have set the correct JTAG clock speed:
+[/"For non-Versal architectures, if your design contains debug cores, ensure that the JTAG clock is 2.5 times slower than the debug hub clock."/](https://www.xilinx.com/content/dam/xilinx/support/documents/sw_manuals/xilinx2022_2/ug908-vivado-programming-debugging.pdf)
+
+-}
+
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveLift #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns#-}
+
+-- See [Note: eta port names for trueDualPortBlockRam]
+{-# OPTIONS_GHC -fno-do-lambda-eta-expansion #-}
+
+module Clash.Cores.Xilinx.Ila
+  ( ila
+
+  -- * Config
+  , IlaConfig(..)
+  , ilaConfig
+  , ProbeType(..)
+  , Depth(..)
+
+  -- * Utilities
+  , Ila(..)
+  ) where
+
+import Clash.Explicit.Prelude
+
+import Clash.Annotations.Primitive (Primitive (InlineYamlPrimitive))
+
+import Data.String.Interpolate (__i)
+
+import Clash.Cores.Xilinx.Ila.Internal
+
+-- | A default ILA config that:
+--
+--  * Configures no pipeline registers
+--  * Stores 4096 samples
+--  * Sets 2 comparators per probe
+--  * Sets all probes to be suited for DATA and TRIGGER
+--  * Enables capture control
+--
+-- See 'IlaConfig' for more information.
+ilaConfig :: Vec n String -> IlaConfig n
+ilaConfig names = IlaConfig
+  { stages = 0
+  , depth = D4096
+  , comparators = Left 2
+  , probeTypes = Left DataAndTrigger
+  , probeNames = names
+  , captureControl = True
+  , advancedTriggers = False
+  }
+
+
+class Ila (dom :: Domain) a where
+  ilaX :: a
+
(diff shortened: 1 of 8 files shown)


clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
```diff
@@ -0,0 +1,360 @@
...
+
+usedArguments :: [Int]
+usedArguments = (ilaConfig : clock : inputProbes)
+ where
+  (    _knownDomain
+    :< _ilaConstraint
+    :< ilaConfig
+    :< clock
+    :< (Infinite.take 8096 -> inputProbes)
+    ) = ((0::Int)...) -- This function is polyvariadic so in theory it supports an
```
DigitalBrains1, line 147: There's no need to specify the type, `(0...)` wil do.

clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
```diff
@@ -0,0 +1,360 @@
...
+
+ilaTF :: (HasCallStack, KnownNat n) => IlaConfig n -> TemplateFunction
+ilaTF config =
+  TemplateFunction
+    usedArguments
+    -- 'validateIlaProbeBCC' already produces string describing
+    -- failing checks, but 'TemplateFunction' cannot handle this
+    -- yet. This is prepared to get updated easily as soon as the
+    -- feature gets implemented in clash-lib.
+    (maybe True error . validateIlaProbeBBC)
```
DigitalBrains1, line 160: Nice, although my suggestion for handling errors is that `TemplateFunction` should be changed to

```haskell
data TemplateFunction where
  TemplateFunction
    :: [Int]
    -- ^ Used arguments
    -> (forall s . Backend s => BlackBoxContext -> ExceptT Text (Lazy.State s Doc))
    -- ^ Render function
    -> TemplateFunction
```
so we can just handle all errors in the render function instead of some here and some there.

clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
```diff
@@ -0,0 +1,360 @@
...
+      '#{name}'. Refusing to instantiate Ila with unreliable probe names.
+  |]
+checkNameCollision _ tExpr = error [__i|
+  Internal error: Expected 'TExpr' with the following form:
+
+    TExpr _ (Identifier _ Nothing)
+
+  got:
+
+    #{ppShow tExpr}
```
DigitalBrains1, line 177: I don't think you should indent your `ppShow`. I'm almost certain this will indent the first line only, having subsequent lines outdent by two spaces again, only making the output more difficult to read.

This happens in multiple places, I'm just commenting once.

martijnbastiaan: Hmm, the quasiquoter I wrote many moons ago handled this correctly. But it's drowned out by the 1001 other QQs so I'm not sure what I'm using now :upside_down_face:

martijnbastiaan: `I.i` handles this :)

clash-cores/src/Clash/Cores/Xilinx/Ila.hs
```diff
@@ -0,0 +1,141 @@
...
+--
+--  * Configures no pipeline registers
+--  * Stores 4096 samples
+--  * Sets 2 comparators per probe
+--  * Sets all probes to be suited for DATA and TRIGGER
+--  * Enables capture control
+--
+-- See 'IlaConfig' for more information.
+ilaConfig :: Vec n String -> IlaConfig n
+ilaConfig names = IlaConfig
```
kleinreact, line 64: I find it strange to have a default configuration, where one of the options still must be provided as an argument. It just makes the usage kind of cumbersome. I would either perfer having some predefined names or sometthing like 
```haskell
ilaConfig = IlaConfig {
  ...
  , probeNames = error "Missing probe names"
  ... 
}

**Review by kleinreact — APPROVED** [review]
LGTM. Only have some minor comments, but nothing critical :smile:.

File: clash-cores/src/Clash/Cores/Xilinx/Ila.hs (hunk 4 above)
Comment (kleinreact) [reply clash-cores/src/Clash/Cores/Xilinx/Ila.hs:64]: Just forgot: It might also be reasonable to change the module name to `Clash.Cores.Xilinx.ILA` to make it consistent with `Clash.Cores.Xilinx.VIO`. Also because a big I and a small l after each other being a hard to read for many font settings :grin:.

Comment (martijnbastiaan) [reply clash-cores/src/Clash/Cores/Xilinx/Ila.hs:64]: Right, we've settled on the other way around :sweat_smile:, because all caps can make it hard to see where the "borders" between concepts lie. I.e., `ILATCL` vs `IlaTcl`. In a future PR we're going to make everything consistent in `clash-cores`.

File: clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs [hunk 5]
```diff
@@ -0,0 +1,327 @@
...
+  -- ^ Probe can only be used to trigger data capture
+  deriving (Eq, Show, Lift, Enum)
+
+-- | Configures the static properties of an 'Clash.Cores.Xilinx.Ila.ila'. Note
+-- that most properties (triggers, number of samples before/after trigger, ...)
+-- are configured at runtime using Vivado.
+--
+-- Use 'Clash.Cores.Xilinx.Ila.ilaConfig' to construct this with some sensible
+-- defaults.
+data IlaConfig n = IlaConfig
```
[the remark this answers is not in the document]
martijnbastiaan: No :-). Let me read up on it.

clash-cores/src/Clash/Cores/Xilinx/Ila.hs
martijnbastiaan: Concerning the config: I don't think we should introduce partial functions / structures. You _have_ to supply it either way - through a record update or by function argument. The latter will give you a type error if you mess up, the first won't.

clash-lib/src/Clash/Core/TermLiteral.hs
```diff
@@ -227,7 +236,7 @@ termToDataError term = bimap err id (termToData term)
 
     In its non-pretty-printed form:
 
-      #{show failedTerm}
+      #{ppShow failedTerm}
```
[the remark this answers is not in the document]
martijnbastiaan: I'll add this info to the commit message.

clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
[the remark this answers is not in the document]
martijnbastiaan: Added!

kleinreact (approved):
Nice to see my suggestions being applied. Looks all good to me then :smile:.

PR #2539, opened 2023-07-13
Add reset-related convenience functions

martijnbastiaan:
## Still TODO:

  - [x] Write a changelog wrt added type families

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Reset.hs
+++ b/clash-prelude/src/Clash/Explicit/Reset.hs
@@ -1,6 +1,6 @@
 {-|
 Copyright  :  (C) 2020-2023, QBayLogic B.V.,
-                  2022     , Google LLC
+                  2022-2023, Google LLC
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -21,6 +21,9 @@ module Clash.Explicit.Reset
   , resetGlitchFilter
   , holdReset
   , convertReset
+  , noReset
+  , andReset, unsafeAndReset
+  , orReset, unsafeOrReset
 
     -- Reexports
   , Reset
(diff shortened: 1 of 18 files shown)


clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -52,6 +55,53 @@ import           GHC.TypeLits (type (+))
...
+-- | A deasserted reset
+noReset :: KnownDomain dom => Reset dom
+noReset = unsafeFromHighPolarity (pure False)
+
+-- | Output reset will be asserted when either one of the input resets is
+-- asserted
+orReset ::
+  forall dom .
+  ( KnownDomain dom
+  , DomainResetKind dom ~ 'Synchronous ) =>
```
DigitalBrains1, line 67: I don't think it's a good idea to solve this with a constraint because it'll need to be copied to all domain-polymorphic encompassing user code. Rather, let's do what we do in `clash-cores:Clash.Cores.Xilinx.DcFifo` and `resetGlitchFilter`:
```haskell
orReset = case resetKind @dom of
            SSynchronous -> unsafeOrReset
            SAsynchronous -> clashCompileError "Computer says No"
```

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -52,6 +55,53 @@ import           GHC.TypeLits (type (+))
...
+  Reset dom ->
+  Reset dom ->
+  Reset dom
+orReset = unsafeOrReset
+
+-- | Output reset will be asserted when either one of the input resets is
+-- asserted. This function is considered unsafe because it can be used on
+-- domains with components whose resets are level sensitive, while use of this
+-- function can introduce glitches.
```
DigitalBrains1, line 76:
```suggestion
-- domains with components with asynchronous resets, where use of this function
-- can introduce glitches triggering a reset.
```

I like consistency in terminology; we mainly call it (a)synchronous resets. Let's not require of our users that they know all the synonyms... also, let's be a bit more verbose about those glitches.

DigitalBrains1 (changes requested):
These are definitely good to have! Let's get rid of "you need `unsafeX` to do this"! I'd like to do it a bit different though. Type safety is great. Amassing type constraints not so much.

clash-prelude/src/Clash/Explicit/Reset.hs
martijnbastiaan: Fair

leonschoorl: We had a long discussion about this in the office, and I think the conclusion was that we do want this at the type level.
But we want to make it more convenient to write `DomainResetKind dom ~ 'Synchronous`, with something like:
```haskell
type KnownSyncDomain (dom :: Domain) = (KnownDomain dom, DomainResetKind dom ~ 'Synchronous)

type DomIsSync (dom :: domain) = DomainResetKind dom ~ 'Synchronous
```

leonschoorl (dismissed):

martijnbastiaan:
@DigitalBrains1 @leonschoorl I've added proposal for the new constraint aliases. I'm reasonably happy with their names and definition (keep in mind the long-term plan is to get rid of `KnownDomain` except for things like `resetGen` and `clockGen`). I also think these two constraints will cover almost all cases, so I don't feel like it's necessary to add constraints for `ActiveEdge` and friends. Lastly, I'm not super-pleased with the `__N.B.__` explanations right now, please make suggestions (unless you guys think it's clear enough!).

martijnbastiaan:
Maybe `HasSynchronousReset dom` / `SetsSynchronousReset dom`? (Not a big fan of contracting it to `Sync` btw, we don't do that anywhere else.)

martijnbastiaan:
I've picked `HasX` in the end. It's a pattern used in the Haskell ecosystem more often, and I feel it pretty much reads as _this domain has X_.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -382,6 +412,20 @@ type DomainResetKind (dom :: Domain) =
...
+-- usage:
+--
+-- @
+-- myFunc :: HasDefinedInitialValues dom => ...
+-- @
+--
+-- Using this type implies 'KnownDomain'.
+--
+-- __N.B.__: There is no @UnknownInitialValues dom@ as a component that works
+--           without initial values will also work if it does have them.
```
DigitalBrains1, line 425: Is this **NB:** or simply "Note:"?

Do you really have to nota this that bene? ;-) NB means "Pay attention please this is important". I think this suffices:
```suggestion
-- Note that there is no @UnknownInitialValues dom@ as a component that works
-- without initial values will also work if it does have them.
```

(Also, in the majority of cases, we don't insert periods in NB. I should probably add the non-conforming cases to a doc polishing PR, but there are more pressing issues first)

martijnbastiaan: Let's get rid of it, I don't like NB anyway.

christiaanb (approved):

DigitalBrains1:
I'd like it if the documentation points out how users can manage these constraints efficiently. We discussed how people could create their own type synonyms in a single design with multiple clock domains like

```haskell
type DesignDomain dom =
  ( HasSynchronousReset dom
  , HasDefinedInitialValues dom
  )
```

and then write all their domain-polymorphic functions in the design as

```haskell
f ::
  DesignDomain dom =>
  Signal dom [...]
```

which is just as much code as `KnownDomain dom`, but correctly provides constraints whenever they use, say, `orReset` somewhere.

DigitalBrains1 (approved):

PR #2540, opened 2023-07-13
Add `unsafe{From,To}Active{Low,High}`

martijnbastiaan:
## Still TODO:

  - [x] Update references
  - [x] Add `unsafeToActive{Low,High}`
  - [x] Change deprecation message to indicate _when_ the old functions get removed

### DIFF
--- a/clash-prelude/src/Clash/Signal/Internal.hs
+++ b/clash-prelude/src/Clash/Signal/Internal.hs
@@ -95,10 +95,10 @@ module Clash.Signal.Internal
   , Reset(..)
   , unsafeToReset
   , unsafeFromReset
-  , unsafeToHighPolarity
-  , unsafeToLowPolarity
-  , unsafeFromHighPolarity
-  , unsafeFromLowPolarity
+  , unsafeToActiveHigh
+  , unsafeToActiveLow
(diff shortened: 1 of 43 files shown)


clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -1128,16 +1134,33 @@ resetPolarityProxy _proxy =
...
+-- loops. In case of synchronous resets it can lead to
+-- <Clash-Explicit-Signal.html#metastability meta-stability> in the presence of
+-- asynchronous resets.
+unsafeToHighPolarity
+  :: forall dom
+   . KnownDomain dom
+  => Reset dom
+  -> Signal dom Bool
+unsafeToHighPolarity = unsafeToActiveHigh
+{-# DEPRECATED unsafeToHighPolarity "Use 'unsafeToActiveHigh' instead. This function will be removed in Clash 1.12." #-}
```
DigitalBrains1, line 1163: I think `INLINE unsafeToHighPolarity` is redundant, but if you do add it you don't need to think, it'll just inline for sure.

leonschoorl (approved):
:heart: the new names, always thought the polarity terminology was odd

PR #2544, opened 2023-07-14
Make three reset glitch filter variants

DigitalBrains1:
The existing code for `resetGlitchFilter` was wrong: the `error` on domains with unknown initial values rendered as an X value in HDL instead of throwing an error.

The new `unsafeResetGlitchFilter` allows to sidestep the restriction if the user is sure this works for them.

The new `resetGlitchFilterWithReset` is the better solution for domains with unknown initial values, relying on an additional power-on reset for which assertion glitches don't occur.

Both `resetSynchronizer` and `resetGlitchFilter` no longer carry a `NOINLINE`/`OPAQUE` annotation. The only reason they had one was for nicer port/signal names, there was no technical reason to put them in a separate HDL file. Furthermore, somehow the names for `resetSynchronizer` were not nice anyway, it would need `-fno-do-lambda-eta-expansion` to work correctly.

## Still TODO:

  - [x] Use new constraints #2539 will introduce

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Reset.hs
+++ b/clash-prelude/src/Clash/Explicit/Reset.hs
@@ -10,6 +10,8 @@ Utilities to deal with resets.
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE TypeFamilies #-}
 
 {-# OPTIONS_GHC -fplugin=GHC.TypeLits.Normalise #-}
@@ -19,6 +21,8 @@ module Clash.Explicit.Reset
   ( -- Defined in this module
     resetSynchronizer
   , resetGlitchFilter
+  , resetGlitchFilterWithReset
+  , unsafeResetGlitchFilter
   , holdReset
   , convertReset
   , noReset
@@ -48,14 +52,14 @@ module Clash.Explicit.Reset
 import           Data.Type.Equality ((:~:)(Refl))
 
 import           Clash.Class.Num (satSucc, SaturationMode(SatBound))
-import           Clash.Explicit.Moore
 import           Clash.Explicit.Signal
+import           Clash.Explicit.Synchronizer (dualFlipFlopSynchronizer)
 import           Clash.Promoted.Nat
 import           Clash.Signal.Internal
 import           Clash.Sized.Index (Index)
 
 import           GHC.Stack (HasCallStack)
-import           GHC.TypeLits (type (+))
+import           GHC.TypeLits (type (+), type (<=))
 
 {- $setup
 >>> import Clash.Explicit.Prelude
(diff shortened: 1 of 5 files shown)


DigitalBrains1:
Review comments on the user documentation are particularly welcome.

There's a circumstance I thought of that I found no good way to put into the documentation. In `unsafeResetGlitchFilter`, about the period when it is still unasserted. I don't know how realistic it is, but during power-on the clock might not have stabilised. It might violate timing requirements if it is pulsing faster than the stable clock rate, which might mess up the Mealy machine in the reset glitch filter, potentially increasing the number of clock cycles spent unasserted. Then again, once the clock is stable, the amount of cycles will not exceed 2 * glitchlessPeriod. Perhaps it is understood that you don't start counting until the clock is stable. Or I'm just thinking too much about it.


Finally, this PR should probably wait for the [discussed changes](https://github.com/clash-lang/clash-compiler/pull/2539#discussion_r1263841356) in #2539.

DigitalBrains1:
The build failure due to a dependency loop would be solved by adding the constraint from #2539, which gets rid of `clashCompileError` in `resetGlitchFilter`, which is responsible for the dependency loop.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -180,23 +184,96 @@ resetSynchronizer clk rst = rstOut
...
 -- === __Example 1__
 -- >>> let sampleResetN n = sampleN n . unsafeToActiveHigh
 -- >>> let resetFromList = unsafeFromActiveHigh . fromList
 -- >>> let rst = resetFromList [True, True, False, False, True, False, False, True, True, False, True, True]
 -- >>> sampleResetN 12 (resetGlitchFilter d2 systemClockGen rst)
 -- [True,True,True,True,False,False,False,False,False,True,True,True]
+resetGlitchFilter =
+  case initBehavior @dom of
+    SDefined -> unsafeResetGlitchFilter
+    SUnknown -> error errMsg
```
christiaanb, line 228: Perhaps use `clashCompileError` so HDL generation stops properly?

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -206,31 +283,43 @@ resetGlitchFilter
...
-  noReset :: Reset dom
-  noReset = unsafeToReset (pure (not asserted))
-
   asserted :: Bool
   asserted =
     case resetPolarity @dom of
       SActiveHigh -> True
       SActiveLow -> False
--- See: https://github.com/clash-lang/clash-compiler/pull/2511
-{-# CLASH_OPAQUE resetGlitchFilter #-} -- Give reset glitch filter its own HDL file
```
christiaanb, line 233: Why don't we want this any longer?

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -180,23 +184,96 @@ resetSynchronizer clk rst = rstOut
 -- See: https://github.com/clash-lang/clash-compiler/pull/2511
 {-# CLASH_OPAQUE resetSynchronizer #-} -- Give reset synchronizer its own HDL file
 
+resetGlitchFilter, unsafeResetGlitchFilter
+  :: forall dom glitchlessPeriod n
+   . ( HasCallStack
+     , KnownDomain dom
+     , glitchlessPeriod ~ (n + 1) )
```
christiaanb, line 191: Why do we do this instead of `1 <= glitchlessPeriod` ?

clash-prelude/src/Clash/Explicit/Reset.hs
christiaanb: Ah, I only read the comment you made last week just now. You can disregard this question.

DigitalBrains1: Quote from above:
> [edit]
I temporarily changed the clashCompileError to error so we can look at, for example, the generated Haddock. This term-level treatment will be replaced by a type-level treatment introduced in https://github.com/clash-lang/clash-compiler/pull/2539 anyway.
[/edit]

;-)

clash-prelude/src/Clash/Explicit/Reset.hs
DigitalBrains1: I asked Martijn why he did this and he said it was just for nicer signal/port names for the filter. You expressed your disapproval about that a while back.

But it should have been mentioned in the commit message, I'll add it.

clash-prelude/src/Clash/Explicit/Reset.hs
DigitalBrains1: I don't know. @martijnbastiaan ?

martijnbastiaan: I think it used to be expressed in terms of `Vec` functions - and `Foldable (Vec n a)` imposes this constraint.

christiaanb: The constraint on Foldable got changed to `1 <= n` in version 0.99 while the `resetGlitchFilter` was introduced in version 1.4.

martijnbastiaan: Ah misread. Dunno in that case. Do we still need this constraint at all?

DigitalBrains1: No, let's drop it. While it makes no sense to filter for a minimum of 0 cycles, it will actually achieve that with the current implementation (unless I'm being really dense), so let's not add unneeded constraints.

DigitalBrains1: Hmmm no, we'd need to test it to see if it works because I am apparently really dense. For instance, `maxBound :: Index 0` is `XException`. And what is `Eq (Index 0)`? There's no concrete value for `Index 0`, it's not even a singleton, it's an empty set.

martijnbastiaan (approved):

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -225,45 +229,164 @@ resetSynchronizer clk rst = rstOut
...
+-- but by doing so it is possible that the design starts up with a period of up
+-- to /2 * glitchlessPeriod/ clock cycles where the reset output is unasserted
+-- (or longer in the case of glitches on the filtered reset input). This can
+-- cause a number of problems. The outputs\/tri-states of the design might
+-- output random things, including coherent but incorrect streams of data. This
+-- might have grave repercussions in the design's environment (sending network
+-- packets, overwriting non-volatile memory, in extreme cases destroying
+-- controlled equipment or causing harm to living beings, ...).
+--
+-- Without initial values, the synthesized result of @unsafeResetGlitchFilter@
```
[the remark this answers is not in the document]
DigitalBrains1: So `'f'` creates a link to the documentation of `f`. Why would you want a link to the documentation of `f` _when you're reading the documentation of `f`_? I admit I actually like the blue color of the link, but the link itself borders on ridiculous. But if you'd rather link a lot, I can change it.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -225,45 +229,164 @@ resetSynchronizer clk rst = rstOut
...
+-- assert, and after the power-on reset deasserts, the reset output will stay
+-- asserted for another /glitchlessPeriod/ clock cycles. This is identical
+-- behavior to 'holdReset' where it concerns the power-on reset, and differs
+-- from the filtered reset, which will only cause an assertion after
+-- /glitchlessPeriod/ cycles.
+--
+-- If @resetGlitchFilterWithReset@ is used in a domain with asynchronous resets
+-- ('Asynchronous'), @resetGlitchFilterWithReset@ will first synchronize the
+-- reset input with 'dualFlipFlopSynchronizer'.
+resetGlitchFilterWithReset
```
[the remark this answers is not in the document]
DigitalBrains1: I think I like `WithReset` better even though it's rather strange. Especially because our rules on letter case make it sound like cut pork (the _k_'s been cut :-P).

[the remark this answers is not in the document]
martijnbastiaan: Cool :)

PR #2545, opened 2023-07-14
Add partial applicative typeclass functions

lmbollen:
We currently have functions like `.==.` that can operate on `Signal`s. However, I find myself often writing `.foo. pure bar` when comparing signals to constants.

With these changes(inspired by @gergoerdi's [retrocomputing lib](https://hackage.haskell.org/package/retroclash-lib-0.1.2.1/docs/RetroClash-Utils.html)) we can write `.foo bar` instead.

## Still TODO:

  - [x] Determine if there are more operators we'd like to add signal functions for.


clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -1436,6 +1436,30 @@ infix 4 .==.
...
+-- | The above type is a generalization for:
+--
+-- @
+-- __(.==)__ :: 'Eq' a => 'Clash.Signal.Signal' a -> a -> 'Clash.Signal.Signal' 'Bool'
+-- @
+--
+-- It is a version of ('==') that allows comparing a @'Clash.Signal.Signal' a@ with a
+-- constant @a@ and returns a 'Clash.Signal.Signal' of 'Bool'
+(.==) :: (Eq a, Applicative f) => f a -> a -> f Bool
+(.==) a b = fmap (==b) a
```
alex-mckenna, line 1449: All of these new functions where only one argument is in a container don't need to be `Applicative`, only `Functor`. You hint at this in the implementations already: only `fmap` is needed to define these functions

martijnbastiaan:
This makes sense to me. Can we also get `(.&&)`, `(&&.)`, `(.||)`, `(||.)`? That seems to cover all the operators we export. Are you interested in moving this out of draft @lmbollen?

DigitalBrains1:
The commit details for the first commit say:

> We already have the ' .foo.' pattern for these typeclasses.
However, you often have to use '.foo . pure bar' if 'bar' is a constant.
After this change, you can use '.foo bar' instead.

Note the stray space between `.foo` and `. pure bar`.

And a bit nitpicky, the commit summary goes over the 50-char limit of the [50/72 rule](https://preslav.me/2015/02/21/what-s-with-the-50-72-rule/). I didn't count the characters, I noticed the GitHub website UI cut off the summary: https://github.com/clash-lang/clash-compiler/pull/2545/commits/24d7979fc58d146f6cf7bc4456fb5971519560ab

lmbollen:
> Are you interested in moving this out of draft @lmbollen?
Yeah I'd really like to have these operators. I hope to be able to update the PR soon.

> Edit: What does bother me a little bit is that we use both `.` for signal lifting (like in this PR) and for `BitVector` operations (like `+>>.`).

Yes, though I feel the `.` is more useful as a lifting indicator since it can be applied a lot more..
For `BitVector` we only have `.<<+` and `+>>.` which are derived from their Vector counterparts `<<+` and `+>>`

martijnbastiaan:
Yeah, I guess my comment wasn't really directed at this PR. I was merely pointing out the fact that `+>>.` is ambiguous: is it `+>>` lifted over `Signal`, or is it the `BitVector` version? We should probably think of an alternative notation for the latter..

DigitalBrains1:
We could just call them `bitShiftInL` and `bitShiftInR` and use them as

```haskell
x `bitShiftInR` y
```

clash-prelude/src/Clash/Signal.hs
```diff
@@ -241,9 +241,9 @@ module Clash.Signal
   , testFor
     -- * Type classes
     -- ** 'Eq'-like
-  , (.==.), (./=.)
+  , (.==.), (.==), (==.), (./=.), (./=), (/=.)
```
martijnbastiaan, line 244: It's further up in this file, but the new functions amongst `(.&&.), (&&.), (.&&), (.||.), (||.), (.||)` should also be imported.

martijnbastiaan (approved):
LGTM modulo missing exports and a CI breakage :)

clash-prelude/src/Clash/Signal.hs
DigitalBrains1: exported*

changelog/2025-03-19T08_52_01+01_00_new_partial_signal_operators
```diff
@@ -0,0 +1 @@
+ADDED: Alongside the existing Eq-like and Ord-like signal operators like `.==.` and `.<=.` etc. There are now new functions for comparing with constants: `.==`, `==.`, `./=`, `/=.`, `.<=`, `<=.`, `.>=`, `>=.`, `.>`, `>.`, `.<`, `<.`, `.&&`, `&&.`, `.||`, `||.`. These are useful for comparing signals with constants in a more readable way. For example, `a .==. pure True` can now be replaced with `a .== True`.
```
DigitalBrains1, line 1:
```suggestion
ADDED: Alongside the existing Eq-like and Ord-like signal operators like `.==.` and `.<=.` etc., there are now new functions for comparing with constants: `.==`, `==.`, `./=`, `/=.`, `.<=`, `<=.`, `.>=`, `>=.`, `.>`, `>.`, `.<`, `<.`, `.&&`, `&&.`, `.||`, `||.`. These are useful for comparing signals with constants in a more readable way. For example, `a .==. pure True` can now be replaced with `a .== True`.
```

PR #2558, opened 2023-08-09
Ditch `ContT` from `clash-ffi`

kleinreact:
This PR simplifies the API of `clash-ffi` by removing the monadic `ContT` transformer reducing the interface to some basic `IO`. 

[`ContT`](https://hackage.haskell.org/package/transformers-0.6.1.1/docs/Control-Monad-Trans-Cont.html#t:ContT) is currently only used in `clash-ffi` for creating a unified interface for [`Foreign.Marshal.Alloc.alloca`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Foreign-Marshal-Alloc.html#v:alloca) (stack allocation) and [`Foreign.Marshal.Alloc.malloc`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Foreign-Marshal-Alloc.html#v:malloc) (heap allocation), which have different interfaces in `base` on purpose for reflecting the different characteristics of the stack and the heap. However, the introduced uniformization does not unify stack and heap usage in `clash-ffi`. The user still has to choose among different API calls depending on where objects have to be allocated. Thus, `ContT` really only provides a uniform type interface for both allocation methods, which is a weak achievement compared to the API modifications resulting from the additional monadic context it introduces.

Overall, the usage of `ContT` comes with the following downsides:
  * A bulky API, as basic operations must be lifted into `ContT` and the environment must be initiated via [`runSimAction`](https://github.com/clash-lang/clash-compiler/blob/682ec4460d146a7ec5754b09632f626d4154d584/clash-ffi/src/Clash/FFI/Monad.hs#L98).
  * `ContT` (as a monad transformer) is limited in preserving several properties of the underlying monad on `lift`. For example, [`ContT` is not a functor on the category of monads](https://hackage.haskell.org/package/transformers-0.6.1.1/docs/Control-Monad-Trans-Cont.html#t:ContT), and thus cannot implement [`MonadBaseControl`](https://hackage.haskell.org/package/monad-control-1.0.3.1/docs/Control-Monad-Trans-Control.html#t:MonadBaseControl).
  * Continuation-passing style is not that popular and thus less intuitive for many users compared to simple monadic IO.

Therefore, we propose to ditch `ContT` from `clash-ffi`. 

This is a breaking API change, which is reflected in a version number increment.


clash-ffi/clash-ffi.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version:      2.2
 
 name:               clash-ffi
-version:            1.5.0
+version:            1.6.0
```
martijnbastiaan, line 4: `clash-ffi` is locked to Clash's versioning (for now). This should have been 1.7.0. (Note that odd numbers indicate development versions, and even numbers released versions - similar to GHC.)

martijnbastiaan (approved):
I don't have a strong opinion on this, so I'm willing to follow @kleinreact on this one - given that you're the one building on top of `clash-ffi`. 

Any opinions @alex-mckenna?

PR #2562, opened 2023-08-16
Allow multiple PrimitiveGuard annotations

### LINKED ISSUE #1436: Multiple PrimitiveGuard annotations makes Clash yield no output HDL
It seems that compilation of any Clash function with (accidentally) multiple PrimitiveGuard annotations does not yield any HDL as output. This may, for example, easily happen if one does not realize that warnAlways and warnNonSynthesizable imply hasBlackBox:
```haskell
module Test where
import Clash.Prelude
import Clash.Annotations.Primitive
import Data.String.Interpolate (i)
import Data.String.Interpolate.Util (unindent)

test :: Bool
test = True
{-# NOINLINE test #-}
{-# ANN test hasBlackBox #-}
{-# ANN test (warnAlways "You should know that ...") #-}
{-# ANN test (InlinePrimitive [VHDL] $ unindent [i|
  [ { "BlackBox" :
      { "name" : "Test.test"
      , "kind" : "Expression"
      , "template" : "true"
      }
    }
  ]
 |]) #-}

topEntity :: Bool
topEntity = test
```

This is a user error, but I think the compiler could warn better about this. Currently it only prints this message, not making clear the consequences:
```The following value has multiple 'PrimitiveGuard' annotations: test```

Tested on Clash 1.2.2 using GHC 8.6.5.

### PR DESCRIPTION
**Pull request by leonschoorl**
Fixes #1436

Due to some refactoring of the error reporting this slightly changes some of the error messages:
```haskell
import Clash.Prelude
topEntity = False
{-# ANN topEntity (defSyn "foo") #-}
{-# ANN topEntity (defSyn "bar") #-}
```
Now generates:
```
    Error processing 'Synthesize' annotations on Main.topEntity:
    A binder can't have more than one 'Synthesize' annotation.
```

Instead of:
```
    The following value has multiple 'Synthesize' annotations: topEntity
```

## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -802,9 +802,7 @@ errOnDuplicateAnnotations nm bndrs anns =
 
 -- | Find annotations by given targets
 findAnnotationsByTargets
-  :: GHC.GhcMonad m
-  => Typeable a
-  => Data a
+  :: (GHC.GhcMonad m, Data a, Typeable a)
   => [Annotations.AnnTarget Name.Name]
   -> m [[a]]
 findAnnotationsByTargets targets =
@@ -816,9 +814,7 @@ findAnnotationsByTargets targets =
 
 -- | Find all annotations of a certain type in all modules seen so far.
 findAllModuleAnnotations
-  :: GHC.GhcMonad m
-  => Data a
-  => Typeable a
+  :: (GHC.GhcMonad m, Data a, Typeable a)
   => m [a]
 findAllModuleAnnotations = do
   hsc_env <- GHC.getSession
@@ -841,9 +837,7 @@ findAllModuleAnnotations = do
 
 -- | Find all annotations belonging to all binders seen so far.
 findNamedAnnotations
-  :: GHC.GhcMonad m
-  => Data a
-  => Typeable a
+  :: (GHC.GhcMonad m, Data a, Typeable a)
   => [CoreSyn.CoreBndr]
   -> m [[a]]
 findNamedAnnotations bndrs =
--- a/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
@@ -29,7 +29,7 @@ where
 #endif
 
 -- External Modules
-import           Clash.Annotations.Primitive     (HDL, PrimitiveGuard)
+import           Clash.Annotations.Primitive     (HDL, PrimitiveGuard(..))
 import           Clash.Annotations.TopEntity     (TopEntity (..))
 import           Clash.Primitives.Types          (UnresolvedPrimitive)
 import           Clash.Util                      (ClashException(..), pkgIdFromTypeable)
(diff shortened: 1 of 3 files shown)


clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -786,25 +787,42 @@ errOnDuplicateAnnotations
   -> [[a]]
   -- ^ Parsed annotations
   -> [(CoreSyn.CoreBndr, a)]
-errOnDuplicateAnnotations nm bndrs anns =
+errOnDuplicateAnnotations nm =
+  combineAnnotationsWith err nm
+ where
+  err _ _ = Left $ "A value can't have more than one '" ++ nm ++ "' annotation."
```
martijnbastiaan, line 793:
```suggestion
  err _ _ = Left $ "A binder can't have more than one '" ++ nm ++ "' annotation."
```

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -855,8 +869,14 @@ findPrimitiveGuardAnnotations
   -> m [(Text.Text, (PrimitiveGuard ()))]
 findPrimitiveGuardAnnotations bndrs = do
   anns0 <- findNamedAnnotations bndrs
-  let anns1 = errOnDuplicateAnnotations "PrimitiveGuard" bndrs anns0
+  let anns1 = combineAnnotationsWith combinePrimGuards "PrimitiveGuard" bndrs anns0
   pure (map (first (qualifiedNameString' . Var.varName)) anns1)
+ where
+  combinePrimGuards a b = case (a,b) of
+    (HasBlackBox x _, HasBlackBox y _) -> Right (HasBlackBox (x++y) ())
```
martijnbastiaan, line 876: Does this concatenate the warnings? In that case I think we should do `unlines [x, y]` instead.

martijnbastiaan (changes requested):
Cool :). Tiny changes requested.

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
leonschoorl: No, there is a little more structure than that, `x,y :: [PrimitiveWarning]` and 
```haskell
data PrimitiveWarning = WarnNonSynthesizable String | WarnAlways String
```

[This `go`](https://github.com/clash-lang/clash-compiler/blob/3a2d3209788aa440356f0c39ae088ef720da6562/clash-lib/src/Clash/Netlist/BlackBox.hs#L382) function goes over that resulting `[PrimitiveWarning]` and calls [`warn`](https://github.com/clash-lang/clash-compiler/blob/3a2d3209788aa440356f0c39ae088ef720da6562/clash-lib/src/Clash/Netlist/BlackBox.hs#L110) on them when applicable.
And `warn` then does a `hPutStLn` for each one.

martijnbastiaan: ahhhhh

clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
leonschoorl: Yes, binder fits better

martijnbastiaan (approved):

PR #2563, opened 2023-08-16
Drop `1 <= n` constraint from `Foldable (Vec n)` and from `Traversable (Vec n)`.

christiaanb:
The `Foldable.{foldr1,foldl1,maximum,minimum}` functions now throw an error at run-/simulation-time, and also at HDL-generation time, for vectors of length zero.

Also moves `clashCompileError` from `Clash.Explicit.Prelude` to `Clash.Magic`.

## Still TODO:

### DIFF
--- a/changelog/2023-08-16T19_26_52+02_00_vec_fold_zero
+++ b/changelog/2023-08-16T19_26_52+02_00_vec_fold_zero
@@ -1,0 +1,3 @@
+CHANGED: The `Foldable (Vec n)` instance and `Traversable (Vec n)` instance no longer have the `1 <= n` constraint. `Foldable.{foldr1,foldl1,maximum,minimum}` functions now throw an error at run-/simulation-time, and also at HDL-generation time, for vectors of length zero.
+CHANGED: The `maximum` and `minimum` functions exported by `Clash.Prelude` work on non-empty vectors, instead of the more generic version from `Data.Foldable`.
+ADDED: `1 <= n => Foldable1 (Vec n)` instance (`base-4.18+` only)
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -1,7 +1,7 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente,
                   2017     , Myrtle Software Ltd
-                  2022     , QBayLogic B.V.
+                  2022-2023, QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
@@ -75,7 +75,7 @@ module Clash.Sized.Vector
   , foldr, foldl, foldr1, foldl1, fold
   , ifoldr, ifoldl
     -- ** Specialized folds
-  , dfold, dtfold, vfold
+  , dfold, dtfold, vfold, maximum, minimum
     -- * Prefix sums (scans)
   , scanl, scanl1, scanr, scanr1, postscanl, postscanr
   , mapAccumL, mapAccumR
@@ -107,6 +107,9 @@ import Data.Constraint.Nat        (leZero)
 import Data.Data
   (Data (..), Constr, DataType, Fixity (..), Typeable, mkConstr, mkDataType)
 import Data.Either                (isLeft)
+#if MIN_VERSION_base(4,18,0)
+import qualified Data.Foldable1   as F1
+#endif
 import Data.Default.Class         (Default (..))
 import qualified Data.Foldable    as F
 import Data.Kind                  (Type)
@@ -129,7 +132,7 @@ import Prelude                    hiding ((++), (!!), concat, concatMap, drop,
                                           repeat, replicate, reverse, scanl,
                                           scanl1, scanr, scanr1, splitAt, tail,
                                           take, unzip, unzip3, zip, zip3, zipWith,
-                                          zipWith3)
+                                          zipWith3, maximum, minimum)
 import qualified Data.String.Interpolate as I
 import qualified Prelude          as P
 import Test.QuickCheck
@@ -138,9 +141,13 @@ import Unsafe.Coerce              (unsafeCoerce)
 
 import Clash.Annotations.Primitive
   (Primitive(InlineYamlPrimitive), HDL(..), dontTranslate, hasBlackBox)
+import Clash.Magic (clashCompileError)
 import Clash.Promoted.Nat
-  (SNat (..), SNatLE (..), UNat (..), compareSNat, leToPlus, pow2SNat,
+  (SNat (..), SNatLE (..), UNat (..), compareSNat, pow2SNat,
    snatProxy, snatToInteger, subSNat, withSNat, toUNat, natToInteger)
+#if MIN_VERSION_base(4,18,0)
+import Clash.Promoted.Nat (leToPlus)
+#endif
 import Clash.Promoted.Nat.Literals (d1)
 import Clash.Sized.Internal.BitVector (Bit, BitVector (..), split#)
 import Clash.Sized.Index          (Index)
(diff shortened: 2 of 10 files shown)


changelog/2023-08-16T19_26_52+02_00_vec_fold_zero
```diff
@@ -0,0 +1 @@
+CHANGED: The `Foldable (Vec n)` instance and `Traversable (Vec n)` instance no longer have the `1 <= n` constraint. `Foldable.{foldr1,foldl1,maximum,minimum}` functions now throw an error at run-/simulation-time, and also at HDL-generation time, for vectors of lenght zero.
```
martijnbastiaan, line 1:
```suggestion
CHANGED: The `Foldable (Vec n)` instance and `Traversable (Vec n)` instance no longer have the `1 <= n` constraint. `Foldable.{foldr1,foldl1,maximum,minimum}` functions now throw an error at run-/simulation-time, and also at HDL-generation time, for vectors of length zero.
```

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -317,25 +318,34 @@ instance KnownNat n => Applicative (Vec n) where
...
+instance KnownNat n => F.Foldable (Vec n) where
+  fold Nil      = mempty
+  fold z@Cons{} = fold mappend z
+  foldMap _ Nil      = mempty
+  foldMap f z@Cons{} = fold mappend (map f z)
   foldr     = foldr
   foldl     = foldl
-  foldr1 f  = leToPlus @1 @n $ foldr1 f
-  foldl1 f  = leToPlus @1 @n $ foldl1 f
+  foldr1 _ Nil      = clashCompileError "foldr1: empty list"
```
martijnbastiaan, line 328:
```suggestion
  foldr1 _ Nil      = clashCompileError "foldr1: empty vector"
```

or

```suggestion
  foldr1 _ Nil      = clashCompileError "foldr1: empty Vec"
```
?

martijnbastiaan (commented):
I'm somewhat hesitant. Is there a `Foldable1` on the horizon? Should we perhaps introduce it ourselves? What motivates this change?

Given that I'm in the middle, I've opted for "Comment". Please sway me.

christiaanb:
It was from a discussion brought on by @gergoerdi on the #clash slack FP channel, so perhaps he can chime in as well.

It has also annoyed me that functions using `sum` or `foldMap` or `traverse` accrue an unnecessary `1 <= n` constraint. But the run-time error turning into `false` in the worst case (Vec 0 Bool for VHDL), or `X` in the “best” case at HDL generation case has always put my off into making the change in this PR. But now that we have `clashCompileError` we won’t get surprising HDL (thanks @leonschoorl for the suggestion)

Note that the `foldr1` and `foldl1` exported by `Clash.Prelude` are still the total ones, so users would have to go out of their way to get the partial versions from `Data.Foldable`. That being said, I should probably add total versions of `maximum` and `minimum` to this PR and have them exported from `Clash.Prelude`.

Finally, `base-4.18` (bundled with GHC 9.6) does introduce a `Data.Foldable1` for non-empty structures. So I could add an instance for that in this PR as well. We will still need the `Data.Foldable` instance though since it is a super-class constraint for `Data.Traversable`.

martijnbastiaan (approved):
I don't understand why `Foldable` is a superclass of `Foldable1`. Doesn't that defeat its purpose..?

In any case, these changes look good to me, but I think should export all the `Foldable1` functions functions instead of the `Foldable` ones. You mentioned `minimum` and `maximum`, but this also goes for: `head`, `last`, `minimumBy`, and `maximumBy`.

gergoerdi:
> I don't understand why `Foldable` is a superclass of `Foldable1`. Doesn't that defeat its purpose..?

I think it makes sense: `Foldable1` means that it can be e.g. `foldr1`'d, i.e. it can be folded when expecting at least one element to be there. `Foldable` means it can be e.g. `foldr`'d, i.e. it can be folded without expecting at least on element.

If you can be folded when there are some extra expectations, then you can be folded when there aren't any.

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -317,25 +324,44 @@ instance KnownNat n => Applicative (Vec n) where
...
-  sum       = leToPlus @1 @n $ fold (+)
-  product   = leToPlus @1 @n $ fold (*)
+  maximum Nil      = clashCompileError "maximum: empty Vec"
+  maximum z@Cons{} = fold (\x y -> if x >= y then x else y) z
+  minimum Nil      = clashCompileError "minimum: empty Vec"
+  minimum z@Cons{} = fold (\x y -> if x <= y then x else y) z
+  sum Nil      = 0
+  sum z@Cons{} = fold (+) z
+  product Nil      = 1
+  product z@Cons{} = fold (*) z
```
DigitalBrains1, line 349: So multiplication on `UFixed 0 n` and `SFixed 0 n` / `SFixed 1 n` is a bit weird to begin with, but it does work. However, 1 is not the identity element of multiplication for those types:
```haskell
>>> :set -XNegativeLiterals
>>> -0.5 :: SFixed 0 8
-0.5
>>> as = -0.5 :> -0.5 :> Nil :: Vec 2 (SFixed 0 8)
>>> product as
0.25
>>> 1 * product as
0.12109375
```
This might for instance be a problem in
```haskell
>>> bs = Nil :: Vec 0 (SFixed 0 8)
>>> product (product as :> product bs :> Nil)
0.12109375
```

The question is, do we care? We seem to be violating the laws of multiplication.

christiaanb:
@martijnbastiaan `maximumBy` and `minimumBy` are not exported by `Prelude`, so there’s no need to override them in `Clash.Prelude`. The non-empty Vec `head` and `last` versions were always exported from `Clash.Prelude`.

martijnbastiaan:
Right @gergoerdi. I was thinking along the lines of "now there's no way you can avoid implementing partial functions", but that's not really relevant.

clash-prelude/src/Clash/Sized/Vector.hs
christiaanb: I guess the question is then more whether `fromInteger` for `SFixed` and `UFixed` should saturate or throw an exception when they go out of bounds.

DigitalBrains1: I'm a bit worried about the hidden character of this change. Before, if someone wrote code that say split a vector up so it would end with a `Vec 0` somewhere, it would not type-check. Now, it will type-check, but somehow the outcome is halved. That's less apparent than someone trying to stuff a 1 in `SFixed 0 8` themselves.

christiaanb: Would your suggestion then be to have `product` return `clashCompileError` for the `Nil` case instead then?

DigitalBrains1: I'm more inclined to change `fromInteger` to use `clashCompileError` for out-of-bounds values, but I'm not saying I have the best answer, just that I think the question is important.

So this is not exactly going to help for this case, but I found it interesting what GHC does for `Data.Int` (I poured a Clash sauce over it to show how it works out for us):
```haskell
module FromInteger where

import Clash.Prelude
import Data.Int

topEntity :: Int8
topEntity = 128
```
gives
```
Loaded package environment from /home/peter/src/clash/clash-compiler/.ghc.environment.x86_64-linux-9.0.2
GHC: Setting up GHC took: 0.219s

FromInteger.hs:7:13: warning: [-Woverflowed-literals]
    Literal 128 is out of the Int8 range -128..127
    If you are trying to write a large negative literal, use NegativeLiterals
  |
7 | topEntity = 128
  |             ^^^

FromInteger.hs:7:13: warning: [-Woverflowed-literals]
    Literal 128 is out of the Int8 range -128..127
    If you are trying to write a large negative literal, use NegativeLiterals
  |
7 | topEntity = 128
  |             ^^^
GHC: Compiling and loading modules took: 0.100s
Clash: Parsing and compiling primitives took 0.181s
GHC+Clash: Loading modules cumulatively took 0.559s
Clash: Compiling FromInteger.topEntity
Clash: Normalization took 0.000s
Clash: Compiling FromInteger.topEntity took 0.006s
Clash: Total compilation took 0.566s
```

PR #2568, opened 2023-08-25
Fix clash normalization benchmark

christiaanb:
Fix fallout from GHC 9.8 upgrade and ensure the benchmark is dynamically linked to avoid very long start times.

### DIFF
--- a/benchmark/benchmark-normalization.hs
+++ b/benchmark/benchmark-normalization.hs
@@ -44,20 +44,20 @@ benchFile idirs src =
   env (setupEnv idirs src) $
     \ ~(clashEnv, clashDesign, supplyN) -> do
       let topEntities = fmap topId (designEntities clashDesign)
-      case topEntities of
-        topEntity:_ ->
-          bench ("normalization of " ++ src)
-                (nfIO
-                  (normalizeEntity
-                    clashEnv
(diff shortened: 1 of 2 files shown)


benchmark/benchmark-normalization.hs
```diff
@@ -44,20 +44,20 @@ benchFile idirs src =
...
+            (nfIO
+              (normalizeEntity
+                clashEnv
+                (designBindings clashDesign)
+                (ghcTypeToHWType (opt_intWidth (envOpts clashEnv)))
+                ghcEvaluator
+                evaluator
+                topEntities
+                supplyN
+                topEntity))
```
leonschoorl, line 60: Does this just clean up the code or fix an actual issue in 9.8?
Because the code before and after seems semantically equivalent to me.

christiaanb: It fixes an issue I introduced in https://github.com/clash-lang/clash-compiler/pull/2560/commits/9a67a4820b430537b2d746345a0884baca885639

The code now doesn’t evaluate `topEntities` before it evaluates  `bench`. This is a requirement of `env` that was being violated.

leonschoorl (approved):

PR #2569, opened 2023-08-25
Move "Work-around for GHC 9.4 bug in ExpWithClashCF"

leonschoorl:
Moves the workaround added in c1ef28f36fbfed5b23de04aa43cf585a8bb83377 from Exp.hs to ExpWithGhcCF.hs

When I updated the CI image from 9.4.3 to 9.4.6 the ExpWithGhcCF test started to fail on the 9.4.6 nightly test with
```
    Other error:
    panic! (the 'impossible' happened)
      GHC version 9.4.6:
    	refineFromInScope
```
I've done some testing and I've found:
### Using ghc <= 9.4.4:
 The -fno-specialise workaround isn't need at all

### Using ghc 9.4.5 and 9.4.6:
 It needs the `-fno-specialise` workaround in ExpWithGhcCF, not in Exp.
 Otherwise the ExpWithGhcCF test fails,
 not the ExpWithClashCF like mentioned in the original commit msg.

## Still TODO:


christiaanb (approved):
LGTM

leonschoorl:
FYI ghc-9.4.7 still hits the same bug without the workaround

PR #2572, opened 2023-08-29
Enable ghc mmap crash workaround for every ghc compiled binary

leonschoorl:
In PRs #2444 we added the `workaround-ghc-mmap-crash` flag on some of our packages.
That flag added `-with-rtsopts=-xm20000000` to some of our binaries.

But other binaries would occasionally still trigger the mmap bug,
most importantly ghc itself.

This commit undoes the previous way of applying the workaround and
applies the same `-xm20000000` RTS option everywhere
via the GHCRTS environment variable.

But for this to work we have to link all binaries with the `-rtsopts`
flag, otherwise they'll just ignore the GHCRTS.

## Still TODO:


martijnbastiaan (approved):
What a mess :(

.ci/gitlab/test.yml
```diff
@@ -177,6 +180,7 @@ ffi:example:
       # Make sure we get custom cabals setting in the clash-example project too
     - cp cabal.project.local clash-ffi/example
   script:
+     - if [[ "$WORKAROUND_GHC_MMAP_CRASH" == "yes" ]]; then sed -i 's/${CABAL} build/${CABAL} build clash-ffi || exit $?\nGHCRTS="" ${CABAL} build/g' clash-ffi/example/run-iverilog.sh; fi
```
DigitalBrains1, line 183: This is pretty brittle; if someone changes `run-iverilog.sh` without realising this CI thing, the `sed` might do weird stuff. I'd rather see something like:
```suggestion
     - ./.ci/clash-ffi-example-build.sh; fi
```

and then that file would be
```sh
#!/bin/sh
set -xeu

# See PR #2572. Undo mmap crash workaround for clash-ffi-example.
cabal build clash-ffi
GHCRTS="" cabal build clash-ffi-example
```

And because `clash-ffi-example` is then already built by the time we `./run-iverilog.sh`, it will be used as-is. It still depends on things being done a certain way in that script, but I think it's way less brittle.

leonschoorl: Thnx

Issue #2574, opened 2023-09-06
ILA ignores custom BitPack instances


kleinreact:
The ILA blackboxes of `clash-cores` use a polyvariadic interface, hency any value can be dumped into an ILA. Now consider some data type with a custom `BitPack` instances, such as
```haskell
data T = A | B deriving (Generic)

instance BitPack T where
   type BitSize T = 1
   pack A = 1
   pack B = 0
   unpack 0 = B
   unpack _ = A
```
Note that this instance intentionally deviates from the default produced by deriving `BitPack`, which would assigns `A` ↦ 0 and `B` ↦ 1.

Now, if we pass a value `x :: T` directly into the ILA, then the ILA still used the implicit encoding as given by the derived `BitPack` instance, e.g. it assigns `A` ↦ 0 and `B` ↦ 1. Only if we pass `pack x :: BitVector (BitSize T)` instead, then it assigns `A` ↦ 1 and `B` ↦ 0.

This can be very confusing, especially if the custom `BitPack` instance only slightly differs against the default, making a wrongly produced ILA dump value hard to catch.

As an ILA user, I just would assume that the ILA automatically uses the custom `BitPack` instance, if it exists. 

The same probably also applies to VIO blackboxes, but I haven't explicitly tested it.

rowanG077:
Isn't this just a consequence that you should never really write a custom `BitPack` instance by hand? You should only derive them OR use the TH generation framework. Does the issue go away if you do:

```
import Clash.Annotations.BitRepresentation

data T = A | B deriving (Generic)

{-# ANN module (DataReprAnn
                  $(liftQ [t|T|])
                  1
                  [ ConstrRepr 'A 0b11 0b1 []
                  , ConstrRepr 'B 0b11 0b0 []
                  ]) #-}
```

If so this is not really specific to the ILA but a general Clash thing.

kleinreact:
Adding a `DataReprAnn` fixes the problem as well, but this means I have to keep the `BitPack` instance and the `DataReprAnn` in sync. I prefer explicitly calling `pack` then instead.

It's just that I need the `BitPack` instance:
* First, because the type is used as part of a larger record, which is dumped via the ILA, so I need the instance to derive `BitPack` for that larger structure. I cannot use a `DataReprAnn` on that record anyway.
* Second, because I read back the dumped data using the same `BitPack` instance (to ensure consistency with respect to the encoded data).

> If so this is not really specific to the ILA but a general Clash thing

[...]

PR #2575, opened 2023-09-06
Clarify `BitPack` is not for creating custom encodings

DigitalBrains1:
It's easy to misconstrue the purpose of `BitPack`

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Class/BitPack/Internal.hs
+++ b/clash-prelude/src/Clash/Class/BitPack/Internal.hs
@@ -1,7 +1,7 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente,
                   2016-2017, Myrtle Software Ltd,
-                  2021-2022  QBayLogic B.V.,
+                  2021-2023  QBayLogic B.V.,
                   2022,      Google Inc.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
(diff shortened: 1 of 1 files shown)


kleinreact (changes requested):
Great to have it mentioned now. I only have some minor points:
* "_**not** the way_" is a little bit weak formulation, because it still suggests that it's fine to define custom instances overall. I would be more strict and use something link "_You should never define a custom instance yourself, but instead use ... . The purpose of `BitPack` is to provide uniform conversion methods only._". I also liked your explanation that "_It allows no creativity._" of #2574.
* I also would mention that `deriving (BitPack)` is fine too. At the moment it may be missinterpreted that defining a custom representation is the only way.
* > Rather than dictate what the bit representation in HDL is, a BitPack instance needs to exactly match that bit representation.

  I also got confused by that sentence, probably because its not immediately clear what "_that bit representation_" refers to.

DigitalBrains1:
> * > Rather than dictate what the bit representation in HDL is, a BitPack instance needs to exactly match that bit representation.
>
>  I also got confused by that sentence, probably because its not immediately clear what "_that bit representation_" refers to.

I hoped it was clear that "that representation" in the second sentence fragment refers to "the representation" in the first sentence fragment, but would you rather see it repeated verbatim?

Rather than dictate what the bit representation in HDL is, a BitPack instance needs to exactly match the bit representation in HDL.

kleinreact:
> Rather than dictate what the bit representation in HDL is, a BitPack instance needs to exactly match the bit representation in HDL.

It still confuses me. Maybe, because "_the bit representation_" is just very generic: so maybe elaborating on what bit representation is meant? I mean: I get what you wanna say with the sentence, but I have to read it at least twice and need to parse it carfully to get the meaning.

DigitalBrains1:
I'll rephrase entirely.

kleinreact (approved):
LGTM now.

DigitalBrains1:
@mergify rebase

Issue #2168, opened 2022-04-12
Uncurried BiSignalIn params lead to faulty Verilog


gergoerdi:
```
module Bi (topEntity) where

import Clash.Prelude
import Clash.Annotations.TH

topEntity
    :: "CLK_50MHZ" ::: Clock System
    -> "RESET"     ::: Reset System
    -> ( "SCL" ::: BiSignalIn 'PullUp System (BitSize Bit)
       , "SDA" ::: BiSignalIn 'PullUp System (BitSize Bit)
       )
    -> ( "SCL" ::: BiSignalOut 'PullUp System (BitSize Bit)
       , "SDA" ::: BiSignalOut 'PullUp System (BitSize Bit)
       )
topEntity clk reset = exposeClockResetEnable run clk reset enableGen

run
    :: (HiddenClockResetEnable dom)
    => ( "SCL_IN"  ::: BiSignalIn 'PullUp dom (BitSize Bit)
       , "SDA_IN"  ::: BiSignalIn 'PullUp dom (BitSize Bit)
       )
    -> ( "SCL_OUT" ::: BiSignalOut 'PullUp dom (BitSize Bit)
       , "SDA_OUT" ::: BiSignalOut 'PullUp dom (BitSize Bit)
       )
run (sclIn, sdaIn) = (sclOut, sdaOut)
  where
    sclOut = writeToBiSignal sclIn sclOut'
    sdaOut = writeToBiSignal sdaIn sdaOut'

    sclOut' = pure $ Just high
    sdaOut' = pure $ Just low

makeTopEntityWithName 'topEntity "bisignal"
```

```
/* AUTOMATICALLY GENERATED VERILOG-2001 SOURCE CODE.
** GENERATED BY CLASH 1.6.1. DO NOT MODIFY.
*/
`timescale 100fs/100fs
module bisignal
    ( // Inputs
      input  CLK_50MHZ // clock
    , input  RESET // reset
    , inout [0:0] SCL
    , inout [0:0] SDA

      // No outputs
    );
  // src/Bi.hs:19:1-9
  wire [0:0] sclIn;
  // src/Bi.hs:19:1-9
  wire [0:0] sdaIn;
  wire [1:0] ds;

  assign ds = {SCL,   SDA};

  assign sclIn = ds[1:1];

  assign sdaIn = ds[0:0];

  // writeToBiSignal# begin
[...]

Issue #1138, opened 2020-03-02
`map writeToBiSignal` doesn't render to HDL


leonschoorl:
```haskell
module Search where
import Clash.Prelude
import Clash.Signal.BiSignal
import Clash.Explicit.Testbench

findSingleJust
  :: forall dom n a. ( n ~ 4, BitPack a, KnownNat n)
  => (a -> Maybe a) -> Signal dom (Vec n a) -> Signal dom a
findSingleJust p xs0 = sigOut
  where
    xs1 :: Vec n (Signal dom a)
    xs1 = unbundle xs0
    xs2 :: Vec n (Signal dom (Maybe a))
    xs2 = map (fmap p) xs1

    bisigOuts :: Vec n (BiSignalOut 'Floating dom (BitSize a))
    bisigOuts = map (writeToBiSignal bisigIn) xs2
    -- bisigOuts =  go xs2a :> go xs2b :> go xs2c :> go xs2d :> Nil
    --   where
    --     go = writeToBiSignal bisigIn
    --     xs2a :> xs2b :> xs2c :> xs2d :> Nil = xs2

    bisigOut :: BiSignalOut 'Floating dom (BitSize a)
    bisigOut = mergeBiSignalOuts bisigOuts

    bisigIn :: BiSignalIn 'Floating dom (BitSize a)
    bisigIn = veryUnsafeToBiSignalIn bisigOut

    sigOut :: Signal dom a
    sigOut = readFromBiSignal bisigIn


predToMaybe :: (a -> Bool) -> (a -> Maybe a)
predToMaybe p = \x -> if p x then Just x else Nothing

topEntity :: Signal System (Vec 4 (Unsigned 4)) -> Signal System (Unsigned 4)
topEntity = findSingleJust (predToMaybe (odd))


{-# NOINLINE topEntity #-}

testBench :: Signal System Bool
testBench = done
  where
    testInput      = stimuliGenerator clk rst input
    expectedOutput = outputVerifier' clk rst expected
    done           = expectedOutput (topEntity testInput)
    clk            = tbSystemClockGen (not <$> done)
    rst            = systemResetGen

input :: Vec 5 (Vec 4 (Unsigned 4))
input =
  (1:>0:>0:>0:>Nil) :>
  (0:>3:>0:>0:>Nil) :>
  (0:>0:>0:>5:>Nil) :>
  (0:>0:>7:>0:>Nil) :>
  (0:>9:>0:>0:>Nil) :>
  Nil
[...]

PR #2584, opened 2023-10-03
Move i2c to clash cores

lmbollen:
Currently the i2c core in the `examples` folder can not be used without copying the files or having a local clone of the reposity.
This pull request moves the core over to `clash-cores` and aims to making it properly documented.

## Still TODO:


clash-cores/test/Test/Cores/I2C.hs
```diff
@@ -23,15 +23,17 @@ system0 clk arst = bundle (regFile,done,fault)
...
 
   rst = liftA2 (<) rstCounter 500
   rstCounter = register clk arst enableGen (0 :: Unsigned 18) (rstCounter + 1)
 -- See: https://github.com/clash-lang/clash-compiler/pull/2511
-{-# CLASH_OPAQUE system0 #-}
+{-# CLASH_OPAQUE testBench0 #-}
 
-{-# ANN system Synthesize { t_name = "system", t_inputs = [], t_output = PortName "" } #-}
-system = system0 systemClockGen resetGen
+{-# ANN testBench Synthesize { t_name = "testBench", t_inputs = [], t_output = PortName "" } #-}
```
DigitalBrains1, line 34: As discussed, some comments:
1. `testBench` is a magic name; I think it's better to give it a different name. The `Synthesize` annotation probably masks the special machinery.
2. The record syntax with a space between the constructor and the `{` is confusing, it looks like function application. How many arguments does `ANN` take? :-)
3. We didn't document this yet, but we did discuss steering people away from the constructor and to using `defSyn` instead. I think mainly so we can add fields without existing use sites suddenly having undefined field contents.

DigitalBrains1: Ah forget about the `defSyn` that doesn't work... \<sad trombone\>

tests/Main.hs
```diff
@@ -158,24 +158,6 @@ runClashTest = defaultMain $ clashTestRoot
     , clashTestGroup "crc32"
         [ runTest "CRC32" def
         ]
-    , clashTestGroup "i2c"
```
christiaanb, line 161: The I2C test is:
1. The only test in the clash testsuite that has multiple synthesize annotations
2. Tests `SimIO` in a common user scenario.

So I would like it if this test wasn't removed. And given that the I2C core in clash cores has a new API, perhaps we can either just:
1. Keep the old domain-monomorphic code in the `examples` directory
2. Or move the old domain-monomorphic code to a test directory
So that we still have a test with multiple Synthesize annotations.

clash-cores/src/Clash/Cores/I2C/ByteMaster.hs
```diff
@@ -146,26 +145,25 @@ byteMasterT s@(ByteS {_srState = ShiftRegister {..}, ..})
...
           byteStateM .= Active
           coreCmd    .= I2Cnop
-          -- generate command acknowledge signal
-          i2cOpAck    .= True
         else do
           byteStateM .= Stop
           coreCmd    .= I2Cstop
       else
-        coreTxd .= bitCoerce ackIn
+        coreTxd .= (bitCoerce $ not ackRead)
```
christiaanb:
```suggestion
        coreTxd .= bitCoerce (not ackRead)
```

clash-cores/src/Clash/Cores/I2C/ByteMaster.hs
```diff
@@ -137,7 +136,7 @@ byteMasterT s@(ByteS {_srState = ShiftRegister {..}, ..})
 
     (Read, _) -> when coreAck $ do
       shiftsr .= True
-      coreTxd .= bitCoerce ackIn
+      coreTxd .= (bitCoerce $ not ackRead)
```
christiaanb:
```suggestion
      coreTxd .= bitCoerce (not ackRead)
```

clash-cores/src/Clash/Cores/I2C/ByteMaster.hs
```diff
@@ -29,10 +29,8 @@ data ByteMasterS
   , _coreTxd    :: Bit              -- coreTxd register
   , _shiftsr    :: Bool             -- shift sr
   , _ld         :: Bool             -- load values in to sr
-  , _i2cOpAck    :: Bool             -- host cmd acknowlegde register
```
christiaanb: Why is the command ack made combinational? The commit message doesn't state why this change was made.

clash-cores/test/Test/Cores/I2C/Config.hs
```diff
@@ -51,88 +52,84 @@ configT s0 (rst,ena,cmdAck,rxAck,al) = do
...
+        pure s { i2cConfStateM = CONFaddrAck
+               , i2cConfOp     = Just (WriteData (unpack i2cSlvAddr))
+               }
 
     CONFaddrAck
       | success
-      -> do display "CONFaddrAck"
-            pure s { i2cConfStateM = CONFreg
-                   , i2cWrite = False
+      -> if rxAck then do
```
christiaanb: Are these changes made because the ack made combinational in the another commit? Perhaps mention that in the commit message then.

clash-cores/test/Test/Cores/I2C/Config.hs
```diff
@@ -22,8 +22,18 @@ data ConfS = ConfS { i2cConfStateM   :: ConfStateMachine
                    , i2cConfFault    :: Bool
                    }
 
-type ConfI = (Bool,Bool,Bool,Bool,Bool)
-type ConfO = (Bool,Maybe I2COperation,Bool,Bool)
+data ConfI = ConfI { i2cRst    :: Bool
```
christiaanb: The commit message just says "refactor", perhaps explain what the refactor does.

clash-cores/test/Test/Cores/I2C/Config.hs
```diff
@@ -49,7 +52,7 @@ configT
   -> SimIO ConfO
 configT s0 ConfI{i2cRst=rst,i2cEna=ena,i2cCmdAck=cmdAck,i2cRxAck=rxAck,i2cAl=al} = do
   s <- readReg s0
-  let ConfS confStateM claim op lutIndex fault = s
+  let ConfS confStateM claim op lutIndex fault debug = s
```
christiaanb: Why are debug messages made conditional? Perhaps that could be mentioned in the commit message.

clash-cores/src/Clash/Cores/Experimental/I2C.hs
```diff
@@ -0,0 +1,87 @@
...
+--
+-- @
+-- i2cComp clk rst ena sclIn sdaIn = (sclOut, sdaOut)
+--  where
+--   sclOut = writeToBiSignal sclIn (mux sclOe (pure $ Just 0) (pure Nothing))
+--   sdaOut = writeToBiSignal sdaIn (mux sdaOe (pure $ Just 0) (pure Nothing))
+--   (sclOe, sdaOe) = unbundle i2cO
+--   i2cIn = bundle (readFromBiSignal sclIn, readFromBiSignal sdaIn)
+--   (dout,i2cOpAck,busy,al,ackWrite,i2cOut) = i2c clk arst rst ena clkCnt claimBus i2cOp ackRead i2cI
+-- @
```
christiaanb, line 36: The outputs for SCL and SDA are `Maybe Bit`, neither the haddock comment, nor the code example seem to get this rights.

christiaanb (changes requested):
Aside from commit messages stating that specific changes were made, they should explain why the changes were made.

tests/Main.hs
DigitalBrains1: 2. Sounds like the best option to me!

clash-cores/src/Clash/Cores/Experimental/I2C.hs
DigitalBrains1: (I did not have a chance to actually look at this code and the changes; and I don't have time anymore to do so now. So I'm making a bunch of assumptions and next week I'll see if I was off base)

SCL and SDA are open-drain I/O pins: they are both input and output, and they are either driven low or high-impedance. I assume the `Maybe Bit` therefore has two states: `Nothing` (Hi-Z) and `Just 0` (drive low). And I suppose this encoding was chosen because in an FPGA, that's how you would drive the tri-state used to implement an open-drain pin on the FPGA. Specifically, the trivial way to implement open-drain with a tri-state is to have the input from the FPGA tied to ground and put your data on the gate: gated when high, outputting when low.

So this `Maybe Bit` is purely pragmatic: it encodes how you drive a tristate, but it only has two states. It's also not entirely practical; we have two tri-state primitives, one for the ICE40 and one for the ECP5, and one of those uses an enable signal to gate, and the other a `Signal Bool` to gate; neither actually _uses_ `Maybe` to gate.

So I'd like to propose we change the SCL and SDA outputs to just be `Bit`. And this would then encode the two possible states: 1 = Hi-Z, 0 = drive low, just like I²C defines it afaik.

And related but not for this PR, I'd like to harmonise our two tri-state primitives, and define wrappers that can be used to implement open-drain and open-source, where the data inputs to those wrappers are just `Bit` and they wrap the underlying tri-state primitives in the logical way (data input tied to rail, gate input tied to data, either inverted or direct). But that's for later; it just clarifies how I see the future for using this I²C core with I/O buffer primitives.

lmbollen: > So this Maybe Bit is purely pragmatic: it encodes how you drive a tristate, but it only has two states

Indeed! It's to show that it drives a tristate.
> It's also not entirely practical; we have two tri-state primitives, one for the ICE40 and one for the ECP5, and one of those uses an enable signal to gate, and the other a Signal Bool to gate; neither actually uses Maybe to gate.

I have never seen these primitives to be honest, the choice was made to be compatible with BiSignals. at least Vivado simply infers the tristate buffer when you use BiSignal. We don't use any explicit primitives in the Bittide implementations.

Personally I prefer the `Maybe Bit` over `Bit.

>And related but not for this PR, I'd like to harmonise our two tri-state primitives, and define wrappers that can be used to implement open-drain and open-source, where the data inputs to those wrappers are just Bit and they wrap the underlying tri-state primitives in the logical way (data input tied to rail, gate input tied to data, either inverted or direct). But that's for later; it just clarifies how I see the future for using this I²C core with I/O buffer primitives.

Personally I don't find it intuitive to make the behavior depend on the outside world, whats the downside of using `Maybe Bit` to you?

tests/Main.hs
lmbollen: Also agree with 2, we can just put the old code in a test directory.

clash-cores/test/Test/Cores/I2C/Config.hs
lmbollen: Because otherwise they flood the terminal when you run `clash-cores:unittests` :)

clash-cores/src/Clash/Cores/Experimental/I2C.hs
rowanG077: `Maybe Bit` can do the wrong thing. You could output `Just 1` which is wrong for i2c. I would like to propose a third option. Have a unit data type `newtype Low = Low` or something like that and output `Maybe Low`.

lmbollen: Indeed the core should never output `Just 1`, and it never does. I don't see a problem in using `Maybe Bit` as tristrate drive type for i2c. It intuitively hints to the user that its mean to be a tristate driver

DigitalBrains1: > at least Vivado simply infers the tristate buffer when you use BiSignal. 

Oh that's good to know! I just immediately thought of those two primitives.

Thinking about it a bit more, a wrapper for a `BiSignal` is less pretty in API than I initially imagined it could be. But still I'm leaning towards it being the right way. It would also hide a bit of the complexity of `BiSignal` for this use.

We could add this to the Prelude:

```haskell
openDrainIO :: BiSignalIn ds dom (BitSize a) -> Signal dom a -> (BiSignalOut ds dom (BitSize a), Signal dom a)
openSourceIO :: BiSignalIn ds dom (BitSize a) -> Signal dom a -> (BiSignalOut ds dom (BitSize a), Signal dom a)
```

and people wouldn't have to do all the `readFromBiSignal` and `writeToBiSignal` themselves.

> Personally I don't find it intuitive to make the behavior depend on the outside world, whats the downside of using `Maybe Bit` to you?

It's semantically wrong. It is a two-valued signal, not a three-valued one. If you want to force people towards a specific implementation, you should perhaps create a new type `OpenDrain` or something like that. So what's above becomes:

```haskell
openDrainIO :: BiSignalIn ds dom (BitSize a) -> Signal dom (OpenDrain a) -> (BiSignalOut ds dom (BitSize a), Signal dom a)
```

(that's my counter-suggestion to Rowan's `Low`. I think it is more obvious in its intention.)

> It intuitively hints to the user that its mean to be a tristate driver

In an FPGA, that is the usual implementation, yes, since it's there anyway. But it's not _supposed_ to be a tri-state driver in my view. It's supposed to be an open-drain driver. But in an FPGA, every I/O pin is connected to a tri-state driver, and that's what you use. A standard totem-pole output is a tri-state that is never Hi-Z, an open-drain output is a tri-state that is never driven high, and an open-source output is a tri-state that is never driven low. Three types of outputs, all of them having two states. Only the true tri-state output has three states.

In an ASIC, you would never use a tri-state to implement an open drain, it would be wastage of die space on a staggering scale, relatively speaking.

rowanG077: Note that I would hint to shy away from `BiSignal` as in API because of https://github.com/clash-lang/clash-compiler/issues/2168 and https://github.com/clash-lang/clash-compiler/issues/1138

DigitalBrains1: I don't intend to codify the `BiSignal` in the I²C stuff; people are free to choose their pin driver. But I think you either use platform-specific primitives or the `BiSignal` stuff. As I said, I don't mean to include these Prelude extensions as part of this PR, it's just how I envision it being used, which supports my argument in favour of just `Bit`.

Issue #978, opened 2019-12-17
Remove the need for `KnownDomain` constraints throughout Clash designs


martijnbastiaan:
Clash primitives currently require `KnownDomain` constraints in order to change their behavior depending on their [synthesis domain](https://clash-lang.org/blog/0005-synthesis-domain/). It's implemented using class constraints as this - apart from all the dormant bugs it exposed -  didn't require adding additional logic to the Clash compiler. An example of a primitive requiring a `KnownDomain` constraint is `register#`:

https://github.com/clash-lang/clash-compiler/blob/9d278c2a7749ce0a419791b000d4e502c73a0d5e/clash-prelude/src/Clash/Signal/Internal.hs#L1078-L1082

To remove the need for this constraint we would need to do two things:

1. On the simulation side of things, we'd need to remove all the `KnownDomain` constraints on Clash primitives. We subsequently need to add them to `clockGen`, `resetGen`, and `enableGen`. These primitives would use `knownDomain` to store a `SDomainConfiguration` inside the constructs they're building. For example, `Clock` might look like:

```haskell
data Clock (dom :: Domain) = 
  Clock 
    (SSymbol dom) 
    (SDomainConfiguration dom (KnownConf dom))
```

(If this doesn't work out, we could go for `VDomainConfiguration` too.) Simulation constructs can now simply extract the synthesis domains from their given clocks. Of course, deconstructing clocks, resets, and enables is still not synthesizable! Thus...

2. To get synthesis working we need domain-extracting primitives for clocks, resets, and enables. For example, for a `Clock` this would look like:

```haskell
clockDomain :: Clock dom -> SDomainConfiguration dom (KnownConf dom)
```

(We might want to wrap this in a typeclass so we can use the same function on clocks, resets, and enables all the same.) Clash would need to collect all instances of `KnownDomain` before synthesis and insert the result of `knownDomain` for every `clockDomain` it sees.

paddytheplaster:
Hi Martijn,


Brilliant idea.

Regards,


Paddy

martijnbastiaan:
Me and @leonschoorl talked about this yesterday. We concluded that the easiest way is probably to add GADT wrappers around `Clock`, `Reset`, `Enable`, and `Signal` that carry the `KnownDomain` dictionary. E.g.:

```haskell
data SClock (dom :: Domain) where
  SClock (Clock dom) :: KnownDomain dom => SClock dom 
```

A few observations:

[...]

PR #2589, opened 2023-10-16
Remove KnowDomain

leonschoorl:
Removing the need for KnownDomain constraints

This an implementation of the idea from #978, removing (most of the need for) the KnownDomain constraint.
The core change of this PR is to hide the `KnownDomain` constraint inside the `Clock` and `Reset` types.

```diff
--- a/clash-prelude/src/Clash/Signal/Internal.hs
+++ b/clash-prelude/src/Clash/Signal/Internal.hs
@@ -931,7 +931,7 @@ enableGen :: Enable dom
 enableGen = toEnable (pure True)

 -- | A clock signal belonging to a domain named /dom/.
-data Clock (dom :: Domain) = Clock
+data Clock (dom :: Domain) = KnownDomain dom => Clock
   { -- | Domain associated with the clock
     clockTag :: SSymbol dom

@@ -1171,7 +1171,8 @@ resetGenN n =
 -- | A reset signal belonging to a domain called /dom/.
 --
 -- The underlying representation of resets is 'Bool'.
-data Reset (dom :: Domain) = Reset (Signal dom Bool)
+data Reset (dom :: Domain) where
+  Reset :: KnownDomain dom => Signal dom Bool -> Reset dom
```

The allows us to drop the `KnownDomain` constraint from every function take takes either a `Clock` or a `Reset`, which is most of them.

The only things that still require a `KnownDomain` constraint are functions that create `Clock`s or `Reset`s:
 * clock and reset generators: `noReset`, `clockGen`, `resetGen` and their variants
 * simulation helpers: `simulate`, `sample` and all their variants [^1]
 * conversion functions creating resets from Bool signals: `unsafeFromActive[High,Low]`, `unsafeToReset`
 * PLLs (and other hardware clock generators): `altpll`, `alteraPll`, `clockWizard`, `clockWizardDifferential`

And some tracing functions:
 * `traceSignal`, `traceVecSignal`

[^1]: Except the Explicit variants of `simulate`, `simulateB`, `sample`, `sampleN`, `sample_lazy` and `sampleN_lazy`.

When using any of these functions that still requires a `KnownDom` there a couple of ways of handling it:

 0. Don't do anything, just propagate the constraint up to your type signature, so your caller will have to solve it.
 1. Make your code monomorphic.
    By specializing to a specific domain, GHC will automatically solve `KnownDomain` constraints for you.
 2. Use the new `provideKnownDomainFrom`
    <details>
    <summary>Click here for its definition</summary>

    ```haskell
    class HasKnownDomain a where
      provideKnownDomainFrom :: a dom -> (KnownDomain dom => r) -> r

    instance HasKnownDomain Clock where ...
    instance HasKnownDomain DiffClock where ...
    instance HasKnownDomain Reset where ...
    ```
    </details>

    ```haskell
    -- foo :: Clock dom -> Signal dom Bool -> Reset dom
    -- foo clk bools = provideKnownDomainFrom clk (unsafeFromActiveHigh bools)

    invertReset :: Reset dom -> Reset dom
    invertReset r = provideKnownDomainFrom r unsafeToReset . fmap not . unsafeFromReset $ r

    ```
    This helper function take something containing a KnownDomain contraint as first argument, extracts it and provides it to the second argument.
 3. Use the new `ExtractClockDom` and `ExtractResetDom` PatternSynonyms to extract the KnownDomain for you
    <details>
    <summary>Click here for its definition</summary>

     ```haskell
     pattern ExtractResetDom
     :: ()              -- constraint required to match the pattern
     => KnownDomain dom -- constraint provided by matching the pattern
     => Reset dom
     pattern ExtractResetDom <- Reset {}
     -- and the same for ExtractClockDom
     {-# COMPLETE ExtractResetDom #-}
     ```
     </details>

     ```haskell
     invertReset :: Reset dom -> Reset dom
     invertReset r@ExtractResetDom = unsafeToReset . fmap not . unsafeFromReset $ r
     ```
     `ExtractResetDom` is just a synonyms for the pattern `Reset{}`. But limited in such a way that it is safe to use, and won't generate broken (V)HDL, and therefore is safe to export from `Clash.Prelude` etc.
 4. Change or make variants of (some of) these that take an additional `Clock dom` argument.

    I'm think this could be convenient for: `noReset`, `traceSignal`, `traceVecSignal` and maybe `unsafeFromActive[High,Low]`.

Options `0.` and `1.` are nothing new, they were already possible before this PR.
Options `2.` and `3.` are new options added in this PR, but I'm unsure if it makes sense to add both, or if that would just clutter up the API unnecessarily.
Also I'm not happy with the current names, they could do with some bike shedding.
Option `4.` I haven't implemented, could be convenient, but would be more incompatible API changes, or add more slightly different variants of existing functions.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Prelude/BlockRam.hs
+++ b/clash-prelude/src/Clash/Prelude/BlockRam.hs
@@ -838,11 +838,28 @@ blockRamPow2 = \cnt rd wrM -> withFrozenCallStack
 
 {- | Create a read-after-write block RAM from a read-before-write one
 
-#if __GLASGOW_HASKELL__ >= 908 && !defined(CLASH_MULTIPLE_HIDDEN)
+#if defined(CLASH_MULTIPLE_HIDDEN)
 >>> :t readNew (blockRam (0 :> 1 :> Nil))
 readNew (blockRam (0 :> 1 :> Nil))
   :: ...
      ...
+     ...
+     ... =>
+     Signal dom addr -> Signal dom (Maybe (addr, a)) -> Signal dom a
+
+#else
+#if __GLASGOW_HASKELL__ >= 908
+>>> :t readNew (blockRam (0 :> 1 :> Nil))
+readNew (blockRam (0 :> 1 :> Nil))
+  :: ...
+     ...
+     ... =>
+     Signal dom addr -> Signal dom (Maybe (addr, a)) -> Signal dom a
+
+#elif __GLASGOW_HASKELL__ >= 902
+>>> :t readNew (blockRam (0 :> 1 :> Nil))
+readNew (blockRam (0 :> 1 :> Nil))
+  :: ...
      ... =>
      Signal dom addr -> Signal dom (Maybe (addr, a)) -> Signal dom a
 
@@ -856,6 +873,7 @@ readNew (blockRam (0 :> 1 :> Nil))
      ... =>
      Signal dom addr -> Signal dom (Maybe (addr, a)) -> Signal dom a
 
+#endif
 #endif
 -}
 readNew
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -48,6 +48,10 @@ let
           "ghc-typelits-natnormalise"
           "${ghc-typelits-natnormalise}"
           { };
+
+      # doctest-parallel = hprev.callHackage "doctest-parallel" "0.3.1" { };
+      doctest-parallel = hprev.callHackageDirect { pkg = "doctest-parallel"; ver = "0.3.1"; sha256 = "sha256-j0mk4RTEau6Cl90E8vMWZiA/ldRRurtLSYbGGE+q/vk="; } {};
+
     };
 
   # An overlay with the packages in this repository.


martijnbastiaan:
Overall this looks good, although it will take me more time to review it properly. Concerning the bike shedding, here are my thoughts:

0. **Don't do anything.** From the looks of it, we have quite a few "normal" functions in `clash-prelude` that want to extract `KnownDomain`s. "Normal" in the sense that these are functions our users would probably write. Its probably best to dogfood a solution and couple them to the changes. Still, this solution doesn't necessarily worsen anything. (I.e., a proposal still propagating `KnownDomain` in some places is still a strict improvement.)

0. **Make your code monomorphic** Similar reasoning to (0), with the addition that you can't really monomorphize library functions.

0. **Use the new `provideKnownDomainFrom`**. I like this. It is explicit, and our users are already exposed to similar functions. I'd prefer the name `withKnownDomain`/`withKnownDomainFrom` to make it fit in with the IP functions.

0. **Use the new `ExtractClockDom` and `ExtractResetDom` `PatternSynonyms` to extract the `KnownDomain` for you**. In general I think pattern synonyms should only be applied whenever the act like a proper constructor (and you never need to access _actual_ constructors). E.g., the way `Seq` does it. I think this, because for any other use it's yet another Haskell concept for users to grasp, which syntactically looks exactly like (de)construction potentially making it very confusing.

0. **Change or make variants of (some of) these that take an additional Clock dom argument.** I've already found it confusing in practice for `unsafeSynchronizer`, so I'd rather not do this. Plus, like you said, this would hard break the current API. (The latter point weighing the most.)

An alternative take on (3) would be to wrap existing types (`Clock`/`Reset`/`Enable`) and make them carry the `KnownDomain` constraint. Their constructors could be safely exported. Though this also feels kinda dirty..

My preference would be (2), (0), (3).

PR #2590, opened 2023-10-16
Remove KnownDomain preparation

leonschoorl:
In preparation for a 1.8 release before merging #2589, this contains the non backwards compatible API changes of #2589.

And some minor cleanup and improvements ([Cleanup clash-lib.cabal](https://github.com/clash-lang/clash-compiler/commit/02015b0c6dd56a60e491f3d91096feae73453319) and [Improve blackbox error message](https://github.com/clash-lang/clash-compiler/commit/930f1ce49c5e6f3f0a9e2743e1b4c96cbe29e273))

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
@@ -185,10 +185,11 @@ verifyBlackBoxContext bbCtx (N.BBTemplate t) =
             Just n ->
               case indexMaybe (bbInputs bbCtx) n of
                 Just _ -> Nothing
-                Nothing ->
-                  Just ( "Blackbox required at least " ++ show (n+1)
-                      ++ " arguments, but only " ++ show (length (bbInputs bbCtx))
-                      ++ " were passed." )
+                Nothing -> do
+                  let str = fromJust (fmap Text.unpack (getAp $ prettyElem e))
+                  Just ( "Blackbox used \"" ++ str ++ "\""
+                      ++ ", but only " ++ show (length (bbInputs bbCtx))
+                      ++ " arguments were passed." )
 
 extractLiterals :: BlackBoxContext
                 -> [Expr]
--- a/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
@@ -57,7 +57,7 @@ import           Text.Read                       (readEither)
 import           Text.Trifecta.Result            hiding (Err)
 
 import           Clash.Backend
-  (Backend (..), Usage (..), AggressiveXOptBB(..), RenderEnums(..))
+  (Backend (..), DomainMap, Usage (..), AggressiveXOptBB(..), RenderEnums(..))
 import           Clash.Netlist.BlackBox.Parser
 import           Clash.Netlist.BlackBox.Types
 import           Clash.Netlist.Types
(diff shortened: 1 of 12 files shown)


clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -1276,7 +1276,8 @@ unsafeFromReset (Reset r) = r
 -- __NB__: You probably want to use 'unsafeFromActiveLow' or
 -- 'unsafeFromActiveHigh'.
 unsafeToReset
-  :: Signal dom Bool
+  :: KnownDomain dom
```
christiaanb, line 1279: The `KnownDomain` constraint is unused. Isn't GHC technically allowed to prune it?

DigitalBrains1: I think that's okay here. As I understand it, this is purely to make the API change now already for 1.8.0 so we can change the implementation later on. So only the user-facing API is what matters here, GHC can prune it without problems. Once the underlying code changes, the constraint will be needed, and it will no longer be pruned.

Huh. Even in general I think it's okay if it is pruned. The constraint is only needed in /Haskell/, and GHC can do what it wants as long as the code continues to work. We worry about stuff being pruned because we'd like to access it during HDL generation, but HDL generation doesn't need the `KnownDomain` for this function, it's a wire, nothing more.

christiaanb: Ah, yes, you're right

clash-prelude/src/Clash/Tutorial.hs
```diff
@@ -1289,16 +1289,16 @@ a general listing of the available template holes:
 * @~TAG[N]@: Name of given domain. Errors when called on an argument which is not
   a 'KnownDomain', 'Reset', or 'Clock'.
 * @~PERIOD[N]@: Clock period of given domain. Errors when called on an argument
-  which is not a 'KnownDomain' or 'KnownConf'.
+  which is not a 'Clock', 'Reset', 'KnownDomain' or 'KnownConf'.
 * @~ISACTIVEENABLE[N]@: Is the @(N+1)@'th argument a an Enable line NOT set to a
   constant True. Can be used instead of deprecated (and removed) template tag
 * @~ISSYNC[N]@: Does synthesis domain at the @(N+1)@'th argument have synchronous resets. Errors
-  when called on an argument which is not a 'KnownDomain' or 'KnownConf'.
+  when called on an argument which is not a 'Reset', 'Clock', 'Enable', 'KnownDomain' or 'KnownConf'.
```
DigitalBrains1, line 1296:
```suggestion
when called on an argument which is not a 'Clock', 'Reset', 'Enable', 'KnownDomain' or 'KnownConf'.
```

leonschoorl: I specifically chose this order. My intent was to put the most logical one first.
`~ISSYNC[n]` returns whether the domain of argument n uses synchronous resets.
And I thought it makes most sense to use it on a Reset.

DigitalBrains1: Ah okay. It just stood out to me as different than in other places.

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -584,52 +585,80 @@ renderElem b (IF c t f) = do
...
+  ClockN dom -> go dom
+  Reset dom  -> go dom
+  Enable dom -> go dom
+  Product _DiffClock _ [Clock dom,_clkN] -> go dom
+  t -> error $ $(curLoc) ++ "Don't know how to get a Domain out of HWType: " <> show t
+ where
+  go dom = do
+    doms <- getDomainMap
+    case HashMap.lookup dom doms of
+      Nothing -> error $ "Can't find domain " <> show dom
```
martijnbastiaan, line 659: This is definitely a bug if this ever happens right? Maybe:

```suggestion
      Nothing -> error $ "Can't find domain " <> show dom <> ". Please report an issue at https://github.com/clash-lang/clash-compiler/issues."
```

martijnbastiaan (approved):

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
leonschoorl: Good idea, thanks

PR #2595, opened 2023-10-27
No more testsuite failures on questa-2023.1 and vivado-2023.1

christiaanb:
Running:
```
cabal run clash-testsuite
```
Used to fail on the following tests:
```
BlackBox/T1524/SystemVerilog/ModelSim
DDR/DDRout/Verilog/tools/Vivado/testBenchUA;testBenchUS;techbenchGA;testbenchGS
HOPrim/Transpose/SystemVerilog/ModelSim
Vector/VMerge/SystemVerilog/ModelSim
```
On my machine, which has:
* Questa Intel FPGA Stater Edition 2023.1
* Xilinx Vivado 2023.1

With the commits in this PR, all tests pass on my machine.

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -669,7 +669,20 @@ runClashTest = defaultMain $ clashTestRoot
                                                         , "testBenchUS"
                                                         , "testBenchGA"
                                                         , "testBenchGS"
-                                                        ]}
+                                                        ]
+                         , hdlLoad = hdlLoad def \\ [Vivado]
+                         , hdlSim = hdlSim def \\ [Vivado]
+                         }
+          in runTest "DDRout" _opts
+        , let _opts = def{ buildTargets = BuildSpecific [ "testBenchUA"
+                                                        , "testBenchUS"
+                                                        , "testBenchGA"
+                                                        , "testBenchGS"
+                                                        ]
+                         , hdlLoad = [Vivado]
+                         , hdlSim = [Vivado]
+                         , clashFlags=["-fclash-hdlsyn", "Vivado"]
+                         }
           in runTest "DDRout" _opts
         ]
       , clashTestGroup "DSignal"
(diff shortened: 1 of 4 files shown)


tests/Main.hs
```diff
@@ -669,7 +669,9 @@ runClashTest = defaultMain $ clashTestRoot
                                                         , "testBenchUS"
                                                         , "testBenchGA"
                                                         , "testBenchGS"
-                                                        ]}
+                                                        ]
+                         , clashFlags=["-fclash-hdlsyn", "Vivado"]
+                         }
```
DigitalBrains1, line 674:
```suggestion
                                                        ]
                         , hdlLoad = hdlLoad def \\ [Vivado]
                         , hdlSim = hdlSim def \\ [Vivado]
                         }
          in runTest "DDRout" _opts
        , let _opts = def{ buildTargets = BuildSpecific [ "testBenchUA"
                                                        , "testBenchUS"
                                                        , "testBenchGA"
                                                        , "testBenchGS"
                                                        ]
                         , hdlLoad = [Vivado]
                         , hdlSim = [Vivado]
                         , clashFlags=["-fclash-hdlsyn", "Vivado"]
                         }
          in runTest "DDRout" _opts
```

This way, we test the other HDL tools without `-fclash-hdlsyn Vivado` and only Vivado with that set. Otherwise, our test bench doesn't exercise the non-Vivado path, and we are free to introduce errors there that will not be spotted by CI...

Slight disadvantage: there now are two tests called `..tests.shouldwork.DDR.DDRout.VHDL.clash (gen)`, same for Verilog and SystemVerilog (and note that the latter is only the `clash (gen)` stage as we don't do SystemVerilog in Vivado yet). If we'd like to deal with this properly, we could special-case this circumstance and add `hdlSyn :: [Sim]` to `TestOptions`, which would instruct the machinery to add `-fclash-hdlsyn X` to every sim `X` listed in `hdlSyn`. Then we can also vary the test names based on that. This might be a good long-term solution, that way we can mark those tests which fail without a proper `-fclash-hdlsyn`. But for now, I don't think we really need it yet, if you'd rather get on with more important stuff :-).

christiaanb: We should probably log this as an improvement on the issue tracker once this PR is merged.

For now, I've added your suggestion to the testsuite.

PR #2608, opened 2023-11-16
Expire artifacts in a week by default

martijnbastiaan:
We're currently using a whopping 523.5 GiB project storage on GitLab.

## Still TODO:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,8 @@
 # Make all tasks interruptible by default
 default:
   interruptible: true
+  artifacts:
+    expire_in: 1 week
 
 include:
   - '/.ci/gitlab/common.yml'


leonschoorl:
While it's probably good to have some sensible default.
I don't think this will change anything for the current situation.
Looking for artifacts configured I found:
 * [.gitlab-ci.yml#L84](https://github.com/clash-lang/clash-compiler/blob/master/.gitlab-ci.yml#L84)
 * [.ci/gitlab/benchmark.yml#L45](https://github.com/clash-lang/clash-compiler/blob/master/.ci/gitlab/benchmark.yml#L45)
 * [.ci/gitlab/publish.yml#L11](https://github.com/clash-lang/clash-compiler/blob/master/.ci/gitlab/publish.yml#L11)
 * [.ci/gitlab/test.yml#L68](https://github.com/clash-lang/clash-compiler/blob/master/.ci/gitlab/test.yml#L68)

All of which already have expiration defined, so those will override the default you add here I assume.

DigitalBrains1 (approved):
If you want this I'm fine with it but I'm also fine with leaving it as it is because it doesn't seem to accomplish much if anything (I didn't check if there's even anything that produces artifacts but does not have an explicit expiry already).

martijnbastiaan:
Well it doesn't seem like the 500 GB is caused by this. The default expiration period of 30 days shouldn't amount to this. It's more likely that it's all the stale branches, like you both noticed. Closing!

PR #2611, opened 2023-11-21
Workaround "Pattern match has inaccessible right hand side" bug

leonschoorl:
When using GHC < 9 `.../ICE40/Blackboxes/IO.hs` sometimes(?) warns:
```
    Pattern match has inaccessible right hand side
    In a pattern binding in
         a pattern guard for
           an equation for ‘sbioTemplate’:
        [_HasCallStack,
         (clk, clkTy, _),
         (en, _enTy, _),
         (pinConfig, _pinTy, pinConfigLiteral -> ()),
         (packagePin, packagePinTy, _),
         (latchInput, Bit, _),
         (dOut0, Bit, _),
         (dOut1, Bit, _),
         (outputEnable, Bool, _)]
        <-
        ...
```
And because we run with `-Werror` on CI this causes compilation to fail. This works around that by explicitly turning it back into a warning.

## Still TODO:

### DIFF
--- a/clash-cores/src/Clash/Cores/LatticeSemi/ICE40/Blackboxes/IO.hs
+++ b/clash-cores/src/Clash/Cores/LatticeSemi/ICE40/Blackboxes/IO.hs
@@ -5,10 +5,18 @@
 
   HDL generation functionality for LATTICE ICE IO primitives.
 -}
+
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ViewPatterns #-}
 
 {-# OPTIONS_HADDOCK hide #-}
+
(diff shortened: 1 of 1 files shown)


leonschoorl:
It's very weird this only happens sometimes.

I've compared the outputs of various `-ddump-...` options between runs where it warns and runs where it doesn't warn, but nothing much changes.
Except for `dump-simpl-iterations` where on a run with warnings `sbioTemplate` has a couple of extra unused local let bindings:
```haskell
k :: Identifier -> Ap (State s) Doc
k :: (Identifier, s) -> Data.Functor.Identity.Identity (Doc, s)
```
But they're unused and get eliminated later on in the simplifier and the final output of the simplifier is identical.
Therefor I believe we can safely ignore the warning.
And because it only seems to happen on GHC < 9, it's probably not worth reporting.

martijnbastiaan (approved):

PR #2624, opened 2023-12-12
Make applyDebug's free variable check take the context into account

### LINKED ISSUE #2623: caseCon introduces free variables because of evaluator changes
```haskell
topEntity = foo @System

foo :: forall dom. Signal dom (Vec 1 (Signed 2)) -> Signal dom Bool
foo = \input ->
    let
      scs :: Signal dom (Vec 1 Bool)
      scs = bundle $ map f $ unbundle input
    in
      fmap bar scs


bar :: (KnownNat n) => Vec (n+1) Bool -> Bool
bar = fold (&&) . map (id)

f :: Signal dom a -> Signal dom Bool
f = const $ pure $ True

```

When compiled with `-fclash-debug-invariants` (`DebugSilent` or higher) this crashes with:
```
    Clash error call:
    Clash.Rewrite.Util(226): Error when applying rewrite caseCon to:
    case Clash.Sized.Vector.map @GHC.Types.Bool[3674937295934324744]
           @GHC.Types.Bool[3674937295934324744]
           @(GHC.TypeNats.+[3674937295934325540] 0 1)
           (GHC.Base.id[8214565720323788222][GlobalId]
              @GHC.Types.Bool[3674937295934324744])
           x[6989586621679027447][LocalId] of
      Clash.Sized.Vector.Cons[8214565720323790062] m[12]
        (_co_[7] :: GHC.Prim.~#[3674937295934324842]
                      GHC.Num.Natural.Natural[3674937295934324786]
                      GHC.Num.Natural.Natural[3674937295934324786]
                      1
                      (GHC.TypeNats.+[3674937295934325540] 0 1))
          (el[6] :: GHC.Types.Bool[3674937295934324744])
          (rest[14] :: Clash.Sized.Vector.Vec[8214565720323790059]
                         0
                         GHC.Types.Bool[3674937295934324744]) ->
        el[6][LocalId]
    Result:
    let
      el[18] :: GHC.Types.Bool[3674937295934324744]
      = (λ(x[8214565720323804840] :: GHC.Types.Bool[3674937295934324744]) ->
          x[8214565720323804840][LocalId])
          ((λ(ds[6341068275337658375] :: Clash.Signal.Internal.Signal[8214565720323790017]
                                           "System"
                                           (Clash.Sized.Internal.Signed.Signed[8214565720323790051] 2)) ->
             Main.f_x[8214565720323803592][GlobalId] @"System")
             (Clash.Sized.Vector.head @0
                @(Clash.Sized.Internal.Signed.Signed[8214565720323790051] 2)
                ds[6341068275337658369][LocalId]))
    in el[18][LocalId]
    It introduces free variables.
    Before: x[6989586621679027447] :: Clash.Sized.Vector.Vec[8214565720323790059]
                                (GHC.TypeNats.+[3674937295934325540] 0 1)
    After: ds[6341068275337658369] :: Clash.Signal.Internal.Signal[8214565720323790017]
                                 "System"
                                 (Clash.Sized.Vector.Vec[8214565720323790059]
                                    1
                                    (Clash.Sized.Internal.Signed.Signed[8214565720323790051] 2))
      error, called at src/Clash/Rewrite/Util.hs:226:15 in clash-lib-1.9.0-inplace:Clash.Rewrite.Util
```
The problem is that since [Give evaluator acces to inscope let-bindings #2571](https://github.com/clash-lang/clash-compiler/pull/2571) the evaluator has access to the let bindings, and they may reference new free variables.

In this case it looks like:
```haskell
\ds[6341068275337658369] ->
  let x[6989586621679027447] = ... ds[6341068275337658369] ...
  in .................. case map id x[6989586621679027447] of ...
                    --  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  term being rewritten
```
The evaluator transforms this into some transformed form of `x` which contains `ds`, and then `checkDebug` complains there is now a new free variable in the expression: `ds`

leonschoorl:
Fixes #2623

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Rewrite/Util.hs
+++ b/clash-lib/src/Clash/Rewrite/Util.hs
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016     , Myrtle Software Ltd,
                     2017     , Google Inc.,
-                    2021-2022, QBayLogic B.V.
+                    2021-2023, QBayLogic B.V.
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -77,7 +77,7 @@ import           Clash.Core.Type             (Type (..), normalizeType)
 import           Clash.Core.Var
   (Id, IdScope (..), TyVar, Var (..), mkGlobalId, mkLocalId, mkTyVar)
 import           Clash.Core.VarEnv
-  (InScopeSet, extendInScopeSet, extendInScopeSetList, mkInScopeSet,
+  (InScopeSet, extendInScopeSet, extendInScopeSetList, mkInScopeSet, notElemInScopeSet,
    uniqAway, uniqAway', mapVarEnv, eltsVarEnv, unitVarSet, emptyVarEnv,
    mkVarEnv, eltsVarSet, elemVarEnv, lookupVarEnv, extendVarEnv, elemVarSet,
    differenceVarEnv)
@@ -173,12 +173,13 @@ apply = \s rewrite ctx expr0 -> do
     return ()
 
   if isDebugging opts
-    then applyDebug s expr0 hasChanged expr1
+    then applyDebug ctx s expr0 hasChanged expr1
     else return expr1
 {-# INLINE apply #-}
 
 applyDebug
-  :: String
+  :: TransformContext
+  -> String
   -- ^ Name of the transformation
   -> Term
   -- ^ Original expression
@@ -187,7 +188,7 @@ applyDebug
   -> Term
   -- ^ New expression
   -> RewriteMonad extra Term
-applyDebug name exprOld hasChanged exprNew = do
+applyDebug ctx name exprOld hasChanged exprNew = do
   nTrans <- Lens.use transformCounter
   opts <- Lens.view debugOpts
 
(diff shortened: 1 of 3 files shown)


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -266,6 +272,17 @@ applyDebug name exprOld hasChanged exprNew = do
...
     after  = showPpr exprNew
 
+-- NOTE [Filter free variables]
+-- Since [Give evaluator acces to inscope let-bindings #2571](https://github.com/clash-lang/clash-compiler/pull/2571)
+-- the evaluator can rewrite expressions using let bindings from the 'TransformContext',
+-- these bindings may reference other things bound in the context which weren't
+-- in the expression before, and in doing so introduces new free variables and
+-- fails this check for new free variables.
+-- To prevent this we filter out all variables from bound in the context.
+-- But only during a caseCon transformation, to not weaken this check unnecessarily.
```
martijnbastiaan, line 282:
```suggestion
-- Since [Give evaluator acces to inscope let-bindings #2571](https://github.com/clash-lang/clash-compiler/pull/2571)
-- the evaluator can rewrite expressions using let bindings from the 'TransformContext'.
-- These bindings may reference other things bound in the context which weren't
-- in the expression before, and in doing so introduces new free variables. Consequently,
-- this fails a check that prevents new free variables from being introduced -
-- as this is a bug in all other cases. To prevent this we filter out all
-- variables from bound in the context. We only do this during a 'caseCon' transformation,
-- to not weaken this check unnecessarily.
```

I've reformatted / broke up sentences to make it easier to read.

I'm not sure about this though:

> We only do this during a 'caseCon' transformation,
> to not weaken this check unnecessarily.

Is this because the evaluator is only called in `caseCon`?

martijnbastiaan (approved):
LGTM in general! Reformat of comment + question about it in the comments.

clash-lib/src/Clash/Rewrite/Util.hs
martijnbastiaan: @christiaanb maybe you could comment on this?

christiaanb: The evaluator is called in multiple places, including `caseCon`. But as @leonschoorl points out, the "introduce free variables" check is quite useful to catch bugs in the Clash compiler, so we want to leave it enabled in as many places as possible. So it's only disabled in `caseCon` as we have a confirmed "false positive" for that transformation.

martijnbastiaan: Okay, this sounds like: we're aware this is an issue, and we're going to wait until another transformation calling the evaluator is going to blow up..

What we should do is check the other transformations calling the evaluator, check whether they suffer from the same issue, and add a comment explaining why we do (not) need a similar approach as introduced in this PR.

PR #2625, opened 2023-12-18
CI: change nix-build log format to be less noisy

leonschoorl:
Change nix-build log format to be less noisy, so we don't overflow the log limit on CI anymore.

Also set max-silent-time to 1h.
This way if one of the derivations hangs, and doesn't output anything anymore for 1 hour,
 nix will detect it and report which one it was.

## Still TODO:


DigitalBrains1 (approved):
LGTM

PR #2637, opened 2024-01-11
clash-prelude: Add ShowX (Proxy a) instance

bgamari:
Both I and `clash-protocols` have found this instance useful, leading to overlapping instances. Let's put it where it belongs.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/XException.hs
+++ b/clash-prelude/src/Clash/XException.hs
@@ -63,6 +63,7 @@ import qualified Data.List.Infinite  as Inf
 import           Data.List.Infinite  (Infinite (..))
 import           Data.List.NonEmpty  (NonEmpty)
 import           Data.Ord            (Down (Down))
+import           Data.Proxy          (Proxy)
 import           Data.Ratio          (Ratio, numerator, denominator)
 import qualified Data.Semigroup      as SG
 import qualified Data.Monoid         as M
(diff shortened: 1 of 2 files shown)


bgamari:
One question is how this should be versioned until the next release is made. I would suggest that the version of `clash-prelude` be bumped such that downstream users can rely on CPP to guard existing orphans.

clash-prelude/src/Clash/XException.hs
```diff
@@ -397,6 +398,8 @@ printX :: ShowX a => a -> IO ()
 printX x = putStrLn $ showX x
 
 instance ShowX ()
+-- @since 1.8.2
```
martijnbastiaan, line 401: To be picked up by Haddock it needs to be:

```suggestion
-- | @since 1.8.2
```

Right?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for the Clash project
 
+## 1.8.2 *TBD*
+
+* Add `ShowX`, `NFDataX` instances for `Proxy`
+
```
martijnbastiaan, line 6: Could you follow the procedure outlined in [`changelog/README.md`](https://github.com/clash-lang/clash-compiler/blob/master/changelog/README.md)?

martijnbastiaan (changes requested):
In general LGTM!

martijnbastiaan:
> One question is how this should be versioned until the next release is made. I would suggest that the version of clash-prelude be bumped such that downstream users can rely on CPP to guard existing orphans.

The next release (for 1.8) will be 1.8.2, which should be guardable using `MIN_VERSION_clash_prelude`.

clash-prelude/src/Clash/XException.hs
bgamari: Good catch!

PR #2651, opened 2024-01-31
Fix Clash.Primitives.DSL.tuple on GHC-9.6

### LINKED ISSUE #2512: Clash generates invalid HDL on GHC 9.6 for `XpmCdcHandshake` primitive
More to follow.. is just want an issue ID for now.

leonschoorl:
It had the module name of tuples hardcoded to GHC.Tuple, but that changed in GHC-9.6.
This would result in type mismatches in the generated VHDL.

Fixes #2512

## Still TODO:


martijnbastiaan (approved):
Nice!

DigitalBrains1:
So the _only_ thing in our test suite that triggered this bug was in a Vivado-specific test? I think we should write a regression test that doesn't rely on Vivado.

Issue #2656, opened 2024-02-04
`unknown RTS option: -xm20000000`


paulyoung:
If I try to build `clash-ghc` using Nix, `clash-prelude` fails in the `checkPhase` with `unknown RTS option: -xm20000000` when trying to run `doctests` and `unittests`.

I'm using an M1 Macbook Pro.

If I do the following in `nix/overlay.nix` the error goes away.

```diff
       clash-prelude =
-        hprev.callCabal2nixWithOptions
+        hprev.callCabal2nix
           "clash-prelude"
           ../clash-prelude
-          "--flag workaround-ghc-mmap-crash"
           { };
```

I've tried to override this using various Nix mechanisms (e.g. `override`, `overrideAttrs`, a custom overlay, `pkgs.haskell.lib.dontCheck`) and while some of these work when building `clash-prelude` in isolation, it doesn't help when building `clash-ghc` for some reason that I don't understand.

paulyoung:
> If I do the following in `nix/overlay.nix` the error goes away.

I spoke too soon. While the build succeeds, actually trying to use `clash` results in a runtime error.

```
nix build .#clash-ghc
./result/bin/clash --help
clash: unknown RTS option: -xm20000000
...
```

paulyoung:
I made an additional change to `nix/overlay.nix` and unsurprisingly the runtime error is gone.

```diff
       clash-ghc =
         let
           unmodified =
-            hprev.callCabal2nixWithOptions
+            hprev.callCabal2nix
               "clash-ghc"
-              ../clash-ghc
-              "--flag workaround-ghc-mmap-crash" {
+              ../clash-ghc {
               inherit (hfinal) clash-lib clash-prelude;
             };
```

```
nix build .#clash-ghc
./result/bin/clash --help
Usage:

    clash [command-line-options-and-input-files]

...
```

paulyoung:
> I've tried to override this using various Nix mechanisms

To clarify, what I meant here is that I have the flake in this repository as an input to my own flake.

DigitalBrains1:
It would seem the `-xm` flag is only on x86_64:
https://gitlab.haskell.org/ghc/ghc/-/blob/0df8ce27f1c418fee1ba860f1c6575f66cae2ca7/rts/RtsFlags.c?page=2#L1756

We had not realised this. We'll need to make it such that the flag is only passed on x86_64.

paulyoung:
[...]

PR #2657, opened 2024-02-04
cabal: Make workaround-ghc-mmap-crash a noop on non-x86_64

rowanG077:
Fixes [#2656](https://github.com/clash-lang/clash-compiler/issues/2656)


DigitalBrains1:
Ah, thank you!

I'm thinking we should (also?) go one layer deeper and have the flag be a no-op on anything but x86_64, either using CPP in all call sites or in all the `.cabal` files if Cabal knows how to deal with these situations.

rowanG077:
I have moved the checks to the cabal files.

DigitalBrains1 (approved):
Makes perfect sense! Thanks!

PR #2662, opened 2024-02-08
VIO and ILA: explicitly check setName

### LINKED ISSUE #2654: `__VOID_TDECL_NOOP__` is not correctly matched
As discovered during the discussion of #2649 by @christiaanb, the following pattern match does not seem to work correctly
https://github.com/clash-lang/clash-compiler/blob/5e01ae2aec13f7b91cc8a6f5475feb4f24bf50b5/clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs#L278 as the following example
```haskell
module Test where

import Clash.Explicit.Prelude
import Clash.Cores.Xilinx.Ila

topEntity :: Clock System -> Signal System ()
topEntity clk =
    hwSeqX (ilaInst (ilaConfig ("a" :> Nil)))
  $ hwSeqX (ilaInst (ilaConfig ("b" :> Nil)))
    pure ()
 where
  ilaInst :: IlaConfig 1 -> Signal System ()
  ilaInst cfg = ila cfg clk (pure 0 :: Signal System (Unsigned 1))
```
still produces 
```
...
topEntity_ila_E981CE816A1232C3 topEntity_probe___VOID_TDECL_NOOP___0
  (.clk (c$pTS0), .probe0 (a));
...
topEntity_ila_E981CE816A1232C3 topEntity_probe___VOID_TDECL_NOOP__
  (.clk (c$pTS0), .probe0 (b));
...
```

DigitalBrains1:
If `Clash.Magic.setName` was used, use that name for the instance. Otherwise, use a fixed default name.

Fixes #2654

## Still TODO:

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
@@ -29,7 +29,7 @@ import Data.List (zip4, group)
 import Data.List.Infinite((...), Infinite((:<)))
 import Data.Proxy (Proxy(..))
 import Data.String.Interpolate (__i)
-import Data.Maybe (isJust)
+import Data.Maybe (isJust, fromMaybe)
 import Data.Text.Prettyprint.Doc.Extra (Doc)
 import GHC.Stack (HasCallStack)
 import GHC.TypeLits (KnownNat, SomeNat(..), someNatVal)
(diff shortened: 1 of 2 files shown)


martijnbastiaan (approved):
But in general LGTM!

clash-cores/src/Clash/Cores/Xilinx/Ila/Internal.hs
```diff
@@ -148,15 +148,18 @@ areEqual :: Eq a => [a] -> Maybe a
 areEqual = \case { [x:_] -> Just x; _ -> Nothing } . group
 
 ilaBBF :: HasCallStack => BlackBoxFunction
-ilaBBF _isD _primName args _resTys = Lens.view tcCache >>= go
+ilaBBF _isD _primName args _resTys = do
+  instName <- fromMaybe "ila_inst" <$> Lens.view setName
```
christiaanb, line 152:
```suggestion
  instName <- fromMaybe "ila_inst" <$> traverse affixName (Lens.view setName)
```
The above really is just a suggestion. Normally we add suffixes and prefixes to `setName`. I know this is not what we want for ILAs, but now the ILA is the odd one out with regards to `setName`.

DigitalBrains1: No, I agree, this was a simple oversight. This is not the place to strip affixes; if we want to do that, we should do it elsewhere.

DigitalBrains1:
> Interesting! Didn't know this was a thing. Should we add a netlist output regression test?

What exactly would it test; that it doesn't regress to issue #2654? So, in essence that the netlist output should not contain `__VOID_TDECL_NOOP__` anywhere?

My tired brain won't come up with an analysis how useful this is, so I'm just going to go with your opinion. If you want it, I'll write it.

christiaanb:
> > Should we add a netlist output regression test?
> 
> What exactly would it test; that it doesn't regress to issue #2654? So, in essence that the netlist output should not contain `__VOID_TDECL_NOOP__` anywhere?
> 
> My tired brain won't come up with an analysis how useful this is, so I'm just going to go with your opinion. If you want it, I'll write it.

We already have HDL tests that check whether something is not in the generated HDL: https://github.com/clash-lang/clash-compiler/blob/22b3a581a87bf079203302e732008219531ad36b/tests/shouldwork/Vector/IterateCF.hs#L30

So I guess it could work like that?

PR #2664, opened 2024-02-08
Drop support for GHC 8.6 and 8.8

martijnbastiaan:
## Still TODO:

  - [x] Remove old code (guarded with CPP)


clash-ghc/clash-ghc.cabal
```diff
@@ -187,10 +183,7 @@ library
                       utf8-string               >= 1.0.0.0  && < 1.1.0.0,
                       vector                    >= 0.11     && < 1.0
 
-  if os(windows)
-    -- 8.8 is broken on Windows - it randomly segfaults
-    Build-Depends: ghc >= 8.6.0 && < 8.8.0 || >= 8.10.0 && < 9.9
-  elif os(darwin)
+  if os(darwin)
     -- 8.10 is broken on macOS - it exits tests with status code -11
     Build-Depends: ghc >= 8.6.0 && < 8.10.0 || >= 9.0.0 && < 9.9
```
leonschoorl, line 188:
```diff
-    Build-Depends: ghc >= 8.6.0 && < 8.10.0 || >= 9.0.0 && < 9.9
+    Build-Depends: ghc >= 9.0.0 && < 9.9
```

README.md
```diff
@@ -65,8 +65,8 @@ Note that release branches might contain non-released patches.
 ## GHC compatibility
 |      | Linux | Windows | macOS | Clash (released) | Clash (development version)
 |------|-------|---------|-------|------------------|--------------------------
-| 8.6  | ✔️     | ✔️       | ✔️     | 1.0 - 1.6        | ✔️
-| 8.8  | ✔️     | ❌      | ✔️     | 1.0 - 1.6        | ✔️
+| 8.6  | ✔️     | ✔️       | ✔️     | 1.0 - 1.6        |  ️
+| 8.8  | ✔️     | ❌      | ✔️     | 1.0 - 1.6        |  ️
```
leonschoorl, line 69: Maybe put an :x: in the last column.
Or drop 8.6 and 8.8 from this table.

.gitlab-ci.yml
```diff
@@ -49,13 +49,8 @@ tests:
...
 
-      - GHC_VERSION: [8.10.7, 8.8.4]
+      - GHC_VERSION: [8.10.7]
         WORKAROUND_GHC_MMAP_CRASH: "yes"
-        RUN_TESTS: "nightly"
-
-      - GHC_VERSION: 8.6.5
-        WORKAROUND_GHC_MMAP_CRASH: "yes"
-        MULTIPLE_HIDDEN: "no"
         RUN_TESTS: "always"
```
DigitalBrains1, line 54: Shouldn't we test without `multiple-hidden` on 8.10.7 here?

README.md
DigitalBrains1: Let's not drop them, not even after we release 1.10. We provide information on historical releases in this table. I do wonder why 1.8 isn't listed in the table, this seems like an accidental omission.

I agree a cross would be better.

DigitalBrains1: Also, for another PR, the footnotes for that table say:
> We use Stack to test on Windows/macOS. There is no resolver yet that uses GHC 9.6, so this remains untested.

This is no longer the case, so we should start testing already!

martijnbastiaan: @DigitalBrains1 Could you check out https://github.com/clash-lang/clash-compiler/pull/2665? 

I originally intended the table to be an easy way to view which historical Clash/GHC versions are compatible with one another to make plotting an upgrade path easier. That is to say, I agree with @DigitalBrains1, let's keep them there. I'll fix the notation though!

.gitlab-ci.yml
martijnbastiaan: You're right

clash-ghc/clash-ghc.cabal
```diff
@@ -187,12 +183,9 @@ library
...
-    -- 8.8 is broken on Windows - it randomly segfaults
-    Build-Depends: ghc >= 8.6.0 && < 8.8.0 || >= 8.10.0 && < 9.9
-  elif os(darwin)
+  if os(darwin)
     -- 8.10 is broken on macOS - it exits tests with status code -11
-    Build-Depends: ghc >= 8.6.0 && < 8.10.0 || >= 9.0.0 && < 9.9
+    Build-Depends: ghc >= 9.0.0 && < 9.9
   else
     -- Unix
     Build-Depends: ghc >= 8.6.0 && < 9.9
```
christiaanb, line 191:
```suggestion
    Build-Depends: ghc >= 8.10.0 && < 9.9
```

christiaanb (approved):
LGTM

PR #2669, opened 2024-02-15
GitLab CI: Simplify finding binaries

DigitalBrains1:
The `build` job creates an artifact that subsequent jobs download to run the tests. The jobs that actually run the tests cannot use `cabal list-bin` because the artifact doesn't include enough of the build data. We can simplify finding the binaries by having the artifact include symlinks to the binaries, created in the `build` job where `cabal list-bin` still works.

The artifact included some bits that seem redundant since PR #2403, they have been removed.

The minimum version of `cabal-install` that contains a properly working `list-bin` command is 3.8.1.0. Because of the fixed security issue, we instead update `cabal-install` to 3.10.2.0 for all GHC's in CI.

GHC versions in CI are updated to the latest point release.

Fix mistake in the Docker image: use the same Ubuntu version for building and for running.

In GitLab CI, the `after_script` cannot use environment variables set in `before_script` or `script`. We use `zstd -T$THREADS` in `after_script`. Before this commit, that rendered to `zstd -T` which happens to mean "use all cores" and worked fine but was unintended.

Triggered child pipelines (`.ci/gitlab/test.yml`) did not have `interruptible: true` set. Move the `default` section into `.ci/gitlab/common.yml`.

## Still TODO:


.ci/docker/Dockerfile
```diff
@@ -121,7 +121,7 @@ RUN curl "https://downloads.haskell.org/~ghcup/$ghcup_version/x86_64-linux-ghcup
  && ghcup install ghc $ghc_version --set \
  && ghcup install cabal $cabal_version --set
 
-ARG UBUNTU_VERSION=focal-20210416
+ARG UBUNTU_VERSION=jammy-20240125
 FROM ubuntu:$UBUNTU_VERSION AS run
```
leonschoorl, line 125: To prevent future mismatches like this, consider removing the value in both places in the docker file, just have two times:
```
ARG UBUNTU_VERSION
```

And then setting the version in one place, in `.ci/docker/build-and-publish-docker-image.sh` with a `--build-arg`

DigitalBrains1: You're quite right.

martijnbastiaan (approved):

PR #2672, opened 2024-02-16
Fix GitHub CI, upgrade actions

DigitalBrains1:
PR #2669 broke GitHub CI "Build and Test" because it still used an older Docker image. This went unnoticed because the test is only run for external contributions and `master`.

Also upgraded some actions because the old ones were deprecated.

## Still TODO:


martijnbastiaan (approved):
Welp..

PR #2673, opened 2024-02-16
Look through annotations in tyfam solving

### LINKED ISSUE #2593: Bug: HDL generation failure with GHC 9.0.2, multiple hidden, synthesis attribute
In Clash `master` and `1.6`, there is a bug when generating HDL under very specific conditions. Only with GHC 9.0.2 (out of the versions we test in CI), with multiple hidden enabled, an annotation from `Clash.Annotations.SynthesisAttributes` and the following reproducer:

```haskell
topEntity ::
  Signal System Bit ->
  Signal System Bit `Annotate` 'StringAttr "break" "me"
topEntity dIn =
  exposeClockResetEnable (register 0) clk rst en dIn
 where
   clk = clockGen
   rst = resetGen
   en = enableGen
```

VHDL generation will produce the error:
```
<no location info>: error:
    Clash error call:
    Clash.Netlist.BlackBox.Util(585): IsActiveEnable: Expected Bool or Enable, not: Void Nothing
      error, called at src/Clash/Netlist/BlackBox/Util.hs:585:13 in clash-lib-1.7.0-inplace:Clash.Netlist.BlackBox.Util
      renderElem, called at src/Clash/Netlist/BlackBox/Util.hs:319:15 in clash-lib-1.7.0-inplace:Clash.Netlist.BlackBox.Util
```

Verilog and SystemVerilog instead say:
```
<no location info>: error:
    Clash error call:
    Clash.Netlist.BlackBox(290): Forced to evaluate untranslatable type: (resetGenN
      @(IfStuck
          TryDomainResult
          [...]
               (TryDomain (Signal "System" Bit -> Signal "System" Bit) (Signal "System" Bit -> Signal "System" Bit))))))
      error, called at src/Clash/Netlist/BlackBox.hs:290:21 in clash-lib-1.7.0-inplace:Clash.Netlist.BlackBox
```
              
Each of the following is enough to make the code produce HDL again:
* Disable multiple-hidden
* Remove `Annotate`
* Add type application/signature to either `clk`, `rst` or `en`
* Add type application to `exposeClockResetEnable`
* Eta-reduce `topEntity` (`topEntity = exposeClockResetEnable ...`)

Type-applying `register`, though, does not help.

I created a branch [issue-2593](https://github.com/clash-lang/clash-compiler/tree/issue-2593) which tests the reproducer in CI. The full file with the reproducer is [T2593.hs](https://github.com/clash-lang/clash-compiler/blob/issue-2593/tests/shouldwork/Issues/T2593.hs)
For 1.6, the branch is [issue-2593-1.6](https://github.com/clash-lang/clash-compiler/tree/issue-2593-1.6).

christiaanb:
1. Annotations are implemented as type synonyms.
2. GHC looks through type synonyms when matching type families
3. So Clash should do the same.

Fixes #2593

## Still TODO:

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -788,6 +788,7 @@ runClashTest = defaultMain $ clashTestRoot
         , outputTest "T2510" def{hdlTargets=[VHDL], clashFlags=["-DNOINLINE=OPAQUE"]}
 #endif
         , outputTest "T2542" def{hdlTargets=[VHDL]}
+        , runTest "T2593" def{hdlSim=[]}
         ] <>
         if compiledWith == Cabal then
           -- This tests fails without environment files present, which are only
--- a/tests/shouldwork/Issues/T2593.hs
+++ b/tests/shouldwork/Issues/T2593.hs
@@ -1,0 +1,14 @@
+module T2593 where
+
+import Clash.Prelude
+import Clash.Annotations.SynthesisAttributes
+
+topEntity ::
+  Signal System Bit ->
+  Signal System Bit `Annotate` 'StringAttr "breaka" "me"
+topEntity dIn =
+  exposeClockResetEnable (register 0) clk rst en dIn
+ where
+   clk = clockGen
+   rst = resetGen
+   en = enableGen
--- a/changelog/2024-02-16T16_42_52+01_00_fix2593
+++ b/changelog/2024-02-16T16_42_52+01_00_fix2593
@@ -1,0 +1,1 @@
+FIXED: HDL generation fails when using multiple-hidden feature in combination with synthesis attributes [#2593](https://github.com/clash-lang/clash-compiler/issues/2593)
--- a/clash-lib/src/Clash/Core/Type.hs
+++ b/clash-lib/src/Clash/Core/Type.hs
@@ -2,7 +2,7 @@
   Copyright   :  (C) 2012-2016, University of Twente,
                      2016     , Myrtle Software Ltd,
                      2017     , Google Inc.
-                     2021     , QBayLogic B.V.
+                     2021-2024, QBayLogic B.V.
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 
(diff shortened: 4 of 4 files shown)


martijnbastiaan (approved):

tests/Main.hs
```diff
@@ -788,6 +788,7 @@ runClashTest = defaultMain $ clashTestRoot
         , outputTest "T2510" def{hdlTargets=[VHDL], clashFlags=["-DNOINLINE=OPAQUE"]}
 #endif
         , outputTest "T2542" def{hdlTargets=[VHDL]}
+        , runTest "T2593" def{hdlSim=[]}
```
DigitalBrains1, line 791:
```suggestion
        , runTest "T2593" def{hdlLoad=[], hdlSim=[]}
```
We don't need to "load" them in HDL tools either, it's enough that it compiles.

tests/shouldwork/Issues/T2593.hs
```diff
@@ -0,0 +1,14 @@
+module T2593 where
+
+import Clash.Prelude
+import Clash.Annotations.SynthesisAttributes
+
+topEntity ::
+  Signal System Bit ->
+  Signal System Bit `Annotate` 'StringAttr "breaka" "me"
```
DigitalBrains1, line 8:
```suggestion
  Signal System Bit `Annotate` 'StringAttr "break" "me"
```
:-)

tests/Main.hs
DigitalBrains1: Ah it occurred to me to check and I see what you did is a pattern that's often used in the test suite... so maybe I was just overthinking it and it should just be as the rest is. Feel free to disregard this suggestion.

tests/shouldwork/Issues/T2593.hs
martijnbastiaan: It's a-me!

christiaanb: `break` is a systemverilog keyword, we cannot use it as a property name.

DigitalBrains1: > It's a-me!

PR #2676, opened 2024-02-21
Bump Verilator to v5.020

martijnbastiaan:
Verilator v5 implements its own (C++ co-routine based) scheduler, making it support delay statements. We can therefore remove all the Verilator specific code from our code base.

## Still TODO:

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -664,13 +664,23 @@ runClashTest = defaultMain $ clashTestRoot
                                                         , "testBenchUS"
                                                         ]}
           in runTest "DDRin" _opts
+
+          -- XXX: `ddrOut` contains a number of (implicit) parallel, coinciding
+          --      processes. Execution for these processes is left undefined in
+          --      the Verilog spec, nor is there a mechanism to get consistent
+          --      behavior (like in VHDL). Therefore, simulators are free to pick
+          --      any execution order they'd like. ModelSim and Verilator pick
+          --      differently, so when a test works in one, it doesn't in the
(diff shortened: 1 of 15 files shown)


tests/clash-testsuite.cabal
```diff
@@ -151,6 +151,7 @@ executable clash-testsuite
 
   build-depends:
     containers,
+    extra,
```
christiaanb, line 154: This change should go in the other commit.

christiaanb (approved):
LGTM

tests/Main.hs
```diff
@@ -662,13 +664,21 @@ runClashTest = defaultMain $ clashTestRoot
                                                         , "testBenchUS"
                                                         ]}
           in runTest "DDRin" _opts
+
+          -- XXX: `ddrOut` contains two concurrent always-blocks in the generated
+          --      Verilog. Execution order is not defined in the Verilog spec, so
+          --      simulators are free to pick an order. ModelSim and Verilator
+          --      disagree on the order, so when a test works in one, it doesn't
+          --      in the other. As a quick "fix" we disable Verilator, though we
```
leonschoorl, line 672: As discussed on slack, the problem isn't with the two always-blocks.
But it's between one of the always blocks and the continuous assignment of the result.
(To add to the confusion in SystemVerilog the assignment of the result is done in a third always-block instead. So there the problem is between two (out of three) always-blocks.)

You should probably make the description here just slightly more general.

martijnbastiaan: Yes, I'll generalize the description. Luckily Nix failed :-)

PR #2682, opened 2024-02-27
Add `Clash.Sized.Vector.ToTuple`

martijnbastiaan:
Newer GHCs (mistakenly) emit warnings when pattern matching on `Vec`s, with neither `:>` nor `Cons` being able to prevent it. `vecToTuple` allows users to convert `Vec`s to tuples, sidestepping GHCs exhaustiveness checker while not compromising on safety.

Nested tuples were considered as an alternative implementation, simplifying implementation and removing the need for `TemplateHaskell`. This would, however, compromise on user comfort while offering no expected real-world benefits.

-------------------

This is clearly a work around. Perhaps we should write a GHC plugin to circumvent these warnings (and depending on the solution upstream)?

## Documentation
Screenshot 2024-02-28 at 09-17-26 Clash Sized Vector ToTuple

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
+++ b/clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
@@ -1,0 +1,98 @@
+{-|
+Copyright  :  (C) 2024, Google LLC
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Tooling to safely work around @incomplete-uni-patterns@ warnings produced by GHC
+9.2 and up, or @incomplete-patterns@ warnings on all GHC versions. See 'vecToTuple'
+for more information and examples.
+-}
+
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilyDependencies #-}
+
+-- Purpose of this module
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
+
+-- For debugging:
+-- {-# OPTIONS_GHC -ddump-splices #-}
+
+module Clash.Sized.Vector.ToTuple (VecToTuple(..)) where
+
+import Clash.CPP
+import Clash.Sized.Vector
+import Clash.Sized.Vector.ToTuple.TH (vecToTupleInstances)
+
+import Data.Tagged (Tagged(..))
+
+#if MIN_VERSION_base(4,18,0)
+import Data.Tuple (Solo(MkSolo))
+#elif MIN_VERSION_base(4,16,0)
+import Data.Tuple (Solo(Solo))
+#endif
+
+{- $setup
+>>> :set -XMonoLocalBinds -XGADTs
+>>> import Clash.Sized.Vector
+-}
+
+class VecToTuple a where
+  type TupType a = r | r -> a
+
+  -- | Given a vector with three elements:
+  --
+  -- >>> myVec = (1 :> 2 :> 3 :> Nil) :: Vec 3 Int
+  --
+  -- The following would produce a warning even though we can be sure
+  -- no other pattern can ever apply:
+  --
+  -- >>> (a :> b :> c :> Nil) = myVec
+  --
+  -- 'vecToTuple' can be used to work around the warning:
+  --
+  -- >>> (a, b, c) = vecToTuple myVec
+  --
+  -- Of course, you will still get an error if you try to match a vector of the
+  -- wrong length:
+  --
+  -- >>> (a, b, c, d) = vecToTuple myVec
+  -- ...
+#if __GLASGOW_HASKELL__ > 900
+  --     • Couldn't match type: (a, b, c, d)
+  --                      with: (Int, Int, Int)
+#else
+  --     • Couldn't match type: (Int, Int, Int)
+  --                             ^
+  --                      with: (a, b, c, d)
+#endif
+  -- ...
+  --
+  vecToTuple ::  a -> TupType a
+
+instance VecToTuple (Vec 0 a) where
+  type TupType (Vec 0 a) = Tagged a ()
+  vecToTuple _ = Tagged ()
+
+#if MIN_VERSION_base(4,18,0)
+-- | Instead of using 'vecToTuple' for @Vec 1 _@, you could also consider using 'head'
+instance VecToTuple (Vec 1 a) where
+  type TupType (Vec 1 a) = Solo a
+  vecToTuple (a1 :> _) = MkSolo a1
+#elif MIN_VERSION_base(4,16,0)
+-- | Instead of using 'vecToTuple' for @Vec 1 _@, you could also consider using 'head'
+instance VecToTuple (Vec 1 a) where
+  type TupType (Vec 1 a) = Solo a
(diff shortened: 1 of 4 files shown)


clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
```diff
@@ -0,0 +1,95 @@
+{-|
+Copyright  :  (C) 2024, Google LLC
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Tooling to work around incomplete pattern match warnings produced by newer
+GHC versions. See 'vecToTuple' for more information and examples.
```
leonschoorl, line 7: Please make "newer GHC versions" specific.
Also in the changelog

clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
```diff
@@ -0,0 +1,95 @@
...
+  vecToTuple (a1 :> Nil) = Solo a1
+#endif
+
+-- | __NB__: The documentation only shows instances up to /3/-tuples. By
+-- default, instances up to and including /12/-tuples will exist. If the flag
+-- @large-tuples@ is set instances up to the GHC imposed limit will exist. The
+-- GHC imposed limit is either 62 or 64 depending on the GHC version.
+instance VecToTuple (Vec 2 a) where
+  type TupType (Vec 2 a) = (a, a)
+  vecToTuple (a1 :> a2 :> Nil) = (a1, a2)
```
leonschoorl, line 93: You could make the functions slightly less strict by replacing the `Nil`s with `_`s.
And `conP 'Nil []` with `wildP` in the TH version.

clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
```diff
@@ -0,0 +1,95 @@
...
+  --
+  vecToTuple ::  a -> TupType a
+
+instance VecToTuple (Vec 0 a) where
+  type TupType (Vec 0 a) = Tagged a ()
+  vecToTuple Nil = Tagged ()
+
+#if MIN_VERSION_base(4,18,0)
+instance VecToTuple (Vec 1 a) where
+  type TupType (Vec 1 a) = Solo a
```
leonschoorl, line 79: Would it make sense to put haddock comment on the `Vec 1` instance that you could also use `head` instead?

clash-prelude/src/Clash/Sized/Vector/ToTuple/TH.hs
```diff
@@ -0,0 +1,50 @@
...
+  vecType = conT ''Vec `appT` litT (numTyLit n) `appT` varT (mkName "a")
+
+  -- associated type
+  tupTypeCon = conT (mkName "TupType")
+  aTypeLhs = tupTypeCon `appT` vecType
+  aTypeRhs = tupT [varT (mkName "a") | _ <- [1..n]]
+
+  -- vecToTuple
+  vecToTupleFunName = mkName "vecToTuple"
+  vecToTuplePat = appPsInfix '(:>) (map varP varNames ++ [wildP])
```
christiaanb, line 47: Do we need the laziness of `:>`? If people need to lazily evaluate the resulting tuple, they can just use a lazy match there. So perhaps we can just use `Cons` here, so we do not get the desugaring to `head` and `tail`.

clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
```diff
@@ -0,0 +1,98 @@
...
+#if MIN_VERSION_base(4,18,0)
+-- | Instead of using 'vecToTuple' for @Vec 1 _@, you could also consider using 'head'
+instance VecToTuple (Vec 1 a) where
+  type TupType (Vec 1 a) = Solo a
+  vecToTuple (a1 :> _) = MkSolo a1
+#elif MIN_VERSION_base(4,16,0)
+-- | Instead of using 'vecToTuple' for @Vec 1 _@, you could also consider using 'head'
+instance VecToTuple (Vec 1 a) where
+  type TupType (Vec 1 a) = Solo a
+  vecToTuple (a1 :> _) = Solo a1
```
christiaanb, line 87: `Solo` is lazy, so I think we can use `Cons` instead of `:>` here.

clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
```diff
@@ -0,0 +1,98 @@
...
+
+instance VecToTuple (Vec 0 a) where
+  type TupType (Vec 0 a) = Tagged a ()
+  vecToTuple _ = Tagged ()
+
+#if MIN_VERSION_base(4,18,0)
+-- | Instead of using 'vecToTuple' for @Vec 1 _@, you could also consider using 'head'
+instance VecToTuple (Vec 1 a) where
+  type TupType (Vec 1 a) = Solo a
+  vecToTuple (a1 :> _) = MkSolo a1
```
christiaanb, line 82: `MkSolo` is lazy, so I think we can use `Cons` instead of `:>` here.

clash-prelude/src/Clash/Sized/Vector/ToTuple.hs
```diff
@@ -0,0 +1,98 @@
...
+  vecToTuple (a1 :> _) = Solo a1
+#endif
+
+-- | __NB__: The documentation only shows instances up to /3/-tuples. By
+-- default, instances up to and including /12/-tuples will exist. If the flag
+-- @large-tuples@ is set instances up to the GHC imposed limit will exist. The
+-- GHC imposed limit is either 62 or 64 depending on the GHC version.
+instance VecToTuple (Vec 2 a) where
+  type TupType (Vec 2 a) = (a, a)
+  vecToTuple (a1 :> a2 :> _) = (a1, a2)
```
christiaanb, line 96: The tuple is lazy, so I think we can use `Cons` instead of `:>` here.

christiaanb (commented):
Was there a reason you use `:>` instead of `Cons` for the pattern match?

clash-prelude/src/Clash/Sized/Vector/ToTuple/TH.hs
martijnbastiaan: I don't think we need the laziness. I was just being lazy, no pun intended, by writing the thing I always write for `Vec` pattern matching.

martijnbastiaan:
> Was there a reason you use :> instead of Cons for the pattern match?

Discussed in the office: there is a difference in behavior. E.g. in the following:

```haskell
vecToTuplePat :: (Vec 2 a) -> (a, a)
vecToTuplePat (a1 :> a2 :> _) = (a1, a2)

-- True
pat :: Bool
pat = case vecToTuplePat (1 :> undefined) of
  ~(a, _) -> a `seq` True

vecToTupleCon :: (Vec 2 a) -> (a, a)
vecToTupleCon (a1 `Cons` a2 `Cons` _) = (a1, a2)

-- Bottom
con :: Bool
con = case vecToTupleCon (1 :> undefined) of
  ~(a, _) -> a `seq` True
```

then `pat ~ True`, while `con ~ _|_`. We're not sure whether this will be an issue in practice, but we've decided to err on the safe side and be lazy.

christiaanb (approved):

martijnbastiaan:
@leonschoorl I applied your suggestions, thanks!

PR #2686, opened 2024-03-01
Add boundary proofs to `dfold` and `smap`

kleinreact:
The PR extends `dfold` with a boundary proof for the folded function and introduces `smapWithBounds` offering a boundary proof as well. `smapWithBounds` is introduced in addition to `smap`, as the served proof may lead to `untouchable` errors requiring additional type annotations. While this is fine for `dfold`, which requires advanced knowledge of the type system for usage anyway, `smap` should stay intuitive for beginners, which is why we introduce an additional version instead.

The PR also renames typenat variables `l` (_small letter `L`_) to `n` to make them better distinguishable from the constant `1`.

### DIFF
--- a/changelog/2024-04-08T06_51_45+00_00_smap_with_bounds
+++ b/changelog/2024-04-08T06_51_45+00_00_smap_with_bounds
@@ -1,0 +1,2 @@
+CHANGED: `dfold` now offers a proof witness for the upper bound of the vector size to the folding function. Note that this change may require additional type annotations, as solutions working in the past may complain with an untouchable type error now.
+ADDED: `smapWithBounds` extending `smap` via offering a proof witness for the upper bound of the vector size to the mapping function.
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
@@ -27,6 +27,7 @@ module Clash.GHC.Evaluator.Primitive
 
 import           Control.DeepSeq            (force)
 import           Control.Exception          (ArithException(..), Exception, tryJust, evaluate)
+import qualified Control.Lens               as Lens
 import           Control.Monad.State.Strict (State, MonadState)
 import qualified Control.Monad.State.Strict as State
 import           Control.Monad.Trans.Except (runExcept)
@@ -92,17 +93,19 @@ import           TysWiredIn          (tupleTyCon)
 import           Clash.Class.BitPack (pack,unpack)
 import           Clash.Core.DataCon  (DataCon (..))
 import           Clash.Core.Evaluator.Types
+import           Clash.Core.FreeVars (typeFreeVars)
 import           Clash.Core.HasType  (piResultTys, applyTypeToArgs)
 import           Clash.Core.Literal  (Literal (..))
 import           Clash.Core.Name
   (Name (..), NameSort (..), mkUnsafeSystemName)
 import           Clash.Core.Pretty   (showPpr)
+import           Clash.Core.Subst    (extendTvSubst, mkSubst, substTy)
 import           Clash.Core.Term
   (IsMultiPrim (..), Pat (..), PrimInfo (..), Term (..), WorkInfo (..), mkApps,
    PrimUnfolding(..), collectArgs)
 import           Clash.Core.Type
   (Type (..), ConstTy (..), LitTy (..), TypeView (..), mkFunTy, mkTyConApp,
-   splitFunForallTy, tyView)
+   normalizeType, splitFunForallTy, tyView)
 import           Clash.Core.TyCon
   (TyConMap, TyConName, tyConDataCons)
 import           Clash.Core.TysPrim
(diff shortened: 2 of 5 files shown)


martijnbastiaan:
I'm curious; what motivated you to implement this? Any concrete example where you need the boundary proof?

kleinreact:
I would use it [here](https://github.com/bittide/bittide-hardware/compare/staging...hitl-topologies#diff-ffba1bb951968297f0a0bf02d7b31bd54f347dba6b054dec31d990152baf29e1R64), but the currently used workaround is also sufficient in that case. It is required to use the `Resize` class in that case.

rowanG077:
I have ran into this a few times. And just now I ran into it again. I was trying to do some type magics to allow dynamic multi lane CRC to work. I tried to do something like this:

```Haskell
data CRCLaneParams (crcWidth :: Nat) (dataWidth :: Nat) (lanes :: Nat) where
  CRCLane0
    :: CircuitCRCParams crcWidth dataWidth
    -> CRCLaneParams crcWidth dataWidth 0
  CRCLaneN
    :: 1 <= lanes
    => SNat lanes
    -> CircuitCRCParams crcWidth ((lanes + 1) * dataWidth)
    -> CRCLaneParams crcWidth dataWidth (lanes - 1)
    -> CRCLaneParams crcWidth dataWidth lanes

foo :: CRCLaneParams 32 8 3
foo = CRCLaneN d3 (mkCircuitCRCParams $ cRC32_ETHERNET SNat)
        $ CRCLaneN d2 (mkCircuitCRCParams $ cRC32_ETHERNET SNat)
        $ CRCLaneN d1 (mkCircuitCRCParams $ cRC32_ETHERNET SNat)
        $ CRCLane0 (mkCircuitCRCParams $ cRC32_ETHERNET d8)

data SNatCMP a b where
  SNatLT2 :: forall a b . a <= (b - 1) => SNatCMP a b
  SNatEQ2 :: forall a b . a ~ b => SNatCMP a b
  SNatGT2 :: forall a b . (b+1) <= a => SNatCMP a b

-- | Get an ordering relation between two SNats
compareSNat2 :: forall a b . SNat a -> SNat b -> SNatCMP a b
compareSNat2 a b =
  case compare (snatToInteger a) (snatToInteger b) of
    LT -> unsafeCoerce (SNatLT2 @0 @1)
    EQ -> unsafeCoerce (SNatEQ2 @0 @0)
    GT -> unsafeCoerce (SNatGT2 @1 @0)

getCircuitParams
  :: lane <= lanes
  => CRCLaneParams crcWidth dataWidth lanes
  -> SNat lane
  -> CircuitCRCParams crcWidth ((lane + 1) * dataWidth)
getCircuitParams (CRCLane0 params) lane
  = case compareSNat2 d0 lane of
      SNatEQ2 -> params
      _ -> error "getCircuitParams: Absurd"
getCircuitParams (CRCLaneN lanes params next) lane
  = case compareSNat2 lane lanes of
      SNatLT2 -> getCircuitParams next lane
      SNatEQ2 -> params
      _ -> error "getCircuitParams: Absurd"

-- Cannot satisfy: l <= 3
engines = smap (\l _ -> getCircuitParams foo l) (repeat 0 :: Vec 4 Int)

```

martijnbastiaan:
Could you write a changelog entry? See `changelog/README.md` for instructions.

changelog/2024-04-08T06_51_45+00_00_smap_with_bounds
```diff
@@ -0,0 +1 @@
+* Add proof witness to `dfold` and `smapWithBounds`
```
martijnbastiaan, line 1:
```suggestion
* CHANGED: Add proof witness to `dfold` and `smapWithBounds`
```

For earlier (Slack?) discussion I understood that this can sometimes cause code not to compile due to ambiguous variables. Is this true? If so, this heads up should be added to the change log.

kleinreact:
>Why did we change `dfold`, but introduced an additional function for `smap`?

The primary argument was that `smap` may be used by beginners, which may be overwhelmed by the extended type signature required for the proof. On the other hand, `dfold` is not beginner friendly in terms of the type signature anyway, which is why it shouldn't hurt to introduce the change directly here.

martijnbastiaan:
Makes sense!

martijnbastiaan (approved):

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -4370,13 +4373,24 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
          0 -> reduce (valToTerm z)
          _ -> let (tyArgs,_)  = splitFunForallTy ty
                   (tyArgs',_) = splitFunForallTy (Either.rights tyArgs !! 2)
-                  TyConApp snatTcNm _ = tyView (Either.rights tyArgs' !! 0)
+                  ubpT = Either.rights tyArgs' !! 0
+                  fTVs = Lens.toListOf typeFreeVars ubpT
+                  Just tvN = List.find ((== "n") . nameOcc . varName) fTVs
+                  subst0 = extendTvSubst (mkSubst is0) tvN k'ty
+                  Just tvK = List.find ((== "k") . nameOcc . varName) fTVs
+                  subst1 = extendTvSubst subst0 tvK (LitTy (NumTy k'))
```
christiaanb, line 4381: `tvN` and `tvK` are probably in `Either.lefts tyArgs'`, no need to "find" them.

clash-lib/src/Clash/Normalize/PrimitiveReductions.hs
```diff
@@ -735,22 +739,35 @@ reduceDFold n aTy _kn _motive fun start arg (TransformContext is0 _ctx) = do
...
               snatTc <- UniqMap.lookup snatTcNm tcm
               Maybe.listToMaybe (tyConDataCons snatTc)
-            lbody = doFold (buildSNat snatDc) (n-1) (NE.toList vars)
+            ubp k = Maybe.fromMaybe
+              (error "reduceDFold: failed to extract upper bound proof") $ do
+                (_ltv:Right ubpT:_,_) <- pure (splitFunForallTy (inferCoreTypeOf tcm fun))
+                -- toListOf does not de-duplicate, but we know that there is only
+                -- one free variable in here, thus, taking the first element is fine
+                let fTVs = Lens.toListOf typeFreeVars ubpT
+                tvN <- find ((== "n") . nameOcc . varName) fTVs
```
christiaanb, line 752: Surely we don't have to "find" this `n` type variable by string name. It should be one of the type variables of `fun`.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
kleinreact: `Either.lefts tyArgs'` only contains `tvN`, while `fTVs`holds both `tvN` and `tvK`, but nothing more. I updated it in the sense that I take `tvN` from `Either.lefts tyArgs'` now and then pick `tvK` as the first non-`tvN` element in `fTVs`.

christiaanb (approved):

PR #2691, opened 2024-03-09
Make `foldl` strict, like `foldl'`

### LINKED ISSUE #2482: nrf and rnfX for Vec have a spaceleak
https://github.com/clash-lang/clash-compiler/commit/48a739f8a947c0877346d500b83d34e2eea779d1 changed the definitions of `deepseq` and `deepseqX` to:

```haskell
rnf = foldl (\() -> rnf) ()
rnfX v =seqX (foldl (\() -> rnfX) () v) ()
```

where both use `foldl` meaning we allocating a chain of `rnf` (cq `rnfX`) computations and only reducing the computation to NF at the very end.

We can see this when we do:
```haskell
{-# LANGUAGE BangPatterns, DataKinds, NoImplicitPrelude, GADTs, ScopedTypeVariables #-}
module Main where

import Clash.Prelude
import Clash.Sized.Vector
import Control.DeepSeq

rnfY :: NFData a => Vec n a -> ()
rnfY = foldl (\() -> rnf) ()

main :: IO ()
main = do
  let !_ = rnfY (repeat @1000000 (0 :: Int))
  return ()
```
```
$ ghc -O -fforce-recomp Main.hs 
Loaded package environment from /home/christiaan/devel/clash-compiler/.ghc.environment.x86_64-linux-9.6.2
[1 of 2] Compiling Main             ( Main.hs, Main.o )
[2 of 2] Linking Main [Objects changed]
$ ./Main +RTS -s
     418,784,560 bytes allocated in the heap
     413,417,200 bytes copied during GC
     102,759,576 bytes maximum residency (9 sample(s))
      21,341,496 bytes maximum slop
             233 MiB total memory in use (0 MiB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0        84 colls,     0 par    0.099s   0.099s     0.0012s    0.0037s
  Gen  1         9 colls,     0 par    0.256s   0.256s     0.0284s    0.0637s

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    0.070s  (  0.070s elapsed)
  GC      time    0.355s  (  0.355s elapsed)
  EXIT    time    0.000s  (  0.005s elapsed)
  Total   time    0.426s  (  0.430s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    5,951,813,489 bytes per MUT second

  Productivity  16.5% of total user, 16.2% of total elapsed
```
versus using a strict fold:
```haskell
{-# LANGUAGE BangPatterns, DataKinds, NoImplicitPrelude, GADTs, ScopedTypeVariables #-}
module Main where

import Clash.Prelude
import Clash.Sized.Vector
import Control.DeepSeq

foldl' :: forall a b n . (b -> a -> b) -> b -> Vec n a -> b
foldl' f = go
 where
  go :: b -> Vec m a -> b
  go z Nil = z
  go z (Cons x xs) =
    let z1 = f z x
     in seq z1 (go z1 xs)

rnfY :: NFData a => Vec n a -> ()
rnfY = foldl' (\() -> rnf) ()

main :: IO ()
main = do
  let !_ = rnfY (repeat @1000000 (0 :: Int))
  return ()
```
```
$ ghc -O -fforce-recomp Main.hs 
Loaded package environment from /home/christiaan/devel/clash-compiler/.ghc.environment.x86_64-linux-9.6.2
[1 of 2] Compiling Main             ( Main.hs, Main.o )
[2 of 2] Linking Main [Objects changed]
$ ./Main +RTS -s
     112,040,744 bytes allocated in the heap
         904,104 bytes copied during GC
          35,984 bytes maximum residency (2 sample(s))
          29,552 bytes maximum slop
               5 MiB total memory in use (0 MiB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0        25 colls,     0 par    0.001s   0.001s     0.0000s    0.0008s
  Gen  1         2 colls,     0 par    0.000s   0.000s     0.0001s    0.0001s

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    0.013s  (  0.013s elapsed)
  GC      time    0.001s  (  0.001s elapsed)
  EXIT    time    0.000s  (  0.005s elapsed)
  Total   time    0.015s  (  0.020s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    8,370,134,387 bytes per MUT second

  Productivity  86.9% of total user, 65.6% of total elapsed
```

We could use `Data.Foldable.foldl'`:
```haskell
{-# LANGUAGE BangPatterns, DataKinds, NoImplicitPrelude, GADTs, ScopedTypeVariables #-}
module Main where

import Clash.Prelude
import Clash.Sized.Vector
import Control.DeepSeq
import Data.Foldable (foldl')

rnfY :: (NFData a, 1 <= n, KnownNat n) => Vec n a -> ()
rnfY = foldl' (\() -> rnf) ()

main :: IO ()
main = do
  let !_ = rnfY (repeat @1000000 (0 :: Int))
  return ()
```
```
$ ./Main +RTS -s
     216,040,760 bytes allocated in the heap
       1,246,288 bytes copied during GC
          35,984 bytes maximum residency (2 sample(s))
          29,552 bytes maximum slop
               6 MiB total memory in use (0 MiB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0        50 colls,     0 par    0.001s   0.001s     0.0000s    0.0011s
  Gen  1         2 colls,     0 par    0.000s   0.000s     0.0001s    0.0001s

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    0.030s  (  0.030s elapsed)
  GC      time    0.001s  (  0.001s elapsed)
  EXIT    time    0.000s  (  0.008s elapsed)
  Total   time    0.033s  (  0.040s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    7,175,807,767 bytes per MUT second

  Productivity  92.4% of total user, 74.5% of total elapsed
```
But it's not as good: it runs twice as long and allocates twice as much on the heap. In addition, Clash will be forced to unroll the usage of `foldr` in the default implementation of `Data.Foldable.foldl'`:
```haskell
foldl' f z0 = \ xs ->
        foldr (\ (x::a) (k::b->b) -> oneShot (\ (z::b) -> z `seq` k (f z x)))
              (id::b->b) xs z0
```

So it is probably worth it to add a `Clash.Sized.foldl'`, including blackbox and evaluator rule, and implement `rnf` and `rnfX` using that.

martijnbastiaan:
https://github.com/hasura/graphql-engine/pull/2933#discussion_r328821960

Fixes #2482

-------

I opted to use `clashSimulation` instead of writing custom black boxes and evaluator rules, because I feel we already have way too many error prone reimplementations of mundane code in `clash-lib`.

## Still TODO:


clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1058,8 +1058,26 @@ foldr f z (x `Cons` xs) = f x (foldr f z xs)
...
+  --   https://well-typed.com/blog/90/
+  --
+  | clashSimulation = go z0 xs0
+  | otherwise = last (scanl f z0 xs0)
+ where
+  go :: forall m. b -> Vec m a -> b
+  go z Nil = z
+  go z (Cons x xs) =
+    let z1 = f z x
+    in z1 `seq` go z1 xs
```
martijnbastiaan, line 1080:
```suggestion
    in z1 `seqX` go z1 xs
```

?

DigitalBrains1: `seqX` would allow using a function that is lazy in its first argument to keep working as they did before `foldl` became strict, but it is also pretty confusing to people comfortable with bottom values that not all bottoms are treated equal in a function explicitly marked as being strict. So I'm inclined to treat all bottoms as producing bottom by definition.

DigitalBrains1: Another option, a `seq` where exceptions in its first argument are simply ignored, would at least treat all bottoms as equal but still violate the expectance that a strict `foldl` would in fact return bottom whenever a bottom is encountered while accumulating. So, once again, I'm inclined to use `seq`.

DigitalBrains1: Maybe we should offer a variant with the behaviour that bottoms don't propagate up, so people can still actually write a function lazy in its first argument producing results.

martijnbastiaan: I'm inclined to use `seqX`, for the reasons you mention:

> `seqX` would allow using a function that is lazy in its first argument to keep working for `XException` as it did before `foldl` became strict,

Because hunting `XException`s is a pretty miserable experience a lot of the time.

DigitalBrains1: We would already differ in the respect that `foldl` is strict _at all_, so I suppose if we clearly document `foldl` as being strict with the exception of ignoring `XException`s in its accumulator, that would also work. But we do need the extra documentation because what we're doing is somewhat strange.

martijnbastiaan: That's fair, but I think that's only because we specialize its types. As mentioned in the documentation of `Data.Foldable`, it never makes sense to use `foldl` on lists. We're somewhat special because we (ab)use bottoms to represent X.

Also I thought newer `base`s already make `foldl` strict for `List`s, but it doesn't seem to be so:

```haskell
foldl k z0 xs =
  foldr (\(v::a) (fn::b->b) -> oneShot (\(z::b) -> fn (k z v))) (id :: b -> b) xs z0
```

(Also if you think that implementation is unreadable, look at the one in `Foldable`:

```haskell
foldl f z t = appEndo (getDual (foldMap (Dual . Endo . flip f) t)) z
```

)

DigitalBrains1: Where does it say it doesn't make sense on lists? They even mention the case we're discussing here as an application:
> This is rarely what you want, but can work well for structures with efficient right-to-left sequencing and an operator that is lazy in its left argument.

DigitalBrains1: Is there a good reason for propagating other bottoms than `XException`? Couldn't we just ignore all bottoms while forcing the accumulator, and have it outwardly appear to work as the lazy version? As soon as the function actually forces its first argument to WHNF, the bottom would still pop out.

martijnbastiaan: I mean that's pretty heavily implied.. In any case, [lexi lambda's](https://github.com/hasura/graphql-engine/pull/2933#discussion_r328821960) comment makes the case pretty convincingly. Unless there is some glaring mistake in her reasoning, I don't think we should rehash the discussion here.

DigitalBrains1: 1. Did you answer to a message I already deleted because it wasn't that important / I changed my mind?
2. I think the case is convincingly made. but we are discussing doing something different. We're discussing ignoring certain bottom values. She doesn't say anything about that, so I don't see what we're rehashing.

martijnbastiaan: Yes! I did respond to the deleted comment. I'll delete mine too (and this one after :-)).

christiaanb: In what case would we even observe the difference? The function would be really weird... where it ignores its arguments in some cases, and return exceptions in others...

```
f 0 y = errorX "QQ"
f _ 2 = 1
f x y = x + y

foldl f 1 [(-1),2]
```
would return `1` if we used `seqX`, and raise an exception if we used `seq`.

I would be nice to have some benchmark numbers to determine whether the performance difference is negligible enough to use `seqX` over `seq`.

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1058,8 +1058,26 @@ foldr f z (x `Cons` xs) = f x (foldr f z xs)
...
+foldl f z0 xs0
+  -- We use 'go' to make 'foldl' strict during Haskell simulation, while using
+  -- 'scanl' to make Clash's life easier during normalization. We make 'foldl'
+  -- strict due to similar reasoning to:
+  --
+  --   https://github.com/hasura/graphql-engine/pull/2933#discussion_r328821960
+  --
+  -- Also see:
+  --
+  --   https://well-typed.com/blog/90/
```
christiaanb, line 1071: But it was posted on April 1st....

Anyhow, I feel this one is the stronger reference to make the change, as it actually argues to make a strict foldl the default. i.e. I would reference it first.

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1058,8 +1058,26 @@ foldr f z (x `Cons` xs) = f x (foldr f z xs)
...
+  --
+  | clashSimulation = go z0 xs0
```
christiaanb, line 1073: This will make it that there could be a HDL-vs-Simulation difference if we don't use `seqX`....

christiaanb (approved):
I would like to see some benchmark numbers on `seq` vs `seqX`.

clash-prelude/src/Clash/Sized/Vector.hs
christiaanb: Yes, that's right.

clash-prelude/src/Clash/Sized/Vector.hs
martijnbastiaan: With

```haskell
{-# LANGUAGE BangPatterns, DataKinds, NoImplicitPrelude, GADTs, ScopedTypeVariables, NumericUnderscores #-}
module Main where

import Clash.Prelude
import Clash.Sized.Vector
import Control.DeepSeq

rnfY :: NFData a => Vec n a -> ()
rnfY = foldl (\() -> rnf) ()

main :: IO ()
main = do
  let !_ = rnfY (repeat @100_000_000 (0 :: Int))
  return ()
```

and

```
cabal build clash-prelude && ghc -fforce-recomp Test.hs
```

and

```
time ./Test && time ./Test && time ./Test && time ./Test && time ./Test && time ./Test
```

I get:

| seq   | seqX  | diff   |
|-------|-------|--------|
| 2.932 | 5.082 | 57.69% |
| 2.932 | 5.152 | 56.91% |
| 2.902 | 5.152 | 56.33% |
| 2.942 | 5.192 | 56.66% |
| 2.932 | 5.182 | 56.58% |
| 2.932 | 5.252 | 55.83% |

martijnbastiaan:
@christiaanb Given the benchmark numbers I think we should pick `seq`. Though I have no evidence, I'm also inclined to say it's unlikely somewhat would write a function where `seq`/`seqX` would make a difference.

PR #2692, opened 2024-03-10
Add `Int*`, `Word*`, `Identity` and `Maybe` instances for `Counter`

### LINKED ISSUE #2534: Expose `Clash.Class.Counter` methods
Currently, functions like `countMin` and `countSuccOverflow` are marked in the documentation as internal implementation details that shouldn't be used by client code. But they (or something similar) would be very useful in a lot of cases where we only want to count "once" and then do something else. 

I realize overflow detection can be achieved "on the outside" by adding one more bit to the counter, without access to these functions by doing something like:

```
let ((bitCoerce @(Index 2) -> overflow), x') = countSucc (0, x)
in x' <$ guard (not overflow)
```

but this is not really readable (it could be improved a bit if we had a `Counter Bool` instance). And I'm not sure what similar trick could apply to `countMin`.

martijnbastiaan:
Fixes #2534

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Class/Counter/Internal.hs
+++ b/clash-prelude/src/Clash/Class/Counter/Internal.hs
@@ -1,5 +1,7 @@
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralisedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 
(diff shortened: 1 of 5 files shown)


martijnbastiaan:
@gergoerdi Could you check whether this does what you want?

clash-prelude/src/Clash/Class/Counter/Internal.hs
```diff
@@ -80,6 +85,56 @@ instance KnownNat n => Counter (Unsigned n)
...
+instance Counter Word32
+-- | @since 1.8.2
+instance Counter Word64
+
+-- | @since 1.8.2
+deriving newtype instance Counter a => Counter (Identity a)
+
+-- | Opiniated instance for 'Maybe'. 'Nothing' is considered the minimum value,
+-- while @'Just' 'countMax'@ is considered the maximum value.
+--
```
leonschoorl, line 118: Doesn't seem too opinionated to me, doesn't Nothing being the minimum follow from the Ord instance of Maybe?

```
ghci> Nothing < Just ()
True
ghci> Just () < Nothing
False

```

martijnbastiaan: Ooh, you're right! I'll amend the PR.

DigitalBrains1: Thanks, Leon, that was what was niggling me, I remembered we discussed this but I couldn't recall what it was! :-D

gergoerdi:
This doesn't seem to address exporting `counterMin`.

clash-prelude/src/Clash/Class/Counter/Internal.hs
christiaanb: I guess it's fair as this class links `Ord` and `Bounded` which are normally explicitly unrelated.

christiaanb (approved):

martijnbastiaan:
> This doesn't seem to address exporting `counterMin`.

You're right, I've exported the internal functions now. We haven't thought of anything that would need changing.

PR #2694, opened 2024-03-10
Clash.Cores: Add CRC module

rowanG077:
This module should provide in most standard CRC needs, with the exception of very high performance (Ex. 25G+ Ethernet) where you really need pipelined CRCs. It took inspiration from a CRC implementation given in amaranth which is available [here](https://github.com/amaranth-lang/amaranth/tree/main/amaranth/lib/crc). Due to this I have added the amaranth contributors to the LICENSE. 

## Still TODO:

  - [x] Compare software with parallel hardware implementation using randomized inputs 
  - [x] Write some convenience functions for the circuit version (possibly using TH)
  - [x] Write examples


rowanG077:
@DigitalBrains1 I tagged you as reviewer. If you think of someone else please don't hesitate to tag that someone.

rowanG077:
The PR is reviewable. I haven't yet made a nice user exposed circuit. I'm not super sure how to structure it yet and that is because of the `residue`. When validating a CRC what is best to do in hardware is feed your message including the given CRC through the CRC pipeline. If the message is valid the `residue` should fall out, a constant value which is cheaper to validate with then comparing two dynamic values.

I want to allow users to do CRC validations using only a single CRC circuit even if they use a `dataWidth` that is a multiple of their native message data width. For example ethernet is byte oriented. In a lot of FPGA ethernet implementations you go wider to allow a slower clock domain to process ethernet frames at line rate. For example you process 4 bytes every clock cycle. This means however that the end of a ethernet frame 1, 2, 3 or 4 bytes are valid. This sucks because in a naive implementation you need to instantiate a CRC engine for each case. But for validation you can cheat a little, You calculate a residue for each case assuming you pad the message with null bytes. This means you can do a CRC validation in this scenario with only a single engine by just introduction a mux at the end to compare to the special residues.  This padded residue can be calculated already using the `residue` in the module. I'm just not sure how to make a nice CRC function that includes this functionality from a user perspective.

rowanG077:
All open points have been done.

DigitalBrains1:
I'm in the process of reviewing this :-)

clash-cores/test/unittests.hs
```diff
@@ -18,7 +19,8 @@ import qualified Test.Cores.Xilinx.DcFifo
 
 tests :: TestTree
 tests = testGroup "Unittests"
-  [ Test.Cores.SPI.tests
+  [ Test.Cores.CRC.tests
```
DigitalBrains1, line 22: Hehe, it looks like the unit tests now take three times as long to run. It's almost at a point where I'd move it to a self-hosted runner rather than a free shared runner provided by GitLab.

DigitalBrains1 (changes requested):
Thanks for this impressive piece of work!

Do you agree that this will be a squash commit, merging the stuff into a single commit on `master`?

All the code looks high quality to me, as usual for your contributions, so I don't have anything else to say about that! Instead, I will discuss API and documentation.

I did a fair amount of copyediting on the Haddock. Instead of making this PR a litany of little change suggestions, I pushed the copyediting to its own branch, [peter/crc](https://github.com/clash-lang/clash-compiler/tree/peter/crc). The first commit is simply replacing all backticks with single quotes; they are synonymous in Haddock but this repository generally uses single quotes and I personally think this should be used throughout. So that is just a consistency detail. The second commit is the actual copyedit. The commits should just be squashed eventually, but they can be fast-forwarded on this branch if you don't object to my changes and agree that this PR will be squashed as a whole anyway.

Then API and structure. We'd like to reduce the amount of `Internal` modules being listed in the Haddock; I thought about whether we can avoid it here. I think in this case, all three modules can be folded into one! I suspect there will be some things we'd like to emit Haddock for even though they are not part of the API; they can just go in a section of the Haddock with the title "Internal". This'll be the last section, so people can stop reading there if they read from top to bottom.

If you don't agree with flattening the module structure completely, perhaps we can still make it such that `Clash.Cores.CRC` documents all things people will be exposed to, and then omit `Clash.Cores.CRC.Internal` from the Haddock?

I don't think the `Clash.Cores.CRC.Derive` module really adds anything in ways of structure; it can just be folded into `Internal` if we keep the latter.

clash-cores/src/Clash/Cores/CRC/Internal.hs
```diff
@@ -0,0 +1,650 @@
...
+            (dataWidth :: Nat)
+            (nLanes :: Nat)
+   . HiddenClockResetEnable dom
+  => HardwareCRC crc dataWidth nLanes
+  => Proxy crc
+  -- ^ The CRC
+  -> Proxy dataWidth
+  -- ^ The width of the words to feed
+  -> Proxy nLanes
+  -- ^ How many words can be fed at maximum in a single cycle
```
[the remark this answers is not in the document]
DigitalBrains1: Erm, I was too quick in scanning tyvars. The `Proxy crc` is still needed, because `crc` is not constrained, it is only _used_ in the form of `CRCWidth crc`. But I think my point stands for `nLanes` and `dataWidth`.

rowanG077:
@DigitalBrains1 Thank you for your review! I'm happy there are no major issues. The main reason for the division is the test code. There you really don't want to use `deriveHardwareCRC` for every single random test that is happening. So there it is needed to construct `HardwareCRCParams` by hand without TH. But a user should never do that. I do agree that `Internal` should not be exposed in haddock.

DigitalBrains1:
Right, not only are the tests a good reason to export internal functions, I just realised we also actually like exporting internal functions anyway so people can use them to construct things not provided for without having to copy-paste code. So the module in itself is a good thing, just rather not in the Haddock. A smallish section named _Internal_ at the end of the Haddock of the `Clash.Cores.CRC` module would also have been fine, but there are so many internal functions here that I think it's nicer to have them in their own module.

I don't think there is a reason to have a separate `Clash.Cores.Derive` though.

I also just realised a naming thing. In our [style guide](https://clash-lang.readthedocs.io/en/latest/hacking-on-clash/style.html#naming) we say:
> For readability reasons, don’t capitalize all letters when using an abbreviation. For example, write `HttpServer` instead of `HTTPServer`. Exception: Two letter abbreviations, e.g. `IO`.

CRC is 50% more letters than two. I'm not too critical about internal function naming, but for the exposed API, it's nice if we at some point reach some consistency X-D. So I'd like it if you could rename them!

rowanG077:
Review comment have been handled.

rowanG077:
Regarding the test times. It may be a good idea to commit most of them out. This was my original intention, have a covering config space. But during development I noticed that quite a few times that just some specific CRC failed so I kept it the full suite.

martijnbastiaan:
> But during development I noticed that quite a few times that just some specific CRC failed so I kept the full suite.

Then IMO we should just keep the full suite on CI too. We can move it to local runners to not be a burden to GitLab, as @DigitalBrains1 already said.

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,688 @@
...
+type instance TryDomain t (SoftwareCrc crcWidth dataWidth) = 'NotFound
+
+-- | Apply function only when bool is True
+applyWhen :: Bool -> (a -> a) -> a -> a
+applyWhen True f x = f x
+applyWhen False _ x = x
+
+-- | Reverse bitvector
+reverseBV :: KnownNat n => BitVector n -> BitVector n
+reverseBV = v2bv . reverse . bv2v
```
christiaanb, line 115: Once this PR is merged could you raise an issue to 
1. add this function to `clash-prelude` and
2. make something with better time complexity such as https://graphics.stanford.edu/~seander/bithacks.html#ReverseParallel

clash-cores/LICENSE
```diff
@@ -1,4 +1,5 @@
 Copyright (c) 2019, Foamspace Corp.
+              2023, Amaranth HDL contributors
```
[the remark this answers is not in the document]
rowanG077: Parts of the implementation are based on their implementation. We should comply with their license. I didn't feel the need to add it to the source code files since the code is written from scratch. If you prefer this I can do that as well.

clash-cores/src/Clash/Cores/Crc/Internal.hs
rowanG077: Sure

DigitalBrains1: Hah I had noted this too, but didn't comment on it. I think I didn't even write it down for myself, which was dumb. Anyway...... the current implementation is fine in hardware, and a quick implementation would probably need a primitive. Unless you use `clashSimulation` from `Clash.Magic` to use the quick implementation in simulation but use `v2bv . reverse . bv2v` when generating HDL. That will write your primitive for you (but like a usual primitive, it's still up to you to make sure simulation and HDL match in behaviour).

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+import qualified Language.Haskell.TH.Syntax as TH
+import           Type.Reflection
+
+-- | Contains all the parameters to generate a CRC implementation
+data CrcParams (crcWidth :: Nat) (dataWidth :: Nat)
+  = CrcParams
+      { _crcWidth         :: SNat crcWidth
+      -- ^ Bit width of CRC word. Also known as "width" in the Williams model.
+      --   See http://techref.massmind.org/techref/method/math/crcguide.html
+      , _crcDataWidth     :: SNat dataWidth
```
christiaanb, line 38: I think this is actually an implementation detail, and should thus go into `SoftwareCRC` and `HardwareCRC` types.

I was also not a particular fan of using:
* `EmptyDataType`
* Proxy arguments
* Type Applications

To pass in a particular crc configuration. I've thus created an slightly altered version that uses `newtype` instead over at: https://github.com/clash-lang/clash-compiler/commit/03c3e3acf315b76288d1f157c88357dba60a787a

Let me know your thoughts. Also @DigitalBrains1 I'd like to hear your thoughts on this.

christiaanb: I do realize you lose canonicity as in principle you could make multiple configurations that would e.g. all be of type `Crc3_gsm`. Meaning that in principle you could end up with one part _encoding_ with a certain `Crc3_gsm` configuration, and another part _decoding_ with a __different__ `Crc3_gsm` configuration. But I don't know how important we think it is to protect ourselves against that vs the aesthetics/usability of the library.

christiaanb: Perhaps we could simplify it even further compared to https://github.com/clash-lang/clash-compiler/commit/03c3e3acf315b76288d1f157c88357dba60a787a and even drop the newtype wrapper and just have:

```
crc3_gsm :: CrcParams 3
```

rowanG077: I think that is too far. I expect users would like to write circuits polymorph in datawidth/lanes but with a fixed CRC.

Something like 

```
foo :: HardwareCrc Crc3_gsm dataWidth nLanes => Crc3_gsm -> ...
```

Then the user of `foo` is responsible for deriving the appropriate `HardwareCrc`. If you don't have this newtype there is no way to communicate and enforce this.

DigitalBrains1: We're defining unique CRC's at the type level, with `deriveHardwareCrc`. This feels like that if there is a term-level representation of such a type, it should be a singleton? What's the purpose of the `crc3_gsm` term, what does it communicate?

If  instead you go the term-route, you could also construct these terms with Template Haskell, and make the term completely contain the `CrcHardwareParams`. No further data at the type level, just make the term contain everything needed.

But I'll just go with what you two can agree on.

DigitalBrains1: > I'm also not super sure the Coercible business doesn't makes things harder to understand rather than easier.

I'd be happier if we hid all that from the user, it definitely makes understanding the stuff harder.

DigitalBrains1: Ah, also, if you base stuff on Christiaans branch, I'd like to have another go at a review. In other words, **changes requested**. :-)

clash-cores/src/Clash/Cores/Crc.hs
```diff
@@ -0,0 +1,96 @@
...
+on the CRC circuit.
+
+>>> :{
+deriveHardwareCrc (Proxy @Crc32_ethernet) d8 d4
+dummy = 1
+:}
+
+>>> crcEngine' = exposeClockResetEnable crcEngine systemClockGen resetGen enableGen
+>>> myEngine = crcEngine' (Proxy @Crc32_ethernet)
+
```
DigitalBrains1, line 58:
```suggestion
>>> :{
deriveHardwareCrc (Proxy @Crc32_ethernet) d8 d4
dummy = 1
:}

>>> crcEngine' = exposeClockResetEnable crcEngine systemClockGen resetGen enableGen
>>> myEngine = crcEngine' (Proxy @Crc32_ethernet)

Note how we hinted to @clashi@ that our multi-line command was a list of
declarations by including a dummy declaration @dummy = 1@. Without this trick,
@clashi@ would expect an expression and the Template Haskell emitted by
@deriveHardwareCrc@ would not work.

```

DigitalBrains1: Hey GitHub does things subtly wrong with trailing empty lines in suggestions... funny.

DigitalBrains1 (commented):
Looking good, just a few last finishing touches as far as I'm concerned :-).

Since I have a day off tomorrow (and the usual day off Monday), I thought it'd be nice if I did the review today. However, I'm _really_ tired. So sorry if I missed things or otherwise made a mess of things. Best effort performance only, no guarantees either express or implied. Due to limited fitness for purpose of author, who is liable to sleep.

clash-cores/LICENSE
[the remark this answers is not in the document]
DigitalBrains1: I think maybe it's better to put them in the source file and explain why they are there. The current solution _does_ mention they have a copyright here, but it's not really clear why. I'm not sure about the legal rules, but it seems fair to mention explicitly that they cooked up part of the CRC solution specifically. And I think it would not be weird to put that in `CRC/Internal.hs` in any way.

clash-cores/src/Clash/Cores/Crc/Internal.hs
rowanG077: `crc3_gsm` in this case would just be a "type hint" to know for which `HardwareCrc` you want to do something. Exactly the same as `Proxy` we had before but without the `Proxy`.

rowanG077: Actually my very first commit reflects this: https://github.com/clash-lang/clash-compiler/pull/2694/commits/e4cfacceda4da54f7cb0556e98a9983ed10d40a2#diff-80045d2b23ddac9d150e56d2c8b51fb3ed39401a4da0da69d6908927acad6761

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+  -> Vec n (Vec p (BitVector q))
+partitionMat SNat mat = fmap (fmap v2bv . unconcatI . bv2v) mat
+
+-- | Flatten vertically partitioned matrix
+unPartitionMat
+  :: KnownNat q
+  => KnownNat p
+  => Vec n (Vec p (BitVector q))
+  -> Vec n (BitVector (p * q))
+unPartitionMat mat = fmap (v2bv . concat . fmap bv2v) mat
```
christiaanb, line 335:
```suggestion
unPartitionMat mat = fmap pack mat
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+    f = postProcess $ bv2v . (\p -> runCrc p 0) <$> onehots (SNat @(CrcWidth crc))
+    g = postProcess $ bv2v . (\p -> runCrc 0 p) <$> onehots dataWidth
+
+-- | Vertically partition a matrix
+partitionMat
+  :: KnownNat q
+  => SNat p
+  -> Vec n (BitVector (p * q))
+  -> Vec n (Vec p (BitVector q))
+partitionMat SNat mat = fmap (fmap v2bv . unconcatI . bv2v) mat
```
christiaanb, line 327:
```suggestion
partitionMat SNat mat = fmap unpack mat
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+  -> Signal dom (Maybe (Index nLanes, Vec nLanes (BitVector dataWidth)))
+  -> Signal dom Bool
+crcValidatorFromParams hwParams@(CrcHardwareParams _ CrcParams{..} _ residues) inDat = match
+    where
+      step = laneStep hwParams (subSNatLe (SNat @nLanes) d1) ()
+      (validLanesX, datX) = unbundle $ fromJustX <$> inDat
+
+      nextCrcState = step <$> crcState <*> datX
+      crcState = regEn _crcInitial (isJust <$> inDat) nextCrcState
+      matches = zipWith (==) residues <$> (fmap pure crcState)
```
christiaanb, line 687:
```suggestion
      matches = zipWith (==) residues <$> fmap pure crcState
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+takeLe SNat vs = leToPlus @n @m $ takeI vs
+
+-- | Apply a step to a CRC and input given by @lanePrev@
+--
+-- This function is meant to be used via 'smap' and crashes if called with
+-- out-of-bounds @lanePrev@.
+laneStep
+  :: CrcHardwareParams crcWidth dataWidth nLanes
+  -> SNat lanePrev
+  -> ()
```
christiaanb, line 575: Remove this please

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+   . HiddenClockResetEnable dom
+  => KnownNat crcWidth
+  => KnownNat dataWidth
+  => KnownNat nLanes
+  => CrcHardwareParams crcWidth dataWidth nLanes
+  -> Signal dom (Maybe (Index nLanes, Vec nLanes (BitVector dataWidth)))
+  -> Signal dom (BitVector crcWidth)
+crcEngineFromParams hwParams@(CrcHardwareParams _ CrcParams{..} _ _) inDat = crcOut
+    where
+      steps = smap (laneStep hwParams) (repeat @nLanes ())
```
christiaanb, line 631:
```suggestion
      steps = smap (\sn _ -> laneStep hwParams sn) (repeat @nLanes ())
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+  => KnownNat crcWidth
+  => KnownNat dataWidth
+  => KnownNat nLanes
+  => 1 <= nLanes
+  => CrcHardwareParams crcWidth dataWidth nLanes
+  -> Signal dom (Maybe (Index nLanes, Vec nLanes (BitVector dataWidth)))
+  -> Signal dom Bool
+crcValidatorFromParams hwParams@(CrcHardwareParams _ CrcParams{..} _ residues) inDat = match
+    where
+      step = laneStep hwParams (subSNatLe (SNat @nLanes) d1) ()
```
christiaanb, line 682:
```suggestion
      step = laneStep hwParams (subSNatLe (SNat @nLanes) d1)
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+mkCrcHardwareParams crc dataWidth nLanes@SNat
+  = hwParams
+    where
+      computeResidue = rawResidue crc . (snatToNum dataWidth *) . fromIntegral
+
+      hwParams = CrcHardwareParams
+        { _crcNlanes = nLanes
+        , _crcBaseParams = crcParams crc dataWidth
+        , _crcLaneParams = mkCrcLaneParams crc dataWidth nLanes
+        , _crcResidues = fmap computeResidue $ reverse indicesI
```
christiaanb, line 480:
```suggestion
        , _crcResidues = computeResidue <$> reverse indicesI
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+mkFGMatrices crc dataWidth@SNat = FGMatrices f g
+  where
+    params = crcParams crc dataWidth
+    newParams = params
+                  { _crcReflectOutput = False
+                  , _crcXorOutput = 0
+                  }
+    withInitCrc i = mkSoftwareCrcFromParams $ newParams { _crcInitial = i }
+    runCrc initial dat = digest $ feed (withInitCrc initial) dat
+    onehots n = (shiftL 1) . fromIntegral <$> (indices n)
```
christiaanb, line 316:
```suggestion
    onehots n = shiftL 1 . fromIntegral <$> indices n
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+  -> BitVector (CrcWidth crc)
+  -- ^ The residue
+rawResidue p nExtra = rawResi
+  where
+    dataWidth = d1
+    CrcParams {..} = crcParams p dataWidth
+    swCrc = mkSoftwareCrc p dataWidth
+    crcVal = applyWhen _crcReflectOutput reverse $ bv2v $ digest swCrc
+    bitsToFeed = fmap (v2bv . reverse . singleton) $
+                    (V.toList crcVal) L.++ (L.replicate nExtra 0)
```
christiaanb, line 253:
```suggestion
                    V.toList crcVal L.++ L.replicate nExtra 0
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,689 @@
...
+feed engine@(SoftwareCrc {..}) word = go _crcParams
+  where
+    CrcParams {..} = _crcParams
+    go (CrcParams SNat SNat _ _ _ _ _) = engine { _crcCurrent = nextCrc }
+      where
+        applyPolyBit crcB
+          | crcB .&. _crcTopBitMask > 0 = (shiftL crcB 1) `xor` _crcpolyShifted
+          | otherwise                   = shiftL crcB 1
+        applyPoly crcW = L.iterate applyPolyBit crcW L.!! snatToNum _crcDataWidth
+        applyCrc crc w = applyPoly $ crc `xor` (shiftL (extend w) (snatToNum _crcWidth))
```
christiaanb, line 195:
```suggestion
        applyCrc crc w = applyPoly $ crc `xor` shiftL (extend w) (snatToNum _crcWidth)
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
christiaanb: Alright, I understand now. In that case, could we move from "an empty data type + proxy" to a "unit data type"? e.g.
```haskell
data Crc3_gsm = Crc3_gsm
```

rowanG077:
All review comments have been handled.

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,704 @@
...
+    dataWidth = d1
+    CrcParams {..} = crcParams crc
+    swCrc = mkSoftwareCrc crc dataWidth
+    crcVal = applyWhen _crcReflectOutput reverse $ bv2v $ digest swCrc
+    bitsToFeed = fmap (v2bv . reverse . singleton) $
+                    V.toList crcVal L.++ L.replicate nExtra 0
+    rawResi = getCrcState $ L.foldl' feed swCrc bitsToFeed
+
+-- | Compute the residue value for these 'CrcParams', which is the value left in the
+--   CRC register after processing any valid codeword.
```
DigitalBrains1, line 262:
```suggestion
--   CRC register after processing any valid codeword.
--
-- __NB__: This function is not synthesizable
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,704 @@
...
+  ( KnownCrc crc, KnownNat dataWidth, 1 <= dataWidth, KnownNat nLanes, 1 <= nLanes)
+  => HardwareCrc (crc :: Type) (dataWidth :: Nat) (nLanes :: Nat) where
+  crcHardwareParams
+    :: crc
+    -- ^ Which CRC
+    -> Proxy dataWidth
+    -- ^ What word size in bits the hardware implementation can handle
+    -> Proxy nLanes
+    -- ^ The number of lanes
+    -> CrcHardwareParams (CrcWidth crc) dataWidth nLanes
```
DigitalBrains1, line 401: I like `SNat (n :: Nat)` better than `Proxy (n :: Nat)`. It's there, why not use it. This seems to compile fine:

```diff
--- a/clash-cores/src/Clash/Cores/Crc/Internal.hs
+++ b/clash-cores/src/Clash/Cores/Crc/Internal.hs
@@ -394,9 +394,9 @@ class
   crcHardwareParams
     :: crc
     -- ^ Which CRC
-    -> Proxy dataWidth
+    -> SNat dataWidth
     -- ^ What word size in bits the hardware implementation can handle
-    -> Proxy nLanes
+    -> SNat nLanes
     -- ^ The number of lanes
     -> CrcHardwareParams (CrcWidth crc) dataWidth nLanes
 
@@ -623,7 +623,7 @@ crcEngine
   --
   -> Signal dom (BitVector (CrcWidth crc))
   -- ^ The resulting CRC. There is a delay of a single cycle from input to output.
-crcEngine crc = crcEngineFromParams $ crcHardwareParams crc Proxy Proxy
+crcEngine crc = crcEngineFromParams $ crcHardwareParams crc SNat SNat
 
 -- | Like 'crcEngine' but uses 'CrcHardwareParams' instead of a 'HardwareCrc' constraint
 crcEngineFromParams
@@ -675,7 +675,7 @@ crcValidator
   -- ^ The input data
   -> Signal dom Bool
   -- ^ Whether the CRC is valid. There is a delay of a single cycle from input to output.
-crcValidator crc = crcValidatorFromParams $ crcHardwareParams crc Proxy Proxy
+crcValidator crc = crcValidatorFromParams $ crcHardwareParams crc SNat SNat
 
 -- | Like 'crcValidator' but uses 'CrcHardwareParams' instead of a 'HardwareCrc' constraint
 crcValidatorFromParams
```

DigitalBrains1 (commented):
I still have some wishes! I discovered some things I hadn't seen yet. I'm sorry I didn't spot it before.

I think this'll be the last one?

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,704 @@
...
+--   in the scenario where where you want to handle more words in a single cycle
+--   than are native to the protocol.
+--
+--   For example ethernet is byte-oriented but you may want to handle 32 bit
+--   in a single cycle to achaive a high throughput.
+--   This means that it is possible for 8, 16, 24 or 32 bits in the last data
+--   word of the stream to be valid.
+--   We can calculate a residue for each case and compare it to the CRC result.
+--   This saves us from having to instantiate four seperate CRCs for the receive path.
+--   The invalid words must be set to zero.
```
DigitalBrains1, line 241:
```suggestion
--   The invalid words must be set to zero.
--
-- __NB__: This function is not synthesizable
```

Weird, this comment was dropped. Must have pressed an oopsie-button.

clash-cores/src/Clash/Cores/Crc/Internal.hs
DigitalBrains1: Ah, I opened the docs and saw that these functions are grouped under _Software_. That explains why they aren't annotated non-synthesisable. The thing is, these functions (or rather a compile-time evaluated version of them through Template Haskell) are really useful in hardware, so people will want to use them anyway. So let's document it.

```suggestion
--   CRC register after processing any valid codeword.
--
--   To use the output of this function as a constant in your design, you can
--   lift it through Template Haskell:
--
-- @
-- import qualified Data.List as L
--
-- res :: BitVector 32
-- res = $(lift $ rawResidue Crc32_ethernet 0)
--
-- ress :: Vec 4 (BitVector 32)
-- ress = $(listToVecTH $ L.map (rawResidue Crc32_ethernet) [0..3])
-- @
--
--   __NB__: This function is not synthesizable
--
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
DigitalBrains1:
```suggestion
--   The invalid words must be set to zero.
--
--   To use the output of this function as a constant in your design, you can
--   lift it through Template Haskell:
--
-- @
-- import qualified Data.List as L
--
-- res :: BitVector 32
-- res = $(lift $ rawResidue Crc32_ethernet 0)
--
-- ress :: Vec 4 (BitVector 32)
-- ress = $(listToVecTH $ L.map (rawResidue Crc32_ethernet) [0..3])
-- @
--
-- __NB__: This function is not synthesizable
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
christiaanb: In addition, you could drop the `KnownNat dataWidth` and `KnownNat nLanes`.

christiaanb (approved):
LGTM modulo the Proxy arguments instead of SNat arguments for the `crcHardwareParams` class.

clash-cores/src/Clash/Cores/Crc/Internal.hs
DigitalBrains1: Heck, if you try hard enough, you can even get rid of the `KnownNat crcWidth`! But is there a way to use `{..}` and pattern matching in one go? It would eliminate the function `go` in this suggestion:

```diff
--- a/clash-cores/src/Clash/Cores/Crc/Internal.hs
+++ b/clash-cores/src/Clash/Cores/Crc/Internal.hs
@@ -657,15 +657,17 @@ crcEngineFromParams
             (dataWidth :: Nat)
             (nLanes :: Nat)
    . HiddenClockResetEnable dom
-  => KnownNat crcWidth
-  => KnownNat dataWidth
-  => KnownNat nLanes
   => CrcHardwareParams crcWidth dataWidth nLanes
   -> Signal dom Bool
   -> Signal dom (Maybe (Index nLanes, Vec nLanes (BitVector dataWidth)))
   -> Signal dom (BitVector crcWidth)
-crcEngineFromParams hwParams@(CrcHardwareParams _ _ CrcParams{..} _ _) resetCrc inDat = crcOut
-    where
+crcEngineFromParams
+  hwParams@(CrcHardwareParams SNat SNat _crcBaseParams@(CrcParams SNat _ _ _ _ _) _ _)
+  resetCrc inDat
+  = go _crcBaseParams
+   where
+    go CrcParams{..} = crcOut
+     where
       steps = smap (\sn _ -> laneStep hwParams sn) (repeat @nLanes ())
       (validLanesX, datX) = unbundle $ fromJustX <$> inDat
       step = fmap (steps !!) validLanesX
@@ -709,16 +711,18 @@ crcValidatorFromParams
             (dataWidth :: Nat)
             (nLanes :: Nat)
    . HiddenClockResetEnable dom
-  => KnownNat crcWidth
-  => KnownNat dataWidth
-  => KnownNat nLanes
   => 1 <= nLanes
   => CrcHardwareParams crcWidth dataWidth nLanes
   -> Signal dom Bool
   -> Signal dom (Maybe (Index nLanes, Vec nLanes (BitVector dataWidth)))
   -> Signal dom Bool
-crcValidatorFromParams hwParams@(CrcHardwareParams _ _ CrcParams{..} _ residues) resetCrc inDat = match
-    where
+crcValidatorFromParams
+  hwParams@(CrcHardwareParams SNat SNat _crcBaseParams@(CrcParams SNat _ _ _ _ _) _ residues)
+  resetCrc inDat
+  = go _crcBaseParams
+   where
+    go CrcParams{..} = match
+     where
       step = laneStep hwParams (subSNatLe (SNat @nLanes) d1)
       (validLanesX, datX) = unbundle $ fromJustX <$> inDat
```

This is all a bit splitting hairs, though, these functions are internal... but I do like to split hairs, so thanks :-)

rowanG077:
All the review comments have been handled. I have also moved the "first fragment" indicator inside the `Maybe` as I discussed on chat with @DigitalBrains1

rowanG077:
CI failure seems to be unrelated.

clash-cores/src/Clash/Cores/Crc/Internal.hs
DigitalBrains1: Did you not like my `ress` hint? Why did you only retain the `res` variant?

rowanG077: I actually overlooked it. Probably due to the similarity to the other comment. I have fixed that oversight.

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,737 @@
...
+--
+-- @nLanes@ indicates the maximum number of @dataWidth@ words you want to
+-- process in a single cycle. For example the stream could be byte-oriented,
+-- but processing is done @n@-bytes at a time.
+--
+-- Use 'Clash.Cores.Crc.deriveHardwareCrc' to create an instance. No instances
+-- should be implemented by hand, because a proper instance requires
+-- compile-time evaluation.
+class
+  ( KnownCrc crc, KnownNat dataWidth, 1 <= dataWidth, KnownNat nLanes, 1 <= nLanes)
```
christiaanb, line 417: The `KnownNat` superclass constraints are no longer needed now that we have `SNat` arguments.

christiaanb: Or were they left intentionally? So that `HardwareCrc` implies `KnownNat dataWidth` and `KnownNat nLanes`.

DigitalBrains1: These constraints mean that whenever a function takes a `HardwareCrc` argument, it is already a given that `(KnownCrc crc, KnownNat dataWidth, 1 <= dataWidth, KnownNat nLanes, 1 <= nLanes)`. This is nice for the functions `crcEngine` and `crcValidator`, which would otherwise still require those constraints in their type signature (well, okay, admittedly, `crcEngine` doesn't need `1 <= nLanes`, but that's really not a problem).

DigitalBrains1: Oh wait, I could have just said "Yes" :-)

rowanG077: Exactly, that's why I kept it in. Just to be clear. There are no longer any open points?

christiaanb: Yeah, I´m good. You too @DigitalBrains1 ?

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,737 @@
...
+-- | Compute the raw residue value for these 'CrcParams', which is the value left in the
+--   CRC register after processing any valid codeword. Raw in this case means
+--   this is the residue before the output reflection and out xor has been applied.
+--
+--   A fixed amount of padding can be given. This is useful during CRC verification
+--   in the scenario where where you want to handle more words in a single cycle
+--   than are native to the protocol.
+--
+--   For example ethernet is byte-oriented but you may want to handle 32-bit
+--   in a single cycle to achieve a high throughput.
```
DigitalBrains1, line 236:
```suggestion
--   For example ethernet is byte-oriented but you may want to handle 32 bit
--   in a single cycle to achieve high throughput.
```

Dropping the article was already part of a previous review comment with suggestion; perhaps the "commit suggestion" function of GitHub is a nice way to apply suggestions so you don't run the risk of accidentally omitting part of the suggestion?

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,737 @@
...
+  deriving (Show, Lift)
+
+-- | No domain in 'CrcHardwareParams'
+type instance TryDomain t (CrcHardwareParams crcWidth dataWidth nLanes) = 'NotFound
+
+-- | This class is used to indicate a CRC has a derived hardware implementation
+--
+-- @nLanes@ indicates the maximum number of @dataWidth@ words you want to
+-- process in a single cycle. For example the stream could be byte-oriented,
+-- but processing is done @n@-bytes at a time.
```
DigitalBrains1, line 411:
```suggestion
-- but processing is done @n@ bytes at a time.
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,737 @@
...
+  --   The @Bool@ must be asserted on the first fragment to start a new CRC
+  --   computation
+  --   @Index nLanes@ indicates how many @dataWidth@ words are
+  --   valid in the vector minus 1.
+  --
+  --     - Ex. 1. @Just (False, 0, 0xDE :> 0xAD :> 0xBE :> 0xEF :> Nil)@
+  --       means this is not the start of a new message and only the first word @0xDE@ is valid
+  --     - Ex. 2. @Just (True, 3, 0xDE :> 0xAD :> 0xBE :> 0xEF :> Nil@
+  --       means this is the start of a new message and all words in the vector are valid
+  --
```
DigitalBrains1, line 650:
```suggestion
  -- ^ The input data
  --
  --   The @Bool@ must be asserted on the first fragment to start a new CRC
  --   computation.
  --
  --   @Index nLanes@ indicates how many @dataWidth@ words are valid in the
  --   vector minus 1.
  --
  --     - Ex. 1. @Just (False, 0, 0xDE :> 0xAD :> 0xBE :> 0xEF :> Nil)@
  --       means this is not the start of a new message and only the first word
  --       @0xDE@ is valid.
  --     - Ex. 2. @Just (True, 3, 0xDE :> 0xAD :> 0xBE :> 0xEF :> Nil@ means
  --       this is the start of a new message and all words in the vector are
  --       valid.
```

If you want multiple paragraphs, you'll need to leave a blank line in between.

I made some more tweaks: periods, trailing blank line, line length.

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,737 @@
...
```
DigitalBrains1, line 708:
```suggestion
  -- ^ The input data
  --
  --   The @Bool@ must be asserted on the first fragment to start a new CRC
  --   computation.
  --
  --   @Index nLanes@ indicates how many @dataWidth@ words are
  --   valid in the vector minus 1.
  --
  --     - Ex. 1. @Just (False, 0, 0xDE :> 0xAD :> 0xBE :> 0xEF :> Nil)@
  --       means this is not the start of a new message and only the first word
  --       @0xDE@ is valid.
  --     - Ex. 2. @Just (True, 3, 0xDE :> 0xAD :> 0xBE :> 0xEF :> Nil@
  --       means this is the start of a new message and all words in the vector
  --       are valid.
```

clash-cores/src/Clash/Cores/Crc/Internal.hs
```diff
@@ -0,0 +1,737 @@
...
+-- and 16 invalid bits.
+--
+-- This rule is __NOT__ checked at the type level because as long you take
+-- care that misalignment does not occur it is not a problem.
+crcEngine
+  :: forall (dom :: Domain)
+            (crc :: Type)
+            (dataWidth :: Nat)
+            (nLanes :: Nat)
+   . HiddenClockResetEnable dom
```
DigitalBrains1, line 635: Didn't you intend to eliminate the reset input of both `crcEngine` and `crcValidator`?

DigitalBrains1 (changes requested):
I see you're still tweaking small aspects of your documentation. So I will also review small aspects, I hope you appreciate it and don't think I'm being fussy :-).

rowanG077:
@DigitalBrains1 All of has been fixed

DigitalBrains1 (approved):
\o/ 🎉 🚀 🥳 ❤️ 💯 👌 🙌 🙆 🦸 🧙

PR #2711, opened 2024-04-19
 Detect and throw error on "broken" GHCs 

martijnbastiaan:
We know some GHC versions interact poorly with Clash on some platforms. Rather than making everyone hunt down the right tickets on their own, this patch makes Clash emit a fat ~~warning~~ error. Because these issues don't _completely_ prevent Clash from working, we allow users to bypass the error by passing a flag `-fclash-ignore-broken-ghcs`.

## Still TODO:

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,18 +66,18 @@ Note that release branches might contain non-released patches.
 |      | Linux | Windows | macOS | Clash (released) | Clash (development version)
 |------|-------|---------|-------|------------------|--------------------------
 | 8.6  | ✔️     | ✔️       | ✔️     | 1.0 - 1.8        | ❌
-| 8.8  | ✔️     | ❌      | ✔️     | 1.0 - 1.8        | ❌
-| 8.10 | ✔️     | ✔️       | ❌    | 1.2 - 1.8        | ✔️
+| 8.8  | ✔️     | ❌       | ✔️     | 1.0 - 1.8        | ❌
+| 8.10 | ✔️     | ✔️       | ❌     | 1.2 - 1.8        | ✔️
 | 9.0  | ✔️     | ✔️²      | ✔️     | 1.4 - 1.8        | ✔️
 | 9.2  | ⚠️¹    | ⚠️¹      | ⚠️¹    | 1.8              | ⚠️¹️
-| 9.4  | ✔️     | ✔️       | ✔️     | 1.8              | ✔️
-| 9.6  | ✔️     | ✔️³      | ✔️     | 1.8              | ✔️
+| 9.4  | ⚠️³ ️   | ⚠️³      | ️⚠️³ ️   | 1.8              | ✔️
+| 9.6  | ✔️³    | ✔️³      | ✔️³    | 1.8              | ✔️
 
 ¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change was reverted in 9.4.
 
-² GHC 9.0.2 on Windows fails to compile `clash-cores`. We therefore don't run the Clash testsuite on CI for this combination.
+² GHC 9.0.2 on Windows fails to compile `clash-cores`. We therefore don't run the Clash test suite on CI for this combination.
 
-³ Clash starts extremely slowly when compiled with 9.6 on Windows. See [#2684](https://github.com/clash-lang/clash-compiler/issues/2684).
+³ Clash starts extremely slowly when compiled with 9.4.8 up to and including 9.6.2. Consider downgrading to 9.4.7 or upgrading to 9.6.3 and up.
 
 ## Cabal
 To use Cabal you need both Cabal and GHC installed on your system. We recommend using [ghcup](https://www.haskell.org/ghcup/). For more information, see [https://www.haskell.org/downloads/](https://www.haskell.org/downloads/).
--- a/clash-lib/src/Clash/Driver/BrokenGhcs.hs
+++ b/clash-lib/src/Clash/Driver/BrokenGhcs.hs
@@ -1,0 +1,161 @@
+{-|
+Copyright   :  (C) 2024, Martijn Bastiaan
+License     :  BSD2 (see the file LICENSE)
+Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Utilities to detect and report GHC / operating system combinations that are
+known to be buggy.
+-}
+
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE QuasiQuotes #-}
+
+module Clash.Driver.BrokenGhcs where
+
+import Data.Maybe (listToMaybe)
+import Data.Version (Version(Version, versionBranch))
+import GHC.Platform (OS(..))
+
+#if __GLASGOW_HASKELL__ > 810
+import System.Info (fullCompilerVersion)
+#endif
+
+import qualified Clash.Util.Interpolate as I
+import qualified System.Info
+
+#if __GLASGOW_HASKELL__ <= 810
+fullCompilerVersion :: Version
+fullCompilerVersion = System.Info.compilerVersion
+#endif
+
+-- | Current OS. Currently only recognizes Linux, Windows, and macOS.
+os :: OS
+os = case System.Info.os of
+  "darwin" -> OSDarwin
+  "linux" -> OSLinux
+  "mingw32" -> OSMinGW32
+  _ -> OSUnknown
+
+-- | What OS GHC is broken on (or all)
+data BrokenOn = All | SomeOs OS
+
+data GhcVersion = Ghc
+  { major0 :: Int
+  , major1 :: Int
+  , patch :: Int
+  }
+  deriving (Eq, Ord)
+
+data GhcRange = GhcRange
+  { from :: GhcVersion
+  -- ^ Start of range, inclusive
+  , to :: GhcVersion
+  -- ^ End of range, exclusive
+  }
+
+-- | Check if a 'GhcVersion' is within a 'GhcRange'
+ghcInRange :: GhcVersion -> GhcRange -> Bool
+ghcInRange ghc GhcRange{from, to} = from <= ghc && ghc < to
+
+-- | Construct a range of all GHC versions matching a major version
+ghcMajor :: Int -> Int -> GhcRange
+ghcMajor major0 major1 = GhcRange
+  { from=Ghc major0 major1 0
+  , to=Ghc major0 (major1 + 1) 0
+  }
+
+data Why = Why
+  { what :: String
+    -- ^ What is broken
+  , solution :: String
+    -- ^ What can be done to work around or solve the issue
+  , issue :: String
+    -- ^ Link to issue
+  , brokenOn :: [(BrokenOn, GhcRange)]
+    -- ^ What operation systems are affected
+  }
+
+-- | Get current GHC version expressed as a triple. It probably does something
+-- non-sensible on unreleased GHCs.
+ghcVersion :: GhcVersion
+ghcVersion = Ghc{major0, major1, patch}
+ where
+  (major0, major1, patch) =
+    case fullCompilerVersion of
+      Version{versionBranch} ->
+        case versionBranch of
+          [] -> (0, 0, 1)
+          [a] -> (a, 0, 1)
+          [a, b] -> (a, b, 1)
+          [a, b, c] -> (a, b, c)
+          (a:b:c:_) -> (a, b, c)
+
+-- | Pretty print 'Why' into an error message
+whyPp :: Why -> String
+whyPp Why{what, solution, issue}= [I.i|
+  Clash has known issues on #{major0}.#{major1}.#{patch} on your current
+  OS. While not completely preventing the compiler from working, we recommend
+  switching to another GHC version. Symptoms:
+
+    #{what}
+
+  Consider the following work around or solution:
+
+    #{solution}
+
+  More information can be found at:
+
+    #{issue}
+
+  If you want to ignore this message, pass the following flag to Clash:
+
+    -fclash-ignore-broken-ghcs
+  |]
+ where
+  Ghc{major0, major1, patch} = ghcVersion
+
+-- | Which GHCs are broken and why
+brokenGhcs :: [Why]
+brokenGhcs = [brokenClashCores, brokenTypeErrors, slowStarts]
+ where
+  brokenClashCores = Why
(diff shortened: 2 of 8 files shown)


clash-lib/src/Clash/Driver/BrokenGhcs.hs
```diff
@@ -0,0 +1,140 @@
...
+    , solution = "Upgrade to GHC 9.4 or downgrade to GHC 9.0"
+    , issue = "<no link>"
+    , brokenOn = [(All, GhcMajor 9 2)]
+    }
+
+  slowStarts = Why
+    { what = "Clash starts really slowly from GHC 9.4.8 up to and including 9.6.2"
+    , solution = "Downgrade to GHC 9.4.7 or upgrade to GHC 9.6.3 or newer"
+    , issue = "https://github.com/clash-lang/clash-compiler/issues/2710"
+    , brokenOn = (All, GhcMinor 9 4 8) : [(All, GhcMinor 9 6 n) | n <- [1, 2]]
```
leonschoorl, line 112: Instead of listing all broken (minor) versions individually, I think it would be better to specify ranges of broken versions.
Something like:
```haskell
   , brokenOn = [(All, GhcRange { from = GhcMinor 9 4 8, to = GhcMinor 9 6 2 }]
```

martijnbastiaan: Done!

clash-lib/src/Clash/Driver/BrokenGhcs.hs
```diff
@@ -0,0 +1,156 @@
...
+  Ghc major minor patch = ghcVersion
+
+-- | Which GHCs are broken and why
+brokenGhcs :: [Why]
+brokenGhcs = [brokenClashCores, brokenTypeErrors, slowStarts]
+ where
+  brokenClashCores = Why
+    { what = "GHC fails to compile clash-cores"
+    , solution = "Upgrade to GHC 9.4 or downgrade to GHC 8.10"
+    , issue = "<no link>"
```
DigitalBrains1, line 121: Given that `clash-cores` is unreleased, this might not tell a user that much. The user might very well conclude that there is some functionality that is not available to them, but they might not know what it is. Interestingly, if I [ask my search engine](https://www.startpage.com/sp/search?query=clash-cores&cat=web&pl=opensearch&language=english) what `clash-cores` means, the results are not very helpful.

So I would prefer if we would elaborate a bit more what we are talking about.

Of course, one way we could to that is to create an issue with a description on GitHub, and link to that.

clash-lib/src/Clash/Driver/BrokenGhcs.hs
```diff
@@ -0,0 +1,156 @@
...
+  brokenTypeErrors = Why
+    { what = "Clash type error messages are indecipherable"
+    , solution = "Upgrade to GHC 9.4 or downgrade to GHC 9.0"
+    , issue = "<no link>"
+    , brokenOn = [(All, ghcMajor 9 2)]
+    }
+
+  slowStarts = Why
+    { what = "Clash starts really slowly from GHC 9.4.8 up to and including 9.6.2"
+    , solution = "Downgrade to GHC 9.4.7 or upgrade to GHC 9.6.3 or newer"
```
DigitalBrains1, line 134:
```suggestion
    , solution = "Upgrade to GHC 9.6.3 or newer, or downgrade to GHC 9.4.7"
```

Ordered from most preferred to least (just like the previous entry).

clash-lib/src/Clash/Driver/BrokenGhcs.hs
```diff
@@ -0,0 +1,156 @@
...
+  -- ^ End of range, exclusive
+  }
+
+-- | Check if a 'GhcVersion' is within a 'GhcRange'
+ghcInRange :: GhcVersion -> GhcRange -> Bool
+ghcInRange ghc GhcRange{from, to} = from <= ghc && ghc < to
+
+-- | Construct a range of all GHC versions matching a major version
+ghcMajor :: Int -> Int -> GhcRange
+ghcMajor major minor = GhcRange{from=Ghc major minor 0, to=Ghc (major + 1) 0 0}
```
DigitalBrains1, line 62: This logic is wrong, and it is because of the mixed terminology. GHC has a 2-part major number. But here it's one part. So what happens is that
```haskell
ghcMajor 9 2
```
becomes
```haskell
GhcRange{from=Ghc 9 2 0, to=Ghc 10 0 0}
```

clash-lib/src/Clash/Driver/BrokenGhcs.hs
```diff
@@ -0,0 +1,156 @@
...
+
+    #{solution}
+
+  More information can be found at:
+
+    #{issue}
+
+  If you want to ignore this message, pass the following flag to Clash:
+
+    -fclash-ignore-broken-ghcs
```
DigitalBrains1, line 109: This looks fine in code: a single blank line before the indented part and a single one after. But I tried it and there is a stray blank line before `#{what}`:
```
<no location info>: error:
    Clash error call:
    Clash has known issues on 9.4.8 on your current OS. While not completely preventing 
    the compiler from working, we recommend switching to another GHC version. Symptoms:
    
    
      Clash type error messages are indecipherable
    
    Consider the following work around or solution:
    
      Upgrade to GHC 9.4 or downgrade to GHC 9.0
    
    More information can be found at:
    
      <no link>
    
    If you want to ignore this message, pass the following flag to Clash:
    
      -fclash-ignore-broken-ghcs
      error, called at src/Clash/Driver/BrokenGhcs.hs:156:15 in clash-lib-1.9.0-inplace:Clash.Driver.BrokenGhcs
```
(this also suffers from the `ghcMajor` function problem, but that's beside the point here)

My suspicion? When a paragraph is more than a single line of text, an extra newline gets tacked at the end of the reflowed paragraph. As that is a clear difference between the first paragraph and the subsequent ones which don't suffer from the issue.

DigitalBrains1 (changes requested):
This is a really nice little feature, but I see one major issue (yes pun intended) and several minor ones! :-)

clash-lib/src/Clash/Driver/BrokenGhcs.hs
martijnbastiaan: 🤦‍♂️ oh wow, I really fooled myself into that one

README.md
```diff
@@ -70,14 +70,14 @@ Note that release branches might contain non-released patches.
...
 | 9.2  | ⚠️¹    | ⚠️¹      | ⚠️¹    | 1.8              | ⚠️¹️
-| 9.4  | ✔️     | ✔️       | ✔️     | 1.8              | ✔️
-| 9.6  | ✔️     | ✔️³      | ✔️     | 1.8              | ✔️
+| 9.4  | ⚠️³ ️   | ⚠️³      | ️⚠️³ ️   | 1.8              | ✔️
+| 9.6  | ✔️³    | ✔️³      | ✔️³    | 1.8              | ✔️
 
 ¹ GHC 9.2 contains a regression, rendering Clash error messages indecipherable. This change was reverted in 9.4.
 
-² GHC 9.0.2 on Windows fails to compile `clash-cores`. We therefore don't run the Clash testsuite on CI for this combination.
+² GHC 9.0.2 on Windows fails to compile `clash-cores`. We therefore don't run the Clash test suite on CI for this combination.
```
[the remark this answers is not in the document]
martijnbastiaan: I think at some point we decided "testsuite" was Dunglish, and that you had to write "test suite".. but now I'm not sure.

README.md
```diff
@@ -70,14 +70,14 @@ Note that release branches might contain non-released patches.
 | 8.10 | ✔️     | ✔️       | ❌    | 1.2 - 1.8        | ✔️
 | 9.0  | ✔️     | ✔️²      | ✔️     | 1.4 - 1.8        | ✔️
 | 9.2  | ⚠️¹    | ⚠️¹      | ⚠️¹    | 1.8              | ⚠️¹️
-| 9.4  | ✔️     | ✔️       | ✔️     | 1.8              | ✔️
-| 9.6  | ✔️     | ✔️³      | ✔️     | 1.8              | ✔️
+| 9.4  | ⚠️³ ️   | ⚠️³      | ️⚠️³ ️   | 1.8              | ✔️
+| 9.6  | ✔️³    | ✔️³      | ✔️³    | 1.8              | ✔️
```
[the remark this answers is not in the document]
martijnbastiaan: yes but _my_ editor....

No joking though, what should we do here? I feel like anytime we change anything here _something_ is going to render missing or extra space.

clash-lib/src/Clash/Driver/BrokenGhcs.hs
martijnbastiaan: Done

clash-lib/src/Clash/Driver/BrokenGhcs.hs
martijnbastiaan: Fixed?

README.md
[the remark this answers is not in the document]
DigitalBrains1: I was kinda thinking that GitHub rendered it as one glyph per character bounding box and so was doing it "correctly". Alternatively, we could decide that we define how GitHub renders it as "correct", at least, hoping that it is browser-independent.

Also, it's not super important.

README.md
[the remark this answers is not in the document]
DigitalBrains1: I agree; it's two words and writing it as one word is probably caused by that being much more common in Dutch.

clash-lib/src/Clash/Driver/BrokenGhcs.hs
DigitalBrains1: I think so. I don't think we need to worry about odd version numbers at all since that is a pretty weird setup, so no need to think about what we want to do with those anyway. And at the end of the day, the user could always use `-fclash-ignore-broken-ghcs`.

martijnbastiaan: The odd numbers are fine, given that the range's upper is exclusive.

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -433,6 +436,7 @@ instance NFData ClashOpts where
     opt_edalize o `deepseq`
     opt_renderEnums o `deepseq`
     opt_timescalePrecision o `deepseq`
+    opt_ignoreBrokenGhcs o `deepseq`
```
leonschoorl, line 439: Not really a comment on this PR, but this feels a little small to open an issue over.

Why aren't we just `deriving NFData ClashOpts`?
This manual instance doesn't evaluate `opt_debug` and `seq`s rather then `deepseq`s `opt_color`, is that by accident, or on purpose?
If so can we probably add a little comment documenting this?

leonschoorl:
We have some special handling of certain flags with respect to HDL caching [over here](https://github.com/clash-lang/clash-compiler/blob/53aabf2b0d069e85ca794572a91372aaa04bcd41/clash-lib/src/Clash/Driver/Manifest.hs#L401-L462).
It might be nice to add this option there too, so adding a `-fclash-ignore-broken-ghcs` doesn't invalidate the HDL cache.
And assuming that the code path producing the "broken ghc error" is way before HDL generation, even removing `-fclash-ignore-broken-ghcs` doesn't have to invalidate the cache.

clash-lib/src/Clash/Driver/Types.hs
martijnbastiaan: I think because `OverridingBool` doesn't have one, but I'd have to check to be sure

clash-lib/src/Clash/Driver/BrokenGhcs.hs
martijnbastiaan: Okay, I've created an issue as this seems out of scope for this PR: https://github.com/clash-lang/clash-compiler/issues/2753. (Though it's probably an easy fix.)

README.md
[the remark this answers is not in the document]
martijnbastiaan: Okay, then I'm considering this resolved :)

clash-lib/src/Clash/Driver/Types.hs
martijnbastiaan: It is indeed `OverridingBool`. I've added an issue https://github.com/clash-lang/clash-compiler/issues/2754.

README.md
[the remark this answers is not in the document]
martijnbastiaan: Gedit renders it the same as GitHub :tada:. Fixed.

clash-lib/src/Clash/Driver/BrokenGhcs.hs
martijnbastiaan: This?

martijnbastiaan:
@DigitalBrains1 @leonschoorl Done.

clash-lib/src/Clash/Driver/BrokenGhcs.hs
DigitalBrains1: The error message:
> GHC fails to compile clash-cores

Without explaining to the user what on Cat's green earth clash-cores is in the first place, users might not know what you're talking about and whether they're affected.

So _this_ = clash-cores.

DigitalBrains1: Where dit the entry go? I no longer see it.

README.md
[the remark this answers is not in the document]
DigitalBrains1: Only now that it _almost_ lines up I can see that the header and horizontal ruler are not aligned
image

clash-lib/src/Clash/Driver/BrokenGhcs.hs
DigitalBrains1: Oh, I get it, instead of explaining what clash-cores is, you just describe that without it, the test suite will not run, without mentioning it by name.

martijnbastiaan: Yes, exactly

README.md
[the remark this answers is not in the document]
martijnbastiaan: It looks okay on my laptop, but on my desktop I'm observing the same as you. Can we just leave this and care about the HTML?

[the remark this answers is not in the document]
martijnbastiaan: You can also force push if you want to make it look okay in your editor/browser.

leonschoorl (approved):

README.md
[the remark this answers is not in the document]
DigitalBrains1: I say let's not sweat this. I hoped we could reach some consistency here but these wide glyphs are just hell in monospaced fonts.

DigitalBrains1 (approved):

PR #2712, opened 2024-04-22
Add `gmiiSgmiiBridge` based on `gig_ethernet_pcs_pma`

lmbollen:
This pull requests adds a wrapper to generate the [LogiCORE IP Ethernet 1000BASE-X PCS/PMA or SGMII](https://docs.amd.com/v/u/en-US/gig_eth_pcs_pma_ds264) as a GMII to SGMII bridge.

It generates a `.tcl` file to be used by the `clashConnector` based infra structure.
For now the configuration is minimal and static.

### DIFF
--- a/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
@@ -1,0 +1,124 @@
+{- |
+Copyright   :  (C) 2024, Google LLC
+License     :  BSD2 (see the file LICENSE)
+Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Contains a wrapper for the Xilinx GMII to SGMII PMA
+
+"LogiCORE IP Ethernet 1000BASE-X PCS/PMA or SGMII"
+
+https://www.xilinx.com/xsw/gig_ethernet_pcs_pma
+
+Currently only the Verilog primitive is supported.
+-}
+
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Clash.Cores.Xilinx.Ethernet.Gmii (
+  -- * GMII SGMII bridge function
+  gmiiSgmiiBridge,
+  -- * GMII interface
+  Gmii (..),
+  -- * GMII SGMII bridge function output
+  BridgeOutput (..),
+  -- ** Status fields
+  Status (..),
+  Pause (..),
+  -- * PHY LVDS signals
+  Lvds (..),
+  -- * Bridge configuration
+  Config (..),
+  AutoNegConfig (..),
+  DuplexMode (..),
+  LinkSpeed (..)) where
+
+import Clash.Explicit.Prelude
+
+import Clash.Cores.Xilinx.Ethernet.Gmii.Internal
+import Clash.Signal.Internal (DiffClock (..))
+
+-- | A subset of the relevant output signals of the GMII to SGMII bridge
+data BridgeOutput gmii125 sgmii625 = BridgeOutput
+  { bridgeLvdsOut :: Signal sgmii625 Lvds
+    -- ^ LVDS output to the PHY
+  , bridgeGmiiRx :: Signal gmii125 Gmii
+    -- ^ GMII output to the MAC
+  , bridgeClk125 :: Clock gmii125
+    -- ^ Clock output for the 125 MHz domain
+  , bridgeRst125 :: Reset gmii125
+    -- ^ Active high reset output for the 125 MHz domain
+  , bridgeStatus :: Signal gmii125 Status
+    -- ^ Status vector
+  }
+
+{- | Wrapper for the LogiCORE IP Ethernet 1000BASE-X PCS/PMA or SGMII, configured to
+function as a GMII to SGMII bridge using LVDS in MAC mode. Currently only the Verilog
+primitive is supported.
+-}
+gmiiSgmiiBridge ::
+  forall sgmii625 gmii125.
+  KnownDomain sgmii625 =>
+  KnownDomain gmii125 =>
+  DomainPeriod sgmii625 ~ Picoseconds 1600 =>
+  DomainPeriod gmii125 ~ Nanoseconds 8 =>
+  DomainActiveEdge sgmii625 ~ 'Rising =>
+  DomainActiveEdge gmii125 ~ 'Rising =>
+  HasAsynchronousReset sgmii625 =>
+  HasSynchronousReset gmii125 =>
+  DomainResetPolarity sgmii625 ~ 'ActiveHigh =>
+  DomainResetPolarity gmii125 ~ 'ActiveHigh =>
+  -- | Reference clock coming from the PHY
+  DiffClock sgmii625 ->
+  -- | Asynchronous reset for the bridge
+  Reset sgmii625 ->
+  -- | Signal detect from the PHY. Either connect to the PHY's signal detect or use
+  -- a constant @True@, otherwise the link will never come up. The IP core considers this
+  -- an asynchronous signal, so synchronisation logic is not needed.
+  Signal sgmii625 Bool ->
+  -- | Configuration for the bridge
+  Signal gmii125 Config ->
+  -- | Auto negotiation configuration for the bridge
+  Signal gmii125 AutoNegConfig ->
+  -- | Restart auto negotiation
+  Signal gmii125 Bool ->
+  -- | LVDS input from the PHY
+  Signal sgmii625 Lvds ->
+  -- | GMII input from the MAC
+  Signal gmii125 Gmii ->
+  -- | Output record, see @BridgeOutput@
+  BridgeOutput gmii125 sgmii625
+gmiiSgmiiBridge refClk refRst signalDetect bridgeConfig anConfig anRestart lvdsIn gmiiTx = BridgeOutput{..}
+ where
+  ( bridgeClk125
+    , activeHighRxRst
+    , lvdsOutP
+    , lvdsOutN
+    , gmiiRxD
+    , gmiiRxDv
+    , gmiiRxEr
+    , fmap fromStatusVector -> bridgeStatus
+    ) =
+      gmiiSgmiiBridgePrim
+        clockP
+        clockN
+        refRst
+        signalDetect
+        bridgeConfig
+        (toAutoNegConfigVector <$> anConfig)
+        anRestart
+        lvdsInP
+        lvdsInN
+        gmiiTxD
+        gmiiTxEn
+        gmiiTxEr
+
+  bridgeRst125 = unsafeFromActiveHigh activeHighRxRst
+  bridgeLvdsOut = Lvds <$> lvdsOutP <*> lvdsOutN
+  (DiffClock clockP clockN) = refClk
+  lvdsInP = pChannel <$> lvdsIn
+  lvdsInN = nChannel <$> lvdsIn
+  gmiiTxD = gmiiData <$> gmiiTx
+  gmiiTxEn = gmiiValid <$> gmiiTx
+  gmiiTxEr = gmiiError <$> gmiiTx
+  bridgeGmiiRx = Gmii <$> gmiiRxD <*> gmiiRxDv <*> gmiiRxEr
--- a/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Internal.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Internal.hs
@@ -1,0 +1,400 @@
+{- |
+  Copyright   :  (C) 2024, Google LLC
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+ Contains the types used in the implementation of the GMII to SGMII bridge wrapper
+-}
+
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE RecordWildCards #-}
+
+{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+module Clash.Cores.Xilinx.Ethernet.Gmii.Internal where
+
+import Clash.Explicit.Prelude hiding ((:<))
+
+import Clash.Annotations.Primitive
+import Clash.Backend (Backend)
+import Clash.Netlist.Types (BlackBoxContext (..), TemplateFunction (..))
+import Clash.Signal.Internal
+import Control.Monad.State (State)
+import Data.List.Infinite ((...), Infinite(..))
+import Data.String.Interpolate
+import Data.Text.Prettyprint.Doc.Extra (Doc)
+import Prettyprinter.Interpolate (__di)
+
+import qualified Clash.Explicit.Prelude as C
+
+-- | A differential pair of Low Voltage Differential Signaling (LVDS) signals
+data Lvds = Lvds {pChannel :: Bit, nChannel :: Bit}
+  deriving (Generic, NFDataX, BitPack)
+
+-- TODO: Move to re-usable module?
+
+-- | A generic GMII interface
+data Gmii = Gmii
+  { gmiiData :: BitVector 8
+  , gmiiValid :: Bit
+  , gmiiError :: Bit
+  }
+  deriving (Generic, BitPack, NFDataX)
+
+-- The order of the constructors of `DuplexMode`, `LinkSpeed` and `Pause` is
+-- important, as their derived `BitPack` instances are used to encode them in
+-- the `AutoNegConfigVector`.
+
+data DuplexMode = HalfDuplex | FullDuplex
+  deriving (Generic, NFDataX, BitPack, Eq, Show)
+
+-- | Link speed in megabits per second.
+data LinkSpeed = Speed10 | Speed100 | Speed1000
+  deriving (Generic, NFDataX, BitPack, Eq, Show)
+
+-- | Whether the link partner supports pause frames.
+data Pause = NoPause | SymmetricPause | AsymmetricPause | SymmetricAsymmetricPause
+  deriving (Generic, NFDataX, BitPack, Eq, Show)
+
+newtype AutoNegConfigVector = AutoNegConfigVector (BitVector 16)
+  deriving (Generic, NFDataX, BitPack, Eq, Show)
+
+-- | A subset of the an_adv_config_vector that is applicable for the GMII to SGMII bridge
+data AutoNegConfig
+  = AutoNegConfig
+  { cAcknowledge :: Bool
+  , cDuplexMode :: DuplexMode
+  , cLinkSpeed :: LinkSpeed
+  , cPhyLinkStatus :: Bool
+  }
+  deriving (Generic, NFDataX, BitPack, Eq, Show)
+
+-- | Converts the AutoNegConfig Records to the an_adv_config_vector format
+toAutoNegConfigVector :: AutoNegConfig -> AutoNegConfigVector
+toAutoNegConfigVector AutoNegConfig{..} = AutoNegConfigVector $ pack cfg
+ where
+  -- See LogiCORE IP Ethernet 1000BASE-X PCS/PMA or SGMII v11.3, page 17,
+  -- table 7: Optional Auto-Negotiation Interface Signal Pinout, signal
+  -- an_adv_config_vector[15:0]
+  -- (SGMII operating in PHY mode)
+  cfg :: Vec 16 Bit
+  cfg =
+    ( bitCoerce cPhyLinkStatus -- Bit 15 - PHY Link Status
+        :> bitCoerce cAcknowledge -- Bit 14 - Acknowledge
+        :> 0 -- Bit 13 - Reserved
+        :> bitCoerce cDuplexMode -- Bit 12 - Duplex Mode
+        :> Nil
+    )
+      ++ bitCoerce cLinkSpeed -- Bit 11:10 - Speed
+      ++ ( repeat 0 -- Bit 9:1 - Reserved
+           C.:< 1 -- Bit 0 - Always 1
+         )
+
+-- | A conservative default configuration for the AutoNegConfig record
+instance Default AutoNegConfig where
+  def = AutoNegConfig{..}
+   where
+    cAcknowledge = False
+    cDuplexMode = HalfDuplex
+    cLinkSpeed = Speed10
+    cPhyLinkStatus = False
+
+-- The order of the fields in the `Config` record is important, as their derived
+-- `BitPack` instance is used as input for the `gmiiSgmiiBridgePrim` primitive.
+-- | A record representation of the configuration_vector
+data Config = Config
+  { cAutoNegEnable :: Bool
+  , cIsolateGmii :: Bool
+  , cPowerDown :: Bool
+  , cLoopback :: Bool
+  , cUnidirectional :: Bool
+  }
+  deriving (Generic, NFDataX, BitPack)
+
+instance Default Config where
+  def = Config{..}
+   where
+    cUnidirectional = False
+    cLoopback = False
+    cPowerDown = False
+    cIsolateGmii = False
+    cAutoNegEnable = False
+
+type StatusVector = BitVector 16
+
+-- | A record representation of the status_vector
+data Status = Status
+  { sLinkStatus :: Bool
+  , sLinkSynchronized :: Bool
+  , sRudiC :: Bool
+  , sRudiI :: Bool
+  , sRudiInvalid :: Bool
+  , sReceivedDisparityError :: Bool
+  , sReceivedInvalidCode :: Bool
+  , sPhyLinkStatus :: Bool
+  , sRemoteFault :: Maybe (BitVector 2)
+  , sLinkSpeed :: LinkSpeed
+  , sDuplexMode :: DuplexMode
+  , sPause :: Pause
+  }
+  deriving (Generic, NFDataX, BitPack)
+
+-- | Transform the status_vector to a Status record
+fromStatusVector :: StatusVector -> Status
+fromStatusVector v = Status{..}
+ where
+  -- There is no BitPack instance for 13-tuples without -flarge-tuples, so we break it up.
+  ( ( sLinkStatus
+      , sLinkSynchronized
+      , sRudiC
+      , sRudiI
+      , sRudiInvalid
+      , sReceivedDisparityError
+      )
+    , sReceivedInvalidCode
+    , sPhyLinkStatus
+    , remoteFaultCode
+    , sLinkSpeed
+    , sDuplexMode
+    , remoteFaultValid
+    , sPause
+    ) = unpack v
+  sRemoteFault = if remoteFaultValid then Just remoteFaultCode else Nothing
+
+{- | Primitive for the LogiCORE IP Ethernet 1000BASE-X PCS/PMA or SGMII, configured
+to function as a GMII to SGMII bridge using LVDS in MAC mode. Currently only the Verilog
+primitive is supported.
+-}
+gmiiSgmiiBridgePrim ::
+  forall sgmii625 gmii125.
+  ( KnownDomain sgmii625
+  , KnownDomain gmii125
+  , DomainPeriod sgmii625 ~ Picoseconds 1600
+  , DomainPeriod gmii125 ~ Nanoseconds 8
+  , DomainActiveEdge sgmii625 ~ 'Rising
+  , DomainActiveEdge gmii125 ~ 'Rising
+  , HasAsynchronousReset sgmii625
+  , HasSynchronousReset gmii125
+  , DomainResetPolarity sgmii625 ~ 'ActiveHigh
+  , DomainResetPolarity gmii125 ~ 'ActiveHigh
+  ) =>
+  -- | P channel of the reference clock coming from the PHY
+  Clock sgmii625 ->
+  -- | N channel of the reference clock coming from the PHY
+  ClockN sgmii625 ->
+  -- | Signal detect from the PHY. Either connect to the PHY's signal detect or use
+  -- a constant @True@, otherwise the link will never come up. The IP core considers this
+  -- an asynchronous signal, so synchronisation logic is not needed.
+  Reset sgmii625 ->
+  -- | Signal detect from the PHY. Either connect to the PHY's signal detect or use
+  -- a constant @True@, otherwise the link will never come up.
+  Signal sgmii625 Bool ->
+  -- | Configuration for the bridge
+  Signal gmii125 Config ->
+  -- | Auto negotiation configuration for the bridge
+  Signal gmii125 AutoNegConfigVector ->
+  -- | Restart auto negotiation
+  Signal gmii125 Bool ->
+  -- | P channel of the LVDS input from the PHY
+  Signal sgmii625 Bit ->
+  -- | N channel of the LVDS input from the PHY
+  Signal sgmii625 Bit ->
+  -- | GMII data input from the MAC
+  Signal gmii125 (BitVector 8) ->
+  -- | GMII valid input from the MAC
+  Signal gmii125 Bit ->
+  -- | GMII error input from the MAC
+  Signal gmii125 Bit ->
+  -- |
+  -- 1. Clock output for the 125 MHz domain
+  -- 2. Active high reset output for the 125 MHz domain
+  -- 3. P channel of the LVDS output to the PHY
+  -- 4. N channel of the LVDS output to the PHY
+  -- 5. GMII data output to the MAC
+  -- 6. GMII valid output to the MAC
+  -- 7. GMII error output to the MAC
+  -- 8. Status vector
+  ( Clock gmii125
+  , Signal gmii125 Bool
+  , Signal sgmii625 Bit
+  , Signal sgmii625 Bit
+  , Signal gmii125 (BitVector 8)
+  , Signal gmii125 Bit
+  , Signal gmii125 Bit
+  , Signal gmii125 StatusVector
+  )
+-- Can't use @deepErrorX@ because one of the results is a clock
+gmiiSgmiiBridgePrim !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
+  (clockGen, undefined, undefined, undefined, undefined, undefined, undefined, undefined)
+{-# CLASH_OPAQUE gmiiSgmiiBridgePrim #-}
+{-# ANN
+  gmiiSgmiiBridgePrim
+  ( let
+      tfName = 'tclTemplateFunction
+      primName = 'gmiiSgmiiBridgePrim
+      -- Constraints
+      _knownDomSgmii
+        :< _knownDomGmii
+        :< _domPeriodSgmii
+        :< _domPeriodGmii1
+        :< _domEdgeSgmii
+        :< _domEdgeGmii
+        :< _hasAsyncRstSgmii
+        :< _hasSyncRstGmii
+        :< _domRstPolSgmii
+        :< _domRstPolgmii
+        :<
+        -- Arguments
+        sgmiiClkP
+        :< sgmiiClkN
+        :< rst
+        :< signalDetect
+        :< pmaConfig
+        :< pmaAnAdvancedConfig
+        :< restartAn
+        :< lvdsInP
+        :< lvdsInN
+        :< gmiiTxD
+        :< gmiiTxDv
+        :< gmiiTxEr
+        :< _ =
+          ((0 :: Int) ...)
+
+      -- Symbols
+      clk125
+        :< rst125
+        :< lvdsOutP
+        :< lvdsOutN
+        :< gmiiRxD
+        :< gmiiRxDv
+        :< gmiiRxEr
+        :< statusVector
+        :< compName
+        :< _ =
+          ((0 :: Int) ...)
+     in
+      InlineYamlPrimitive
+        [Verilog]
+        [__i|
+      BlackBox:
+        kind: Declaration
+        name: #{primName}
+        template: |-
+          wire ~GENSYM[pmaClk125][#{clk125}];
+          wire ~GENSYM[pmaRst125][#{rst125}];
+          wire ~GENSYM[lvdsOutP][#{lvdsOutP}];
+          wire ~GENSYM[lvdsOutN][#{lvdsOutN}];
+          wire [7:0] ~GENSYM[gmiiRxD][#{gmiiRxD}];
+          wire ~GENSYM[gmiiRxDv][#{gmiiRxDv}];
+          wire ~GENSYM[gmiiRxEr][#{gmiiRxEr}];
+          wire [15:0] ~GENSYM[pmaStatusVector][#{statusVector}];
+
+          ~INCLUDENAME[0] ~GENSYM[gig_ethernet_pcs_pma][#{compName}] (
+            .refclk625_p(~ARG[#{sgmiiClkP}]),
+            .refclk625_n(~ARG[#{sgmiiClkN}]),
+            .reset(~ARG[#{rst}]),
+            .speed_is_100(0),
+            .speed_is_10_100(0),
+            .signal_detect(~ARG[#{signalDetect}]),
+            .configuration_vector(~ARG[#{pmaConfig}]),
+            .an_adv_config_vector(~ARG[#{pmaAnAdvancedConfig}]),
+            .an_restart_config(~ARG[#{restartAn}]),
+            .rxp(~ARG[#{lvdsInP}]),
+            .rxn(~ARG[#{lvdsInN}]),
+            .gmii_txd(~ARG[#{gmiiTxD}]),
+            .gmii_tx_en(~ARG[#{gmiiTxDv}]),
+            .gmii_tx_er(~ARG[#{gmiiTxEr}]),
+            .clk125_out(~SYM[#{clk125}]),
+            .rst_125_out(~SYM[#{rst125}]),
+            .txp(~SYM[#{lvdsOutP}]),
+            .txn(~SYM[#{lvdsOutN}]),
+            .gmii_rxd(~SYM[#{gmiiRxD}]),
+            .gmii_rx_dv(~SYM[#{gmiiRxDv}]),
+            .gmii_rx_er(~SYM[#{gmiiRxEr}]),
+            .status_vector(~SYM[#{statusVector}])
+          );
+          assign ~RESULT =
+            { ~SYM[#{clk125}]
+            , ~SYM[#{rst125}]
+            , ~SYM[#{lvdsOutP}]
+            , ~SYM[#{lvdsOutN}]
+            , ~SYM[#{gmiiRxD}]
+            , ~SYM[#{gmiiRxDv}]
+            , ~SYM[#{gmiiRxEr}]
+            , ~SYM[#{statusVector}]
+            };
+        includes:
+          - extension: clash.tcl
+            name: gig_ethernet_pcs_pma
+            format: Haskell
+            templateFunction: #{tfName}
+
+          |]
+  )
+  #-}
+
+tclTemplateFunction :: TemplateFunction
+tclTemplateFunction = TemplateFunction used valid tclTemplate
+ where
+  used = [sgmiiClkP .. gmiiTxEr] -- use all arguments
+  valid = const True
+  -- Constraints
+  _knownDomSgmii
+    :< _knownDomGmii
+    :< _domPeriodSgmii
+    :< _domPeriodGmii
+    :< _domEdgeSgmii
+    :< _domEdgeGmii
+    :< _hasAsyncRstSgmii
+    :< _hasSyncRstGmii
+    :< _domRstPolSgmii
+    :< _domRstPolGmii
+    :<
+    -- Arguments
+    sgmiiClkP
+    :< _sgmiiClkN
+    :< _rst
+    :< _signalDetect
+    :< _pmaConfig
+    :< _pmaAnAdvancedConfig
+    :< _restartAn
+    :< _lvdsInP
+    :< _lvdsInN
+    :< _gmiiTxD
+    :< _gmiiTxDv
+    :< gmiiTxEr
+    :< _ =
+      (0...)
+
+tclTemplate :: Backend s => BlackBoxContext -> State s Doc
+tclTemplate bbCtx
+  | [compName] <- bbQsysIncName bbCtx =
+      let
+        bbText =
+          [__di|
+      namespace eval $tclIface {
+        variable api 1
+        variable scriptPurpose createIp
+        variable ipName {#{compName}}
+        proc createIp {ipName0 args} {
+          create_ip -name gig_ethernet_pcs_pma -vendor xilinx.com -library ip \\\&
+            -version 16.2 -module_name $ipName0
+          set_property -dict [list \\\&
+            CONFIG.LvdsRefClk {625} \\\&
+            CONFIG.Standard {SGMII} \\\&
+            CONFIG.Physical_Interface {LVDS} \\\&
+            CONFIG.Management_Interface {false} \\\&
+            CONFIG.Ext_Management_Interface {false} \\\&
+            CONFIG.SGMII_PHY_Mode {false} \\\&
+            CONFIG.SupportLevel {Include_Shared_Logic_in_Core}] [get_ips $ipName0]
+          return
+        }
+      }
+      |]
+       in
+        pure bbText
+  | otherwise = error "gmiiSgmiiBridgePrim: Expected exactly one component name"
--- a/clash-cores/clash-cores.cabal
+++ b/clash-cores/clash-cores.cabal
@@ -147,6 +147,8 @@ library
     Clash.Cores.Xilinx.Floating.BlackBoxes
     Clash.Cores.Xilinx.Floating.Explicit
     Clash.Cores.Xilinx.Floating.Internal
+    Clash.Cores.Xilinx.Ethernet.Gmii
+    Clash.Cores.Xilinx.Ethernet.Gmii.Internal
     Clash.Cores.Xilinx.Ila
     Clash.Cores.Xilinx.Ila.Internal
     Clash.Cores.Xilinx.Internal
@@ -206,6 +208,7 @@ test-suite unittests
     Test.Cores.Xilinx.BlockRam
     Test.Cores.Xilinx.DcFifo
     Test.Cores.Xilinx.DnaPortE2
+    Test.Cores.Xilinx.Ethernet.Gmii
 
   build-depends:
     clash-cores,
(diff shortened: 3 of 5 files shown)


clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+
+-- | Black box for the GMII to SGMII bridge.
+gmiiSgmiiBridgeBb ::
+  -- | P channel of the reference clock coming from the Phy.
+  Clock sgmii625 ->
+  -- | N channel of the reference clock coming from the Phy.
+  ClockN sgmii625 ->
+  -- | Asynchronous active high reset for the bridge.
+  Signal asyncDom Bool->
+  -- | Whether the link should be configured for 100 Mbit/s.
```
christiaanb, line 89: Since this function is exported, the illegal settings should at least be documented, but preferably checked.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+  Signal gmiiRx125 Speed_100 ->
+  -- | Whether the link should be configured for 10 or 100 Mbit/s.
+  Signal gmiiRx125 Speed_10_100 ->
+  -- | Signal detect from the Phy.
```
christiaanb, line 93: See my earlier comment on the `SignalDetect` argument.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,181 @@
...
+data Gmii = Gmii
+ { gmiiData :: BitVector 8
+ , gmiiValid :: Bit
+ , gmiiError :: Bit
+ } deriving (Generic, BitPack, NFDataX)
+
+type SignalDetect = Bool
+type RestartAutoNeg = Bool
+type ConfigValid = Bool
+type BridgeAnConfigValid = Bool
```
christiaanb, line 34: If these names are important to keep around, then use newtypes instead of type synonyms. Type synonyms are easily lost in type inference and error messages. Otherwise simply drop them and use `Bool` at their current use-sites instead.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+  , Signal gmiiRx125 (BitVector 8)
+  , Signal gmiiRx125 Bit
+  , Signal gmiiRx125 Bit
+  , Clock gmiiRx125
+  , Signal gmiiRx125 Bool
+  , Signal gmiiRx125 StatusVector
+  )
+
+gmiiSgmiiBridgeBb !_ !_  !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
+  (undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined)
```
christiaanb, line 132: Why not `deepErrorX`?

christiaanb (dismissed):

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
lmbollen: One of the results is a `Clock`, added a comment.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+  , Clock gmiiRx125
+  , Signal gmiiRx125 Bool
+  , Signal gmiiRx125 StatusVector
+  )
+
+-- Can't use `deepErrorX' because one of the results is a clock.
+gmiiSgmiiBridgeBb !_ !_  !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
+  (undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined)
+
+{-# OPAQUE gmiiSgmiiBridgeBb #-}
```
DigitalBrains1, line 134:
```suggestion
{-# CLASH_OPAQUE gmiiSgmiiBridgeBb #-}
```

(and don't forget `Language CPP`)

More in general, there are still things preventing succesful compilation, see CI.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,161 @@
...
+  } deriving (Generic, NFDataX, BitPack)
+
+-- | Converts the AutoNegConfig Records to the an_adv_config_vector format.
+toAutoNegConfigVector :: AutoNegConfig -> AutoNegConfigVector
+toAutoNegConfigVector AutoNegConfig{..} = bv4
+  where
+    (linkSpeed0, linkSpeed1) = case cLinkSpeed of
+      Speed10 -> (0, 0)
+      Speed100 -> (0, 1)
+      Speed1000 -> (1, 0)
```
DigitalBrains1, line 55: It feels like this could just be done with `pack @LinkSpeed` when the constructors are ordered suitably.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,161 @@
...
+    , remoteFaultCode
+    , linkSpeedBools
+    , duplexModeBool
+    , remoteFaultValid
+    , pauseBools) = unpack v
+  sRemoteFault = if remoteFaultValid then Just remoteFaultCode else Nothing
+  sLinkSpeed = case linkSpeedBools of
+    (True, False) -> Speed1000
+    (False, True) -> Speed100
+    (False, False) -> Speed10
```
DigitalBrains1, line 145: It would be nice if this could be done with `unpack @LinkSpeed`, but failing that, surely it could be done with `linkSpeedBV :: BitVector 2` rather than booleans?
```suggestion
  sLinkSpeed = case linkSpeedBV of
    0 -> Speed10
    1 -> Speed100
    2 -> Speed1000
```

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+  , KnownDomain gmiiRx125
+  , KnownDomain gmiiTx125
+  , DomainPeriod sgmii625 ~  Picoseconds 1600
+  , DomainPeriod gmiiRx125 ~ Nanoseconds 8
+  , DomainPeriod gmiiTx125 ~ Nanoseconds 8
+  , DomainActiveEdge sgmii625 ~ Rising
+  , DomainActiveEdge gmiiRx125 ~ Rising
+  , DomainActiveEdge gmiiTx125 ~ Rising
+  ) =>
+  -- | Reference clock coming from the Phy.
```
DigitalBrains1, line 41:
```suggestion
  -- | Reference clock coming from the PHY
```

This is in more places: you write MAC all-caps, which I agree with, but you don't do it for PHY.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
```diff
@@ -0,0 +1,129 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+module Clash.Cores.Xilinx.Ethernet.Gmii.BlackBoxes where
```
DigitalBrains1, line 3:
```suggestion
{-# LANGUAGE QuasiQuotes #-}

{-# OPTIONS_HADDOCK hide #-}

module Clash.Cores.Xilinx.Ethernet.Gmii.BlackBoxes where
```

DigitalBrains1 (commented):
Nice job figuring out the primitive!

But I do think there's room for improvement. CI has some errors for you as well.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,161 @@
...
+    (False, True) -> SymmetricPause
+    (True, False) -> AsymmetricPause
+    (True, True) -> SymmetricAsymmetricPause
+
+-- | A subset of the relevant output signals of the GMII to SGMII bridge.
+data BridgeOutput gmiiRx125 sgmii625 = BridgeOutput
+  { bridgeLvdsOut :: Signal sgmii625 Lvds
+  , bridgeGmiiRx :: Signal gmiiRx125 Gmii
+  , bridgeClk125 :: Clock gmiiRx125
+  , bridgeRst125 :: Reset gmiiRx125
```
DigitalBrains1, line 159: Is this reset synchronous or asynchronous? We probably need a constraint somewhere so it can only be used properly.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
```diff
@@ -0,0 +1,129 @@
...
+            name: gig_ethernet_pcs_pma
+            format: Haskell
+            templateFunction: #{tfName}
+
+          |]
+
+binaryTclTF :: TemplateFunction
+binaryTclTF = TemplateFunction used valid tclTemplate
+ where
+  used = [0..11] -- use all arguments
```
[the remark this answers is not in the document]
lmbollen: If I look at other implementations I see the magic constants as pattern, do you know a way to circumvent that?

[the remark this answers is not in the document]
DigitalBrains1: We did use to do that, yes, just like we used to just write the argument number in blackbox template `~ARG` and friends. But for both we now use the same solution:
https://github.com/clash-lang/clash-compiler/blob/a26bb9ef3cb57f24daec3da30db9fb54af91297a/clash-cores/src/Clash/Cores/Xilinx/BlockRam/BlackBoxes.hs#L35-L55

So
```haskell
 where
  _smiiClkP :<
    _smiiClkN :<
    _rst :<
    _signalDetect :<
    _pmaConfig :<
    _pmaAnAdvancedConfig :<
    _restartAn :<
    _lvdsInP :<
    _lvdsInN :<
    _gmiiTxD :<
    _gmiiTxDv :<
    _gmiiTxEr :< _
    = ((0 :: Int)...)
  used = [0..gmiiTxEr] -- use all arguments
```

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
```diff
@@ -0,0 +1,129 @@
...
+  | [compName] <- bbQsysIncName bbCtx
+  =
+  let
+    bbText = [__di|
+      namespace eval $tclIface {
+        variable api 1
+        variable scriptPurpose createIp
+        variable ipName {#{compName}}
+        proc createIp {ipName0 args} {
+          create_ip -name gig_ethernet_pcs_pma -vendor xilinx.com -library ip -version 16.2 \\
```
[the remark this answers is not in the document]
DigitalBrains1: Hmmmm. Actually, I've decided I would prefer 80 column Tcl... but it doesn't matter because it boils down to the same thing in this case! X-D

```haskell
          create_ip -name gig_ethernet_pcs_pma -vendor xilinx.com -library ip \\
            -version 16.2 -module_name $ipName0
```

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
```diff
@@ -0,0 +1,129 @@
...
+      rst125 :<
+      statusVector :<
+      compName :< _
+      = ((0 :: Int)...)
+    in InlineYamlPrimitive [Verilog, SystemVerilog] [__i|
+      BlackBox:
+        name: Clash.Cores.Xilinx.Ethernet.Gmii.gmiiSgmiiBridgeBb
+        kind: Declaration
+        template: |-
+          wire ~GENSYM[lvdsOutP][#{lvdsOutP}];
```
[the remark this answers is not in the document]
lmbollen: Refactored it, ptal

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
lmbollen: What is the `OPTIONS_HADDOCK` pragma for?

DigitalBrains1: Hiding the completely undocumented module from Haddock. Even if there was documentation it still wouldn't need to be in the Haddock doc.

But I'll also propose a change to the module structure today.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
lmbollen: I get an `Unrecognised pragma` warning on both `LANGUAGE CPP` and the `CLASH_OPAQUQE` annotation. Do you perhaps know why?

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+  activeHighRefRst = unsafeToActiveHigh refRst
+  bridgeRst125 = unsafeFromActiveHigh activeHighRxRst
+  bridgeLvdsOut = Lvds <$> lvdsOutP <*> lvdsOutN
+  (DiffClock clockP clockN) = refClk
+  (lvdsInP, lvdsInN) = (fmap lvdsP lvdsIn, fmap lvdsN lvdsIn)
+  (gmiiTxD, gmiiTxEn, gmiiTxEr) = unbundle $ fmap (\Gmii{..} -> (gmiiData, gmiiValid, gmiiError)) gmiiTx
+  bridgeGmiiRx = Gmii <$> gmiiRxD <*> gmiiRxDv <*> gmiiRxEr
+
+-- | Black box for the GMII to SGMII bridge.
+gmiiSgmiiBridgeBb ::
```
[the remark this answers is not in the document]
lmbollen: Unified the constraints

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+  , DomainPeriod gmiiRx125 ~ Nanoseconds 8
+  , DomainPeriod gmiiTx125 ~ Nanoseconds 8
+  , DomainActiveEdge sgmii625 ~ Rising
+  , DomainActiveEdge gmiiRx125 ~ Rising
+  , DomainActiveEdge gmiiTx125 ~ Rising
+  ) =>
+  -- | Reference clock coming from the Phy.
+  DiffClock sgmii625 ->
+  -- | Asynchronous reset for the bridge.
+  Reset asyncDom ->
```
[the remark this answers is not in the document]
lmbollen: Added the correct constraints

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,136 @@
...
+{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
+
+module Clash.Cores.Xilinx.Ethernet.Gmii where
+
+import Clash.Explicit.Prelude
+
+import Clash.Annotations.Primitive
+import Clash.Cores.Xilinx.Ethernet.Gmii.BlackBoxes
+import Clash.Cores.Xilinx.Ethernet.Gmii.Types
+import Clash.Explicit.Prelude
```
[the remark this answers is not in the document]
lmbollen: Odd, does not show up in my code atm

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
lmbollen: The Auto negotiation configuration vector and status vector both contain link speed information, but their formats differ.

I'd like to not rely implicitly on the order of the constructors for the correct behavior.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,161 @@
...
+
+    bv0 = 0x4001 -- How the bridge is configured in MAC mode.
+
+    -- Apply the configuration.
+    bv1 = replaceBit (12 :: Int) (bitCoerce $ cDuplexMode == FullDuplex) bv0
+    bv2 = replaceBit (10 :: Int) linkSpeed0 bv1
+    bv3 = replaceBit (11 :: Int) linkSpeed1 bv2
+
+    -- Set link status
+    bv4 = setBit bv3 (15:: Int)
```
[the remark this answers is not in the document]
lmbollen: I understand your impression when reading this code. It was written to stay close to the documentation.
Please see the definition of `an_adv_config_vector` page 17 of https://docs.amd.com/v/u/en-US/gig_eth_pcs_pma_ds264. After that we can discuss the most correct / transparant way to implement this

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,161 @@
...
+  { cUniDirectional :: Bool
+  , cLoopback :: Bool
+  , cPowerDown :: Bool
+  , cIsolateGmii :: Bool
+  , cAutoNegEnable :: Bool
+  } deriving (Generic, NFDataX, BitPack)
+
+-- | Transform the Config record to a configuration_vector.
+toConfigVector :: Config -> ConfigVector
+toConfigVector Config{..} = pack
```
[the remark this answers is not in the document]
lmbollen: Like the other `pack` suggestion, personally I'd prefer to not implicitly rely on the order of constructors, or in this case, field definitions.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
lmbollen: The constraints have been added to the functions. Would you like to also put them on the record?

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
DigitalBrains1: It works for me, you must have gotten a detail wrong. I tested it with GHC 9.6.4, I'm now testing with 9.0.2 to be completely certain, but I'd be flabbergasted if it failed.

```diff
--- a/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
@@ -9,8 +9,10 @@
 -}
 
 {-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ViewPatterns #-}
+
 {-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
 
 module Clash.Cores.Xilinx.Ethernet.Gmii where
@@ -131,6 +133,6 @@ gmiiSgmiiBridgeBb ::
 gmiiSgmiiBridgeBb !_ !_  !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
   (undefined, undefined, undefined, undefined, undefined, undefined, undefined, undefined)
 
-{-# OPAQUE gmiiSgmiiBridgeBb #-}
+{-# CLASH_OPAQUE gmiiSgmiiBridgeBb #-}
 {-# ANN gmiiSgmiiBridgeBb hasBlackBox #-}
 {-# ANN gmiiSgmiiBridgeBb (veriBinaryPrim) #-}
```

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
DigitalBrains1: No, I think it's good enough to have it just on the functions.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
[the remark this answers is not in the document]
lmbollen: Am I correct in that constraints are not counted as arguments in this sense?

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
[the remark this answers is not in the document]
DigitalBrains1: To that I have two counterarguments:
1. I don't think it's actually implicit. I mean, our documentation doesn't do enough to call it out, but ADTs always generate `BitPack` instances that follow [`ConstructorType Binary`](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Annotations-BitRepresentation-Deriving.html#t:ConstructorType) and [`FieldsType OverlapL`](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Annotations-BitRepresentation-Deriving.html#t:FieldsType).
2. Making your ADTs match some predetermined layout is good practice because it saves LUTs. In the `Config` case, it renders to wires, but for instance in the case of `LinkSpeed` it actually leads to LUTs that are completely superfluous. And the alternative of explicitly generating a `BitRepresentation` that matches what it automatically would be anyway is just unnecessary complexity, both to the compiler as well as to the programmer.

I'm of the belief that labeling this "implicitly relying on order of constructors" is a misconception and an impediment to good design. But we can widen the discussion up and ask more of our colleagues what they think.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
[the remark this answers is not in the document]
DigitalBrains1: No, they're counted. So you'll have to change the 0 as well.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
DigitalBrains1: Ah, that is a shame. I think it's best to make the `BitPack` match one of those and use explicit conversion for the other (see elsewhere for why I still want to use `BitPack`). That way you at least save some LUTs.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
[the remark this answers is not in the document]
lmbollen: > ... but ADTs always generate BitPack instances that follow [ConstructorType Binary](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Annotations-BitRepresentation-Deriving.html#t:ConstructorType) and [FieldsType OverlapL](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Annotations-BitRepresentation-Deriving.html#t:FieldsType).

So the order of constructors determines the meaning as interpreted by the IP, which is what I wanted to prevent.

> Making your ADTs match some predetermined layout is good practice because it saves LUTs.
I do agree with you, perhaps we can do it that way instead and create some small Hunit assertions that verify the correctness?

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
lmbollen: I get
```hs
src/Clash/Cores/Xilinx/Ethernet/Gmii.hs:12:1: warning: [-Wunrecognised-pragmas]
    Unrecognised pragma
   |
12 | {-# LANGAUGE CPP #-}
   | ^^^
src/Clash/Cores/Xilinx/Ethernet/Gmii.hs:215:1: warning: [-Wunrecognised-pragmas]
    Unrecognised pragma
    |
215 | {-# CLASH_OPAQUE gmiiSgmiiBridgePrim #-}
    | ^^^
```
Let's debug this face to face :)

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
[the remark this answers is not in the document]
DigitalBrains1: > So the order of constructors determines the meaning as interpreted by the IP, which is what I wanted to prevent.

I'm not sure I follow, but that sounds in reverse. The interpretation by the IP determines the order you specify the constructors. The IP is what's fixed, it's the determinant.

> perhaps we can do it that way instead and create some small Hunit assertions that verify the correctness?

That'd be nice, yes.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
DigitalBrains1: My face is 170 km away, but that won't be necessary. langAUge isn't language. It's perhaps an indicator of the speed of your local area network, but it's not a word nor is it the correct word.

:-P

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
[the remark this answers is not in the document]
DigitalBrains1: Hmmmm I'm thinking
```haskell
  let
    tfName = 'tclTemplateFunction

    -- Constraints
    _knownDomSgmii :<
      _knownDomGmii :<
      _domPeriodsgmii :<
      _domPeriodgmii1 :<
      _domEdgesgmii :<
      _domEdgegmii :<
      _domRstKindasync :<
      _domRstKindgmii :<
      _domRstPolasync :<
      _domRstPolgmii :<
    -- Arguments
      smiiClkP :<
      smiiClkN :<
      rst :<
      signalDetect :<
      pmaConfig :<
      pmaAnAdvancedConfig :<
      restartAn :<
      lvdsInP :<
      lvdsInN :<
      gmiiTxD :<
      gmiiTxDv :<
      gmiiTxEr :< _
      = ((0 :: Int)...)

    -- Symbols
    clk125 :<
      rst125 :<
      lvdsOutP :<
      lvdsOutN :<
      gmiiRxD :<
      gmiiRxDv :<
      gmiiRxEr :<
      statusVector :<
      compName :< _
      = ((0 :: Int)...)

    in InlineYamlPrimitive [Verilog, SystemVerilog] [__i|
```

I like your idea of adding comments to delineate constraints from arguments. But the symbols are a separate namespace and that list is also a numbering of _symbols_, not of _results_, not just in general but also in this specific case of `compName` which is a symbol but not a result.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,151 @@
...
+
+-}
+
+{-# LANGAUGE CPP #-}
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
+
+module Clash.Cores.Xilinx.Ethernet.Gmii where
```
DigitalBrains1, line 18:
```suggestion
{-# LANGUAGE BangPatterns #-}
{-# LANGUAGE CPP #-}
{-# LANGUAGE RecordWildCards #-}
{-# LANGUAGE ViewPatterns #-}

{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}

module Clash.Cores.Xilinx.Ethernet.Gmii where
```

Typo, sorting, bit more vertical whitespace for ease of comprehension.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,151 @@
...
+  , Signal gmii125 StatusVector
+  )
+
+-- Can't use @deepErrorX@ because one of the results is a clock.
+gmiiSgmiiBridgePrim !_ !_  !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
+  (clockGen, undefined, undefined, undefined, undefined, undefined, undefined, undefined)
+
+{-# CLASH_OPAQUE gmiiSgmiiBridgePrim #-}
+{-# ANN gmiiSgmiiBridgePrim hasBlackBox #-}
+{-# ANN gmiiSgmiiBridgePrim (verilogPrim) #-}
```
DigitalBrains1, line 151:
```suggestion
-- Can't use @deepErrorX@ because one of the results is a clock.
gmiiSgmiiBridgePrim !_ !_  !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
  (clockGen, undefined, undefined, undefined, undefined, undefined, undefined, undefined)
{-# CLASH_OPAQUE gmiiSgmiiBridgePrim #-}
{-# ANN gmiiSgmiiBridgePrim verilogPrim #-}
```


I think annotations for both `hasBlackBox` as well as `InlineYamlPrimitive` is overkill. `hasBlackBox` is more for detached primitives to make sure Clash picked them up. Let's drop `hasBlackBox`.

No blank line between definition and annotations.

You also forgot to remove the parentheses.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
```diff
@@ -0,0 +1,154 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# OPTIONS_HADDOCK hide #-}
+module Clash.Cores.Xilinx.Ethernet.Gmii.BlackBoxes where
```
DigitalBrains1, line 4:
```suggestion
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE QuasiQuotes #-}

{-# OPTIONS_HADDOCK hide #-}

module Clash.Cores.Xilinx.Ethernet.Gmii.BlackBoxes where
```

A bit more breathing room is pleasant on the eyes.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
```diff
@@ -0,0 +1,154 @@
...
+import Clash.Backend (Backend)
+import Clash.Netlist.Types (BlackBoxContext(..), TemplateFunction(..))
+import Control.Monad.State (State)
+import Data.List.Infinite (Infinite(..), (...))
+import Data.String.Interpolate
+import Data.Text.Prettyprint.Doc.Extra (Doc)
+import Prettyprinter.Interpolate (__di)
+
+primName :: String
+primName = "Clash.Cores.Xilinx.Ethernet.Gmii.gmiiSgmiiBridgePrim"
```
DigitalBrains1, line 18: Noooo.... don't hardcode this please!

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,160 @@
+-- | Contains the types used in the implementation of the GMII to SGMII bridge wrapper.
+{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}
+{-# LANGUAGE RecordWildCards #-}
+module Clash.Cores.Xilinx.Ethernet.Gmii.Types where
```
DigitalBrains1, line 4:
```suggestion
-- | Contains the types used in the implementation of the GMII to SGMII bridge wrapper.

{-# LANGUAGE RecordWildCards #-}

{-# OPTIONS_GHC -fconstraint-solver-iterations=10 #-}

module Clash.Cores.Xilinx.Ethernet.Gmii.Types where
```

DigitalBrains1 (changes requested):
First, it occurred to me you don't provide a VHDL primitive. If this isn't a simple oversight, it should be documented.

I was thinking about the module structure. `Clash.Cores.Xilinx.Ethernet.Gmii` is almost empty; it's just two definitions. If people want to do anything with the core other than just passing through inputs and outputs literally, they need to import `Clash.Cores.Xilinx.Ethernet.Gmii.Types`.

Maybe you are accomodating the module name for future additions of other cores that could be considered GMII?

I wonder if this even needs to be more than one module. We could have the `Clash.Cores.Xilinx.Ethernet.Gmii` module as follows:

* `gmiiSgmiiBridge`
* `Gmii`
* `DuplexMode`
* `LinkSpeed`
* `Pause`
* `AutoNegConfig`
* `Config`
* `Status`
* `BridgeOutput`

And then a section _Internal_ in the Haddock with the following:
* `gmiiSgmiiBridgePrim`
* `AutoNegConfigVector`
* `toAutoNegConfigVector` (if `pack` is insufficient)
* `ConfigVector`
* `toConfigVector` (if `pack` is insufficient)
* `StatusVector`
* `fromStatusVector` (if `unpack` is insufficient)
* `tclTemplateFunction`
* `tclTemplate`

In this configuration, you'd just put the `InlineYamlPrimitive` implementation directly inside the annotation as the GHC staging restriction prevents you from using a function defined in the same module.

Also, this is just in the order that it is currently in the Haddock; I'm not proposing an ordering here. Do that as you think makes sense to the reader.d suggest sorting them in a logical fashion. T

Alternatively, all the things that are marked _Internal_ above could go into a separate module `Clash.Cores.Xilinx.Ethernet.Gmii.Internal` that is hidden from Haddock (but still exposed in the library! We like giving everybody all functions to do with as they please, as long as they understand it's all internal.)

I'm also wondering a bit about the very generic name of the module. What if some day we want to add another thing that is GMII, but has a different definition for `Config`, `Status`, `LinkSpeed`, ...? Maybe the module name should be `GmiiSgmiiBridge`. It's a bit long, but it does enable adding other cores later without getting them all mixed up.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
DigitalBrains1: I'm glad you split that out because it was a bit more hidden first and I had missed it in the first review.

DigitalBrains1: With the current module layout you could do this:
```diff
--- a/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
@@ -148,4 +148,4 @@ gmiiSgmiiBridgePrim !_ !_  !_ !_ !_ !_ !_ !_ !_ !_ !_ !_ =
 
 {-# CLASH_OPAQUE gmiiSgmiiBridgePrim #-}
 {-# ANN gmiiSgmiiBridgePrim hasBlackBox #-}
-{-# ANN gmiiSgmiiBridgePrim (verilogPrim) #-}
+{-# ANN gmiiSgmiiBridgePrim (verilogPrim 'gmiiSgmiiBridgePrim) #-}
diff --git a/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs b/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
index a493157f0..2b812f897 100644
--- a/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
+++ b/clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/BlackBoxes.hs
@@ -12,13 +12,11 @@ import Control.Monad.State (State)
 import Data.List.Infinite (Infinite(..), (...))
 import Data.String.Interpolate
 import Data.Text.Prettyprint.Doc.Extra (Doc)
+import Language.Haskell.TH (Name)
 import Prettyprinter.Interpolate (__di)
 
-primName :: String
-primName = "Clash.Cores.Xilinx.Ethernet.Gmii.gmiiSgmiiBridgePrim"
-
-verilogPrim :: Primitive
-verilogPrim =
+verilogPrim :: Name -> Primitive
+verilogPrim primName =
   let
     tfName = 'tclTemplateFunction
 
@@ -151,4 +149,4 @@ tclTemplate bbCtx
       }
     |]
   in pure bbText
-  | otherwise = error $ primName <> ": Expected exactly one component name"
+  | otherwise = error "gmiiSgmiiBridgePrim: Expected exactly one component name"
```

If you put the primitive in the same module as the black box, you could do this:
```haskell
{-# ANN gmiiSgmiiBridgePrim (
  let
    primName = 'gmiiSgmiiBridgePrim
    tfName = 'tclTemplateFunction
    ...
  in InlineYamlPrimitive [Verilog, SystemVerilog] [__i|
    BlackBox:
      name: #{primName}
      ...
```

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,151 @@
...
+  , DomainResetPolarity async ~ 'ActiveHigh
+  , DomainResetPolarity gmii125 ~ 'ActiveHigh
+  ) =>
+  -- | Reference clock coming from the PHY.
+  DiffClock sgmii625 ->
+  -- | Asynchronous reset for the bridge.
+  Reset async ->
+  -- | Signal detect from the PHY. Either connect to the PHY's signal detect or use
+  -- a constant @True@, otherwise the link will never come up.
+  Signal async Bool ->
```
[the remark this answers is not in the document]
DigitalBrains1: Ah wait, no, that doesn't work for `Reset gmii125` because it already has a synchronous reset (output).

The primitive is rather like our clock generator functions: they get a clock input and a reset that is so asynchronous it isn't even synchronously deasserted, and they output a clock with a reset (that reset, in the case of the clock generators, is optionally synchronous).

The type signature for the clock generators makes sense here as well:

```haskell
DiffClock sgmii625 -> Reset sgmii625 -> ... -> ...{Clock gmii125, Reset gmii125, ...}
```

So both `refRst` and `signalDetect` should be `sgmii625`, IMO.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
```diff
@@ -0,0 +1,160 @@
...
+    _ -> errorX "fromStatusVector: invalid link speed"
+  sDuplexMode = if duplexModeBool then FullDuplex else HalfDuplex
+  sPause = case pauseBools of
+    (False, False) -> NoPause
+    (False, True) -> SymmetricPause
+    (True, False) -> AsymmetricPause
+    (True, True) -> SymmetricAsymmetricPause
+
+-- | A subset of the relevant output signals of the GMII to SGMII bridge
+data BridgeOutput gmiiRx125 sgmii625 = BridgeOutput
```
DigitalBrains1, line 154: `gmiiRx125` is called `gmii125` elsewhere, please rename it to `gmii125` here.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
DigitalBrains1: You really need 10 iterations? GHC 9.6.4 seems to need a mere godly 7.

I do realise we are engineers, but that doesn't mean always only using orders of magnitude :-D

lmbollen: I like to keep a bit more overhead, otherwise I might have to bump to 8 on the next change :)

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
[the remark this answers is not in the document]
lmbollen: I decided to just expose the `Acknowledge` and `Phy Link Status` fields

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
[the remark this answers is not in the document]
lmbollen: The reason why I made the `async` domain is because `refRst` and `signalDetect` are not associated with any domain according to the documentation (as indicated by `n/a` under `Clock Domain` in table 2 of DS264.

It's truly a shame we don't have a straightforward way to create truly asynchronous signals at the moment. 
I do feel like associating them with `sgmii625` gives off the false sense that they are synchronized to that domain.

Though if you feel strongly on this, I'm willing to annotate them that way. I also do see that a polymorphic `async` domain that falsely relates a reset and a signal also may cause problems.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
DigitalBrains1: It's not what I would do, but if that's how you want to do it I'm fine with it!

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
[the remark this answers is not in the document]
DigitalBrains1: Asynchronous signals on the output of a function are the truly insidious case; those functions probably all need to have the word _unsafe_ in their name. It is way too easy to mishandle those signals. Luckily, here, `signalDetect` is an input; the primitive doesn't care any way whatsoever. I suspect the `signalDetect` signal will be synchronised inside the primitive, and it wouldn't surprise me if it even didn't care about any glitches. Ethernet links don't go up and down that quickly. And it will probably just come from an FPGA pin connected to the PHY, without any user-defined circuitry in Clash.

So I think it's not super critical what domain you declare on `signalDetect`. If it were up to me, I'd declare it `sgmii625` with some documentation that the signal need not be synchronised. But you could keep the `async` domain.

For `refRst`, I really feel `Reset sgmii625` makes a lot of sense. We already declare it asynchronous (and active high). Clash doesn't differentiate between asynchronous and synchronous deassertion; in Clash, the usual behaviour of an asynchronous reset is that it asserts asynchronously but deasserts synchronously. Like for the clock generators, that's not the case here, and like the clock generators, the primitive's timing is determined by a clock input with a reset input that is completely asynchronous. Since we have the `HasAsynchronousReset`, it makes a lot of sense to put the domain `sgmii625` on that reset input. The output clock and reset are derived from the input clock and reset. They are pretty much grouped.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Types.hs
martijnbastiaan: On the 7 vs 10: the limit is just there to prevent the type checkers from going into an infinite loop. It doesn't actually cause more work. IMO the default should be such that users basically never see it.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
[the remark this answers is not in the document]
lmbollen: I'll change the domains to `sgmii625`

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,117 @@
...
+  DuplexMode (..),
+  Gmii (..),
+  LinkSpeed (..),
+  Lvds (..),
+  Nanoseconds,
+  Pause (..),
+  Picoseconds,
+  Status (..),
+  --- Gmii to Sgmii bridge
+  gmiiSgmiiBridge,
```
DigitalBrains1, line 29: I don't understand the structure of the Haddock; it seems primarily based on alphabetic order? I think this is quite an improvement:
```suggestion
  -- * GMII SGMII bridge function
  gmiiSgmiiBridge,
  -- * GMII interface
  Gmii (..),
  -- * GMII SGMII bridge function output
  BridgeOutput (..),
  -- ** Status fields
  Status (..),
  Pause (..),
  -- * PHY LVDS signals
  Lvds (..),
  -- * Bridge configuration
  Config (..),
  AutoNegConfig (..),
  DuplexMode (..),
  LinkSpeed (..),
```
Note that I no longer export `Nanoseconds` and `Picoseconds` because I dont' see why you'd do that and I didn't want to complicate the job of organising the Haddock even more by including stuff that's not going to be included anyway.

I also removed the two `---` comments as I don't know what they are, as I note in a different review comment.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,117 @@
...
+import Clash.Signal.Internal (DiffClock (..))
+
+-- | A subset of the relevant output signals of the GMII to SGMII bridge
+data BridgeOutput gmii125 sgmii625 = BridgeOutput
+  { bridgeLvdsOut :: Signal sgmii625 Lvds
+  , bridgeGmiiRx :: Signal gmii125 Gmii
+  , bridgeClk125 :: Clock gmii125
+  , bridgeRst125 :: Reset gmii125
+  , bridgeStatus :: Signal gmii125 Status
+  }
```
DigitalBrains1, line 45:
```suggestion
-- | A subset of the relevant output signals of the GMII to SGMII bridge
data BridgeOutput gmii125 sgmii625 = BridgeOutput
  { bridgeLvdsOut :: Signal sgmii625 Lvds
    -- ^ LVDS output to the PHY
  , bridgeGmiiRx :: Signal gmii125 Gmii
    -- ^ GMII output to the MAC
  , bridgeClk125 :: Clock gmii125
    -- ^ Clock output for the 125 MHz domain
  , bridgeRst125 :: Reset gmii125
    -- ^ Active high reset output for the 125 MHz domain
  , bridgeStatus :: Signal gmii125 Status
    -- ^ Status vector
  }
```

I just copied these descriptions from the `Internal` module; it seems like they are also useful to users of the public API.

DigitalBrains1 (changes requested):
I think this concludes my review, but mostly I just want to go home now :-).

DigitalBrains1 (changes requested):
Christiaans very first review comment was
> The module documentation header should be at the very top, before the LANGUAGE and OPTIONS_GHC pragmas

but I suddenly realise that the fix for that evidently did not survive the reorganisation of the files...

DigitalBrains1:
Also, you forgot to hide the `Internal` module from Haddock.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Internal.hs
```diff
@@ -0,0 +1,372 @@
...
+
+          |]
+  )
+  #-}
+
+tclTemplateFunction :: TemplateFunction
+tclTemplateFunction = TemplateFunction used valid tclTemplate
+ where
+  used = [10 .. 21] -- use all arguments
+  valid = const True
```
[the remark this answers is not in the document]
DigitalBrains1: The only way would be to move it to another module so you can satisfy the staging restriction for the `InlineYamlPrimitive` annotation.

Also, we're comparing it to hardcoding magic numbers, so it's already a huge win.

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
```diff
@@ -0,0 +1,117 @@
...
+{-# LANGUAGE ViewPatterns #-}
+
+{- |
+Copyright   :  (C) 2024, Google LLC
+License     :  BSD2 (see the file LICENSE)
+Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Contains a wrapper for the Xilinx GMII to SGMII PMA core :
+ "LogiCORE IP Ethernet 1000BASE-X PCS/PMA or SGMII"
+ https://www.xilinx.com/xsw/gig_ethernet_pcs_pma
```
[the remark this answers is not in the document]
DigitalBrains1: If you don't agree with my review suggestion, please discuss instead of hitting _Resolve conversation_.

DigitalBrains1 (commented):
You didn't apply all my suggestions. If you disagree with them, could you please discuss instead of simply presenting it again for review?

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii.hs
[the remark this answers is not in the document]
DigitalBrains1: Also, Haddock has mangled the URL.

DigitalBrains1 (approved):
LGTM!

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Internal.hs
```diff
@@ -0,0 +1,400 @@
...
+          set_property -dict [list \
+            CONFIG.LvdsRefClk {625} \
+            CONFIG.Standard {SGMII} \
+            CONFIG.Physical_Interface {LVDS} \
+            CONFIG.Management_Interface {false} \
+            CONFIG.Ext_Management_Interface {false} \
+            CONFIG.SGMII_PHY_Mode {false} \
+            CONFIG.SupportLevel {Include_Shared_Logic_in_Core}] [get_ips $ipName0]
+        return
+          }
```
DigitalBrains1, line 395: These backslashes need to be `\\\&` as I think they were in an earlier version. Currently the generated Tcl probably works but it looks horrendous:
```tcl
  proc createIp {ipName0 args} {
    create_ip -name gig_ethernet_pcs_pma -vendor xilinx.com -library ip             -version 16.2 -module_name $ipName0

    set_property -dict [list             CONFIG.LvdsRefClk {625}             CONFIG.Standard {SGMII}             CONFIG.Physical_Interface {LVDS}             CONFIG.Management_Interface {false}             CONFIG.Ext_Management_Interface {false}             CONFIG.SGMII_PHY_Mode {false}             CONFIG.SupportLevel {Include_Shared_Logic_in_Core}] [get_ips $ipName0]







  return
    }
```

DigitalBrains1 (dismissed):

clash-cores/src/Clash/Cores/Xilinx/Ethernet/Gmii/Internal.hs
```diff
@@ -0,0 +1,400 @@
...
+        }
```
DigitalBrains1, line 396: This indentation is messed up
```suggestion
          return
        }
      }
```

DigitalBrains1:
@lmbollen Could you open a PR with this change over on [clash-lang/clash-cores](https://github.com/clash-lang/clash-cores) so we can merge it there? This PR hit in an unfortunate time of limbo where this code still exists in `clash-compiler` but we already split off to the new repo before this PR.

lmbollen:
@DigitalBrains1 https://github.com/clash-lang/clash-cores/pull/11 :)
I simply added `clash-compiler` as remote locally and cherry-picked https://github.com/clash-lang/clash-compiler/commit/20c17495e821faa2b8969db5cb3b3ebdc322b7f7

Which is the only commit that touches the code.

PR #2715, opened 2024-05-05
Bump `lens` dependency bounds

martijnbastiaan:


leonschoorl:
Those macOS failures are being tracked at https://github.com/haskell-actions/setup/issues/77

martijnbastiaan:
Should I just disable those macOS versions for now?

DigitalBrains1:
It says
> Known workarounds
> Use `macOS-13` instead of `macOS-latest`.

So I think we should try that...

PR #2720, opened 2024-05-17
Adjust `BiSignalDefault` doc to clarify affects simulation only

lmbollen:
No logic is added here, it only represents the state of the wire when no-one is driving it.

### DIFF
--- a/clash-prelude/src/Clash/Signal/BiSignal.hs
+++ b/clash-prelude/src/Clash/Signal/BiSignal.hs
@@ -136,8 +136,8 @@ import           GHC.TypeLits               (KnownNat, Nat)
 import           GHC.Stack                  (HasCallStack)
 import           Data.Reflection            (Given (..))
 
--- | Used to specify the /default/ behavior of a \"BiSignal\", i.e. what value is
--- read when no value is being written to it.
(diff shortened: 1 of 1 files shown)


clash-prelude/src/Clash/Signal/BiSignal.hs
```diff
@@ -133,8 +133,8 @@ import           GHC.TypeLits               (KnownNat, Nat)
 import           GHC.Stack                  (HasCallStack)
 import           Data.Reflection            (Given (..))
 
--- | Used to specify the /default/ behavior of a \"BiSignal\", i.e. what value is
--- read when no value is being written to it.
+-- | Used to specify the /default/ behavior of a \"BiSignal\" in simulation, i.e.
+-- what value is read when no value is being written to it.
```
martijnbastiaan, line 137: It only applies to Haskell simulation (at the moment :eyes:).

martijnbastiaan (changes requested):

martijnbastiaan:
That's correct @DigitalBrains1. I do think this PR is a strict improvement, so @lmbollen could you please apply my suggestion, then we can merge IMO.

martijnbastiaan (approved):

martijnbastiaan:
@DigitalBrains1 I've tracked your comment in https://github.com/clash-lang/clash-compiler/issues/2798

PR #2731, opened 2024-05-31
Fix BitVector shifts

### LINKED ISSUE #2730: Shifting bits to 0-length `BitVector`s fails
Shifting a bit into a 0-length BitVector with `(+>>.)` or `(.<<+)` currently fails with an out of range exception.

We should either define the operations to keep empty bit vectors unchanged or exclude them from the interface via adding some `1 <= n` constraints.

The former would be the way to go in order to stay consistent with `(+>>)` and `(<<+)`:
```haskell
clashi> high +>> (Nil :: Vec 0 Bit)
Nil
clashi> (Nil :: Vec 0 Bit) <<+ low
Nil
```

kleinreact:
Fixes #2730 such that `(+>>.)` and `(.<<+)` are compliant with `(+>>)` and `(<<+)` for bit vectors of zero length, respectively.

## Still TODO:

### DIFF
--- a/changelog/2024-05-31T20_14_29+02_00_fix_bitvector_shifts
+++ b/changelog/2024-05-31T20_14_29+02_00_fix_bitvector_shifts
@@ -1,0 +1,1 @@
+FIXED: (+>>.) and (.<<+) such that they are compliant with (+>>) and (<<+) for vectors of zero length in the sense that the input vector is kept unchanged.
--- a/clash-prelude/src/Clash/Sized/BitVector.hs
+++ b/clash-prelude/src/Clash/Sized/BitVector.hs
@@ -1,11 +1,12 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente
-                  2022     , Google Inc.
+                  2022-2024, Google Inc.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 -}
 
 {-# LANGUAGE Trustworthy #-}
+{-# LANGUAGE TypeFamilies #-}
 
 {-# OPTIONS_HADDOCK show-extensions #-}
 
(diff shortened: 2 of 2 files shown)


DigitalBrains1:
Thanks!

We usually use [`Clash.Promoted.Nat.compareSNat`](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Promoted-Nat.html#v:compareSNat) for this. I don't know if there are any downsides to picking `GHC.TypeNats.cmpNat`, though. Both apparently provide proof for the comparison result, but the proof wasn't needed here in the first place since the old code type-checked.

That raises a separate point: perhaps we need to look at `Bits (BitVector n)` or `Enum (BitVector n)` if apparently one of those has some partial functions? We might need a `1 <= n` on one of those. So it won't type-check anymore.

And since this is a fix that pertains to a released version of Clash, it should have a changelog entry.

DigitalBrains1:
Ah no, wait, it _was_ in the bit shifters themselves. They call
```
>>> replaceBit# (0 :: BitVector 0) (-1) 1
*** Exception: replaceBit: -1 is out of range [-1..0]
  error, called at src/Clash/Sized/Internal/BitVector.hs:1143:12 in clash-prelude-1.8.1-72gLWAPSXgPDozrrxr9IhV:Clash.Sized.Internal.BitVector
```
for `n = 0`. It's a pity that that error message is a bit garbled. We should fix that as well, beause it's just not very nice.

changelog/2024-05-31T20_14_29+02_00_fix_bitvector_shifts
```diff
@@ -0,0 +1 @@
+Fixes (+>>.) and (.<<+) such that they are compliant with (+>>) and (<<+) for vectors of zero length.
```
DigitalBrains1, line 1: Could you please describe what they do for vectors of zero length? It's fine to note that they do the same as their vector counterparts, but the natural question that would boil up is: so what do they do? :-)

DigitalBrains1 (approved):
LGTM!

DigitalBrains1:
Ah ideally it'd have `FIXED:` as the first word of the changelog; this helps the person writing the actual changelog for a release in quickly putting it under the right heading. And we'd like to automate a part of this one day; now it's still all manual labor.

PR #2734, opened 2024-06-18
Add utilities to represent time

lmbollen:
This pull requests adds a new module that offers a user friendly API to reason about time.
Clash stores periods in picoseconds, which is a detail users should not have to know to reason about time in their system.

For example:
If one would like to have a free running counter that counts up to 1 second, you could know use the type:
```hs
Index (PeriodToCycles dom (Second 1))
```
Which works even if you change the clock frequency.

Open questions:
* Are there more useful types that we're missing?
* Is the module naming and re-export correct?
## Still TODO:


clash-prelude/src/Clash/Signal/Time.hs
```diff
@@ -0,0 +1,40 @@
...
+-}
+module Clash.Signal.Time
+ ( DomainFrequency
+ , Microseconds
+ , Milliseconds
+ , Nanoseconds
+ , PeriodToCycles
+ , Picoseconds
+ , Seconds
+ ) where
```
DigitalBrains1, line 17: Same thing really, this order doesn't really make sense for the readers of our Haddock.

changelog/2024-06-18T11_16_48+02_00_add_time_types
```diff
@@ -0,0 +1 @@
+ADDED: Introduced a new module to more easily represent time and its relation to your clock domain.
```
DigitalBrains1, line 1: Please name the module you added.

clash-prelude/src/Clash/Signal/Time.hs
```diff
@@ -0,0 +1,40 @@
+{-|
+Copyright  :  (C) 2024 QBayLogic B.V.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+User friendly utilities to express time in your digital design.
+
+-}
+module Clash.Signal.Time
```
DigitalBrains1, line 9: Do we really need a whole module for this? This makes the table of contents of the Haddock larger, so there's a trade-off. I feel they can go in `Clash.Signal.Internal`.

DigitalBrains1 (changes requested):
It's definitely nice to have these. I do feel a bit of déjà vu, but I couldn't find an open PR or issue.

DigitalBrains1 (changes requested):
The déjà vu is of course from Gergő's book. And that immediately brings me to another point. Let's please name things the same as Gergő does! It's here: [hackage](https://hackage.haskell.org/package/retroclash-lib-0.1.2.2/docs/RetroClash-Clock.html) [source](https://github.com/gergoerdi/retroclash-lib/blob/master/src/RetroClash/Clock.hs).

changelog/2024-06-18T11_16_48+02_00_add_time_types
DigitalBrains1: Or if you move them to an existing module, name the types.

clash-prelude/src/Clash/Signal.hs
```diff
@@ -97,7 +97,6 @@ module Clash.Signal
   , ResetKind(..)
   , SResetKind(..)
   , ResetPolarity(..)
-  , SResetPolarity(..)
```
DigitalBrains1, line 100: I completely missed this. I think it was a mistake to remove this export? If not, it should be documented.

DigitalBrains1:
> Let's please name things the same as Gergő does!

Ah, I mistakenly thought `PeriodToCycles` and `DomainFrequency` corresponded to types defined by Gergő. This is not the case, but what about adding those definitions that Gergő defined?

gergoerdi:
I'm all for it -- `retroclash-lib` is MIT so please do feel free to plunder it!

lmbollen:
I added @gergoerdi to the copyright, but left the code as BSD2 in line with the rest of `clash-prelude`

clash-prelude/src/Clash/Signal.hs
```diff
@@ -107,6 +106,13 @@ module Clash.Signal
   , DomainInitBehavior
   , DomainResetPolarity
     -- *** Convenience types
+  , DomainFrequency
+  , Microseconds
+  , Milliseconds
+  , Nanoseconds
+  , PeriodToCycles
+  , Picoseconds
+  , Seconds
```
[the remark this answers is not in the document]
lmbollen: I simply added two new sections

changelog/2024-06-18T11_16_48+02_00_add_time_types
lmbollen: All types?

clash-prelude/src/Clash/Signal/Time.hs
lmbollen: Moved them :)

clash-prelude/src/Clash/Signal.hs
lmbollen: Yup, thanks!

changelog/2024-06-18T11_16_48+02_00_add_time_types
DigitalBrains1: You can also link to the relevant section of the Haddock. Or just say "`Seconds` and similar types", phrased a bit better. It's just that now you say "we added something to more easily represent time". But if I'm a Clash user and I'm reading the changelog, how am I going to find these new toys? There's no clue where I should look. If you mention `Seconds`, they'd probably find the rest.

DigitalBrains1: Ah now you say "added to `Clash.Signal.Internal`" but that's not really helpful either. First off, we shouldn't be directing users to plow trough internal modules. Secondly, that module is fairly large, so it's still a hunt.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -439,6 +451,33 @@ type HasDefinedInitialValues (dom :: Domain) =
...
+-- | Gets time in 'Picoseconds' from time in 'Milliseconds'.
+type Milliseconds (ms :: Nat) = Microseconds (1000 * ms)
+-- | Gets time in 'Picoseconds' from time in 'Microseconds'.
+type Microseconds (us :: Nat) = Nanoseconds  (1000 * us)
+-- | Gets time in 'Picoseconds' from time in 'Nanoseconds'.
+type Nanoseconds  (ns :: Nat) = Picoseconds  (1000 * ns)
+-- | Gets time in 'Picoseconds' from time in 'Picoseconds', essentially 'id'.
+type Picoseconds  (ps :: Nat) = ps
+
+-- | Converts a frequency in Hz to a period in picoseconds. This might lead to rounding
```
DigitalBrains1, line 466:
```suggestion
-- | Converts a frequency in hertz to a period in picoseconds. This might lead to rounding
```

It just looks better if you write out picoseconds...

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -439,6 +451,33 @@ type HasDefinedInitialValues (dom :: Domain) =
...
+-- errors.
+type HzToPeriod hz = Seconds 1 `Div` hz
+
+-- | The domain's clock frequency in Hertz, calculated based on the period stored in ps.
+-- This might lead to rounding errors.
```
DigitalBrains1, line 471:
```suggestion
-- | The domain's clock frequency in hertz, calculated based on the period
-- stored in picoseconds. This might lead to rounding errors.
```

SI says: if you write out unit names, don't capitalise.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -67,12 +67,22 @@ module Clash.Signal.Internal
...
+  , Seconds
+  , Milliseconds
+  , Microseconds
+  , Nanoseconds
+  , Picoseconds
+  -- **** Time conversions
+  , DomainToHz
+  , HzToPeriod
+  , PeriodToCycles
+  , PeriodToHz
```
DigitalBrains1, line 85:
```suggestion
  , HzToPeriod
  , PeriodToHz
  , PeriodToCycles
```

clash-prelude/src/Clash/Signal.hs
```diff
@@ -113,6 +113,17 @@ module Clash.Signal
...
+ , Milliseconds
+ , Microseconds
+ , Nanoseconds
+ , Picoseconds
+ -- **** Time conversions
+ , DomainToHz
+ , HzToPeriod
+ , PeriodToCycles
+ , PeriodToHz
```
DigitalBrains1, line 126:
```suggestion
 , HzToPeriod
 , PeriodToHz
 , PeriodToCycles
```

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -439,6 +451,33 @@ type HasDefinedInitialValues (dom :: Domain) =
...
+-- | Gets time in 'Picoseconds' from time in 'Microseconds'.
+type Microseconds (us :: Nat) = Nanoseconds  (1000 * us)
+-- | Gets time in 'Picoseconds' from time in 'Nanoseconds'.
+type Nanoseconds  (ns :: Nat) = Picoseconds  (1000 * ns)
+-- | Gets time in 'Picoseconds' from time in 'Picoseconds', essentially 'id'.
+type Picoseconds  (ps :: Nat) = ps
+
+-- | Converts a frequency in Hz to a period in picoseconds. This might lead to rounding
+-- errors.
+type HzToPeriod hz = Seconds 1 `Div` hz
```
DigitalBrains1, line 468:
```suggestion
type HzToPeriod (hz :: Nat) = Seconds 1 `Div` hz
```

We're not really completely consistent with this, but we usually do do it. As you do with most tyargs here as well.

And it "leaks" to the Haddock, it's directly user visible.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -439,6 +451,33 @@ type HasDefinedInitialValues (dom :: Domain) =
...
+
+-- | The domain's clock frequency in Hertz, calculated based on the period stored in ps.
+-- This might lead to rounding errors.
+type DomainToHz dom = PeriodToHz (DomainPeriod dom)
```
DigitalBrains1, line 472:
```suggestion
type DomainToHz (dom :: Domain) = PeriodToHz (DomainPeriod dom)
```

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -439,6 +451,33 @@ type HasDefinedInitialValues (dom :: Domain) =
...
+
+-- | Number of clock cycles required at the clock frequency of @dom@ before a minimum
+-- @period@ has passed. Is always at least one.
+type PeriodToCycles dom period =  period `DivRU` DomainPeriod dom
```
DigitalBrains1, line 476:
```suggestion
type PeriodToCycles (dom :: Domain) (period :: Nat) =  period `DivRU` DomainPeriod dom
```

DigitalBrains1 (changes requested):
I think the two commits should be squashed to one, because the first commit doesn't compile:
```
src/Clash/Cores/UART.hs:77:33: error:
    Ambiguous occurrence ‘PeriodToHz’
    It could refer to
       either ‘Clash.Prelude.PeriodToHz’,
              imported from ‘Clash.Prelude’ at src/Clash/Cores/UART.hs:35:1-20
              (and originally defined in ‘Clash.Signal.Internal’)
           or ‘Clash.Cores.UART.PeriodToHz’,
              defined at src/Clash/Cores/UART.hs:46:1
   |
77 |               ':<>: ('ShowType (PeriodToHz (DomainPeriod dom)))
   |                                 ^^^^^^^^^^
```
etcetera.

For most of the following change requests, I could have already requested them in the previous round. It is apparently rather hard for me to actually pick everything out on the first go...

DigitalBrains1:
> Personally I found `ClockDivider` a bit unclear, `PeriodToCycles` is more consistent with the other types. If people prefer it, I'm okay with changing it though.

I think `PeriodToCycles` is more clear, but there's another concern here. If we use Gergő's name, `retroclash-lib` can drop the definition and people who learn Clash from Gergő's book can readily apply what they learned to their designs. Personally, I think that concern trumps the clarity argument, and I'd rather have `ClockDivider`.

clash-prelude/src/Clash/Signal/Internal.hs
DigitalBrains1: Also, why not simply mirror the description of `HzToPeriod`? Why are the texts so different? I like consistency.

DigitalBrains1:
> I added @gergoerdi to the copyright

Did you, though?

lmbollen:
> I just pushed some improvements. First of all, I fixed the formatting of my _Simplifying_, as a separate commit. But I also include a `fixup` commit to be squashed into the other one if you agree with my tweaks.
> 
> My `fixup` commit also includes the fix you just pushed, minutes before I pushed.
> 
> If you agree with my `fixup` commit, either of us can squash it. And then either merge or wait for the other reviews you requested; your choice.
> 
> I thought your changelog entry would take up too much vertical space in our changelog, so I put it on one line. Note that isn't the only change to the changelog; I also changed the reference to an internal module to `Clash.Signal` as I [mentioned in an earlier change request](https://github.com/clash-lang/clash-compiler/pull/2734#discussion_r1647247907).

I agree with your fixup, I'll squash them into my commit.

DigitalBrains1:
Ah mannnn.... the changelog should also mention `ClockDivider` as a service to people coming from the RetroClash book.

martijnbastiaan (approved):
Nice review work @DigitalBrains1 :D. LGTM. (But let's wait for @DigitalBrains1 too.)

DigitalBrains1 (approved):
We first need to appease the CI gods by no longer being ambiguous, though.

DigitalBrains1:
The initial version of this PR defined:
```haskell
type PeriodToCycles dom period = Max 1 (DivRU period (Max 1 (DomainPeriod dom)))
```

but later you changed it to:
```haskell
type PeriodToCycles (dom :: Domain) (period :: Nat) =  period `DivRU` DomainPeriod dom
```

Now, the outer `Max 1` might be unneeded... I think it would only catch the case that someone passed in 0 as a period, which is perhaps not the concern to deal with here. If someone wants a period 0, let them have it.

But the inner `Max 1`, at the divisor, seems to be really useful. Because anytime you try to really do anything with this type, you're going to need a `1 <= DomainPeriod dom` for the solver to be happy with it in the position of the divisor:

```haskell
f ::
  KnownDomain dom =>
  Proxy dom ->
f Proxy = SNat
```
gives
```
[...]: error: [GHC-64725]
    • Cannot satisfy: 1 <= Clash.Signal.Internal.DomainConfigurationPeriod
                             (KnownConf dom)
    • In the expression: SNat
      In an equation for ‘f’: f Proxy = SNat
```

So I think we should at least make sure that the divisor is strictly positive. Either with `Max 1` in the divisor, or what would be even better, fundamentally in the definition of a domain. A domain with period 0 is an invalid domain, so might as well catch this real early. I just couldn't get it to work. I tried to use GADT-syntax for the definition of `DomainConfiguration` but my best wasn't good enough for GHC.

DigitalBrains1:
Although the outer `Max 1` might also be nice... because if you need the result to be strictly positive, you're a bit hindered in providing proof. Currently, the solver does not understand that `1 <= n, DivRU n (Max 1 m)` is enough proof. Of course, it would be good if we could teach the solver this, but that's a different story.

This is not enough:
```haskell
type PeriodToCycles0 dom period = period `DivRU` Max 1 (DomainPeriod dom)

g ::
  HiddenClockResetEnable dom =>
  1 <= period =>
  Proxy dom ->
  SNat period ->
  Signal dom (Index (PeriodToCycles0 dom period))
g Proxy SNat = let o = register maxBound (satPred SatBound <$> o) in o
```
(note that we need `instance (KnownNat n, 1 <= n) => SaturatingNum (Index n)`)
Clash complains:
```
[...]: error: [GHC-64725]
    • Cannot satisfy: 1 <= Div
                             (period
                              + (Max
                                   1
                                   (Clash.Signal.Internal.DomainConfigurationPeriod (KnownConf dom))
                                 - 1))
                             (Max
                                1
                                (Clash.Signal.Internal.DomainConfigurationPeriod (KnownConf dom)))
    • In the first argument of ‘(<$>)’, namely ‘satPred SatBound’
      In the second argument of ‘register’, namely
        ‘(satPred SatBound <$> o)’
      In the expression: register maxBound (satPred SatBound <$> o)
```

but if we provide:
```haskell
type PeriodToCycles0 dom period = Max 1 (period `DivRU` Max 1 (DomainPeriod dom))

g ::
  HiddenClockResetEnable dom =>
  Proxy dom ->
  SNat period ->
  Signal dom (Index (PeriodToCycles0 dom period))
g Proxy SNat = let o = register maxBound (satPred SatBound <$> o) in o
```
then Clash is happy.

So I say, let's add them both for comfort.

lmbollen:
@DigitalBrains1 I'll make a PR and tag you :)

TristanCacqueray:
Hello, it looks like the new `ClockDivider` using `DivRU` breaks the serialRX implementation from retroclash-lib [RetroClash.SerialRX](https://github.com/gergoerdi/retroclash-lib/blob/master/src/RetroClash/SerialRx.hs#L63). In https://github.com/TristanCacqueray/retroclash-lib/commit/be5ffa2262d01fa703d5e7c9ef2c8f520ca0eeb0 I'm updating retroclash to the latest clash-prelude-1.10, and that mostly works fine, but when sending more than a few bytes through the UART, the serialRX starts producing garbage (using a 9MHz clock at 9600 or 57600 baud). 

Not sure if this is expected and how to update the bitDuration in retroclash, but presently, here is the fix needed to make RetroClash UART code work properly:

```haskell
module RetroClash.SerialRx where

import Clash.Prelude hiding (ClockDivider)

type ClockDivider dom ps = ps `Div` DomainPeriod dom
```

rowanG077:
clash defines `ClockDivider` as :

```
ps `DivRU` DomainPeriod dom
```

So that could explain the mismatch. If anything I would say this UART has more problems. So perhaps it would be better to move towards the [Clash cores](https://github.com/clash-lang/clash-cores/blob/main/src/Clash/Cores/Uart.hs) one. That one is a lot more robust.

TristanCacqueray:
I did try to use clash-cores in https://codeberg.org/TristanCacqueray/clash-serial-echo , but couldn't get it to work for my serial over USB use-case.

rowanG077:
You are misusing both uarts. You need to keep the byte to transmit stable until acked/ready is set.
So you want something like:

```Haskell
(ibyte, tx, ack) = Clash.Cores.UART.uart uartSpeed rx buffered
buffered = register
  Nothing
  (mux (ack .||. (isNothing <$> buffered)) ibyte buffered)
```

This keeps the output stable until acked but drops incoming bytes if the current byte as not yet been acked. In practice people usually put a FIFO on both directions of a UART so bursts can be absorbed.

TristanCacqueray:
@rowanG077 thank you very much for having a look and fixing my code! I thought that because the RX and TX speed matched I could discard the ack, but I now see that the clash-cores uartTx only advance when the baud generator is enabled.

PR #2738, opened 2024-06-21
Add 8b/10b encoder and decoder to `clash-cores`

jvnknvlgl:

Add 8b/10b encoding and decoding support to `clash-cores`. This is based on a lookup table which is adapted from the lookup table from [LibSV](https://github.com/bensampson5/libsv).

## Still TODO:


clash-cores/src/Clash/Cores/EbTb.hs
```diff
@@ -0,0 +1,117 @@
...
+  deriving (Generic, NFDataX, Eq, Show)
+
+-- | Function to check whether a 'DataWord' results in a data word
+isDw :: DataWord -> Bool
+isDw (Dw _) = True
+isDw _ = False
+
+-- | Function to check whether a 'DataWord' results in a data word or
+--   control word
+isValidDw :: DataWord -> Bool
```
leonschoorl, line 31: I feel this needs a better name.
I read it as "is valid dataword", ie when this returns `True`, I'd expect is to be valid and be a data word.

clash-cores/src/Clash/Cores/EbTb.hs
```diff
@@ -0,0 +1,117 @@
...
+--   8b/10b encoding and decoding functions
+module Clash.Cores.EbTb where
+
+import qualified Clash.Cores.EbTb.Decoder as Dec
+import qualified Clash.Cores.EbTb.Encoder as Enc
+import Clash.Prelude
+
+-- | Data type that contains a 'BitVector 8' with the corresponding error
+--   condition of the decode function
+data DataWord
```
leonschoorl, line 17: The datatype could probably do with a better name, to avoid confusion between the type DataWord and it being a "data word" (vs a control word).

clash-cores/src/Clash/Cores/EbTb.hs
```diff
@@ -0,0 +1,117 @@
...
+  = Dw (BitVector 8)
+  | Cw (BitVector 8)
+  | DwError (BitVector 8)
+  | RdError (BitVector 8)
```
leonschoorl, line 21: I guess `RdError` is (Running) Disparity error, but what's DwError?

clash-cores/src/Clash/Cores/EbTb/Decoder.hs
```diff
@@ -0,0 +1,2070 @@
...
+--                      2024     , QBayLogic B.V.
+--   License     :  BSD2 (see the file LICENSE)
+--   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+--
+--   8b/10b decoding look-up table
+module Clash.Cores.EbTb.Decoder where
+
+import Clash.Prelude
+
+-- https://github.com/bensampson5/libsv/blob/main/libsv/coders/decoder_8b10b.sv
```
leonschoorl, line 12: Please link to the commit hash that you used instead of `main`.
That way the link won't break when libsv changes things.
And if they change things we can diff things to see what changed and if we should copy any of those changes.

clash-cores/src/Clash/Cores/EbTb.hs
jvnknvlgl: I mostly want to have a distinction between "is something valid (so none of the error types)" and "is something a data word", but I would say naming it `isValid` is probably _too_ generic. Do you have any suggestions?

clash-cores/src/Clash/Cores/EbTb.hs
jvnknvlgl: `DwError` is when a 10b code group does not correspond to any data word, which is possible. Is there a way to properly document this?

clash-cores/src/Clash/Cores/EbTb/Decoder.hs
jvnknvlgl: I wanted to remove this, as 8b/10b is pretty much fixed anyway. I just copied this so I didn't have to write or generate the whole table myself. I did add the original author to the copyright notice.

clash-cores/src/Clash/Cores/EbTb.hs
jvnknvlgl: By the way, the reason these still contain `BitVector`s is because @christiaanb suggested to do the decoding on a best-effort basis, to still be able to see what happens when a `DataWord` is not correct.

rowanG077:
For an algorithm to calculate these values look at litex 8b10b encoder/decoder: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/code_8b10b.py

I have no clue whether or even how it works but I do know it's used in liteeth for sgmii. It seems to use some smaller tables(??) then this implementation. I'm not suggesting to rewrite it like liteeth. Just a datapoint if someone find this back latter wondering if maybe we could make the table a bit more clear.

DigitalBrains1:
Since 8b10b is the concatenation of a 5b/6b and a 3b/4b encoding, it should be possible to do it with smaller tables. Unless of course there is a snag I am unaware of (such as the possibility of multiple concrete encodings being referred to as 8b10b and me talking about the wrong one).

jvnknvlgl:
> It seems to use some smaller tables(?? maybe not, it just looks like that from the code) then this implementation.

It splits the 8b/10b encoder into a 3b/4b encoder and a 5b/6b encoder (8b/10b is essentially a concatenation of these two schemes), which makes for the smaller hardcoded tables. Combining this leads to a table with 8 * 32 = 256 entries. The table I use adds the new running disparity and an indicator for invalid words to that, which gives 10 bits of data (and thus 1024 entries).

clash-cores/src/Clash/Cores/EbTb.hs
DigitalBrains1: These 10-bit quantities are often referred to as _symbols_, which seems like a good alternative to _word_.

jvnknvlgl: I have now renamed `DataWord` to `Symbol8b10b`. Does this sound better?

clash-cores/src/Clash/Cores/EbTb.hs
jvnknvlgl: I have added a few lines to explain these different constructors.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -88,22 +88,16 @@ encode8b10b ::
...
-    _ -> 0
+  cw = isValidSymbol dw && not (isDw dw)
 
   (statusBits, cg) =
     splitAt d2
       $ bv2v
       $ asyncRomBlobPow2
         $(memBlobTH Nothing Enc.encoderLut)
-      $ unpack (cw ++# bitCoerce rd ++# fromDw dw)
+      $ unpack (bitCoerce cw ++# bitCoerce rd ++# fromDw dw)
```
DigitalBrains1, line 98:
```suggestion
      $ unpack (pack cw ++# pack rd ++# pack dw)
```
This is more idiomatic. Another trick would be:
```
$ bitCoerce (cw, rd, dw)
```
using the fact that `++#` and tuple elements have an identical effect on what the `BitVector` looks like.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,103 @@
...
+    splitAt d4
+      $ bv2v
+      $ asyncRomBlobPow2
+        $(memBlobTH Nothing Dec.decoderLut)
+      $ unpack (bitCoerce rd ++# cg)
+
+  rdEr = bitCoerce $ head statusBits
+  cgEr = bitCoerce $ statusBits !! (1 :: Index 4)
+  cw = bitCoerce $ statusBits !! (2 :: Index 4)
+  rdNew = bitCoerce $ last statusBits
```
DigitalBrains1, line 74:
```suggestion
decode8b10b rd cg = (rdNew, sym)
 where
  sym
    | cgEr = DwError w
    | rdEr = RdError w
    | cw = Cw w
    | otherwise = Dw w

  (rdEr, cgEr, cw, rdNew, w) =
    unpack
      $ asyncRomBlobPow2
        $(memBlobTH Nothing Dec.decoderLut)
      $ unpack (pack rd ++# cg)
```

So `MemBlob` loses type information; the only thing it remembers about the type of the data elements is their bit size. But you would usually just `unpack` it back from the data output of the ROM. In this case, since the data inside the `MemBlob` has the type `(Bit, Bit, Bit, Bit, BitVector 8)`, you would usually `unpack` it back into that type. However, I presume that  `Bit` is used here only for the ease of the copy-paste from the source. They are in fact all four `Bool`s, and from a Haskell standpoint, it would have made more sense to write the table with the type `(Bool, Bool, Bool, Bool, BitVector 8)`. So that's what we'll `unpack` it to here. We don't even need to specify that type; it follows from how the resulting bindings are used.

(From a Haskell standpoint, a direct encoding as `(Bool, Symbol8b10b)` would have made even more sense, but  that is even further from the copy-paste).

And on a totally different note, I think your use of the term _code group_ is wrong here. I'm pretty sure that a 10-bit entity is not a code group, but that it consists of two code groups, the first being 5b6b and the second being 3b4b.

We already use the word _symbol_ for things of type `Symbol8b10b`, so let's not call the 10-bit quantity that as well. Let's call it a _code_. I quickly scoured some standards and it seems a not uncommon term. Likewise, I think that the 5b6b section is a _code_ and the 3b4b section is a _code_, and that those together form two _code groups_. So the "group" isn't inherent to the parts, it's because they are grouped together in the 10-bit quantity.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,103 @@
...
+
+  (statusBits, cg) =
+    splitAt d2
+      $ bv2v
+      $ asyncRomBlobPow2
+        $(memBlobTH Nothing Enc.encoderLut)
+      $ unpack (pack cw ++# pack rd ++# fromDw dw)
+
+  rdNew = bitCoerce $ last statusBits
+  out = if isValidSymbol dw then (rdNew, pack cg) else (rd, 0)
```
DigitalBrains1, line 101:
```suggestion
encode8b10b rd sym = out
 where
  (_er, rdNew, cg) =
    unpack @(Bool, Bool, BitVector 10)
      $ asyncRomBlobPow2
        $(memBlobTH Nothing Enc.encoderLut)
      $ unpack (pack (isCw sym) ++# pack rd ++# fromDw sym)

  out = if isValidSymbol sym then (rdNew, cg) else (rd, 0)
```

Here we needed some form of type signature so it knows what the type of `_er`  is. It's not going to be clever enough to just infer we're only interested in the top 11 bits of the `BitVector` and the type of `_er` doesn't matter :-).

I thought this solution looked nicer than
```haskell
  (_er :: Bool, rdNew, cg) =
    unpack
```
but that would have worked as well.

clash-cores/test/Test/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,115 @@
...
+-- 'encode8b10b'.
+prop_decode8b10bCheckNothing :: H.Property
+prop_decode8b10bCheckNothing = H.withTests 1000 $ H.property $ do
+  inp <- H.forAll genDefinedBitVector
+  let out = isValidSymbol dw1 && isValidSymbol dw2
+       where
+        (_, dw1) = decode8b10b False $ snd $ encode8b10b False (Dw inp)
+        (_, dw2) = decode8b10b True $ snd $ encode8b10b True (Dw inp)
+
+  H.assert out
```
DigitalBrains1, line 64:
```suggestion
  H.assert $ isValidSymbol $ snd $ decode8b10b False $ snd $ encode8b10b False (Dw inp)
  H.assert $ isValidSymbol $ snd $ decode8b10b True $ snd $ encode8b10b True (Dw inp)
```

The nice thing about this is the formatting of a counter-example. Suppose I corrupt the test by deliberately mismatching the running disparity, Tasty will tell me:

```
peter@gronau:~/clash/clash-cores$ cabal run unittests -- -p LineC -j12
Unittests
  Test.Cores.LineCoding8b10b
    decode8b10bCheckNothing: FAIL (0.03s)
        ✗ <interactive> failed at test/Test/Cores/LineCoding8b10b.hs:59:3
          after 1 test and 2 shrinks.
          shrink path: 1:a2
        
             ┏━━ test/Test/Cores/LineCoding8b10b.hs ━━━
          56 ┃ prop_decode8b10bCheckNothing :: H.Property
          57 ┃ prop_decode8b10bCheckNothing = H.withTests 1000 $ H.property $ do
          58 ┃   inp <- H.forAll genDefinedBitVector
             ┃   │ 0 b0000_0000
          59 ┃   H.assert $ isValidSymbol $ snd $ decode8b10b True $ snd $ encode8b10b False (Dw inp)
             ┃   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          60 ┃   H.assert $ isValidSymbol $ snd $ decode8b10b True $ snd $ encode8b10b True (Dw inp)
        
          This failure can be reproduced by running:
          > recheckAt (Seed 17468430201590260638 5530981405269087381) "1:a2" <property>
        
      Use "--pattern '$NF ~ /decode8b10bCheckNothing/' --hedgehog-replay '1:a2 Seed 17468430201590260638 5530981405269087381'" to reproduce from the command-line.
      
      Use -p '/LineC/&&/decode8b10bCheckNothing/' to rerun this test only.
```

It pinpoints the exact test that failed.

clash-cores/test/Test/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,115 @@
...
+
+-- | Encode the input signal and check whether it is a valid value. It should be
+--   valid for every possible input.
+prop_encode8b10bCheckNothing :: H.Property
+prop_encode8b10bCheckNothing = H.withTests 1000 $ H.property $ do
+  inp <- H.forAll genDefinedBitVector
+  let out = 0 /= snd (encode8b10b False (Dw inp))
+
+  H.assert out
```
DigitalBrains1, line 73:
```suggestion
  snd (encode8b10b False (Dw inp)) H./== 0
```

I don't think the doc for this function really matches what it does, though. It just checks that it never produces a 10-bit zero. There's a whole bunch of other invalid codes that are not checked.

clash-cores/test/Test/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,115 @@
...
+       where
+        o = g False inp
+
+        g rd i = if isValidSymbol dw then snd $ encode8b10b rd dw else i
+         where
+          dw = snd $ decode8b10b rd i
+
+      expected = inp
+
+  out H.=== expected
```
DigitalBrains1, line 112: I thought this test was flawed. If `encode8b18b False dw` did _not_ result in a valid data word, we don't try it with `encode8b10b True dw`, we just immediately return success. So I rewrote it to:
```suggestion
  let out = case (o1valid, o2valid) of
              (True, _) -> o1
              (_, True) -> o2
              _ -> inp
       where
        (o1valid, o1) = g False inp
        (o2valid, o2) = g True inp

        g rd i = (isValidSymbol dw, snd $ encode8b10b rd dw)
         where
          dw = snd $ decode8b10b rd i

      expected = inp

  out H.=== expected
```

but now I got counterexamples. They truly are counterexamples. For instance, the code 0b11_0000_1001 represents D16.3 if the running disparity is `True`, but the code 0b00_1100_1001 also represents D16.3 under the same conditions. So we get:

```
    decodeEncode8b10b:       FAIL (0.02s)
        ✗ <interactive> failed at test/Test/Cores/LineCoding8b10b.hs:110:7
          after 11 tests and 3 shrinks.
          shrink path: 11:mBa

              ┏━━ test/Test/Cores/LineCoding8b10b.hs ━━━
           91 ┃ prop_decodeEncode8b10b :: H.Property
           92 ┃ prop_decodeEncode8b10b = H.withTests 1000 $ H.property $ do
           93 ┃   inp <-
           94 ┃     H.forAll
              ┃     │ 0 b11_0000_1001
           95 ┃       (Gen.filter checkBitSequence genDefinedBitVector)
           96 ┃   let out = case (o1valid, o2valid) of
           97 ┃               (True, _) -> o1
           98 ┃               (_, True) -> o2
           99 ┃               _ -> inp
          100 ┃        where
          101 ┃         (o1valid, o1) = g False inp
          102 ┃         (o2valid, o2) = g True inp
          103 ┃
          104 ┃         g rd i = (isValidSymbol dw, snd $ encode8b10b rd dw)
          105 ┃          where
          106 ┃           dw = snd $ decode8b10b rd i
          107 ┃
          108 ┃       expected = inp
          109 ┃
          110 ┃   out H.=== expected
              ┃   ^^^^^^^^^^^^^^^^^^
              ┃   │ ━━━ Failed (- lhs) (+ rhs) ━━━
              ┃   │ - 0 b00_1100_1001
              ┃   │ + 0 b11_0000_1001
```

because it picked the other alternative for D16.3.

So let's discuss how we should improve the test.

DigitalBrains1 (changes requested):
Thanks for the PR! I think there are some things that are important enough that I'd like to see them changed. I know this is just supposed to do its job and not be efficient.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,103 @@
+{-# LANGUAGE CPP #-}
```
DigitalBrains1, line 1: I'm pretty sure there's no CPP in this file...

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,103 @@
...
+
+-- | Function to check whether a 'Symbol8b10b' results in a data word
+isDw :: Symbol8b10b -> Bool
+isDw (Dw _) = True
+isDw _ = False
+
+-- | Function to check whether a 'Symbol8b10b' is not an error value
+isValidSymbol :: Symbol8b10b -> Bool
+isValidSymbol (Cw _) = True
+isValidSymbol dw = isDw dw
```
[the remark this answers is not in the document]
jvnknvlgl: Correct, fixed.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,103 @@
...
+
+-- | Function to convert a 'Symbol8b10b' to a plain 'BitVector 8'
+fromDw :: Symbol8b10b -> BitVector 8
```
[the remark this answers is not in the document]
jvnknvlgl: Also correct, fixed (hopefully, I think I removed al occurences of "dw", at least from this branch)

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
```diff
@@ -0,0 +1,103 @@
...
+fromDw dw = case dw of
+  Dw _dw -> _dw
+  Cw _dw -> _dw
+  _ -> 0
```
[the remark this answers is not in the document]
jvnknvlgl: That's reasonable, I took your suggestion. The thing is that you don't really _want_ to do anything with a value that has an `DwError` or `RdError`, so that's why I set it to be 0, but I guess it doesn't hurt if I don't.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
jvnknvlgl: Also reasonable, thank you!

As for the naming, IEEE 802.3 consistently calls these 10-bit things a "code group", so I took that. I'd personally prefer to stick to that.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
jvnknvlgl: Also took this, thanks!

clash-cores/test/Test/Cores/LineCoding8b10b.hs
jvnknvlgl: Nice, I wasn't aware this was possible. Thanks!

clash-cores/test/Test/Cores/LineCoding8b10b.hs
jvnknvlgl: Interesting, I will look at this tomorrow!

clash-cores/test/Test/Cores/LineCoding8b10b.hs
jvnknvlgl: Hmmm, that might be true. I will also take a look at this tomorrow.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
jvnknvlgl: `CLASH_OPAQUE` requires this, right? Or at least the HLS is bugging me about it.

clash-cores/test/Test/Cores/LineCoding8b10b.hs
jvnknvlgl: Ah, got it, it's because of this line line in `encode8b10b`:

`out = if isValidSymbol sym then (rdNew, cg) else (rd, 0)`

So if `sym` is not a valid symbol, the output will always be 0. The description should then be:

```
-- | Encode the input signal and check whether it is a valid value. It should be
--   valid for every valid input symbol.
```

jvnknvlgl: And of course in this case every input value is indeed valid, as it's hard-coded to be a `Dw`.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
DigitalBrains1: Right, 802.3, the 5600-page tome I should have checked instead of elsewhere.

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
DigitalBrains1: Woops, it completely slipped my mind...

clash-cores/src/Clash/Cores/LineCoding8b10b.hs
jvnknvlgl: 7025-page 😉

DigitalBrains1: We should compare copies :-P. Mine is 802.3-2018. I got it legally and everything. It's exactly 5600 pages, nice round number.

jvnknvlgl: Hahaha ah yes, mine is 802.3-2022. Downloaded in the first week of my internship, also legal as I can just turn on EduVPN and click "Download".

clash-cores/test/Test/Cores/LineCoding8b10b.hs
jvnknvlgl: Alright, I decided to add an extra decoding step for both the input and the output to be able to check whether they are the same without having to deal with the running disparity. Does this look OK?

DigitalBrains1: I think the test as I suggested is correct but that the table returns `Dw`s when in fact they are errors. By hand I now think I found these 5:
```diff
--- a/clash-cores/src/Clash/Cores/LineCoding8b10b/Decoder.hs
+++ b/clash-cores/src/Clash/Cores/LineCoding8b10b/Decoder.hs
@@ -1495,7 +1495,7 @@ decoderLut =
   , (0, 0, 0, 1, 0b11111100) -- D28.7
   , (1, 1, 0, 1, 0b11100000) -- D00.7
   , (0, 0, 0, 1, 0b11101111) -- D15.7
-  , (0, 0, 0, 1, 0b11100000) -- D00.7
+  , (0, 1, 0, 1, 0b11100000) -- D00.7
   , (1, 0, 0, 0, 0b11100111) -- D07.7
   , (1, 1, 0, 1, 0b11100000) -- D00.7
   , (0, 0, 0, 1, 0b11110000) -- D16.7
@@ -1814,11 +1814,11 @@ decoderLut =
   , (1, 1, 0, 0, 0b01100000) -- D00.3
   , (0, 0, 0, 0, 0b01111100) -- D28.3
   , (1, 1, 0, 0, 0b01100000) -- D00.3
-  , (0, 0, 0, 0, 0b01101111) -- D15.3
-  , (0, 0, 0, 0, 0b01100000) -- D00.3
+  , (0, 1, 0, 0, 0b01101111) -- D15.3
+  , (0, 1, 0, 0, 0b01100000) -- D00.3
   , (0, 0, 0, 1, 0b01100111) -- D07.3
   , (1, 1, 0, 0, 0b01100000) -- D00.3
-  , (0, 0, 0, 0, 0b01110000) -- D16.3
+  , (0, 1, 0, 0, 0b01110000) -- D16.3
   , (0, 0, 0, 0, 0b01111111) -- D31.3
   , (0, 0, 0, 1, 0b01101011) -- D11.3
   , (0, 0, 0, 0, 0b01111000) -- D24.3
@@ -1942,7 +1942,7 @@ decoderLut =
   , (1, 1, 0, 1, 0b11100000) -- D00.7
   , (0, 0, 1, 1, 0b11111100) -- K28.7
   , (1, 1, 0, 1, 0b11100000) -- D00.7
-  , (0, 0, 0, 1, 0b11101111) -- D15.7
+  , (0, 1, 0, 1, 0b11101111) -- D15.7
   , (0, 0, 0, 1, 0b11100000) -- D00.7
   , (1, 0, 0, 0, 0b11100111) -- D07.7
   , (1, 1, 0, 1, 0b11100000) -- D00.7
```

but this is not fun by hand and easily automated.

However, you were trying to be quickly done with this part of the work to concentrate on the interesting stuff instead, and now we're spending time on this part anyway. The errors I found were all non-critical.¹ So I think I'll suggest merging as-is, and making an issue that says a) this can be so much less resource-intensive and b) it appears to have mistakes, so it should just be replaced by a better implementation (not to be part of your internship, unless of course you find some extra time, but that is not something that happens in an internship I think).

¹ It will decode _some_ garbage as valid data, which is "be liberal in what you accept" taken too far, but in practice will not quickly lead to issues.

clash-cores/test/Test/Cores/LineCoding8b10b.hs
jvnknvlgl: (We have decided to scrap this test as it does not really make a lot of sense)

PR #2740, opened 2024-06-24
Only domains with a positive period are valid

christiaanb:
An alternative means to achieve https://github.com/clash-lang/clash-compiler/pull/2734#issuecomment-2182893091

```haskell
import Clash.Prelude
import Data.Proxy

f ::
  forall dom .
  KnownDomain dom =>
  Proxy dom ->
f Proxy = case knownDomain @dom of
    SDomainConfiguration {} -> SNat
```
now typechecks.

## Still TODO:

### DIFF
--- a/changelog/2024-07-05T14_06_50+02_00_require_positive_period
+++ b/changelog/2024-07-05T14_06_50+02_00_require_positive_period
@@ -1,0 +1,20 @@
+CHANGED: You can now only create an `SDomainConfiguration` when the `period` of the domain is at least `1`.
+Pattern matching on an `SDomainConfiguration` bring the `1 <= period` into scope.
+This in turns enables the following code to typecheck:
+```
+import Clash.Prelude
+import Data.Proxy
+
+f ::
+  forall dom .
+  KnownDomain dom =>
+  Proxy dom ->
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
I like this solution, thanks for giving this the needed amount of thought!

We should then document this feature both in a place where people will find it without specifically looking for the time conversion types and at the time conversion types. Something on the order of _If your code needs a `1 <= DomainPeriod dom` constraint, you can satisfy this with ... so you don't have to add the constraint to your function._ I'm sure we can think of a better way to phrase it.

martijnbastiaan (approved):
This is the best solution to this saga by a mile, IMO. Could you write a changelog?

DigitalBrains1:
Could you add an issue that we should still document this? Just a changelog isn't enough.

christiaanb:
@DigitalBrains1 Document it where though? In the tutorial? as part of the description `KnownDomain`? of `PeriodToCycles`?

DigitalBrains1:
I'm not sure; the one thing I _am_ sure of is it needs to be documented, probably linked or described in multiple places, so people will find it when they need it.

DigitalBrains1:
I'm fine with taking a shot at writing proper documentation myself next week.

PR #2741, opened 2024-06-24
Only domains with a positive period are valid, alternative

christiaanb:
An alternative means to achieve https://github.com/clash-lang/clash-compiler/pull/2734#issuecomment-2182893091
```haskell
import Clash.Prelude
import Data.Proxy

f ::
  forall dom .
  KnownDomain dom =>
  Proxy dom ->
f Proxy = SNat
```
now typechecks.

Though at the very high cost of needing to add `@dom` to any use of `hideClockResetEnable` which might affect user code out in the wild. https://github.com/clash-lang/clash-compiler/pull/2740 is probably a more desirable alternative.

## Still TODO:


DigitalBrains1:
I agree the other alternative is more desirable, all the more since the error message you get when you forget to explicitly type-apply it is not all that clear.

PR #2742, opened 2024-06-26
clash-cores: Add missing `TryDomain` for CRCs in the catalog

rowanG077:


DigitalBrains1:
We _just_ ran into this, `clash-cores:doctests` did not work with _multiple hidden_, which we tripped over in `nix-shell`.

We should add `clash-cores:doctests` to CI (using Cabal).

And I wonder why the `nix-build` CI job runs, to name one, `clash-prelude:unittests`, but not `clash-cores:doctests`.

rowanG077:
I know that motivated this PR :)

DigitalBrains1:
You're basically our chef Nix; do you have the time to see why our `nix-build` CI job doesn't seem to run for `clash-cores:doctests` even though it runs for, for instance, `clash-prelude:doctests`? I'm just testing locally on an Ubuntu machine, invoking `nix-build`, but that's what I'm seeing if I deliberately introduce failing tests.

DigitalBrains1:
Thanks! We'll need to decide whether that is the behaviour we are looking for.

What would be the deciding factor for me is: would building and testing under Nix have a fair chance of finding mistakes that would not be found by the tests we do with Cabal? If yes, let's test more under Nix. Otherwise, it might not be a good ratio between CPU time for the test and the odds of it finding actual issues.

rowanG077:
I''m happy with nix build. You can specify the target you want to build explicitly if you want. `nix build .#clash-cores` for example will build clash-cores. It's just  that we currently only build the default package which is set as clash-ghc.

For the shell I'm less happy. It requires the build of allmost packages inside the repo which takes ages. It makes little sense that if I want to develop the prelude I need to first compile prelude before I get into the shell for example. Ideally I would want it to just build the dependencies. And you can then build any of the packages in clash-compiler.

DigitalBrains1 (approved):
LGTM!

PR #2743, opened 2024-06-26
Add `clash-cores:doctests` to GitLab CI

DigitalBrains1:
Fixed minor details in `clash-cores` CRC core.

With `multiple-hidden` enabled, the following doctest line:

`crcEngine' = exposeClockResetEnable crcEngine systemClockGen resetGen enableGen`

worked fine on GHC 9.4.8 and 9.6.{4,5}. But on GHC 8.10.7, 9.0.2 and 9.2.8, GHC wanted to check `WithSingleDomain` then and there, but could not because the `crc` term to `crcEngine` could not be inspected through `TryDomain`. The GHC's where it did work postponed this check to the `myEngine` binding in the next doctest line, where it could be resolved.

However, this `exposeClockResetEnable` is superfluous anyway as the `sampleN` function later on deals with such implicit parameters fine on its own. So the code is changed to no longer use `exposeClockResetEnable` and instead specify the domain to `sampleN`.

Additionally, a function for which no documentation was even generated had Haddock documentation on a constraint. Haddock on GHC 8.10.7 cannot handle this, breaking the doctests for clash-cores.

The documentation of that constraint has simply been removed.

With these details fixed, we can run `clash-cores:doctests` in GitLab CI. Up until now, we only ran it for external PR's in the GitHub CI _Build and Test_ job, which was an accidental disparity.

GitHub CI also received several correctness and performance tweaks.

## Still TODO:

  - [x] Check copyright notices are up to date in edited files (did not add copyright for minuscule changes)

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -28,6 +28,7 @@ cabal v2-build all --write-ghc-environment-files=always
 
 # Put all the test binaries in a predictable location
 TESTS="
+clash-cores:doctests
 clash-cores:unittests
 clash-cosim:test
 clash-ffi:ffi-interface-tests
--- a/.ci/gitlab/test.yml
+++ b/.ci/gitlab/test.yml
@@ -79,6 +79,11 @@ build:
 
 # Tests run on shared runners:
 
+cores:doctests:
+  extends: .test-nocache
+  script:
+    - bin/clash-cores:doctests -j$THREADS
+
 cores:unittests:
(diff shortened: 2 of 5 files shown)


DigitalBrains1:
I should take a look at CI for 1.8 to see if the tweaks here also make sense there.

martijnbastiaan (approved):

PR #2750, opened 2024-07-07
Remove "multiple hidden"

martijnbastiaan:
Due to worsened type inference, this is unlikely to ever get out of experimental. At the same time, due to its heavy reliance on CPP, it is a burden to maintain.

## Still TODO:


clash-prelude/tests/Clash/Tests/TopEntityGeneration.hs
```diff
@@ -324,10 +315,6 @@ tests =
           $(unTypeQ $ maybeBuildTopEntity Nothing 'topEntityFailure5) @?= failed
       , testCase "topEntityFailure6" $
           $(unTypeQ $ maybeBuildTopEntity Nothing 'topEntityFailure6) @?= failed
-#if MULTIPLE_HIDDEN
-      , testCase "topEntityFailure7" $
-          $(unTypeQ $ maybeBuildTopEntity Nothing 'topEntityFailure7) @?= failed
-#endif
       , testCase "topEntityFailure8" $
```
leonschoorl, line 318: You could drop the `{-# LANGUAGE CPP #-}` from this file

clash-prelude/tests/Clash/Tests/Signal.hs
```diff
@@ -192,24 +71,6 @@ tests =
...
-
-test4ok_0 = withReset resetGen (test4 @System @System @System @System)
-test4ok_1 = withSpecificReset (resetGen @System) (test4 @System @_ @_ @_)
-test4ok_2 = withSpecificReset (resetGen @System) (test4 @_ @System @_ @_)
-test4ok_3 = withSpecificReset (resetGen @System) (test4 @_ @_ @System @_)
-
-test5ok_0 = withSpecificReset (resetGen @System) (test5 @System @_)
-#endif
-
 -- | Asserts that static clocks behave the same as dynamic clocks with a static
```
leonschoorl, line 74: You could now remove the `{-# LANGUAGE CPP #-}`

leonschoorl (commented):
A little sad, but that does clean up rather nicely.

I don't fully understand when `TryDomain`/`HasDomain` were needed, but looking at the uart and crc cores: they didn't seem to be using multiple hidden domains, yet still had instances defined.
Do you think that's common in user code?
If so, should we, for now, deprecated rather then outright remove those type classes, so code out there with such instances, could keep compiling on clash versions pre and post this change?

DigitalBrains1:
> I don't fully understand when `TryDomain`/`HasDomain` were needed, but looking at the uart and crc cores: they didn't seem to be using multiple hidden domains, yet still had instances defined.

Yes, that's correct. Suppose you have this code:

```haskell
f ::
  HiddenClock dom =>
  T ->
  [...]
```
and you wish to use a function which needs `With{Specific,Single}Domain` (such as `exposeClock`) on `f`, then `T` will need a `TryDomain` instance. This is just when `multiple-hidden` is enabled regardless of whether you actually use _multiple_ hidden clocks or just a single one.

> Do you think that's common in user code? If so, should we, for now, deprecated rather then outright remove those type classes, so code out there with such instances, could keep compiling on clash versions pre and post this change?

Good point! `multiple-hidden` has always been experimental, however, even without `multiple-hidden` enabled, these types have been exposed in our Prelude. Ergo, removing them is an API change and maybe it's better to keep the types around in a deprecated state for a major release cycle. They would no longer serve a purpose but just be there so existing code will keep compiling for another major release (at least for this specific API change).

I think it would not be _common_ per se in user code, but people who experimented with `multiple-hidden` might have haibtually added the types to their personal libraries so they _could_ experiment.

In the end, I think it's a judgement call whether to keep them around or not.

martijnbastiaan:
After talking to @DigitalBrains1: he has a slight preference towards keeping+deprecating, I've got a slight preference towards removing them. I'll ask @leonschoorl in the office tomorrow: if he's got a preference, I'll do that, if he doesn't I'll implement my preference.

martijnbastiaan:
@leonschoorl preferred keeping the classes around for now. I've set them to be removed in Clash 1.12, i.e. users get two releases to remove their references similar to `unsafeFromLowPolarity` and friends.

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
```diff
@@ -173,6 +173,7 @@ type instance TryDomain t Double                = 'NotFound
 type instance TryDomain t (Maybe a)             = TryDomain t a
 type instance TryDomain t (Either a b)          = Merge t a b
 
+{-# DEPRECATED HasSingleDomain "Experimental feature multiple hidden has been removed. This type family will therefore be removed in Clash 1.12." #-}
```
DigitalBrains1, line 176: Please also slap `DEPRECATED` on `TryDomain`, so people who declare type instances for that family will also get a deprecation notice.

I wonder if we shouldn't just slap it on all types; I don't know which types people would actually _use_ in their code; I am certain that this goes for `TryDomain` but why not slap it on every type we will remove?

DigitalBrains1 (changes requested):

clash-prelude/src/Clash/Class/HasDomain/HasSingleDomain.hs
martijnbastiaan: Done.

martijnbastiaan: I've slapped the deprecation on the whole module, which I should have done in the first place.

DigitalBrains1 (approved):

PR #2764, opened 2024-07-21
Add own version of `OverridingBool`

### LINKED ISSUE #2754: `OverridingBool` prevents auto derivation of various classes
See https://github.com/clash-lang/clash-compiler/pull/2711#discussion_r1670894452. We could remove a whole bunch of code if we were to define our own (isomorphic) version of `OverridingBool`.

martijnbastiaan:
Fixes #2754

Backporting not strictly necessary, but this will reduce effort for future backports.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver/Bool.hs
+++ b/clash-lib/src/Clash/Driver/Bool.hs
@@ -1,0 +1,34 @@
+{-|
+  Copyright   :  (C) 2024, Martijn Bastiaan
+  License     :  BSD2 (see the file LICENSE)
+  Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveAnyClass #-}
+
+module Clash.Driver.Bool where
+
+import Control.DeepSeq (NFData)
+import Data.Hashable (Hashable)
+import GHC.Generics (Generic)
+
+#if MIN_VERSION_ghc(9,4,0)
+import qualified GHC.Data.Bool as Ghc
+#elif MIN_VERSION_ghc(9,0,0)
+import qualified GHC.Utils.Misc as Ghc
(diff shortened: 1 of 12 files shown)


clash-lib/src/Clash/Driver/Bool.hs
```diff
@@ -0,0 +1,34 @@
...
+
+#if MIN_VERSION_ghc(9,4,0)
+import GHC.Data.Bool (OverridingBool(..))
+#elif MIN_VERSION_ghc(9,0,0)
+import GHC.Utils.Misc (OverridingBool(..))
+#else
+import Util (OverridingBool(..))
+#endif
+
+data COverridingBool = CAuto | CNever | CAlways
```
christiaanb, line 23: Can’t we just name it the same way as GHC? Instead of using this 'C' prefix for both the type and data constructors.

And the name the conversion functions: from/toGHCOverridingBool.

We have a lot of the same names as the GHC code base, so it’s not like this would be the first time we would do this.

christiaanb (changes requested):

clash-lib/src/Clash/Driver/Bool.hs
martijnbastiaan: Done (with slightly different capitalization due to our style guide)

christiaanb: But the type constructor still has the 'C' prefix, can't we remove that as well?

martijnbastiaan: Darn it, missed that with renaming

martijnbastiaan: Done

christiaanb: Thanks

christiaanb (approved):

PR #2768, opened 2024-07-21
Add GHC 9.8 to CI

martijnbastiaan:
## Still TODO:

  - [X] ~~Add issue mentioning we need to add GHC 9.8 to GitHub Actions (once Stackage has a stable release with 9.8?)~~ Fixed, thanks @christiaanb!

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -67,13 +67,7 @@ optional-packages:
 allow-newer:
   brick:base,
   brick:deepseq,
-  vector-binary-instances:base,
-  cryptohash-sha256:base,
   hashable,
-  haskell-src-meta:template-haskell,
-  string-interpolate:template-haskell,
-  string-interpolate:text,
-  tasty-hedgehog:tasty,
   hedgehog-fakedata:hedgehog,
   rewrite-inspector:binary,
   rewrite-inspector:containers,
--- a/cabal.project
+++ b/cabal.project
@@ -66,13 +66,20 @@ optional-packages:
 
 allow-newer:
   brick:base,
+  brick:deepseq,
   vector-binary-instances:base,
   cryptohash-sha256:base,
   hashable,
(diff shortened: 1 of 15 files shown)


cabal.project
```diff
@@ -66,13 +66,20 @@ optional-packages:
 
 allow-newer:
   brick:base,
+  brick:deepseq,
```
DigitalBrains1, line 69: I think we can drop the following ones:
```
vector-binary-instances:base,
cryptohash-sha256:base,
haskell-src-meta:template-haskell,
string-interpolate:template-haskell,
string-interpolate:text,
tasty-hedgehog:tasty,
```

DigitalBrains1: Perhaps that also goes for `clash-ffi/example/cabal.project`? I did not test it.

christiaanb: Thanks, I've made the allow-newer lists smaller.

DigitalBrains1 (approved):
LGTM, and I'm fine with merging as-is, but I wonder if it should also include dropping the following env var:

https://github.com/clash-lang/clash-compiler/blob/aba55fed9f45711c8336935721a43d243f7f78c1/tests/Main.hs#L66-L70

Because apparently it does not matter that `~/.cabal/store/ghc-9.8.2` does not exist.

Hmmm, wait, I now see that it also ends in `:`! This means that all standard paths are appended. So maybe we shouldn't _drop_ the env var, but modify it to only include `dist-newstyle/packagedb/ghc-9.8.2` which does in fact exist; it turns out the GHC ABI is not appended in that dir (at least not for me, locally).

martijnbastiaan:
Wow, that's weird @DigitalBrains1. I think we should make an issue of it.

I'm happy merging too. Thanks for all the work @christiaanb!

PR #2780, opened 2024-08-02
Add `maybeResize` (and friends)

### LINKED ISSUE #2779: Add `maybeResize` (and friends)
We have `resize`, a function that can truncate by throwing away bits. We also have `checkedResize`, a function that does the same, but throws an error in simulation if the value of the argument does not fit in the result type. Ideally, we'd also have a total function that returns a `Maybe` instead.

jvnknvlgl:

Add `maybeResize` and `maybeTruncateB`. Resolves #2779.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Class/Resize.hs
+++ b/clash-prelude/src/Clash/Class/Resize.hs
@@ -6,7 +6,11 @@ License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
-{-# LANGUAGE NoGeneralizedNewtypeDeriving #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE MonoLocalBinds #-}
+#if __GLASGOW_HASKELL__ == 902
+{-# LANGUAGE GADTs #-}
+#endif
 
 {-# LANGUAGE Safe #-}
 
@@ -20,6 +24,8 @@ module Clash.Class.Resize
  , checkedResize
  , checkedFromIntegral
  , checkedTruncateB
+ , maybeResize
+ , maybeTruncateB
  ) where
 
 import Data.Kind (Type)
@@ -28,6 +34,12 @@ import GHC.Stack (HasCallStack)
 import GHC.TypeLits (Nat, KnownNat, type (+))
 
 import Clash.Sized.Internal (formatRange)
+
+#if MIN_VERSION_base(4,16,0)
+import GHC.TypeLits (OrderingI(EQI, GTI), cmpNat)
+#else
+import Clash.Promoted.Nat (natToNatural)
+#endif
 
 -- | Coerce a value to be represented by a different number of bits
 class Resize (f :: Nat -> Type) where
@@ -117,3 +129,38 @@ checkedTruncateB ::
   f (a + b) -> f a
 checkedTruncateB v =
   checkIntegral (Proxy @(f a)) v `seq` truncateB v
+
+-- | Like 'resize', but returns 'Nothing' if the argument is out of bounds for
+-- the result type.
+maybeResize ::
+  forall a b f.
+  ( Resize f
+  , KnownNat a, Integral (f a)
+  , KnownNat b, Integral (f b), Bounded (f b) ) =>
+  f a -> Maybe (f b)
+maybeResize v =
+#if MIN_VERSION_base(4,16,0)
+  case Proxy @a `cmpNat` Proxy @b of
+    GTI | v > resize (maxBound @(f b)) -> Nothing
+    GTI | v < resize (minBound @(f b)) -> Nothing
+    EQI -> Just v
+    _ -> Just (resize v)
+#else
+  case natToNatural @a `compare` natToNatural @b of
+    GT | v > resize (maxBound @(f b)) -> Nothing
+    GT | v < resize (minBound @(f b)) -> Nothing
+    _ -> Just (resize v)
+#endif
+
+-- | Like 'truncateB', but returns 'Nothing' if the argument is out of bounds for
+-- the result type.
+maybeTruncateB ::
+  forall a b f.
+  ( Resize f
+  , KnownNat b, Integral (f (a + b))
+  , KnownNat a, Integral (f a), Bounded (f a) ) =>
+  f (a + b) -> Maybe (f a)
+maybeTruncateB v
+  | v > resize (maxBound @(f a)) = Nothing
+  | v < resize (minBound @(f a)) = Nothing
+  | otherwise = Just (truncateB v)
--- a/clash-prelude/tests/Clash/Tests/Resize.hs
+++ b/clash-prelude/tests/Clash/Tests/Resize.hs
@@ -1,18 +1,37 @@
+{-# LANGUAGE NumericUnderscores #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TypeFamilies #-}
 
-module Clash.Tests.Resize (tests) where
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.Extra.Solver #-}
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.Normalise #-}
+{-# OPTIONS_GHC -fplugin=GHC.TypeLits.KnownNat.Solver #-}
 
+module Clash.Tests.Resize (tests, main) where
+
+import Clash.Prelude (Unsigned, Signed, BitVector, Index)
+import Clash.Promoted.Nat (SNat(..))
+import Clash.XException (XException)
 import Control.DeepSeq (NFData)
 import Control.Exception (SomeException, try, evaluate)
-import Clash.XException (XException)
 import Data.Either (isLeft)
 import Data.Proxy (Proxy(Proxy))
-import GHC.TypeNats (KnownNat, type (<=))
-import Test.Tasty (TestTree, testGroup)
-import Test.Tasty.QuickCheck
+import GHC.TypeNats (KnownNat, SomeNat(..), type (<=), type (+), Nat, someNatVal)
+import Hedgehog ((===))
+import Numeric.Natural (Natural)
+import Test.Tasty (TestTree, testGroup, defaultMain)
+import Test.Tasty.Hedgehog (testPropertyNamed)
+import Test.Tasty.QuickCheck (testProperty, discard, ioProperty, counterexample)
 
 import qualified Clash.Class.Resize as Resize
-import Clash.Sized.Index
+import qualified Hedgehog as H
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+import qualified Test.Tasty.QuickCheck as Q
+
+withSomeSNat :: Natural -> (forall (n :: Nat). SNat n -> r) -> r
+withSomeSNat n f = case someNatVal n of
+  SomeNat (_ :: Proxy n) -> f (SNat @n)
 
 -- | Anything that's in bounds should not cause an error
 indexProp ::
@@ -26,7 +45,7 @@ indexProp Proxy v =
 indexFailProp ::
   forall a b.
   ((b <= a), KnownNat a, KnownNat b) =>
-  Proxy b -> Index a -> Property
+  Proxy b -> Index a -> Q.Property
 indexFailProp Proxy v =
   let checked = Resize.checkedResize @a @b v in
   if toInteger v > toInteger (maxBound @(Index b)) then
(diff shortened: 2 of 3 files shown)


clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -117,3 +120,57 @@ checkedTruncateB ::
...
+  a -> Maybe b
+maybeFromIntegral v =
+  if maybeIntegral (Proxy @b) v
+    then Just (fromIntegral v)
+    else Nothing
+
+-- | Like 'resize', but returns 'Nothing' if /f a/ is out of bounds for /f b/.
+-- Useful when you do not know /f a/ can be out of bounds, and would like to
+-- have your assumptions checked.
+maybeResize ::
```
rowanG077, line 153: This will break for anything larger than native word size (64 bit). At least mention this limitation in the doc.

rowanG077: Oh in fact you can do a short circuit on the type level if `BitSize a >= BitSize b` which means this function will be free in that case in hardware.

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -117,3 +120,57 @@ checkedTruncateB ::
...
+  Proxy b ->
+  a -> Bool
+maybeIntegral Proxy v =
+     toInteger v < toInteger (maxBound @b)
+  && toInteger v > toInteger (minBound @b)
+
+-- | Like 'fromIntegral', but returns 'Nothing' if /a/ is out of bounds for /b/.
+-- Useful when you do not know /a/ can be out of bounds, and would like to have
+-- your assumptions checked.
+
```
DigitalBrains1, line 137: The Haddock for `maybeFromIntegral` is not generated, and that's probably because of this blank line. This blank line means that what follows the blank line is a regular Haskell comment rather than a Haddock comment. I would have expected the Haddock to still be associated with the function, but apparently it's cut off and lost?

TL;DR: Remove the blank line :-). And always check the Haddock output :-).

To view the Haddock, you can click on the _Details_ button next to the _ci/gitlab/gitlab.com_ CI test result, and then click on _haddock_ (in the column _test_), and then click the big _Browse_ button in the rightmost column of that webpage.

Or just generate Haddock on your own system of course, before you run CI.

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -117,3 +120,57 @@ checkedTruncateB ::
...
+  (Integral a, Integral b, Bounded b) =>
+  a -> Maybe b
+maybeFromIntegral v =
+  if maybeIntegral (Proxy @b) v
+    then Just (fromIntegral v)
+    else Nothing
+
+-- | Like 'resize', but returns 'Nothing' if /f a/ is out of bounds for /f b/.
+-- Useful when you do not know /f a/ can be out of bounds, and would like to
+-- have your assumptions checked.
```
DigitalBrains1, line 152: I think all these `maybe` variants can be described much shorter:

> Like `resize`, but returns `Nothing` if _f a_ is out of bounds for _f b_.

The rest seems unnecessary, and also it mentions assumptions, but we don't go in with assumptions. This function is total; the assumption with the `checked` variant was that it would fit, but we want to error when that assumption is wrong. This is different.

Also, Martijn and I discussed that we'd best rephrase all these functions a bit, but that can be done in a followup PR. Or I can tack a commit onto this PR...

DigitalBrains1 (changes requested):
I have some thoughts on the Haddock

martijnbastiaan:
I've added HDL friendly versions and added tests @DigitalBrains1. Could you take another look? I've removed `maybeFromIntegral`, it seemed too big of a footgun. (Don't mind the CI changes, I'll struggle through all the old GHC versions at a later time..)

clash-prelude/tests/Clash/Tests/Resize.hs
```diff
@@ -1,18 +1,37 @@
...
+
+import qualified Test.Tasty.QuickCheck as Q
+import qualified Hedgehog as H
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
 
 import qualified Clash.Class.Resize as Resize
-import Clash.Sized.Index
+import Clash.Promoted.Nat (SNat(..))
+import Clash.Prelude (Unsigned, Signed, BitVector, Index)
```
DigitalBrains1, line 30: Did you intend to sort the imports like this? I got the sense you were separating non-qualified and qualified imports but here it's non-qualified again?

clash-prelude/tests/Clash/Tests/Resize.hs
```diff
@@ -26,16 +45,146 @@ indexProp Proxy v =
...
+    maxFa = fromIntegral (maxBound @(f a)) :: Integer
+    minFb = fromIntegral (minBound @(f b)) :: Integer
+    maxFb = fromIntegral (maxBound @(f b)) :: Integer
+
+  H.footnote $ "minFa: " <> show minFa
+  H.footnote $ "maxFa: " <> show maxFa
+  H.footnote $ "minFb: " <> show minFb
+  H.footnote $ "maxFb: " <> show maxFb
+
+  input <- H.forAll $ Gen.integral (Range.linear minFa maxFa)
```
DigitalBrains1, line 83: Wouldn't these be rather biased to zero? I.e., not producing interesting cases?
```suggestion
  input <- H.forAll $ Gen.integral (Range.constant minFa maxFa)
```

clash-prelude/tests/Clash/Tests/Resize.hs
```diff
@@ -26,16 +45,146 @@ indexProp Proxy v =
...
+  let
+    minFa = fromIntegral (minBound @(f a)) :: Integer
+    maxFa = fromIntegral (maxBound @(f a)) :: Integer
+    minFab = fromIntegral (minBound @(f (a + b))) :: Integer
+    maxFab = fromIntegral (maxBound @(f (a + b))) :: Integer
+
+  H.footnote $ "minFa: " <> show minFa
+  H.footnote $ "maxFa: " <> show maxFa
+  H.footnote $ "minFab: " <> show minFab
+  H.footnote $ "maxFab: " <> show maxFab
```
DigitalBrains1, line 146: This is a truly fabulous footnote!

(Also, thanks for adding footnotes, it's very nice in the case a counterexample is found)

clash-prelude/tests/Clash/Tests/Resize.hs
```diff
@@ -26,16 +45,146 @@ indexProp Proxy v =
...
+
+  input <- H.forAll $ Gen.integral (Range.linear minFa maxFa)
```
DigitalBrains1, line 148: Same comment about bias

```suggestion
  input <- H.forAll $ Gen.integral (Range.constant minFa maxFa)
```

clash-prelude/tests/Clash/Tests/Resize.hs
```diff
@@ -49,5 +198,16 @@ tests = testGroup "Resize"
...
+    , testPropertyNamed "maybeResizeBitVectorProp" "maybeResizeBitVectorProp" maybeResizeBitVectorProp
+    , testPropertyNamed "maybeTruncateUnsignedProp" "maybeTruncateUnsignedProp" maybeTruncateUnsignedProp
+    , testPropertyNamed "maybeTruncateSignedProp" "maybeTruncateSignedProp" maybeTruncateSignedProp
+    , testPropertyNamed "maybeTruncateIndexProp" "maybeTruncateIndexProp" maybeTruncateIndexProp
+    , testPropertyNamed "maybeTruncateBitVectorProp" "maybeTruncateBitVectorProp" maybeTruncateBitVectorProp
     ]
   ]
+
+main :: IO ()
+main = defaultMain $ adjustOption (\_ -> HedgehogTestLimit (Just 1_000_000)) $ tests
```
DigitalBrains1, line 213: Did you intend to include this in the PR or was it just for local testing?

DigitalBrains1 (changes requested):
Thanks for going forward with this! I still have some thoughts.

clash-prelude/tests/Clash/Tests/Resize.hs
martijnbastiaan: Yes, it would. I wasn't aware of `constant`! Actually, they both suck. They basically have zero chance of producing interesting edge cases.

clash-prelude/tests/Clash/Tests/Resize.hs
martijnbastiaan: No rhyme or reason really :see_no_evil:. I'll fix it.

martijnbastiaan: I want [`by-scope-then-qualified`](https://github.com/fourmolu/fourmolu/pull/403) and never manually sort imports again :heart_eyes:

clash-prelude/tests/Clash/Tests/Resize.hs
DigitalBrains1: We know what the edge cases are, so we could produce primarily edge cases. [`Gen.frequency`](https://hackage.haskell.org/package/hedgehog-1.5/docs/Hedgehog-Gen.html#v:frequency).

You know, now that I look at it, the doc for `frequency` only says it shrinks towards the first generator. I really thought it also only considers earlier items on the list for small sizes. If I now look at the implementation, it would appear I am wrong about that.

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -117,3 +120,30 @@ checkedTruncateB ::
   f (a + b) -> f a
 checkedTruncateB v =
   checkIntegral (Proxy @(f a)) v `seq` truncateB v
+
+-- | Like 'resize', but returns 'Nothing' if /f a/ is out of bounds for /f b/.
```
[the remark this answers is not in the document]
DigitalBrains1: Actually, this goes for the whole file, not just the addition.

Isn't this what I [referred to above](https://github.com/clash-lang/clash-compiler/pull/2780#discussion_r1705517838) by
>  Also, Martijn and I discussed that we'd best rephrase all these functions a bit, but that can be done in a followup PR. Or I can tack a commit onto this PR...

clash-prelude/tests/Clash/Tests/Resize.hs
martijnbastiaan: Both! I'll make it less opinionated, but it's really nice for test modules to have `main`s. Then I can do:

```
./repld p:tests -W -T Clash.Tests.Resize.main
```

clash-prelude/tests/Clash/Tests/Resize.hs
martijnbastiaan: Would you mind if I skip this (=> `frequency`. I've changed to `constant`.)? I don't want to spend much more effort. Or rather, I think the current situation is "good enough".

martijnbastiaan:
@DigitalBrains1 I think I've applied all your suggestions.

clash-prelude/tests/Clash/Tests/Resize.hs
DigitalBrains1: Go ahead. It's good enough for me with `constant`, I just continued on from your assertion that it still sucked :-D.

clash-prelude/tests/Clash/Tests/Resize.hs
DigitalBrains1: ... _Or_ ...

We extend `repld` to allow passing [command line arguments](https://github.com/ndmitchell/ghcid?tab=readme-ov-file#how-do-i-run-pass-command-arguments-with---test) to `clash-prelude/tests/unittests.hs`, including arbitrrary Tasty options. You could do

```
./repld p:tests -W -T unittests -S Resize -A "--hedgehog-tests 1000000"
```
(S for _specific test_ and A for _additional arguments_)
and it would invoke
```
ghcid ... -T Main.main --setup ":set args -p Resize --hedgehog-tests 1000000"
```

clash-prelude/src/Clash/Class/Resize.hs
```diff
@@ -117,3 +129,38 @@ checkedTruncateB ::
...
+  f a -> Maybe (f b)
+maybeResize v =
+#if MIN_VERSION_base(4,16,0)
+  case Proxy @a `cmpNat` Proxy @b of
+    GTI | v > resize (maxBound @(f b)) -> Nothing
+    GTI | v < resize (minBound @(f b)) -> Nothing
+    EQI -> Just v
+    _ -> Just (resize v)
+#else
+  case natToNum @a @Int `compare` natToNum @b @Int of
```
DigitalBrains1, line 149: What about `Index (2^65)`? (Do we still do 32-bit machine compatibility? If so, what about `Index (2^33)`?)

I feel it would be better to use `Integer` or `Natural` here.

DigitalBrains1:
```haskell
case natToNatural @a `compare` natToNatural @b of
```

martijnbastiaan: Oh wow, I didn't even consider that. I guess the tests should pick from `[1..2^128]` too!

clash-prelude/tests/Clash/Tests/Resize.hs
martijnbastiaan: Oh huh, I didn't know you could do that. Are you insisting on removing the `main`? I don't really mind it existing, even if the same thing can be achieved in another way.

martijnbastiaan:
@jvnknvlgl I've removed `maybeFromIntegral`, as I think it is dangerous due to it having no good HDL interpretation. I wanted to explicitly mention it, because I feel we're close to merging this. If you feel like removing it was a mistake, please feel free to make a new PR/issue where we can discuss this.

DigitalBrains1 (approved):

jvnknvlgl:
I agree with whatever changes you have made -  to be honest I completely forget about this PR anyway. Thanks for doing the work!

clash-prelude/tests/Clash/Tests/Resize.hs
DigitalBrains1: Oh no, I just wondered if there wasn't a "more nice" way to achieve the result, and this seemed like one. Go ahead and keep it as is.

PR #2784, opened 2024-08-07
Add `BitPack` instance for `Index 0`

kleinreact:
This PR gets rid of the additional `1 <= n` constraint for the `BitPack` instance of `Index n`.

### Background:
* `Index 0` and `Void` are empty types → are isomorphic to the empty set
* `Index 1` and `()` are singleton types → are isomorphic to a singleton set

Hence, we only need to question: how many bits do you need to distinguish between different elements of these types / sets? In both cases the answer clearly is: 0. Witnessing, `BitSize (Index 0) = 0` makes sense.

### Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Sized/Internal/Index.hs
+++ b/clash-prelude/src/Clash/Sized/Internal/Index.hs
@@ -9,6 +9,7 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RoleAnnotations #-}
 {-# LANGUAGE TemplateHaskell #-}
@@ -21,6 +22,7 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise       #-}
 
 {-# OPTIONS_HADDOCK show-extensions not-home #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Clash.Sized.Internal.Index
   ( -- * Datatypes
@@ -75,8 +77,10 @@ import Prelude hiding             (even, odd)
 
 import Control.DeepSeq            (NFData (..))
 import Data.Bits                  (Bits (..), FiniteBits (..))
+import Data.Constraint            (Dict(..))
 import Data.Data                  (Data)
 import Data.Default.Class         (Default (..))
+import Data.Proxy                 (Proxy(..))
 import Text.Read                  (Read (..), ReadPrec)
 import Text.Printf                (PrintfArg (..), printf)
 import Data.Ix                    (Ix(..))
@@ -93,9 +97,11 @@ import Language.Haskell.TH        (TypeQ)
 import GHC.Generics               (Generic)
 import GHC.Natural                (Natural, naturalFromInteger)
 import GHC.Natural                (naturalToInteger)
+import GHC.Num.Integer            (integerLog2)
 import GHC.Stack                  (HasCallStack)
 import GHC.TypeLits               (KnownNat, Nat, type (+), type (-),
                                    type (*), type (<=), natVal)
+import GHC.TypeLits.KnownNat      (KnownNat1(..), SNatKn(..), nameToSymbol)
 import GHC.TypeLits.Extra         (CLog)
 import Test.QuickCheck.Arbitrary  (Arbitrary (..), CoArbitrary (..),
                                    arbitraryBoundedIntegral,
@@ -111,9 +117,10 @@ import Clash.Class.BitPack.BitIndex (replaceBit)
 import Clash.Sized.Internal       (formatRange)
 import {-# SOURCE #-} Clash.Sized.Internal.BitVector (BitVector (BV), high, low, undefError)
 import qualified Clash.Sized.Internal.BitVector as BV
-import Clash.Promoted.Nat         (SNat(..), snatToNum, natToInteger, leToPlusKN)
+import Clash.Promoted.Nat         (SNat(..), SNatLE(..), snatToNum, natToInteger, leToPlusKN, compareSNat)
 import Clash.XException
   (ShowX (..), NFDataX (..), errorX, showsPrecXWith, rwhnfX, seqX)
+import Unsafe.Coerce              (unsafeCoerce)
 
 {- $setup
 >>> import Clash.Sized.Internal.Index
@@ -184,10 +191,28 @@ instance NFData (Index n) where
   -- NOINLINE is needed so that Clash doesn't trip on the "Index ~# Integer"
   -- coercion
 
-instance (KnownNat n, 1 <= n) => BitPack (Index n) where
-  type BitSize (Index n) = CLog 2 n
-  pack   = packXWith pack#
-  unpack = unpack#
+instance KnownNat n => BitPack (Index n) where
+  type BitSize (Index n) = BitSizeIndex n
+  pack   = case compareSNat (SNat @1) (SNat @n) of
+    SNatGT                   -> const 0
+    SNatLE | Dict <- fact @n -> packXWith pack#
+  unpack = case compareSNat (SNat @1) (SNat @n) of
+    SNatGT                   -> const undefined
+    SNatLE | Dict <- fact @n -> unpack#
+
+type family BitSizeIndex (n :: Nat) where
+  BitSizeIndex 0 = 0
+  BitSizeIndex n = CLog 2 n
+
+instance KnownNat n => KnownNat1 $(nameToSymbol ''BitSizeIndex) n where
+  natSing1 = let n = natVal (Proxy @n)
+              in SNatKn $ if n == 0
+                          then 0
+                          else fromInteger $ toInteger $ integerLog2 n
+  {-# INLINE natSing1 #-}
+
+fact :: forall n. 1 <= n => Dict (CLog 2 n ~ BitSizeIndex n)
+fact = unsafeCoerce (Dict :: Dict (0~0))
 
 -- | Safely convert an `SNat` value to an `Index`
 fromSNat :: (KnownNat m, n + 1 <= m) => SNat n -> Index m


clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -184,10 +191,28 @@ instance NFData (Index n) where
...
-  type BitSize (Index n) = CLog 2 n
-  pack   = packXWith pack#
-  unpack = unpack#
+instance KnownNat n => BitPack (Index n) where
+  type BitSize (Index n) = BitSizeIndex n
+  pack   = case compareSNat (SNat @1) (SNat @n) of
+    SNatGT                   -> const 0
+    SNatLE | Dict <- fact @n -> packXWith pack#
+  unpack = case compareSNat (SNat @1) (SNat @n) of
+    SNatGT                   -> const undefined
```
martijnbastiaan, line 200: Could you replace this with an `errorX` call with a nice error message?

clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -184,10 +191,28 @@ instance NFData (Index n) where
...
+type family BitSizeIndex (n :: Nat) where
+  BitSizeIndex 0 = 0
+  BitSizeIndex n = CLog 2 n
+
+instance KnownNat n => KnownNat1 $(nameToSymbol ''BitSizeIndex) n where
+  natSing1 = let n = natVal (Proxy @n)
+              in SNatKn $ if n == 0
+                          then 0
+                          else fromInteger $ toInteger $ integerLog2 n
+  {-# INLINE natSing1 #-}
```
martijnbastiaan, line 212: Could we perhaps move this to a submodule `Orphans` (and import it here) such that we can eliminate the need for `-wno-orphans` on this whole module?

martijnbastiaan (changes requested):
Review together with @DigitalBrains1 

Nice, this would get rid of a lot of unwanted `1 <= n` constraints.

> Add [this fact](https://github.com/clash-lang/clash-compiler/pull/2784/files#diff-c5d06203ed828f905a88d010d96c98851be5ca4cee3abc46a05b3131f5161080R214) to the typelits plugins.

I doubt reasoning about type families is in-scope for the plugins, but you might want to take that up with the maintainers :-).

I think we should change the cover letter's motivation to be something along the lines of "this aligns clash-the-compiler's thoughts on `Index 0` with `BitPack (Index 0)`".

A changelog is needed, but that's also in the TODOs.

Other than that LGTUs.

clash-prelude/src/Clash/Sized/Internal/Index.hs
rowanG077: I vote for `clashCompileError`

martijnbastiaan: Is it? We want Clash to always throw constructs like this away (because netlist says it is zero bits).

rowanG077: Ahh yes, this case can happen and must be undefined. Yes I agree with `errorX`.

DigitalBrains1:
It seems to still need something more though:

```
src/Clash/Explicit/Synchronizer.hs:126:35: error: [GHC-25897]
    • Could not deduce ‘Clash.Sized.Internal.Index.BitSizeIndex
                          (2 ^ addrSize)
                        ~ addrSize’
      from the context: (KnownDomain wdom, KnownDomain rdom, NFDataX a,
                         KnownNat addrSize, 1 <= addrSize)
```

<details><summary>Full message</summary>

```
src/Clash/Explicit/Synchronizer.hs:126:35: error: [GHC-25897]
    • Could not deduce ‘Clash.Sized.Internal.Index.BitSizeIndex
                          (2 ^ addrSize)
                        ~ addrSize’
      from the context: (KnownDomain wdom, KnownDomain rdom, NFDataX a,
                         KnownNat addrSize, 1 <= addrSize)
                   fifoMem :: forall (wdom :: Clash.Signal.Internal.Domain)
                                     (rdom :: Clash.Signal.Internal.Domain) a
                                     (addrSize :: GHC.TypeNats.Nat).
                              (KnownDomain wdom, KnownDomain rdom, NFDataX a, KnownNat addrSize,
                               1 <= addrSize) =>
                              Clock wdom
                              -> Clock rdom
                              -> Enable wdom
                              -> Enable rdom
                              -> Signal wdom Bool
                              -> Signal rdom (BitVector addrSize)
                              -> Signal wdom (BitVector addrSize)
                              -> Signal wdom (Maybe a)
                              -> Signal rdom a
        at src/Clash/Explicit/Synchronizer.hs:(100,1)-(115,18)
      Expected: BitVector addrSize
                -> Clash.Sized.Internal.Index.Index (2 ^ addrSize)
        Actual: BitVector
                  (Clash.Class.BitPack.Internal.BitSize
                     (Clash.Sized.Internal.Index.Index (2 ^ addrSize)))
                -> Clash.Sized.Internal.Index.Index (2 ^ addrSize)
      ‘addrSize’ is a rigid type variable bound by
        the type signature for:
          fifoMem :: forall (wdom :: Clash.Signal.Internal.Domain)
                            (rdom :: Clash.Signal.Internal.Domain) a
                            (addrSize :: GHC.TypeNats.Nat).
                     (KnownDomain wdom, KnownDomain rdom, NFDataX a, KnownNat addrSize,
                      1 <= addrSize) =>
                     Clock wdom
                     -> Clock rdom
                     -> Enable wdom
                     -> Enable rdom
                     -> Signal wdom Bool
                     -> Signal rdom (BitVector addrSize)
                     -> Signal wdom (BitVector addrSize)
                     -> Signal wdom (Maybe a)
                     -> Signal rdom a
        at src/Clash/Explicit/Synchronizer.hs:(100,1)-(115,18)
    • In the first argument of ‘fmap’, namely ‘unpack’
      In the second argument of ‘(<$>)’, namely ‘fmap unpack waddr’
      In the first argument of ‘(<*>)’, namely
        ‘RamWrite <$> fmap unpack waddr’
    • Relevant bindings include
        portB :: Signal wdom (RamOp (2 ^ addrSize) a)
          (bound at src/Clash/Explicit/Synchronizer.hs:125:4)
        waddr :: Signal wdom (BitVector addrSize)
          (bound at src/Clash/Explicit/Synchronizer.hs:116:38)
        raddr :: Signal rdom (BitVector addrSize)
          (bound at src/Clash/Explicit/Synchronizer.hs:116:32)
        fifoMem :: Clock wdom
                   -> Clock rdom
                   -> Enable wdom
                   -> Enable rdom
                   -> Signal wdom Bool
                   -> Signal rdom (BitVector addrSize)
                   -> Signal wdom (BitVector addrSize)
                   -> Signal wdom (Maybe a)
                   -> Signal rdom a
          (bound at src/Clash/Explicit/Synchronizer.hs:116:1)
    |
126 |                (RamWrite <$> fmap unpack waddr <*> fmap fromJustX wdataM)
    |                                   ^^^^^^
```
</details>

Experienced with GHC 9.6.6.

kleinreact:
It's because of the new `BitSizeIndex` type family introduced and can be fixed by teaching the typelits plugins about it's properties in particular. Unfortunately, I also don't know about any better solution at the moment. I tried several other things to help the type checker in acting in any smarter way, but that's the best solution I could find.

clash-prelude/src/Clash/Sized/Internal/Index.hs
kleinreact: Done.

clash-prelude/src/Clash/Sized/Internal/Index.hs
kleinreact: Not required any more. The type family is now embedded into the `ghc-typelits-extra` plugin.

leonschoorl:
I understand the desire to get rid of that `1 <= n` constraint in a polymorphic setting.

But lets say someone is (accidentally) calling `unpack @(Index 0)`.
This PR changes this from a compile time error to a run time error.
Which feels to me like a step in the wrong direction.

kleinreact:
> This PR changes this from a compile time error to a run time error.

I don't think that this is the right way to look at it. Consider the following example:
```haskell
topEntity ::
  HiddenClockResetEnable System =>
  Signal System (Index 0)
topEntity = pure $ unpack @(Index 0) $ resize (0 :: BitVector 1)
```
If that description is turned into Verilog with `clash`, without the changes of this PR, then the example produces a compile time error. However, it produces perfectly valid Verilog after the fixes introduced by this PR.

I agree that the user might end up with a run time error in simulation, if trying to inspect what comes out of applying `unpack @(Index 0)`, but that is fine, as there is no output `unpack @(Index 0)` can ever produce. Thus, the error is not that `unpack @(Index 0)` now is a valid object, which can be created at runtime, but more that the user tries to get something out of it, which just is not possible. 

Note that the type system never was intended to prevent us from this, in the same way as it cannot prevent us from ever having a look at the contents of `undefined`.

DigitalBrains1 (changes requested):
This isn't a full review yet; I'll wait until you have it working.

But browsing a bit I did notice you expanded the scope of the PR a lot. You've removed a lot of `Index n, 1 <= n` constraints even if they don't involve `BitPack`. If you want to address different cases of `Index` bounds as well, I think it's better treated in a new PR because it'll require some good thought.

For instance, you're removing the `1 <= n` from the `resetGlitchFilter` functions, but now it is just completely broken:

```haskell
>>> printX $ sampleN 10 $ unsafeFromReset $ resetGlitchFilter d2 systemClockGen resetGen
[True,True,True,True,True,False,False,False,False,False]
>>> printX $ sampleN 10 $ unsafeFromReset $ resetGlitchFilter d0 systemClockGen resetGen
[True,True,True,True,undefined,undefined,undefined,undefined,undefined,undefined]
```

changelog/2026-01-03T06_22_01+01_00_add_bitpack_index_0
```diff
@@ -0,0 +1,2 @@
+ADDED: `Index 0` instances for `BitPack`, `Bits`, `Counter`, `FiniteBits`, `MaybeNumConvert`, `NumConvert`, `Partiy`, and `SaturatingNum`.
+ADDED: Zero-sized block RAM implementations for `BlockRam1` and `BlockRamU`.
```
DigitalBrains1:
```suggestion
ADDED: Zero-sized block RAM implementations for `blockRam1` and `blockRamU`.
```

clash-lib/prims/commonverilog/Clash_Sized_Internal_Index.primitives.yaml
```diff
@@ -1,16 +1,16 @@
 - BlackBox:
     name: Clash.Sized.Internal.Index.pack#
     kind: Expression
-    type: 'pack# :: Index
-      n -> BitVector (CLog 2 n)'
+    type: 'pack# :: Index n ->
+      BitVector (CLogWZ 2 n 0)'
```
DigitalBrains1:
```suggestion
    type: |-
      pack# :: Index n -> BitVector (CLogWZ 2 n 0)
```

Please note the previous contents was generated by a conversion script. Now that we're back to human writing, let's make it less ugly. However, I understand this is a paper cut and if you don't want to do this, feel free to ignore. I'd like to see it, though. GitHub has an _add suggestion to batch_ feature that would make it less effort to do. You might need to switch out of the GitHub "new experience preview" for the button to appear.

clash-lib/prims/commonverilog/Clash_Sized_Internal_Index.primitives.yaml
```diff
@@ -1,16 +1,16 @@
...
     template: ~ARG[0]
     workInfo: Never
 - BlackBox:
     name: Clash.Sized.Internal.Index.unpack#
     kind: Expression
-    type: 'unpack# :: (KnownNat
-      n, 1 <= n) => BitVector (CLog 2 n) -> Index n'
-    template: ~ARG[2]
+    type: 'unpack# :: KnownNat n =>
+      BitVector (CLogWZ 2 n 0) -> Index n'
```
DigitalBrains1:
```suggestion
    type: |-
      unpack# :: KnownNat n => BitVector (CLogWZ 2 n 0) -> Index n
```

clash-lib/prims/vhdl/Clash_Sized_Internal_Index.primitives.yaml
```diff
@@ -1,16 +1,16 @@
 - BlackBox:
     name: Clash.Sized.Internal.Index.pack#
     kind: Expression
-    type: 'pack# :: Index
-      n -> BitVector (CLog 2 n)'
+    type: 'pack# :: Index n ->
+      BitVector (CLogWZ 2 n 0)'
```
DigitalBrains1:
```suggestion
    type: |-
      pack# :: Index n -> BitVector (CLogWZ 2 n 0)
```

clash-lib/prims/vhdl/Clash_Sized_Internal_Index.primitives.yaml
```diff
@@ -1,16 +1,16 @@
...
     template: std_logic_vector(~ARG[0])
     workInfo: Never
 - BlackBox:
     name: Clash.Sized.Internal.Index.unpack#
     kind: Expression
-    type: 'unpack# :: (KnownNat
-      n, 1 <= n) => BitVector (CLog 2 n) -> Index n'
-    template: unsigned(~ARG[2])
+    type: 'unpack# :: KnownNat n =>
+      BitVector (CLogWZ 2 n 0) -> Index n'
```
DigitalBrains1:
```suggestion
    type: |-
      unpack# :: KnownNat n => BitVector (CLogWZ 2 n 0) -> Index n
```

clash-prelude/src/Clash/Class/Counter/Internal.hs
```diff
@@ -79,7 +85,7 @@ class Counter a where
     | a == minBound = (True, countMax)
     | otherwise = (False, pred a)
 
-instance (1 <= n, KnownNat n) => Counter (Index n)
+instance KnownNat n => Counter (Index n)
```
DigitalBrains1: Is the difference in behaviour between `Index 0` and `Index 1` intended? I would have thought a tuple with `Index 0` elements would just count past the `Index 0` but it causes `XException`s instead...

```haskell
>>> x = countMin :: (Index 2, Index 1, Index 3)
>>> P.take 10 $ P.map (\(a,b,c) -> (a,c)) $ P.iterate countSucc x
[(0,0),(0,1),(0,2),(1,0),(1,1),(1,2),(0,0),(0,1),(0,2),(1,0)]
>>> y = countMin :: (Index 2, Index 0, Index 3)
>>> P.take 10 $ P.map (\(a,b,c) -> (a,c)) $ P.iterate countSucc y
[(0,0),(0,1),(0,2),(*** Exception: X: maxBound of 'Index 0' is undefined
  errorX, called at src/Clash/Sized/Internal/Index.hs:272:10 in clash-prelude-1.9.0-inplace:Clash.Sized.Internal.Index
```

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
```diff
@@ -82,13 +83,13 @@ class MaybeNumConvert a b where
 instance (KnownNat n, KnownNat m) => MaybeNumConvert (Index n) (Index m) where
   maybeNumConvert !a = maybeResize a
 
-instance (KnownNat n, KnownNat m, 1 <= n) => MaybeNumConvert (Index n) (Unsigned m) where
-  maybeNumConvert !a = maybeResize $ bitCoerce @_ @(Unsigned (CLog 2 n)) a
+instance (KnownNat n, KnownNat m) => MaybeNumConvert (Index n) (Unsigned m) where
+  maybeNumConvert !a = maybeResize $ bitCoerce @_ @(Unsigned (CLogWZ 2 n 0)) a
```
DigitalBrains1: Shouldn't the result be `Just undefined` instead of `undefined` for `Index 0`?

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
```diff
@@ -82,13 +83,13 @@ class MaybeNumConvert a b where
...
-instance (KnownNat n, KnownNat m, 1 <= n) => MaybeNumConvert (Index n) (Signed m) where
-  maybeNumConvert !a = maybeNumConvert $ bitCoerce @_ @(Unsigned (CLog 2 n)) a
+instance (KnownNat n, KnownNat m) => MaybeNumConvert (Index n) (Signed m) where
+  maybeNumConvert !a = maybeNumConvert $ bitCoerce @_ @(Unsigned (CLogWZ 2 n 0)) a
 
-instance (KnownNat n, KnownNat m, 1 <= n) => MaybeNumConvert (Index n) (BitVector m) where
+instance (KnownNat n, KnownNat m) => MaybeNumConvert (Index n) (BitVector m) where
   maybeNumConvert !a = maybeResize $ pack a
 
 instance (KnownNat n, KnownNat m) => MaybeNumConvert (Unsigned n) (Index m) where
```
DigitalBrains1: Wouldn't it be nicer if `maybeNumConvert` returned `Nothing` when the result type is `Index 0`?

```haskell
>>> printX (maybeNumConvert (2 :: Unsigned 4) :: Maybe (Index 1))
Nothing
>>> printX (maybeNumConvert (2 :: Unsigned 4) :: Maybe (Index 0))
undefined
```

Although I note that
```haskell
>>> printX (maybeNumConvert (undefined :: Index 1) :: Maybe (Unsigned 4))
undefined
>>> printX (maybeNumConvert (undefined :: Index 0) :: Maybe (Unsigned 4))
undefined
```
even though it could have been `Just undefined` with more knowledge of the domain. This is however universally the case, whereas the other way around we see a difference between `Index 1` and `Index 0` that I think is undesirable.

clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -343,29 +343,32 @@ minus# (I a) (I b) =
...
-    case natToInteger @n of
-      1 -> a +# b
-      _ -> leToPlusKN @1 @n $
-        case plus# a b of
-          z | let m = fromInteger# (natToInteger @n)
-            , z >= m -> resize# (z - m)
-          z -> resize# z
-  satAdd SatZero a b =
-    leToPlusKN @1 @n $
+
```
DigitalBrains1: A spurious blank line has been added.

clash-prelude/src/Clash/Class/Counter/Internal.hs
kleinreact: In my opinion, we should **not** offer a `Counter` instance for `Index 0`, because `countMin` and `countMax` don't exist. However, `clash-prelude` also offers a `Bounded` instance for `Index 0`, which similarly throws `XException`s for `minBound` and `maxBound`. Hence, in terms of consistency, also exposing the instance for `Counter` makes sense to me.

The alternative would be to add a `1 <= n` constraint to the `Bounded` instances of `Index`. Nevertheless, introducing a constraint can be a more problematic API change, while just dropping a constraint usually won't cause any trouble.

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
kleinreact: I don't have a strong opinion on that, but I'd be happy to get some feedback from the inventor of `NumConvert`. 

@martijnbastiaan any particular thoughts?

clash-prelude/src/Clash/Class/Counter/Internal.hs
DigitalBrains1: I think the `Counter` instance is much more problematic, though, as the `XException`s bleed to the other parts of the tuple. Please note that it is the `Index 2` with the `XException` in the output I pasted above. I think this is much more likely to trip up a user than the `Bounded` one.

From a usability standpoint, I'd like to see it skip over the `Index 0`, but I'm not against just keeping the constraint if you feel strongly that that isn't correct.

DigitalBrains1: I'd also like to note that in the PR that added `Counter`: https://github.com/clash-lang/clash-compiler/pull/1763 , @martijnbastiaan chose to include the `1 <= n` for `Index` even though he could have omitted it and it would have worked the same as it does know in this PR. So it likely was a design decision, not a necessity.

kleinreact: > Is the difference in behaviour between Index 0 and Index 1 intended?

Yes, it is very much intended. May it be that there is a conceptual misunderstanding, as void types and singleton types are very different objects indeed?

> From a usability standpoint, I'd like to see it skip over the Index 0, but I'm not against just keeping the constraint if you feel strongly that that isn't correct.

I don't get that point. 

In terms of the number of elements a type captures, composition of product types is multiplicative, not additive, e.g., `(Index 2, Index 0, Index 3)` captures `2 * 0 * 3 = 0` elements,  while `(Index 2, Index 1, Index 3)` captures `2 * 1 * 3 = 6` elements. 

Hence, the `y` that you pick in your example
```haskell
>>> y = countMin :: (Index 2, Index 0, Index 3)
>>> showX y
"(0,undefined,0)"
```
does not exist and therefore very much makes sense to be an exception, right?

DigitalBrains1: > May it be that there is a conceptual misunderstanding, as void types and singleton types are very different objects indeed?

No, I understand it. I just think allowing a `Counter` instance for something containing `Index 0` is a bad idea from a usability standpoint. Even though your reasoning may be sound, it's the end result that's not desirable. So can we just keep the `1 <= n` constraint then? You yourself said we should not offer a `Counter` instance for `Index 0`!

> I don't get that point.

Is it that I'm not explaining what I mean by usability? I mean that I think it's likely a user will not expect nor desire the `Counter` behaviour for `Index 0` and accidentally create an error in their design that way, leading to tedious and hard "chase the `XException`" debugging of their design.

kleinreact: > Is it that I'm not explaining what I mean by usability?

Yes, I was confused by that, because it sounded to me like you want to have some particular use case for `(Index 2, Index 0, Index 3)` iterating over the type's elements  with `countSucc` somehow. 

However, if I understand you correctly now, then you are more worried about the user accidentally introducing some `Index 0` as part of a larger `Counter` structure, which I agree: can lead to unexpected behavior the user might not be aware of at first.

I still consider the introduction of a `1 <= n` constraint to the `Bounded` instances of `Index` to be a cleaner solution, nevertheless, because then the defaulting behavior will automatically exclude any implementation of `Counter` for `Index 0` without the need for this particular extra constraint.

You hopefully remember our internal discussion on how much pain in the neck all these extra `1 <= n` constraints can be, hence eliding as many as possible is a desirable goal to achieve imo.

DigitalBrains1: > because then the defaulting behavior will automatically exclude any implementation of Counter for Index 0 without the need for this particular extra constraint.

Are you sure? As far as I'm aware, even if you use defaulting, you still need to write out the constraints needed. That's also why you see
```haskell
instance KnownNat n => Counter (Unsigned n)
```
That's because `Enum` and `Bounded` need `KnownNat n`, so we need to specify it for the `Counter` instance using the default methods as well. It's not _automatically_ excluded; it just doesn't type-check if you don't specify it. So it doesn't constrain it any _further_, but it still needs to be _specified_.

Anyway, introducing a `1 <= n` constraint to `Bounded` is out of scope for this PR, and if a new PR doing that would be accepted,¹ it would revert the `Counter (Index 0)` instance to what it already was, with the constraint. Both semantically and syntactically. What is the purpose of removing the constraint in this PR then?

All of this feels rather academical. I get the feeling we agree `Counter (Index n)` needs a `1 <= n` constraint, let's just keep it that way and call it a day?

I think I've made my point clear, I'd like to finish this discussion please.

¹ My first thought is I would not be in favour

kleinreact: > Are you sure? As far as I'm aware, even if you use defaulting, you still need to write out the constraints needed.

You are right, I missed that particular consequence at this point. I reverted it back to the original.

> All of this feels rather academical.

Where I am very happy that we can have these types of discussion in our community as well.

DigitalBrains1: Thanks for reverting the definition!

> > All of this feels rather academical.
> 
> Where I am very happy that we can have these types of discussion in our community as well.

Are we using the same definiton for the word academic? (When looking it up I noticed I meant _academic_, the word _academical_ does not appear to exist, but _academically_ does so maybe my ESL brain got confused by that)

<a href="https://www.oxfordlearnersdictionaries.com/definition/english/academic_1">:link:
image
<hr>
image</a>

kleinreact: The word `academical` seems to have a German correspondent at least, which reflects the meaning I had in mind when reading your sentence: https://dict.leo.org/englisch-deutsch/academical

DigitalBrains1:
Oops, I really did not expect my git to choose your branch to push to when I executed the command
```
$ git push -u origin peter/add-bitpack-for-index-zero
```
It probably had something to do with a gaffe in setting up that branch, sorry.

Anyway, on [peter/add-bitpack-for-index-zero](https://github.com/clash-lang/clash-compiler/tree/peter/add-bitpack-for-index-zero) I added a commit to have `maybeNumConvert @_ @(Index 0)` always emit `Nothing` as I suggested above. This seems a much better result than always having it return `XException`. I'd like to suggest integrating the contents of that commit in this PR.

I added a `[skip ci]` commit to prevent a costly GitLab CI pipeline, but if I subsequently screw up this branch and restore it, the CI will run for this branch anyway. Sorry, wasted electrons, I salute thee.

[edit]
Ooh, actually, I suspect GitLab CI will not run, only GitHub CI. The revert to the original commit was probably quick enough to put everything in order before the earlier push had a chance to trigger a run. Saved electrons, you may do a useful job later on, I salute thee.
[/edit]

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
martijnbastiaan: It seems to me that introducing `Index 0` breaks almost all laws laid out by both `NumConvert` and `MaybeNumConvert`. Even with a favorable interpretation where undefinedness is treated as "neither violates nor follows" it seems iffy to me. If a conversion is in `NumConvert` it strongly suggest (and is maybe even implied by the laws) that `MaybeNumConvert` would return a `Just`. However, the bound laws on `MaybeNumConvert` suggest it should return `Nothing`. Which leads me to the Python Zen:

> In the face of ambiguity, refuse the temptation to guess.

I.e., don't implement it / constrain it by types. If I _had_ to pick one, I would return `Nothing`. But again, I strongly believe the laws/ideas to be self-contradicting when introducing empty types. 

Edit: it was pointed out to me that there was still an open discussion here due to my new PR on `NumConvert`. I didn't see it at the time.

kleinreact: > It seems to me that introducing `Index 0` breaks almost all laws laid out by both `NumConvert` and `MaybeNumConvert`it

@martijnbastiaan Can you give a concrete example where you see things breaking?

There are only two types of changes regarding `NumConvert` in this PR:

1. **Conversion to** `Index m`, e.g., 
    ```diff
    -instance (KnownNat n, KnownNat m, 1 <= m, 2 ^ n <= m) => NumConvert (Unsigned n) (Index m) where
    +instance (KnownNat n, KnownNat m, 2 ^ n <= m) => NumConvert (Unsigned n) (Index m) where
    ```
    In all updated cases the dropped `1 <= m` constraint is obsolete, as `1 <= 2 ^ n` (by definition) and `2 ^ n <= m` (as required). Thus, the constraint already follows via transitivity of `<=`.

    The only reason for the constraint being required before was the missing `BitPack` instance for `Index 0` added in this PR.

2. **Conversion from** `Index n`, e.g.,
    ```diff
    -instance (KnownNat n, KnownNat m, 1 <= n, n <= 2 ^ m) => NumConvert (Index n) (BitVector m) where
    +instance (KnownNat n, KnownNat m, n <= 2 ^ m) => NumConvert (Index n) (BitVector m) where
    ```
    where for `Index 0` we have `numConvert` specializing to  
    ```haskell
    numConvert :: KnownNat m => Index 0 -> BitVector m
    ```
    *  Thus, this function only can be applied to values of type `Index 0`.
    *  There are no values that are legitimate to be typed `Index 0`. 
    * Hence, this particular instance of `numConvert` is impossible to ever cause any harm.

    In other words, it's just another classical example for [ex falso quodlibet](https://en.wikipedia.org/wiki/Ex_falso_quodlibet), which makes it logically sound to have such functions in the first place (also see [absurd](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Void.html#v:absurd) in `Data.Void`).

   In particular, this means that in both cases, `NumConvert` and `MaybeNumConvert`, logic allows us to return whatever we want when converting from some `Index 0` to any type.

   > In the face of ambiguity, refuse the temptation to guess.

   There is no ambiguity here. We just have a free choice and no matter how we choose the answer is correct.

martijnbastiaan: Ah, I missed that `1 <= m` follows from `2 ^ n <= m`. If there is no way for `numConvert` to return an `Index 0`, it's fine by me. Accepting an `Index 0` as an argument is also fine by me, though this should yield the original `XException` given to it if called, as all other instances do.

For `maybeNumConvert`: if it needs to yield a `Maybe (Index 0)`, it seems more sensible to me to yield a `Nothing`. This is more in line with the spirit of the law, so to speak. If it needs to accept an `Index 0`, it should return the `XException` given to it (as it does for any other bottom given to it).

Edit: Unless I'm misunderstanding what you're saying, this behavior/implementation should be fine according to your reasoning, right?

DigitalBrains1: The current PR does break a law though:

> Instances should make sure their constraints are as "tight" as possible. I.e., if an instance's constraints cannot be satisfied, then for the same types `maybeNumConvert` should return `Nothing` for one or more values in the domain of the source type `a`:
> ```
> L.any isNothing (L.map (maybeNumConvert @a @b) [minBound ..])
> ```

For `b ~ Index 0`, the current PR always returns `XException`, never `Nothing`. My patch fixes that. It's not just sensible as Martijn says in the message directly above, it is required to satisfy the laws.

While reviewing this PR, I neglected to read the laws before this morning. That's why I didn't mention this before: I only discovered it this morning.

DigitalBrains1: And it also breaks this law:

> A conversion is either successful or it fails gracefully. I.e., it does not produces produce errors (also see `Clash.XException`).

This law is relaxed by this statement:

> For the time being, if the input is an `XException`, then the output is too. This property might be relaxed in the future.

but that relaxation does not allow:

```haskell
>>> printX (maybeNumConvert (2 :: Unsigned 4) :: Maybe (Index 0))
undefined
```

The two-sentence description of the `MaybeNumConvert` class already expresses this, by the way:

> A failure is expressed by returning `Nothing`, never by an `XException`.

martijnbastiaan: Ah, great catch @DigitalBrains1. When I said:

> This is more in line with the spirit of the law, so to speak. 

I was going on a somewhat fuzzy interpretation of:

> `isJust (maybeNumConvert @a @b x) == (i x >= i (minBound @b) && i x <= i (maxBound @b))`

But you're right: the laws do force our hands to `Nothing`.

kleinreact: Thanks for the feedback and the suggested fix. I cherry-picked @DigitalBrains1's improvement.

changelog/2026-01-03T06_22_01+01_00_add_bitpack_index_0
```diff
@@ -0,0 +1,2 @@
+ADDED: `Index 0` instances for `BitPack`, `Bits`, `Counter`, `FiniteBits`, `MaybeNumConvert`, `NumConvert`, `Partiy`, and `SaturatingNum`.
```
DigitalBrains1:
```suggestion
ADDED: `Index 0` instances for `BitPack`, `Bits`, `Counter`, `FiniteBits`, `Parity`, and `SaturatingNum`.
```

I noticed a typo in the word _parity_. But then it also dawned on me that we have not released a Clash version with `NumConvert` yet, so you can't actually log a change in something that doesn't exist yet.

changelog/2026-01-03T06_22_01+01_00_add_bitpack_index_0
```diff
@@ -0,0 +1,2 @@
...
+ADDED: Zero-sized block RAM implementations for `blockRam1` and `blockRamU`.
```
DigitalBrains1, line 2:
```suggestion
ADDED: Zero-sized block RAM implementations for `blockRam1` and `blockRamU`.
ADDED: `Counter a => Counter (Vec 0 a)` instance.
```

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
```diff
@@ -98,19 +100,23 @@ maybeNumConvert a =
...
-  maybeNumConvertCanonical !a = maybeResize $ bitCoerce @_ @(Unsigned (CLog 2 n)) a
+instance (KnownNat n, KnownNat m) => MaybeNumConvertCanonical (Index n) (Unsigned m) where
+  maybeNumConvertCanonical !a = maybeResize $ bitCoerce @_ @(Unsigned (CLogWZ 2 n 0)) a
 
-instance (KnownNat n, KnownNat m, 1 <= n) => MaybeNumConvertCanonical (Index n) (Signed m) where
-  maybeNumConvertCanonical !a = maybeNumConvertCanonical $ bitCoerce @_ @(Unsigned (CLog 2 n)) a
+instance (KnownNat n, KnownNat m) => MaybeNumConvertCanonical (Index n) (Signed m) where
+  maybeNumConvertCanonical !a = maybeNumConvertCanonical $ bitCoerce @_ @(Unsigned (CLogWZ 2 n 0)) a
 
 instance (KnownNat n, KnownNat m, 1 <= n) => MaybeNumConvertCanonical (Index n) (BitVector m) where
```
DigitalBrains1:
```suggestion
instance (KnownNat n, KnownNat m) => MaybeNumConvertCanonical (Index n) (BitVector m) where
```

kleinreact:
> Could you please elaborate what you mean?

I say that it is impossible to create any wrong hardware this way, because the Clash model still protects you from breaking things after HDL generation. The only case, where you can observe a difference is in simulation, if returning an `Index 0` as part of your top entity result and trying to read that result after simulating. In any other case, lazy evaluation protects you from the obvious, e.g. as the following won't cause any problems, neither in simulation nor in HDL:

```haskell
topEntity :: HiddenClockResetEnable System => Signal System Bool -> Signal System Bool
topEntity = mealy (\i b -> (if b then i + 1 else i, b)) (undefined :: Index 0)
```

```
clashi> sampleN @System 3 $ topEntity $ fromList [True, False, True]
[True,False,True]
```

In other words, we only give the user a foot gun for the specific case, where he wants to read some `Index 0` as a simulation result, which is perfectly fine to lead to an exception, because that means that the simulation test is already set up in a wrong way.

The current situation however, instead imposing that `1 <= n` constraint, completely forbids the user to generate that nice and perfectly valid HDL, which is causing all the problems we see with these constraints in our projects regularly.

christiaanb:
The PR still introduced a run-time exception by dropping a `1 <= n` constraint, so it’s not that much different. And the reason why that PR uses clashCompileError is because generating HDL for `sum Nil` would actually be bad, which is why we couldn’t get by with a simple `errorX`

kleinreact:
> The delay is absolutely not due to the reviewers, which you seem to imply.

There never was any intention that would imply such allegation. 

I am just tiered of running into similar appearing discussions over and over again with individual opinions about the "dos and don'ts" swapping back and forth all the time. My question simply was about "What do you like to see as the outcome?", because if we don't even agree on the outcome, then it doesn't make  any sense to further put effort into changing the code.

I opened this PR, because multiple users at QBayLogic reported to be restricted by the `1 <= n` constraints for no real reason, which I also could confirm and see independently in the projects I worked on.

I first want confirmation now, whether maintainers agree with this update or not, where agreement also means for me to not raise any further arguments against the update, which is merely counter-productive. 

Before agreeing on the outcome, I don't see why I should put any more effort into further touching this PR.

christiaanb:
I agree @kleinreact i will merge this PR

Issue #2787, opened 2024-08-17
`Counter` for `Vec`tors


gergoerdi:
How about something like this?

```
data Endian = BigEndian | LittleEndian

newtype Counters (endianness :: Endian) n a = Counters{ getCounters :: Vec n a }
    deriving (Generic, NFDataX, Eq, Show)

ripple :: ((Bool -> a -> (Bool, a)) -> Bool -> Vec n a -> (Bool, Vec n a)) -> (a -> (Bool, a)) -> Vec n a -> (Bool, Vec n a)
ripple mapAccum f = mapAccum step True
  where
    step carry x = if carry then f x else (False, x)

instance (Counter a, KnownNat n, 1 <= n) => Counter (Counters BigEndian n a) where
    countMin = Counters $ repeat countMin
    countMax = Counters $ repeat countMax

    countSuccOverflow = fmap Counters . ripple mapAccumR countSuccOverflow . getCounters
    countPredOverflow = fmap Counters . ripple mapAccumR countPredOverflow . getCounters

instance (Counter a, KnownNat n, 1 <= n) => Counter (Counters LittleEndian n a) where
    countMin = Counters $ repeat countMin
    countMax = Counters $ repeat countMax

    countSuccOverflow = fmap Counters . ripple mapAccumL countSuccOverflow . getCounters
    countPredOverflow = fmap Counters . ripple mapAccumL countPredOverflow . getCounters
```

Mostly I just want to be able to use a `Counter BigEndian n (Index 10)` as the world's least-effort BCD counter...

martijnbastiaan:
Is there a reason we can't build `Counter a => Counter (Vec n a)`? Or do you want to introduce `Counters` as a means to introduce endianness?

gergoerdi:
Just neither direction seemed more natural to me than the other, so I went with both.

martijnbastiaan:
Yeah, I've had similar thoughts when writing instances for tuples. I see you made https://github.com/clash-lang/clash-compiler/pull/2788 marking this issue as fixed; feel free to open an additional issue proposing to extend `Counter` with endianness if that's important to you.

PR #2788, opened 2024-08-19
Add `Counter a => Counter (Vec n a)` instance #2787

gergoerdi:
Fixes: #2787

### DIFF
--- a/clash-prelude/src/Clash/Class/Counter/Internal.hs
+++ b/clash-prelude/src/Clash/Class/Counter/Internal.hs
@@ -14,6 +14,7 @@ import Clash.Sized.BitVector (BitVector, Bit)
 import Clash.Sized.Index (Index)
 import Clash.Sized.Signed (Signed)
 import Clash.Sized.Unsigned (Unsigned)
+import Clash.Sized.Vector as Vec (Vec, repeat, mapAccumR)
 
 import Data.Bifunctor (bimap)
 import Data.Functor.Identity (Identity(..))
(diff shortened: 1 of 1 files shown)


leonschoorl:
I think it would be good if the instance came with a comment about then endianness.
Something like https://github.com/clash-lang/clash-compiler/blob/96f741d01d9ad22ecb5635c862e3a8b640027c89/clash-prelude/src/Clash/Class/Counter/Internal.hs#L156-L168

I like having examples like that, because I can never which way round is supposed to be big or little.

leonschoorl:
Also this endianness you implemented now is the "right" one I think.
Because its consistent with the view encoded in `BitPack`/[`Clash.Class.BitPack.BitIndex`](https://hackage.haskell.org/package/clash-prelude-1.8.1/docs/Clash-Class-BitPack-BitIndex.html):

```
clashi> countSucc @(Vec 3 Bit) (0 :> 0 :> 0 :> Nil)
0 :> 0 :> 1 :> Nil
clashi> msb @(Vec 3 Bit) (0 :> 0 :> 1 :> Nil)
0
clashi> lsb @(Vec 3 Bit) (0 :> 0 :> 1 :> Nil)
1
```

martijnbastiaan:
@kloonbot run_ci ca3cba089d317a1cd517421c177345ab786fdacc

martijnbastiaan (approved):
Agreed with @leonschoorl on the examples / doctests. Other than that, LGTM.

gergoerdi:
Where are we on this?

martijnbastiaan:
@kloonbot run_ci ec242f66a60af2955c44901d4fb15c2a5ac6edab

martijnbastiaan:
I'm okay with this if CI is.

PR #2816, opened 2024-10-01
Fix (System)Verilog literal rendering of Index values

### LINKED ISSUE #2813: The power of `Index`
The following code

```haskell
module Top where

import Clash.Prelude

topEntity :: HiddenClockResetEnable System => Signal System (Index (2 ^ 61))
topEntity = pure 0
```

causes `clash`  from master (currently [3b755b9](https://github.com/clash-lang/clash-compiler/commit/3b755b900810c727833875b1a15222d69d37029e)) to get stuck and to eat up all my 64G of memory after about 15 minutes. Just run 

```
cabal run -- clash-ghc:clash Top.hs  --verilog
```

from within the repository for reproduction. There seems to be an exponential rise in compile time that is directly related to the chosen exponent. Here are some compile times, produced via replacing the `61` in the code above with the given values.

| Exponent | Time (s) |
| -------- | ---- |
| 24 | 0.698 |
| 25 | 0.754 |
| 26 | 0.890 |
| 27 | 1.204 |
| 28 | 1.763 |
| 29 | 3.191 |
| 30 | 6.050 |
| 31 | 11.397 |
| 32 | 25.777 |

Everything works, if I compile the file with my system `clash` tough, which still is at `v1.8.1`. I tested the behavior with GHC `9.6.4` and `9.8.2`.

leonschoorl:
There was confusion between the size in bits and the type level argument to Index.
Causing it to use way too big numbers in the overflow calculation.

Fixes #2813

## Still TODO:


leonschoorl:
This fixes the hang that Felix experienced.

But I've still left this as a draft, as the overflow behaviour still isn't right, or at least not the same as in Haskell.

```haskell
topEntity = 6 :: Index 5
```

Evaluating this in Haskell produces:
```
*** Exception: X: Clash.Sized.Index: result 6 is out of bounds: [0..4]
```

But clash will just generate HDL with the 6 in it:
```verilog
  assign result = 3'd6;
```

Should it just produce the same HDL as if one would call `errorX`?
So:
```verilog
  assign result = {3 {1'bx}};
```
or:
```vhdl
  result <= Index_topEntity_types.index_5'(0 to 2 => '-');
```

DigitalBrains1:
> But I've still left this as a draft, as the overflow behaviour still isn't right, or at least not the same as in Haskell.
> 
> Should it just produce the same HDL as if one would call `errorX`? So:

Let me just note that the contract is "When Clash produces an `XException`, the HDL can produce any value whatsoever". So it's law-abiding.

I'll not touch the issue whether it's _desired_.

leonschoorl:
I've now added the overflow detection too

clash-lib/src/Clash/Backend/SystemVerilog.hs
```diff
@@ -1228,7 +1228,9 @@ expr_ _ (BlackBoxE pNm _ _ _ _ bbCtx _)
 expr_ _ (BlackBoxE pNm _ _ _ _ bbCtx _)
   | pNm == "Clash.Sized.Internal.Index.fromInteger#"
   , [Literal _ (NumLit n), Literal _ i] <- extractLiterals bbCtx
-  = exprLitSV (Just (Index (fromInteger n),fromInteger n)) i
+  , Just k <- clogBase 2 n
+  , let k' = max 1 k
```
DigitalBrains1, line 1232: When would `k` be zero in this code path?

I feel like that only happens for types like `Index 1` which shouldn't render in the first place?

leonschoorl: I was basically copying what the VHDL backend does:
https://github.com/clash-lang/clash-compiler/blob/3b755b900810c727833875b1a15222d69d37029e/clash-lib/src/Clash/Backend/VHDL.hs#L1767-L1772

leonschoorl: But I think you're right

christiaanb: That code was added in version 0.6.15 https://github.com/clash-lang/clash-compiler/commit/6b365dc082e78ed8c0f54f77faf30f62ae260a4a#diff-d636015ead3ec91b11144d7f18c90be9133810ab83457b3649a91e046cfae73fR647

Which might be from a time before we elided 0-bit values?

christiaanb (approved):
LGTM

changelog/2024-10-01T14_41_36+02_00_fix_index_literal_rendering
```diff
@@ -0,0 +1,2 @@
+FIXED: Clash hanging when rendering `Index n` literals, for large values of `n` [#2813](https://github.com/clash-lang/clash-compiler/issues/2813)
+FIXED: Render overflowed Index literals as errors
```
DigitalBrains1, line 2:
```suggestion
FIXED: Render overflowed Index literals as don't-cares in HDL
```
?

Maybe that's not good either, but I worry people might not understand what "render as an error" means. I'd also like to emphasise that we're talking about HDL here for clarity.

clash-lib/src/Clash/Backend/SystemVerilog.hs
leonschoorl: Yeah probably.

I had seen that, at least in simple cases clash, removes `Index 1` literals because they're 0 bits wide values.
But I'm wasn't sure if there was some code path I didn't foresee. And assumed the author of that code in VHDL.hs knew what they were doing.
 
You want me to take one the `clogBase`?

PR #2833, opened 2024-10-26
Fix DDR primitives

DigitalBrains1:
All of our DDR primitives turned out to have some issues. This PR initially started as fixing some issues in Xilinx `oddr` but quickly turned into fixing more bugs and extending the Haddock documentation.

* `Clash.Explicit.DDR`:
  * `ddrOut`: VHDL: Fix incorrect usage of `Enable` signal. With asynchronous resets, the `Enable` was not connected to the registers. Furthermore, the `Enable` was combinatorially(!) connected to the mux that selects the output. The mux should not be affected by the `Enable` at all (with the current Haskell simulation model).
  * `ddrIn`: VHDL: Remove bogus sensitivity list. In the initial version of this primitive, the version for asynchronous resets included the data input in its sensitivity lists for the registers. A later commit broke this, making it render the empty string for `~VARS`. But the input should not be in the sensitivity list anyway, so let's just remove the `~VARS` without actually changing the rendering in HDL.
    Additionally, the register outputting the data clocked in on the edge other than the active edge (`ddrIn_neg_latch`) also had its data input in its sensitivity list like the other two registers. This is also bogus and has been removed.
* `Clash.Xilinx.DDR`:
  * `oddr`: Verilog correctly picked one of the possible arguments to inspect for `ResetKind`, but the others picked the `KnownNat` constraint. This caused Clash to error out during HDL generation.
  * Both in and out prims: Renamed symbols that were incorrectly copy-pasted from the Intel prims.
* `Clash.Intel.DDR`:
  * `altddioIn`:
    * VHDL: Fix `~ISACTIVEENABLE` parameter reference that caused HDL generation to error out.
    * Verilog: Fix width parameter; it accidentally always rendered as 1.
    * SystemVerilog: Accidental comma generated invalid HDL.
  * `altddioOut`: VHDL: Several numeric identifiers were incorrect, referring to wrong arguments or symbols. HDL generation completed without error, but the resulting VHDL was completely broken.
  * Both prims: evaluate the `SSymbol` in Haskell to fix HDL generation warning about using an argument that is unused in Haskell.

<br>

* Add test benches for Xilinx DDR primitives
  A variant for Intel is also included, but we don't have a simulator in CI that supports Intel IP, so it can only be run manually for now. If the Verilog code is to be tested in Questa Intel FPGA Edition, one needs to define the preprocessor macro `INTEL_VERILOG`, which will skip verifying the output of the `Clash.Explicit.DDR.ddrOut` primitive as Questa trips on this: [`VendorDDR.hs` comment](https://github.com/clash-lang/clash-compiler/pull/2833/files#diff-cc19e30df0bd8ce972aaf795bf279b3ff015944190d5c5d0638f5613b6ea38c6R269-R294). SystemVerilog doesn't need this. Vivado is also fine with the Verilog code, so we actually test the Verilog variant of the vendor-neutral output DDR primitive in our nightlies now.
  The new test bench for vendor primitives tests the vendor-neutral primitives more extensively than the existing test bench did. For this reason, we also no longer run the existing test bench in Vivado, as it would not increase test coverage.
  The new test bench is very slow in generating HDL; this is due to some unfortunate GHC transformations that are hard on Clash and not easily fixed. Compensating a bit: Vivado runs all test bench configurations in parallel, saving a lot of time spent in Vivado.

<br>

* The primitives for Xilinx and Intel only support the rising edge as the active edge of the domain. This is now enforced by a constraint, instead of silently leading to a malfunctioning implementation.
* The existing constraints on the domains could be redefined in terms of `DomainPeriod`.     Since the virtual DDR domain only exists in simulation, it really doesn't matter if it doesn't match with the real domain in all its parameters, so we can drop those and make the constraints more intuitive to read.
* The primitives for Xilinx and Intel now support any data type that has a `BitPack` instance, instead of being restricted to `BitVector`.
* The Haddock for the functions is extended, explaining the purpose of the tuple elements with data.
* The internal functions for Xilinx and Intel are now exported (under the heading _Internal_) in the modules, as a general principle that we give people all the pieces so they can use them if they want to do something special.

I have written a variant for Clash 1.8 that does not change the API, but uses `clashCompileError` to prevent people instantiating the vendor primitives using a domain where the falling edge is the active edge. Once this PR has passed review, I'll open a new PR for that one (incorporating changes that are the result of review here).

Several of the bugs were introduced in 2019 in commit 0c38d6568 part of PR #527. We could not test the vendor primitives, so it went unnoticed. Now we at least test automatically in Vivado, and manually testing in Questa Intel FPGA Edition is straightforward.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Explicit/DDR.hs
+++ b/clash-prelude/src/Clash/Explicit/DDR.hs
@@ -50,8 +50,9 @@ import Clash.Signal.Internal
 >>> import Clash.Explicit.Prelude
 >>> import Clash.Explicit.DDR
 >>> :{
-instance KnownDomain "Fast" where
-  type KnownConf "Fast" = 'DomainConfiguration "Fast" 5000 'Rising 'Asynchronous 'Defined 'ActiveHigh
+type DDR = "DDR" :: Domain
+instance KnownDomain "DDR" where
+  type KnownConf "DDR" = 'DomainConfiguration "DDR" 5000 'Rising 'Asynchronous 'Defined 'ActiveHigh
   knownDomain = SDomainConfiguration SSymbol SNat SRising SAsynchronous SDefined SActiveHigh
 :}
 
@@ -62,24 +63,37 @@ instance KnownDomain "Fast" where
 -- Consumes a DDR input signal and produces a regular signal containing a pair
 -- of values.
 --
--- >>> printX $ sampleN 5 $ ddrIn systemClockGen systemResetGen enableGen (-1,-2,-3) (fromList [0..10] :: Signal "Fast" Int)
+-- Data is clocked in on both edges of the clock signal. We can discern the
+-- /active edge/ of the clock and the /other edge/. When the domain has the
+-- rising edge as the active edge (which is the most common), this means that
+-- the /rising/ edge is the /active/ edge and the /falling/ edge is the /other/
+-- edge.
+--
+-- Of the output pair @(o0, o1)@, @o0@ is the data clocked in on the /other/
+-- edge and @o1@ is the data clocked in on the /active/ edge, and @o0@ comes
+-- before @o1@ in time. With a domain where the rising edge is the active edge,
+-- this means @o0@ is clocked in on the falling clock edge and @o1@ is clocked
+-- in on the rising clock edge. For a domain with the falling edge as the active
+-- edge, this is the other way around, but @o0@ still comes before @o1@ in time.
+--
+-- >>> sampleN 5 $ ddrIn @Int @System @DDR clockGen resetGen enableGen (-1,-2,-3) (fromList [0..10])
 -- [(-1,-2),(-1,-2),(-3,2),(3,4),(5,6)]
 ddrIn
-  :: ( HasCallStack
-     , NFDataX a
-     , KnownConfiguration fast ('DomainConfiguration fast fPeriod edge reset init polarity)
-     , KnownConfiguration slow ('DomainConfiguration slow (2*fPeriod) edge reset init polarity) )
-  => Clock slow
-  -- ^ clock
-  -> Reset slow
-  -- ^ reset
-  -> Enable slow
+  :: forall a dom domDDR
+   . HasCallStack
+  => NFDataX a
+  => KnownDomain dom
+  => KnownDomain domDDR
+  => DomainPeriod dom ~ (2 * DomainPeriod domDDR)
+  => Clock dom
+  -> Reset dom
+  -> Enable dom
   -> (a, a, a)
-  -- ^ reset values
-  -> Signal fast a
+  -- ^ Reset values
+  -> Signal domDDR a
   -- ^ DDR input signal
-  -> Signal slow (a, a)
-  -- ^ normal speed output pairs
+  -> Signal dom (a, a)
+  -- ^ Normal speed output pair @(o0, o1)@
 ddrIn clk rst en (i0,i1,i2) =
   withFrozenCallStack $ ddrIn# clk rst en i0 i1 i2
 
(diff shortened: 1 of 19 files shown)


DigitalBrains1:
> Without Vivado, we can't actually test these primitives. 

Of course, while we didn't have Vivado back when the bug was introduced, we do now. Maybe it's even worth it to make a combined test with IDDR and ODDR prims, but just a "does this generate HDL" test is much more quickly written.

Anyway, the names in the VHDL are a bit weird:
```vhdl
        D1 => dataout_l(i),    -- 1-bit data input (positive edge)
        D2 => dataout_h(i),    -- 1-bit data input (negative edge)
```
Why is the data for the positive edge labeled L and the data for the negative edge labeled H? Note that `Clash.Explicit.DDR.ddrOut` labels them `data_Pos` and `data_Neg` respectively, which makes more sense.

christiaanb:
I think they're called like that due to copy/paste from the Intel primitives where the output ports of the Intel primivite are called `dataout_l` and `dataout_h`: https://github.com/clash-lang/clash-compiler/blob/5706eafca799ae04fda0ee7d666a40b6c0e7f22b/clash-lib/prims/vhdl/Clash_Intel_DDR.primitives.yaml#L43-L44

DigitalBrains1:
Shall I name them more obviously in this PR as well? I'm thinking `dataout_p` and `dataout_n`, after the edge that they are presented.

christiaanb (approved):
Yes, please clean up the names, and also it would be nice if we tested these primitives on Vivado at least in nightly builds. I mean... we already do that for the generic DDRout primitives: https://github.com/clash-lang/clash-compiler/blob/10f26ff16fec579545e0f2333e9fdd650bbac8e6/tests/Main.hs#L517-L526

DigitalBrains1:
Note to self: to do: run the nightly tests on a temporary branch to be absolutely sure they did not break. The changed test benches were run succesfully under a single configuration on my laptop, but CI will test a whole bunch on configurations.

tests/shouldwork/DDR/VendorDDR.hs
```diff
@@ -0,0 +1,335 @@
...
+    clkSlow clkSlow noReset (map fst expIn0) $
+      ignoreFor clkSlow noReset enableGen d2 0 vendorInOut1
+  done6 = outputVerifierWith (\clk rst -> assert clk rst "vendorInOut2")
+    clkSlow clkSlow noReset (map snd expIn0) vendorInOut2
+  doneFast = done1 `strictAnd` done2
+  doneSlow = done3 `strictAnd` done4 `strictAnd` done5 `strictAnd` done6
+  done = doneFast `strictAnd` unsafeSynchronizer clkSlow clkFast doneSlow
+  notDone = not <$> done
+  clk4 = tbClockGen $ unsafeSynchronizer clkFast clk4 notDone
+  -- {-# NOINLINE clk4 #-}
```
DigitalBrains1, line 313: Ah these can still be removed. They were part of an experiment to see if I could improve HDL generation times.

clash-prelude/src/Clash/Explicit/DDR.hs
```diff
@@ -62,46 +63,60 @@ instance KnownDomain "Fast" where
...
+-- edge.
+--
+-- Of the output pair @(o0, o1)@, @o0@ is the data clocked in on the /other/
+-- edge and @o1@ is the data clocked in on the /active/ edge, and @o0@ comes
+-- before @o1@ in time. With a domain where the rising edge is the active edge,
+-- this means @o0@ is clocked in on the falling clock edge and @o1@ is clocked
+-- in on the rising clock edge. For a domain with the falling edge as the active
+-- edge, this is the other way around, but @o0@ still comes before @o1@ in time.
+--
+-- >>> printX $ sampleN 5 $ ddrIn @Int @System @DDR clockGen resetGen enableGen (-1,-2,-3) (fromList [0..10])
```
DigitalBrains1, line 79: `printX` is unneeded

tests/shouldwork/DDR/VendorDDR.hs
```diff
@@ -0,0 +1,335 @@
...
+outAsyncReset es =
+  let
+    es0 = P.take 50 es P.++ P.replicate 7 0 P.++ P.drop 57 es
+  in P.take 11 es0 P.++ P.replicate 8 0 P.++ P.drop 19 es0
+
+genOutEnable :: [D] -> [D]
+genOutEnable es =
+  let
+    es0 = P.take 69 es P.++ P.take 8 (cycle [133, 233]) P.++ P.drop 77 es
+  in P.take 31 es0 P.++ P.take 8 (cycle [114, 214]) P.++ P.drop 39 es0
```
martijnbastiaan, line 169: Wow that's a lot of magic numbers! How did they come to be?

DigitalBrains1: Because I definitely did not feel like writing this in a comprehensible way which would be _way_ much more work! :-D

They come from looking at waveforms, understanding them, agreeing with them, and stuffing them in a list of expected values here.

This kind of code would be inexcusible in `src/` but we have more instances of magic numbers in `tests/`.

Also, don't bother running the Haskell simulation against them. I still need to file some bugs that our simulation doesn't agree with HDL simulation. Although I have to say the Xilinx DDR primitive has extremely surprising timing regarding resets and enables, I actually wonder whether _their_ simulation matches hardware.

DigitalBrains1: (also, note that 69 plus 8 equals 77 and 31 plus 8 equals 39, these do not actually change the length of the list of samples, they overwrite samples that were affected by a deasserted enable)

martijnbastiaan (approved):

PR #2835, opened 2024-10-29
Reduce `splitAt` to undefined in illegal contexts

### LINKED ISSUE #2831: Prelude.!!: index to large
The following

```haskell
f :: forall n. SNat n -> Unsigned 4
f n@SNat = case compareSNat n (SNat @15) of
  SNatLE -> at @n @(16 - n - 1) SNat vec
  SNatGT -> 0
 where
  vec :: Vec 16 (Unsigned 4)
  vec = repeat 0

topEntity :: HiddenClockResetEnable System => Signal System (Unsigned 4)
topEntity = pure $ f d17
```

results in

```
<no location info>: error:
    Clash error call:
    Prelude.!!: index too large
      error, called at libraries/ghc-internal/src/GHC/Internal/List.hs:1689:14 in ghc-internal:GHC.Internal.List
      tooLarge, called at libraries/ghc-internal/src/GHC/Internal/List.hs:1699:50 in ghc-internal:GHC.Internal.List
      !!, called at src-ghc/Clash/GHC/Evaluator/Primitive.hs:3833:61 in clash-ghc-1.9.0-inplace:Clash.GHC.Evaluator.Primitive
```

when compiling with `--verilog`.

christiaanb:
Previously, the Clash compiler would try to reduce

```
splitAt d1 Nil
```
to something of type
```
(Vec 1 a, Vec (0-1) a)
```
by trying to project the head and the tail out of the `Nil` constructor. This of course does not work resulting in an out-of-bounds indexing error reported in:

https://github.com/clash-lang/clash-compiler/issues/2831

The compiler now reduces above expressions to:

```
undefined :: (Vec 1 a, Vec (0-1) a)
```

Which is morally equivalent to the run-time exception Haskell evaluation would have thrown if the circuit description was evaluated like a regular Haskell program.

Fixes #2831

## Still TODO:

### DIFF
--- a/changelog/2024-10-29T14_54_54+01_00_fix2831
+++ b/changelog/2024-10-29T14_54_54+01_00_fix2831
@@ -1,0 +1,1 @@
+FIXED: Clash errors out when `Clash.Sized.Vector.splitAt` is compile-time evaluated in an illegal context
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
@@ -3827,12 +3827,23 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
          -- (x:xs) <- v
          m' | DC _ vArgs <- last args
             -- (x:fst (splitAt (m-1) xs),snd (splitAt (m-1) xs))
-            -> reduce $
-               mkApps (Data tupDc) $ (map Right tyArgs) ++
-                 [ Left (mkVecCons consCon aTy m' (Either.lefts vArgs !! 1)
-                           (splitAtSelR (Either.lefts vArgs !! 2) m1VecTy [lAlt]))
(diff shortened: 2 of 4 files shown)


rowanG077:
Perhaps reduce to `clashCompileError`? So that if it would latter end up in the circuit it would error out.

christiaanb:
> Perhaps reduce to `clashCompileError`? So that if it would latter end up in the circuit it would error out.

Why wouldn't we want to generate HDL for something like `splitAt d1 (unsafeCoerce Nil :: Vec 1 Bool)` ?

christiaanb:
I do not want the compiler to generate `clashCompileError`. Let's say the compiler misses some optimization opportunity, and the `clashCompileError` is not elided before we get to HDL generation. In that case we don't generate the HDL just because of some (performance) bug in the compiler. Anyone hitting that bug then has to wait for someone to fix it and use what's probably a development version of the compiler to continue.

If we just generate `undefined`, the downstream logic synthesis tools will very likely optimize the logic away and neither the functionality nor the size of the circuit are affected. Users can now just happily synthesize the generated HDL and not bother with having to wait for some fix and run a development version of the compiler.

PR #2843, opened 2024-11-15
Elide unsafeCoerce from promoted nats

kleinreact:
The PR elides the need of `unsafeCoerce` in `Clash.Promoted.Nat` via using the tools provided by `base` (>= 4.16.0.0, GHC >= 9.2) instead. ~Furthermore, it elides the need for `unsafeCoerce` in `Clash.Sized.Vector`~ (shipped via #2873 instead).

## Still TODO:

  - [x] Add a working fallback implementation for the old versions of GHC

### DIFF
--- a/clash-prelude/src/Clash/Promoted/Nat.hs
+++ b/clash-prelude/src/Clash/Promoted/Nat.hs
@@ -1,7 +1,7 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente,
                   2016     , Myrtle Software Ltd
-                  2022     , QBayLogic B.V.
+                  2022-2025, QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
@@ -70,10 +70,22 @@ module Clash.Promoted.Nat
   )
 where
 
+#if MIN_VERSION_base(4,16,0)
+import Data.Constraint    (Dict(..), (:-)(..))
+import Data.Constraint.Nat (euclideanNat)
+#endif
 import Data.Kind          (Type)
+#if MIN_VERSION_base(4,16,0)
+import Data.Type.Equality ((:~:)(..))
+import Data.Type.Ord      (OrderingI(..))
+#endif
 import GHC.Show           (appPrec)
 import GHC.TypeLits       (KnownNat, Nat, type (+), type (-), type (*),
-                           type (^), type (<=), natVal)
+                           type (^), type (<=),
+#if MIN_VERSION_base(4,16,0)
+                           cmpNat, sameNat,
+#endif
+                           natVal)
 import GHC.TypeLits.Extra (CLog, FLog, Div, Log, Mod, Min, Max)
 import GHC.Natural        (naturalFromInteger)
 import Language.Haskell.TH (appT, conT, litT, numTyLit, sigE)
@@ -82,7 +94,9 @@ import Language.Haskell.TH.Syntax (Lift (..))
 import Language.Haskell.TH.Compat
 #endif
 import Numeric.Natural    (Natural)
+#if !MIN_VERSION_base(4,16,0)
 import Unsafe.Coerce      (unsafeCoerce)
+#endif
 
 import Clash.Annotations.Primitive (hasBlackBox)
 import Clash.XException   (ShowX (..), showsPrecXWith)
@@ -184,11 +198,20 @@ instance KnownNat n => ShowX (UNat n) where
 --
 -- __NB__: Not synthesizable
 toUNat :: forall n . SNat n -> UNat n
+#if MIN_VERSION_base(4,16,0)
+toUNat p@SNat = case cmpNat (SNat @1) p of
+  LTI -> USucc (toUNat @(n - 1) (predSNat p))
+  EQI -> USucc UZero
+  GTI -> case sameNat p (SNat @0) of
+    Just Refl -> UZero
+    _ -> error "toUNat: impossible: 1 > n and n /= 0 for (n :: Nat)"
+#else
 toUNat p@SNat = fromI @n (snatToInteger p)
   where
     fromI :: forall m . Integer -> UNat m
     fromI 0 = unsafeCoerce @(UNat 0) @(UNat m) UZero
     fromI n = unsafeCoerce @(UNat ((m-1)+1)) @(UNat m) (USucc (fromI @(m-1) (n - 1)))
+#endif
 
 -- | Convert a unary-encoded natural number to its singleton representation
 --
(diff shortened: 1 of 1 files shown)


changelog/2024-11-16T11_59_31+01_00_avoid_using_unsafecoerce
```diff
@@ -0,0 +1,2 @@
+CHANGED: Several primitives in `Clash.Promoted.Nat` and `Clash.Sized.Vector` have been rephrased such that they avoid the usage of `unsafeCoerce`, where possible. The changes are only applied for `base >= 4.16.0.0` / `GHC >= 9.2`, as the use of `unsafeCoerce` cannot be avoided with lower versions, unfortunately.
```
DigitalBrains1, line 1: In what sense is this user-visible? For example, did something crash before that now works? If the users do not notice anything different at all, I think a Changelog entry is unnecessary. If something did change for them, perhaps the entry could be expanded to note what is changed.

DigitalBrains1: Also, I think _`base >= 4.16.0.0` / `GHC >= 9.2`_ is a tautology, I'd prefer it just says `GHC >= 9.2`.

clash-prelude/src/Clash/Promoted/Nat.hs
```diff
@@ -560,3 +611,15 @@ leToPlusKN
   -> r
 leToPlusKN r = r @(n - k)
 {-# INLINE leToPlusKN #-}
+
+-- | Change a function that has an argument with an @(n <= 0)@ constraint to a
+-- function with an argument that has an @(n ~ 0)@ constraint.
+leZeroToEqZero :: forall n r. (KnownNat n, n <= 0) => (n ~ 0 => r) -> r
```
rowanG077, line 617: As per discussion in chat I think this function should not exist currently due to the bug in the natnormalise plugin: https://github.com/clash-lang/ghc-typelits-natnormalise/issues/85

Essentially if you have `n <= 0` constraint you are on the way to mount doom of bugs currently. Having this function indicates that this will work at all which is the wrong message to send now. So I would prefer to keep this function out until that bug is fixed.

changelog/2024-11-16T11_59_31+01_00_avoid_using_unsafecoerce
kleinreact: I removed the changelog entry, as the PR now contains no user-visible changes any more.

clash-prelude/src/Clash/Promoted/Nat.hs
kleinreact: I removed the function again. Thanks for the feedback and the discussion.

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1687,18 +1686,24 @@ at n xs = head $ snd $ splitAt n xs
 -- 2 :> 4 :> 6 :> Nil
 -- >>> select d1 d2 d3 (1:>2:>3:>4:>5:>6:>7:>8:>Nil)
 -- 2 :> 4 :> 6 :> Nil
-select :: (CmpNat (i + s) (s * n) ~ 'GT)
+select :: forall i s n f a. s * n + 1 <= i + s
```
DigitalBrains1, line 1689: I think this constitutes an API change. The order of the tyvars is unchanged, but the constraint has changed. API changes should get a Changelog entry.

If we look at the origin of the uncommon constraint that this had, it's [this commit](https://github.com/clash-lang/clash-compiler/commit/461e14982a0c85bd22b5d9358fb67a3ec1ed8033). It seems to me the suggested change in this PR is an improvement that satisfies the exact same constraint, unless I'm missing some shortcoming in our solver plugins... however, the solver can't prove the equality of the old and new constraints, so users might have to rephrase the constraints in their designs.

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -1708,7 +1713,7 @@ select f s n xs = select' (toUNat n) $ drop f xs
 --
 -- >>> selectI d1 d2 (1:>2:>3:>4:>5:>6:>7:>8:>Nil) :: Vec 2 Int
 -- 2 :> 4 :> Nil
-selectI :: (CmpNat (i + s) (s * n) ~ 'GT, KnownNat n)
+selectI :: (1 <= s, s * n + 1 <= i + s, KnownNat n)
```
DigitalBrains1, line 1716: Same comment as for `select` above.

clash-prelude/src/Clash/Sized/Vector.hs
kleinreact: I added a corresponding changelog entry for it.

martijnbastiaan (approved):
I have no objection to this, though I do feel we're trading one thing (`unsafeCoerce`) for another (`error`) and I don't see a major benefit right away. Will this make your ~~format~~ formal work easier?

DigitalBrains1:
The only reason I hadn't fully approved this yet is I wondered how much test coverage there is and hadn't investigated that yet.

I'm not 100% sure more testing is needed other than "do these functions still return a sane result for a few numbers", but I wondered to what extent they were tested already.

Martijn, I don't think any `error` here is in reachable code? It's just there to satisfy the incomplete pattern checks. The advantage of this approach is that the code is type checked instead of relying on us as devs deciding for Haskell that, yes, our coercion is correct.

clash-prelude/src/Clash/Promoted/Nat.hs
```diff
@@ -404,14 +437,28 @@ showBNat = go []
...
 -- __NB__: Not synthesizable
-toBNat :: SNat n -> BNat n
+toBNat :: forall n. SNat n -> BNat n
+#if MIN_VERSION_base(4,16,0)
+toBNat s@SNat = case cmpNat (SNat @1) s of
+  LTI -> case euclideanNat @2 @n of
+    Sub Dict -> case sameNat (SNat @(n `Mod` 2)) (SNat @0) of
+      Just Refl -> B0 (toBNat (SNat @(n `Div` 2)))
+      Nothing -> case sameNat (SNat @(n `Mod` 2)) (SNat @1) of
+        Just Refl -> B1 (toBNat (SNat @(n `Div` 2)))
```
DigitalBrains1, line 447: Maybe you could share the `mod 2` and `div 2` binders between these assignments to halve the time the plugin will spend on these computations? (I'm assuming this isn't done automagically, which is an assumption without a lot of clout behind it, I'll admit it)

martijnbastiaan:
> Well, admittedly, the assertion that these errors are unreachable are in fact also us saying so instead of a checked fact. But the proof required that this is the case is more obvious than that the unsafeCoerce was correct. It's a bit halfway still.

Yeah, those were my thoughts. When looking at the "proofs" for both branches I feel like it's a toss-up which is easier to read/understand. TBF, the explicit error messages with reasoning why the code is unreachable is very nice!

> Do you mean formal with an ell instead of a tee?

Very much.

> I'm not 100% sure more testing is needed other than "do these functions still return a sane result for a few numbers", but I wondered to what extent they were tested already.

Agreed. It should be easy enough to write some tests for them, though a few doctests would suffice too. Given that there weren't any tests before (?) and the proofs seem simple enough I'm also inclined to not burden this PR with it though.. @kleinreact Would you be up for writing a few basic tests for this?

kleinreact:
> Well, admittedly, the assertion that these errors are unreachable are in fact also us saying so instead of a checked fact.

The major difference that I see here is that if an error would be reachable then the user at least gets some feedback due to the error raised, while with an error and `unsafeCoerce` it will just silently break the integrity of the code.

> @kleinreact Would you be up for writing a few basic tests for this?

I don't see any benefit which would legitimate the extra effort of adding such test cases. Everything is basically happening on the type level, so GHC already takes care quite much. Note that this is also a difference against the prior state of using `unsafeCoerce`.

DigitalBrains1:
We have some coverage... [these doctests](https://github.com/clash-lang/clash-compiler/blob/6a26f6b912c50fd2011d06d0ddf6558cc220946d/clash-prelude/src/Clash/Sized/RTree.hs#L437-L440) test `toUNat` and so do several in `Clash.Sized.Vector`.

But I don't think any errors in `compareSNat` and `toBNat` would be caught.

I prefer code to have a test, even if it's a basic one. I feel the original code needed some tests already. But since no test was already the status quo, this PR feels like an improvement.

martijnbastiaan:
Agreed @DigitalBrains1

PR #2844, opened 2024-11-18
Do not inline let-bound recursive calls

christiaanb:
This prevents `recToLetRec` from turning a globally recursive function into a local let-bound recursive function which can be synthesized to hardware.

Fixes #2839

## Still TODO:

  - [x] Write a changelog entry (see changelog/README.md)
  - [x] Check copyright notices are up to date in edited files

Issue #2570, opened 2023-08-27
Clock blackbox primitive is duplicated


rowanG077:
Consider the following `TopEntity`:

```Haskell
{-# LANGUAGE QuasiQuotes #-}

module TopEntity where

import qualified Clash.Prelude as P
import Clash.Explicit.Prelude
import Clash.Annotations.Primitive
import Data.String.Interpolate (i)

oscg :: Clock System
oscg = clockGen
{-# ANN oscg (InlineYamlPrimitive [Verilog] [i|
  BlackBox:
    name: TopEntity.oscg
    kind: Declaration
    template: |-
      // OSCG begin
      OSCG \#( .DIV(6) ) ~GENSYM[oscillator][1] ( .OSC(~RESULT) );
      // OSCG end
  |]) #-}
{-# NOINLINE oscg #-}

topEntity = unsafeFromActiveHigh s1
  where
    clk = oscg

    f = P.exposeClockResetEnable (P.register False) clk noReset enableGen
    s0 = f $ dflipflop clk (pure True)
    s1 = f s0
```

And running `clash TopEntity.hs --verilog` we see that `OSCG` primitive is duplicated 3 times:

```Verilog
/* AUTOMATICALLY GENERATED VERILOG-2001 SOURCE CODE.
** GENERATED BY CLASH 1.7.0. DO NOT MODIFY.
*/
`default_nettype none
`timescale 100fs/100fs
module topEntity
    ( // No inputs

      // Outputs
      output wire  result // reset
    );
  // TopEntity.hs:27:5
  wire  rst;
  reg  result_1 = 1'b0;
  wire  c$app_arg;
  // TopEntity.hs:27:5
  wire  rst_0;
  reg  result_2 = 1'b0;
  wire  c$app_arg_0;
  reg  c$app_arg_1 = ({1 {1'bx}});
  wire  c$app_arg_2;

  assign rst = 1'b0;

  // register begin
  always @(posedge c$app_arg or  posedge  rst) begin : result_1_register
    if ( rst) begin
      result_1 <= 1'b0;
    end else begin
      result_1 <= result_2;
    end
  end
  // register end

  // OSCG begin
  OSCG #( .DIV(6) ) oscillator ( .OSC(c$app_arg) );
  // OSCG end

  assign rst_0 = 1'b0;

  // register begin
  always @(posedge c$app_arg_0 or  posedge  rst_0) begin : result_2_register
    if ( rst_0) begin
[...]

PR #2846, opened 2024-11-18
No inlining let-bound global vars with clock types

### LINKED ISSUE #2845: Clash generates invalid verilog names
Tracking the invalid name generation in a separate issue

> I think I have two more reproducers that lead to heaps of duplication, with a clock primitive being involved. I'm using GHC 9.8.2. They generate funky names as well: names like `Duplicate.topEntity_cntr` as well as `\Duplicate.topEntity_cntr`. The latter might technically be valid Verilog, but it seems Vivado isn't too happy about it anyway.

The first variant is without `NOINLINE cntr`:

```haskell
module Duplicate where

import Clash.Explicit.Prelude

topEntity ::
  Signal System (Unsigned 8)
topEntity = cntr + x
 where
  cntr = register clk noReset enableGen 0 0
  x = register clk noReset enableGen 100 0
  done = (== 100) <$> cntr
  clk = tbClockGen $ not <$> done
```

This generates the following Verilog:
<details><summary>Duplicate.hs Verilog</summary>

```verilog
** GENERATED BY CLASH 1.9.0. DO NOT MODIFY.
*/
`default_nettype none
`timescale 100fs/100fs
module topEntity
    ( // No inputs

      // Outputs
      output wire [7:0] result
    );
  reg [7:0] c$app_arg = 8'd100;
  // Duplicate.hs:10:3-6
  reg [7:0] \Duplicate.topEntity_cntr  = 8'd0;
  // Duplicate.hs:13:3-5
  wire  \Duplicate.topEntity_clk ;
  // Duplicate.hs:10:3-6
  reg [7:0] Duplicate.topEntity_cntr = 8'd0;
  // Duplicate.hs:13:3-5
  wire  Duplicate.topEntity_clk;
  wire  c$rst;
  wire  c$rst_0;
  wire  c$rst_1;

  assign c$rst = (1'b0);

  // register begin
  always @(posedge Duplicate.topEntity_clk or  posedge  c$rst) begin : DuplicatetopEntity_cntr_register
    if ( c$rst) begin
      Duplicate.topEntity_cntr <= 8'd0;
    end else begin
      Duplicate.topEntity_cntr <= 8'd0;
    end
  end
  // register end

  // tbClockGen begin
  // pragma translate_off
  reg  clk;
  // 1 = 0.1ps
  localparam half_period = (100000 / 2);
  always begin
    // Delay of 1 mitigates race conditions (https://github.com/steveicarus/iverilog/issues/160)
    #1 clk =  0 ;
    #1000000 forever begin

      if (~ (~ (Duplicate.topEntity_cntr == 8'd100))) begin
        `ifdef VERILATOR
          $c("std::exit(0);");
        `endif
        $finish(0);
      end

      clk = ~ clk;
      #half_period;
      clk = ~ clk;
      #half_period;
    end
  end

  assign Duplicate.topEntity_clk = clk;
  // pragma translate_on
  // tbClockGen end

  assign c$rst_0 = (1'b0);

  // register begin
  always @(posedge Duplicate.topEntity_clk or  posedge  c$rst_0) begin : c$app_arg_register
    if ( c$rst_0) begin
      c$app_arg <= 8'd100;
    end else begin
      c$app_arg <= 8'd0;
    end
  end
  // register end

  assign c$rst_1 = (1'b0);

  // register begin
  always @(posedge \Duplicate.topEntity_clk  or  posedge  c$rst_1) begin : DuplicatetopEntity_cntr_register_0
    if ( c$rst_1) begin
      \Duplicate.topEntity_cntr  <= 8'd0;
    end else begin
      \Duplicate.topEntity_cntr  <= 8'd0;
    end
  end
  // register end

  // tbClockGen begin
  // pragma translate_off
  reg  clk_0;
  // 1 = 0.1ps
  localparam half_period_0 = (100000 / 2);
  always begin
    // Delay of 1 mitigates race conditions (https://github.com/steveicarus/iverilog/issues/160)
    #1 clk_0 =  0 ;
    #1000000 forever begin

      if (~ (~ (\Duplicate.topEntity_cntr  == 8'd100))) begin
        `ifdef VERILATOR
          $c("std::exit(0);");
        `endif
        $finish(0);
      end

      clk_0 = ~ clk_0;
      #half_period_0;
      clk_0 = ~ clk_0;
      #half_period_0;
    end
  end

  assign \Duplicate.topEntity_clk  = clk_0;
  // pragma translate_on
  // tbClockGen end

  assign result = \Duplicate.topEntity_cntr  + c$app_arg;


endmodule
```
</details>

Al least the following things are duplicated:
```verilog
  reg [7:0] \Duplicate.topEntity_cntr  = 8'd0;
  wire  \Duplicate.topEntity_clk ;
  reg [7:0] Duplicate.topEntity_cntr = 8'd0;
  wire  Duplicate.topEntity_clk;

  assign Duplicate.topEntity_clk = clk;

  assign \Duplicate.topEntity_clk  = clk_0;
```
along with 2 instead of 1 clock primitive, and 2 register blocks for `cntr` instead of 1.

[edit]
I noticed I could get rid of the second clock and still have the duplication. The one with the second clock might still be interesting because of other weirdness, such as the `done` signal in HDL being `not <$> done` in Haskell?
```verilog
  // tbClockGen begin
  [...]
      if (~ Duplicate.topEntity_done) begin
        `ifdef VERILATOR
          $c("std::exit(0);");
        `endif
        $finish(0);
      end

  assign Duplicate.topEntity_done = ~ (Duplicate.topEntity_cntr == 8'd100);
```
<details><summary>This is output generated by the following code:</summary>

```haskell
module Duplicate where

import Clash.Explicit.Prelude

topEntity ::
  Signal System (Unsigned 8)
topEntity = cntr + posD
 where
  cntr = register clk1 noReset enableGen 0 0
  posD = register clk2 noReset enableGen 0 0
  done = (== 100) <$> cntr
  (clk1, clk2) = biTbClockGen $ not <$> done
```
</details>
[/edit]

_Originally posted by @DigitalBrains1 in https://github.com/clash-lang/clash-compiler/issues/2570#issuecomment-2451635677_

christiaanb:
The global vars are usually backed by a clock generator that are not work-free.

In addition, when these global vars are recursively defined, they can mess up the post-normalization flattening stage which then violates certain invariants of the netlist generation stage. This then causes the netlist generation stage to generate bad Verilog names.

Fixes #2845

## Still TODO:

### DIFF
--- a/changelog/2024-11-18T14_59_34+01_00_fix2845
+++ b/changelog/2024-11-18T14_59_34+01_00_fix2845
@@ -1,0 +1,1 @@
+FIXED: Clash generates illegal Verilog names [#2845](https://github.com/clash-lang/clash-compiler/issues/2845)
--- a/clash-lib/src/Clash/Rewrite/WorkFree.hs
+++ b/clash-lib/src/Clash/Rewrite/WorkFree.hs
@@ -155,7 +155,18 @@ isWorkFreeClockOrResetOrEnable tcm e =
   if isClockOrReset tcm eTy || isEnable tcm eTy then
     case collectArgs e of
       (Prim p,_) -> Just (primName p == Text.showt 'removedArg)
-      (Var _, []) -> Just True
+      -- Only local variables with a clock type are work-free. When it is a global
+      -- variable, it is probably backed by a clock generator, which is definitely
+      -- not work-free.
+      --
+      -- Inlining let-bindings referencing a global variable with a clock type
+      -- can sometimes lead to the post-normalization flattening stage to generate
+      -- code that violates the invariants of the netlist generation stage.
+      -- Especially when this global binder is defined recursively such as when
+      -- using `tbClockGen`.
+      -- This then ultimately leads to bad verilog names being generated as
+      -- reported in: https://github.com/clash-lang/clash-compiler/issues/2845
+      (Var v, []) -> Just (isLocalId v)
       (Data _, [_dom, Left (stripTicks -> Data _)]) -> Just True -- For Enable True/False
       (Literal _,_) -> Just True
       _ -> Just False
--- a/changelog/2024-11-18T13_43_58+01_00_fix2839
+++ b/changelog/2024-11-18T13_43_58+01_00_fix2839
@@ -1,0 +1,1 @@
+FIXED: Clash errored saying it cannot translate a globally recursive function in code that originally only contains let-bound (local) recursion [#2839](https://github.com/clash-lang/clash-compiler/issues/2839).
--- a/clash-lib/src/Clash/Normalize/Transformations/Inline.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Inline.hs
@@ -120,7 +120,14 @@ bindConstantVar = inlineBinders test
       True -> return (i `notElemFreeVars` e)
       _    -> do
         tcm <- Lens.view tcCache
-        case isWorkFreeIsh tcm e of
+        (fn,_) <- Lens.use curFun
+        -- Don't inline things that perform work, it increases the circuit size.
+        --
+        -- Also don't inline globally recursive calls, it prevents the
+        -- recToLetRec transformation from transforming global recursion to
+        -- local recursion.
+        -- See https://github.com/clash-lang/clash-compiler/issues/2839
+        case isWorkFreeIsh tcm e && not (e == Var fn) of
           True -> Lens.view inlineConstantLimit >>= \case
             0 -> return True
             n -> return (termSize e <= n)
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -635,6 +635,8 @@ runClashTest = defaultMain $ clashTestRoot
         , runTest "T2781" def{hdlLoad=[],hdlSim=[],hdlTargets=[VHDL]}
         , runTest "T2628" def{hdlTargets=[VHDL], buildTargets=BuildSpecific ["TACacheServerStep"], hdlSim=[]}
         , runTest "T2831" def{hdlLoad=[],hdlSim=[],hdlTargets=[VHDL]}
+        , runTest "T2839" def{hdlLoad=[],hdlSim=[],hdlTargets=[VHDL]}
+        , runTest "T2845" def{hdlSim=[],hdlTargets=[Verilog]}
         ] <>
         if compiledWith == Cabal then
           -- This tests fails without environment files present, which are only
--- a/tests/shouldwork/Issues/T2839.hs
+++ b/tests/shouldwork/Issues/T2839.hs
@@ -1,0 +1,12 @@
+module T2839 where
+
+import Clash.Explicit.Prelude
+import Clash.Explicit.Testbench
+
+topEntity ::
+  Signal System (Unsigned 8)
+topEntity = register clk noReset enableGen 100 0
+ where
+  cntr = register clk noReset enableGen (0 :: Unsigned 8) 0
+  done = (== 100) <$> cntr
+  clk = tbClockGen $ not <$> done
--- a/tests/shouldwork/Issues/T2845.hs
+++ b/tests/shouldwork/Issues/T2845.hs
@@ -1,0 +1,13 @@
+module T2845 where
+
+import Clash.Explicit.Prelude
+import Clash.Explicit.Testbench
+
+topEntity ::
+  Signal System (Unsigned 8)
+topEntity = cntr + x
+ where
+  cntr = register clk noReset enableGen 0 0
+  x = register clk noReset enableGen 100 0
+  done = (== 100) <$> cntr
+  clk = tbClockGen $ not <$> done


changelog/2024-11-18T14_59_34+01_00_fix2845
```diff
@@ -0,0 +1 @@
+FIXED: Clash generates illegal Verilog names
```
martijnbastiaan, line 1: Could you mention the issue here? That way it will get picked up by https://github.com/clash-lang/clash-compiler/blob/master/changelog/comment-gh.py.

christiaanb: Ah, will do.

DigitalBrains1 (approved):
I think we should backport this to 1.8?

Furthermore, it wouldn't surprise me if the reproducer in `clash-testsuite` that is added here no longer reproduces the issue once issue #2570 is truly fixed. This is just something I realised and wanted to point out.

clash-lib/src/Clash/Rewrite/WorkFree.hs
```diff
@@ -155,7 +155,18 @@ isWorkFreeClockOrResetOrEnable tcm e =
   if isClockOrReset tcm eTy || isEnable tcm eTy then
     case collectArgs e of
       (Prim p,_) -> Just (primName p == Text.showt 'removedArg)
-      (Var _, []) -> Just True
+      -- Only local variables with a clock type are work-free. When it is a global
+      -- variable, it is probably backed by a clock generator, which is definitely
+      -- not work-free.
```
leonschoorl, line 160: The local variable distinction feel arbitrary to me.

Given:
```haskell
clk = clockGen

clkGlobal = clk

f =
 let
   clkLocal = clk
   clkA = clkLocal
   clkB = clkGlobal
 in [...]
```

As I understand it, you're saying here it is ok for bindConstantVar to replace `clkA` inside of `[...]` with `clkLocal`, but not `clkB` with `clkGlobal`.
Is that what you're saying?
And if so, why?
They seem equivalent to me.

clash-lib/src/Clash/Rewrite/WorkFree.hs
```diff
@@ -155,7 +155,18 @@ isWorkFreeClockOrResetOrEnable tcm e =
...
+      --
+      -- Inlining let-bindings referencing a global variable with a clock type
+      -- can sometimes lead to the post-normalization flattening stage to generate
+      -- code that violates the invariants of the netlist generation stage.
```
leonschoorl, line 164: Do you remember which invariants?
And are these invariants actually documented anywhere?

leonschoorl (commented):
Are you aware that `bindConstantVar` already checks for (vars with) local ids?
https://github.com/clash-lang/clash-compiler/blob/10f26ff16fec579545e0f2333e9fdd650bbac8e6/clash-lib/src/Clash/Normalize/Transformations/Inline.hs#L118-L120

I think that means that your `(isLocalId v)` will be `False`, except possibly when we recurse inside `isWorkFreeIsh`.

clash-lib/src/Clash/Rewrite/WorkFree.hs
christiaanb: Yes, that's what I'm saying. Inlining `clkA` everywhere will not make the circuit `f` any larger (perform work). Inlining `clkB` everywhere will make the circuit `f` larger, because it will have duplicated calls to what will ultimately be `clockGen`.

clash-lib/src/Clash/Rewrite/WorkFree.hs
christiaanb: We cannot have let-expressions appear in the argument position of an application.

christiaanb:
> Are you aware that `bindConstantVar` already checks for (vars with) local ids?
> 
> https://github.com/clash-lang/clash-compiler/blob/10f26ff16fec579545e0f2333e9fdd650bbac8e6/clash-lib/src/Clash/Normalize/Transformations/Inline.hs#L118-L120
> 
> I think that means that your `(isLocalId v)` will be `False`, except possibly when we recurse inside `isWorkFreeIsh`.

I am aware. But prior to this commit/PR, the compiler would inline let-bindings binding global variables with a clock type, which is bad because it duplicates work/increases the size of the circuit..

christiaanb:
I think @leonschoorl comments correctly identify a missed CSE oppertunity in that the current implementation will, for:
```haskell
clk = clockGen

clkGlobal = clk

f =
 let
   clkLocal = clk
   clkA = clkLocal
   clkB = clkGlobal
 in [...]
```
not transform `f` such that ultimately there is only one application/call of the `clockGen` primitive. While it is most likely desirable in increase sharing in this case, that has never been a guarantee/must of the Clash compiler. That's because CSE is not always desirable as it increases fanout. Decreasing sharing should however be considered a bug, which this PR actually fixes.

That being said, once this PR is merged, we should probably open an "enhancement" issue that Clash is missing a desirable CSE opportunity.

DigitalBrains1:
I'd like to call your attention to the point that there can ever only be one clock primitive in one domain.¹ Two clock primitives might generate the same frequency, but they are probably not perfectly phase-aligned, making them separate clock domains. So with constructions like `clkGlobal = clk`, you have to wonder: is this ever actually going to occur in a valid design? Or is this already user error since they are now putting things in the same domain even though they are not synchronised to the same clock? If I want stuff to be in the same clock domain, I pass them all the exact same bind of the clock variable, not `clk` to one and `clkGlobal` to another.

¹ The other way around is fine: you could have one clock primitive which outputs multiple clocks.

DigitalBrains1:
Mergify was not opening a backport PR so I tried to delete and re-add the label. Didn't work; but I remembered you can look at Actions for Mergify stuff. Mergify reports:

```
error: commit 6f9b7300da535bae01da67eeac99dd82f2cac965 is a merge but no -m option was given.
fatal: cherry-pick failed
```

and boy _something has gone wrong_ with this PR. Just admire this nice shape:

```
$ git log --graph --oneline --no-show-signature origin/master
* 8a827e1e2 (origin/master) GitLab CI: Use Stackage Docker image (#2805)
*   47b8fa7c8 Merge pull request #2846 from clash-lang/fix_2845
|\  
| *   6f9b7300d Merge pull request #2844 from clash-lang/fix_2839
| |\  
* | | 8cded4cbe Move block RAM reset value into the `ClearOnReset` constructor of `ResetStrategy`, (#2849)
* | |   25a7bc18f Merge pull request #2844 from clash-lang/fix_2839
|\ \ \  
| |/ /  
|/| /   
| |/    
| * 00722413f Do not inline let-bound recursive calls
* | f13b8534e Add `Distributive` and `Representable` instances to `Vec n` (#2829)
|/  
* 10f26ff16 (gronau/master) Reduce `splitAt` to undefined in illegal contexts (#2837 
```

I don't know whatever happened here, but yeah, I don't blame Mergify not being able to backport this... D-:

DigitalBrains1:
Commit [b13b96f](https://github.com/clash-lang/clash-compiler/commit/b13b96fa791cd95ebe4b8b34035da8d983381d42) still looks fine, but then it got force-pushed to [6f9b730](https://github.com/clash-lang/clash-compiler/commit/6f9b7300da535bae01da67eeac99dd82f2cac965) and that is a merge commit with contents. It merges `master` and the commit from #2844, but its tree contains the changes of this PR instead of just being the combination of its parent commits.

christiaanb:
I have no idea what happened here. My regular way of working is:
- `git pull —rebase origin/master`
And then after solving conflicts do a 
- `git push -f`
I guess I should have done a
- `git fetch origin`
Before the rebase?

DigitalBrains1:
`pull` should already include a `fetch`.

I also don't know what went wrong, and we actually lost your commit message of this PR in the process. I'm currently handcrafting a backport PR.

DigitalBrains1:
The intended commit message of this PR is identical to [the PR cover letter](https://github.com/clash-lang/clash-compiler/pull/2846#issue-2668676656).

PR #2849, opened 2024-11-25
Move block RAM reset value into`ClearOnReset` 

gergoerdi:
_As seen on ~TV~ the Clash Stack_.

### DIFF
--- a/changelog/2024-11-25T20_38_27+08_00_reset_strategy
+++ b/changelog/2024-11-25T20_38_27+08_00_reset_strategy
@@ -1,0 +1,1 @@
+CHANGED: `ResetStrategy` now contains the reset function, to avoid dummy arguments on `NoClearOnReset`
--- a/clash-prelude/src/Clash/Explicit/BlockRam.hs
+++ b/clash-prelude/src/Clash/Explicit/BlockRam.hs
@@ -854,9 +854,9 @@ blockRamPow2 = \clk en cnt rd wrM -> withFrozenCallStack
   (blockRam clk en cnt rd wrM)
 {-# INLINE blockRamPow2 #-}
 
-data ResetStrategy (r :: Bool) where
-  ClearOnReset :: ResetStrategy 'True
-  NoClearOnReset :: ResetStrategy 'False
+data ResetStrategy (r :: Bool) a where
+  ClearOnReset :: a -> ResetStrategy 'True a
+  NoClearOnReset :: ResetStrategy 'False a
 
 -- | A version of 'blockRam' that has no default values set. May be cleared to
 -- an arbitrary state using a reset function.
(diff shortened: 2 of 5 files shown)


clash-prelude/src/Clash/Explicit/BlockRam.hs
```diff
@@ -968,7 +966,7 @@ blockRam1
   -- for the BRAM to be reset to its initial state.
   -> Enable dom
   -- ^ 'Enable' line
-  -> ResetStrategy r
+  -> ResetStrategy r a
```
rowanG077, line 969: To keep the same behavior as is done elsewhere in clash keep the initial value the same as the reset value. Which means the `a` argument to this function disappears completely. Or keep it and use `ResetStrategy r ()` whatever you prefer.

DigitalBrains1: This function is a bit different. The initial value is undefined, but you can specify a function which will be used to fill the RAM with data when you assert reset long enough (this is an actual circuit running, filling one data position every cycle, so you need to have a running clock and keep reset asserted while the clock ticks for as many cycles as the RAM is large).

This behaviour isn't changed by this PR; it just moves the argument that holds the function to populate the RAM if you assert reset.

DigitalBrains1: I don't think we have a block RAM that has both an initial value and a reset value like you propose... we have block RAMs with initial values, but those are lost if you write to them, you can't restore the initial value on reset. If you needed both an initial value and a reset value in an FPGA, you're pretty much going to need two block RAMs: one as a ROM to restore the initial value from, and one that is your actual writable block RAM.

It's quite different from a set or clear input to a register where the bitstream just selects one of those to use as initial/reset.

rowanG077: I don't think the initial value is undefined. This is the type of `blockRam1`

```- | A version of 'blockRam' that is initialized with the same value on all
-- memory positions
blockRam1
   :: forall n dom a r addr
   . ( KnownDomain dom
     , HasCallStack
     , NFDataX a
     , Enum addr
     , NFDataX addr
     , 1 <= n )
  => Clock dom
  -- ^ 'Clock' to synchronize to
  -> Reset dom
  -- ^ 'Reset' line. This needs to be asserted for at least /n/ cycles in order
  -- for the BRAM to be reset to its initial state.
  -> Enable dom
  -- ^ 'Enable' line
  -> ResetStrategy r a
  -- ^ Whether to clear BRAM on asserted reset ('ClearOnReset') or
  -- not ('NoClearOnReset'). The reset needs to be asserted for at least /n/
  -- cycles to clear the BRAM.
  -> SNat n
  -- ^ Number of elements in BRAM
  -> a
  -- ^ Initial content of the BRAM (replicated /n/ times)
  -> Signal dom addr
  -- ^ Read address @r@
  -> Signal dom (Maybe (addr, a))
  -- ^ (write address @w@, value to write)
  -> Signal dom a
  -- ^ Value of the BRAM at address @r@ from the previous clock cycle
```

So it has an initial value `a` and a resetValue `a` inside `ResetStrategy`. If you look at the actual current implementation:

```
    ClearOnReset a' ->
      -- Use reset infrastructure
      blockRam1# clk en n a rd1 we1 wa1 w1
      where
        rd1 = mux rstBool 0 (fromEnum <$> rd0)
        we1 = mux rstBool (pure True) we0
        wa1 = mux rstBool (fromInteger . toInteger <$> waCounter) (fromEnum <$> wa0)
        w1  = mux rstBool (pure a') w0
```


It uses the original `a` as the initial repeated contents of the blockram. And then writes the reset value into on reset via `w1`.

DigitalBrains1: Well I was talking about `blockRamU`. I now see you commented on `blockRam1`; I had misinterpreted the collapsed diff and not noticed that `blockRam1` changed at all.

I agree that it doesn't make sense for `blockRam1`.

DigitalBrains1: In fact, I don't see why `blockRam1` needs a `ResetStrategy`. If you don't want a reset to do anything, just don't assert the reset?

Initially I thought "well maybe you want the pipeline registers cleared but not the contents", but this code doesn't reset the pipeline registers _ever_?

DigitalBrains1: It's just that for the version with implicit resets, it's easier to pass `NoClearOnReset` than get rid of the implicit reset...

rowanG077: (I only noticed this because I had a hunch and checked the full code because you can't tell this without the code around it)

The perhaps unsatisfying reason for the existence `ResetStrategy` is that many people using clash will simply not thing about and.

Now that I'm reading the code I think the reset is unsafe anyway if the reset is async:

```
  rstInv = invertReset rst0

  waCounter :: Signal dom (Index n)
  waCounter = register clk rstInv en 0 (satSucc SatBound <$> waCounter)
```

Even if `rst0` is synchronously de-asserted. `rstInv` is not. In practice you are probably not running at such a high speed and PnR will put them reasonably close. But in theory this is a ticking time bomb if you do rely on refreshed contents on reset.

rowanG077: Anyway that's all out of scope of this PR. For now I would say just use `ResetStrategy r ()`.

changelog/2024-11-25T20_38_27+08_00_reset_strategy
```diff
@@ -0,0 +1,2 @@
+CHANGED: `ResetStrategy` now contains the reset function/value, to avoid
+dummy arguments on `NoClearOnReset`
```
DigitalBrains1, line 2:
```suggestion
CHANGED: `ResetStrategy` now contains the reset function, to avoid dummy arguments on `NoClearOnReset`
```

Since we now do `()` on `blockRam1`, I feel the message should omit _value_.

Incidentally, it wasn't something that would block approval and I would have happily merged it as-is, but ideally a changelog message is one line per entry (since this is one entry, it means all on one line) and should have a final newline. Since building the Changelog is still a manual process, it's all not super strict, but it's nice to have to reduce effort in building the real Changelog.

PR #2858, opened 2024-12-30
Add Finite class

kleinreact:
The PR adds the `Finite` class as well as supplemental instances for most of the standard types.

`Finite` is a class for types with only finitely many inhabitants and can be considered a more hardware-friendly alternative to `Bounded` and `Enum`, utilizing `Index` instead of `Int` and vectors instead of lists. 

Have a look at the haddock documentation for further insights.

## Requires

  - #2871 

## ~~Still TODO:~~

### DIFF
--- a/clash-prelude/src/Clash/Class/Finite.hs
+++ b/clash-prelude/src/Clash/Class/Finite.hs
@@ -1,0 +1,43 @@
+{-|
+Copyright  :  (C) 2024-2025, QBayLogic B.V.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+The class of types holding only a finite number of elements. The
+'Finite' class offers type level access to the number of elements @n@
+and defines a total order on the elements via indexing them from @0@
+to @n-1@. Therewith, it gives access to the vector of all inhabitants
+of the type and allows to iterate over them in order or to map them back
+and forth between their associated indices.
+
+The class can be considered as a more hardware-friendly alternative to
+'Bounded' and 'Enum', utilizing 'Clash.Sized.Index.Index' instead of
+'Int' and vectors instead of lists.
+
+In comparison, 'Finite' is well suited for types holding finitely many
+elements, while the 'Enum' class is better suited for types with
+infinitely many inhabitants. The type of @'succ', 'pred' :: 'Enum' a
+=> a -> a@ clearly reflects this design choice, as it assumes that
+every element has a successor and predecessor, which makes perfect
+sense for infinite types, but requires finite types to error on
+certain inputs. Wrapping behavior is forbidden according to the
+documentation of 'Enum' (assuming that finite types usually have a
+'Bounded' instance) such that 'Enum' instances must ship partial
+functions for most finite types. Likewise, the number of inhabitants
+does not align with the number of indices offered by 'Int' for most
+types, which 'Finite' resolves by using @'Clash.Sized.Index.Index' n@
+instead.
+-}
+module Clash.Class.Finite
+  ( -- * Finite Class
+    Finite(..)
+    -- * Extensions
+  , GenericReverse(..)
+  , WithUndefined(..)
+    -- * Deriving Helpers
+  , FiniteDerive(..)
+  , BoundedEnumEq(..)
+  )
+where
+
+import Clash.Class.Finite.Internal
--- a/clash-prelude/src/Clash/Class/Finite/Internal.hs
+++ b/clash-prelude/src/Clash/Class/Finite/Internal.hs
@@ -1,0 +1,1009 @@
+{-|
+Copyright  :  (C) 2024-2025, QBayLogic B.V.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+
+{-# LANGUAGE ConstrainedClassMethods #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE EmptyCase #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise #-}
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.KnownNat.Solver #-}
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.Extra.Solver #-}
+
+{-# OPTIONS_HADDOCK not-home #-}
+
+module Clash.Class.Finite.Internal
+  ( Finite(..)
+  , GFinite(..)
+  , BoundedEnumEq(..)
+  , FiniteDerive(..)
+  , GenericReverse(..)
+  , WithUndefined(..)
+  )
+where
+
+import Prelude hiding ((++), (!!), concatMap, foldl, foldr, repeat, reverse)
+
+import Control.Applicative (Alternative(..))
+import Control.Arrow (second)
+import Data.Bits (Bits(..), FiniteBits(..))
+import Data.Coerce (coerce)
+import Data.Constraint (Dict(..))
+import Data.Functor.Compose (Compose(..))
+import Data.Functor.Const (Const(..))
+import Data.Functor.Identity (Identity(..))
+import Data.Functor.Product (Product)
+import Data.Functor.Sum (Sum)
+import Data.Int (Int8, Int16, Int32, Int64)
+import Data.Kind (Type)
+import Data.Maybe (fromJust)
+#if MIN_VERSION_base(4,15,0)
+import Data.Ord (Down(..))
+#endif
+import Data.Proxy (Proxy(..))
+import Data.Singletons (Apply, TyFun)
+import Data.Type.Bool (If)
+import Data.Type.Equality ((:~:)(..), (:~~:)(..), type (~~))
+import Data.Void (Void)
+import Data.Word (Word8, Word16, Word32, Word64)
+import GHC.Generics
+  (Generic(..), Rep, V1, U1(..), M1(..), K1(..), (:+:)(..), (:*:)(..))
+import GHC.TypeNats
+  ( Nat, KnownNat
+  , type (^), type (<=), type (<=?), type (*), type (+), type (-)
+  )
+import GHC.TypeLits (KnownSymbol)
+
+import Clash.Class.Num (SaturatingNum(..), SaturationMode(..))
+import Clash.Class.Finite.Internal.Dictionaries
+  ( powPositiveIfPositiveBase, powPositiveImpliesPositiveBase
+  , mulPositiveImpliesPositiveOperands, zeroLeAdd, powMonotone1
+  , powLawsRewrite
+#if !MIN_VERSION_base(4,16,0)
+  , pow2CLogDual, leqOnePlusMinus
+#endif
+  )
+import Clash.Class.Finite.Internal.TH (deriveFiniteTuples)
+import Clash.Class.Resize (Resize(..))
+import Clash.Class.BitPack (BitPack(..), bitCoerce)
+import Clash.Promoted.Nat
+  ( SNat(..), UNat(..), SNatLE(..)
+  , toUNat, fromUNat, natToNum, snatToNum, compareSNat
+  )
+import Clash.Promoted.Symbol (SSymbol(..))
+import Clash.Sized.Index (Index)
+import Clash.Sized.Internal.BitVector (BitVector(..), Bit(..), high, low)
+import Clash.Sized.Internal.Unsigned (Unsigned(..))
+import Clash.Sized.Signed (Signed)
+import Clash.Sized.RTree (RTree(..), tdfold, tfold, trepeat)
+import Clash.Sized.Vector
+  ( Vec(..), (++), (!!), concatMap, bv2v, dfold, foldl, foldr
+  , ifoldr, indicesI, iterateI, unfoldrI, repeat, reverse, replace
+  )
+import Clash.XException (ShowX, NFDataX, errorX)
+
+import qualified Data.List as List (iterate)
+
+{- $setup
+>>> :m -Prelude
+>>> :set -XDeriveAnyClass
+>>> import Clash.Prelude
+>>> import Data.Ord
+-}
+
+-- * Finite Class
+
+-- | The class of types holding only a finite number of elements.
+--
+-- The class supports generic deriving, i.e., for custom types the
+-- class instances can be derived via @derive (Generic, Finite)@
+-- requiring that all inner types of the type declaration have
+-- @Finite@ instances as well.
+--
+-- >>> data T = B Bit | D (Index 2) (Signed 1) deriving (Generic, Finite, Show)
+-- >>> natToNum @(ElementCount T)
+-- 6
+-- >>> elements @T
+-- B 0 :> B 1 :> D 0 -1 :> D 0 0 :> D 1 -1 :> D 1 0 :> Nil
+-- >>> lowestMaybe @T
+-- Just (B 0)
+-- >>> highestMaybe @T
+-- Just (D 1 0)
+-- >>> succMaybe (B 1)
+-- Just (D 0 -1)
+-- >>> predMaybe (B 0)
+-- Nothing
+-- >>> index (D 0 0)
+-- 3
+-- >>> ith @T 5
+-- D 1 0
+--
+-- Any definition must satisfy the following laws (automatically
+-- ensured when generic deriving the instance):
+--
+-- [Index Order]
+--   @ index '<$>' elements = 'indicesI' @
+-- [Forward Iterate]
+--   @ 'iterateI' ('>>=' succMaybe) (lowestMaybe \@a)
+--      = 'Just' '<$>' (elements \@a) @
+-- [Backward Iterate]
+--   @ 'iterateI' ('>>=' predMaybe) (highestMaybe \@a)
+--      = 'Just' '<$>' 'reverse' (elements \@a) @
+-- [Index Isomorphism]
+--   @ith (index x) = x@
+-- [Minimum Predecessor]
+--   @ lowestMaybe '>>=' predMaybe = 'Nothing' @
+-- [Maximum Successor]
+--   @ highestMaybe '>>=' succMaybe = 'Nothing' @
+-- [Extremes]
+--   @ ElementCount a = 0:
+--       lowestMaybe \@a  = 'Nothing',
+--       highestMaybe \@a = 'Nothing' @
+--  [ ] @ ElementCount a > 0:
+--       lowestMaybe \@a  = 'Just' lowest,
+--       highestMaybe \@a = 'Just' highest @
+--
+-- If @a@  has a 'Bounded' instance, it further must satisfy:
+--
+-- [Bounded Compatibility]
+--   @ ElementCount a > 0: lowest \@a
+--      = 'minBound', highest \@a = 'maxBound' @
+--
+-- If @a@ has an 'Enum' instance, it further must satisfy:
+--
+-- [Enum Compatibility]
+--   @ succMaybe x = Just y: 'succ' x = y @
+--  [ ] @ predMaybe x = Just y: 'pred' x = y @
+--
+class KnownNat (ElementCount a) => Finite a where
+  -- | The number of elements of the type.
+  type ElementCount a :: Nat
+  type ElementCount a = GElementCount (Rep a)
+
+  -- | The elements of the type.
+  elements :: Vec (ElementCount a) a
+  default elements ::
+    ( Generic a, GFinite (Rep a)
+    , ElementCount a ~ GElementCount (Rep a)
+    ) => Vec (ElementCount a) a
+  elements = to <$> gElements
+
+  -- | The element at index @0@.
+  lowest :: 1 <= ElementCount a => a
+  default lowest ::
+    (Generic a, GFinite (Rep a), ElementCount a ~ GElementCount (Rep a)) =>
+    1 <= ElementCount a => a
+  lowest = to gLowest
+
+  -- | Just the element at index 0. Nothing, if @ElementCount a = 0@.
+  lowestMaybe :: Maybe a
+  default lowestMaybe ::
+    ( Generic a, GFinite (Rep a)
+    ) => Maybe a
+  lowestMaybe = to <$> gLowestMaybe
+
+  -- | The element at index @(ElementCount a - 1)@.
+  highest :: 1 <= ElementCount a => a
+  default highest ::
+    (Generic a, GFinite (Rep a), ElementCount a ~ GElementCount (Rep a)) =>
+    1 <= ElementCount a => a
+  highest = to gHighest
+
+  -- | Just the element at index @(ElementCount a - 1)@. Nothing, if
+  -- @ElementCount a = 0@.
+  highestMaybe :: Maybe a
+  default highestMaybe ::
+    ( Generic a, GFinite (Rep a)
+    ) => Maybe a
+  highestMaybe = to <$> gHighestMaybe
+
+  -- | Just the element before the given one according to the
+  -- associated index order with the lowest one being the only element
+  -- that has no predecessor.
+  predMaybe :: a -> Maybe a
+  default predMaybe ::
+    ( Generic a, GFinite (Rep a)
+    ) => a -> Maybe a
+  predMaybe = fmap to . gPredMaybe . from
+
+  -- | Just the element after the given one according to the
+  -- associated index order with the highest one being the only
+  -- element that has no successor.
+  succMaybe :: a -> Maybe a
+  default succMaybe ::
+    ( Generic a, GFinite (Rep a)
+    ) => a -> Maybe a
+  succMaybe = fmap to . gSuccMaybe . from
+
+  -- | Maps from an index to the associated element.
+  ith :: Index (ElementCount a) -> a
+  default ith ::
+    ( Generic a, GFinite (Rep a)
+    , ElementCount a ~ GElementCount (Rep a)
+    ) => Index (ElementCount a) -> a
+  ith = to . gIth
+
+  -- | Maps an element of the type to it's associated index.
+  index :: a -> Index (ElementCount a)
+  default index ::
+    ( Generic a, GFinite (Rep a)
+    , ElementCount a ~ GElementCount (Rep a)
+    ) => a -> Index (ElementCount a)
+  index = gIndex . from
+
+  -- | Returns the suffix slice of 'elements' starting at the index
+  -- provided via the @SNat@ argument.
+  elementsFrom ::
+    n + 1 <= ElementCount a =>
+    SNat n -> Vec (ElementCount a - n) a
+  elementsFrom sn@SNat =
+    iterateI (fromJust . succMaybe) (ith $ snatToNum sn)
+
+  -- | Returns the infix slice of 'elements' from the index provided
+  -- via the first @SNat@ argument to the index provided via the
+  -- second one.
+  elementsFromTo ::
+    (n + 1 <= ElementCount a, n <= m, m + 1 <= ElementCount a) =>
+    SNat n -> SNat m -> Vec (m - n + 1) a
+  elementsFromTo sn@SNat SNat =
+    iterateI (fromJust . succMaybe) (ith $ snatToNum sn)
+
+class KnownNat (GElementCount rep) => GFinite rep where
+  type GElementCount rep :: Nat
+  gElements     :: Vec (GElementCount rep) (rep a)
+  gLowest       :: 1 <= GElementCount rep => rep a
+  gLowestMaybe  :: Maybe (rep a)
+  gHighest      :: 1 <= GElementCount rep => rep a
+  gHighestMaybe :: Maybe (rep a)
+  gPredMaybe    :: rep a -> Maybe (rep a)
+  gSuccMaybe    :: rep a -> Maybe (rep a)
+  gIth          :: Index (GElementCount rep) -> rep a
+  gIndex        :: rep a -> Index (GElementCount rep)
+
+instance GFinite V1 where
+  type GElementCount V1 = 0
+  gElements     = Nil
+  gLowest       = (\case{} :: Dict (1 <= 0) -> a) Dict
+  gLowestMaybe  = Nothing
+  gHighest      = (\case{} :: Dict (1 <= 0) -> a) Dict
+  gHighestMaybe = Nothing
+  gPredMaybe    = const Nothing
+  gSuccMaybe    = const Nothing
+  -- GHC has no knowledge about Index 0 being an uninhabited
+  -- type. Hence, we need to throw an error here although there
+  -- provably are no values that can ever be passed to gIth.
+  gIth a        = errorX $ "Index 0 cannot contain any values like " <> show a
+  gIndex        = \case {}
+
+instance GFinite U1 where
+  type GElementCount U1 = 1
+  gElements     = U1 :> Nil
+  gLowest       = U1
+  gLowestMaybe  = Just U1
+  gHighest      = U1
+  gHighestMaybe = Just U1
+  gPredMaybe    = const Nothing
+  gSuccMaybe    = const Nothing
+  gIth          = const U1
+  gIndex        = const 0
+
+instance Finite a => GFinite (K1 i a) where
+  type GElementCount (K1 _ a)  = ElementCount a
+  gElements     = K1 <$> elements
+  gLowest       = K1 lowest
+  gLowestMaybe  = K1 <$> lowestMaybe
+  gHighest      = K1 highest
+  gHighestMaybe = K1 <$> highestMaybe
+  gPredMaybe    = fmap K1 . predMaybe . unK1
+  gSuccMaybe    = fmap K1 . succMaybe . unK1
+  gIth          = K1 . ith
+  gIndex        = index . unK1
+
+instance GFinite a => GFinite (M1 i v a) where
+  type GElementCount (M1 _ _ a) = GElementCount a
+  gElements     = M1 <$> gElements
+  gLowest       = M1 gLowest
+  gLowestMaybe  = M1 <$> gLowestMaybe
+  gHighest      = M1 gHighest
+  gHighestMaybe = M1 <$> gHighestMaybe
+  gPredMaybe    = fmap M1 . gPredMaybe . unM1
+  gSuccMaybe    = fmap M1 . gSuccMaybe . unM1
+  gIth          = M1 . gIth
+  gIndex        = gIndex . unM1
+
+instance (GFinite a, GFinite b) => GFinite (a :*: b) where
+  type GElementCount (a :*: b) = GElementCount a * GElementCount b
+  gElements = concatMap (\a -> (a :*:) <$> gElements @b) (gElements @a)
+
+  gLowest
+    | Dict <- mulPositiveImpliesPositiveOperands @(GElementCount a) @(GElementCount b)
+    = gLowest :*: gLowest
+  gLowestMaybe = (:*:) <$> gLowestMaybe  <*> gLowestMaybe
+
+  gHighest
+    | Dict <- mulPositiveImpliesPositiveOperands @(GElementCount a) @(GElementCount b)
+    = gHighest :*: gHighest
+  gHighestMaybe = (:*:) <$> gHighestMaybe <*> gHighestMaybe
+
+  gPredMaybe (a :*: b) =
+        (:*:)             a <$> gPredMaybe b
+    <|> (:*:) <$> gPredMaybe a <*> gHighestMaybe
+  gSuccMaybe (a :*: b) =
+        (:*:)            a <$> gSuccMaybe b
+    <|> (:*:) <$> gSuccMaybe a <*> gLowestMaybe
+
+  gIth x = gIth (resize $ x `div` m) :*: gIth (resize $ x `mod` m)
+   where
+    m = natToNum @(GElementCount b)
+
+  gIndex (a :*: b) =
+      resize (gIndex a) * natToNum @(GElementCount b)
+    + resize (gIndex b)
+
+instance (GFinite a, GFinite b) => GFinite (a :+: b) where
+  type GElementCount (a :+: b) = GElementCount a + GElementCount b
+  gElements = (L1 <$> gElements @a) ++ (R1 <$> gElements @b)
+
+  gLowest = case compareSNat (SNat @1) (SNat @(GElementCount a)) of
+    SNatLE -> L1 gLowest
+    SNatGT -> case compareSNat (SNat @1) (SNat @(GElementCount b)) of
+      SNatLE -> R1 gLowest
+      SNatGT -> case zeroLeAdd @(GElementCount a) @1 of
+        Dict -> case zeroLeAdd @(GElementCount b) @1 of {}
+  gLowestMaybe = L1 <$> gLowestMaybe  @a <|> R1 <$> gLowestMaybe  @b
+
+  gHighest = case compareSNat (SNat @1) (SNat @(GElementCount b)) of
+    SNatLE -> R1 gHighest
+    SNatGT -> case compareSNat (SNat @1) (SNat @(GElementCount a)) of
+      SNatLE -> L1 gHighest
+      SNatGT -> case zeroLeAdd @(GElementCount a) @1 of
+        Dict -> case zeroLeAdd @(GElementCount b) @1 of {}
+  gHighestMaybe = R1 <$> gHighestMaybe @b <|> L1 <$> gHighestMaybe @a
+
+  gPredMaybe = \case
+    L1 x -> L1 <$> gPredMaybe x
+    R1 x -> R1 <$> gPredMaybe x <|> L1 <$> gHighestMaybe
+
+  gSuccMaybe = \case
+    R1 x -> R1 <$> gSuccMaybe x
+    L1 x -> L1 <$> gSuccMaybe x <|> R1 <$> gLowestMaybe
+
+  gIth x
+    | e x < n   = L1 $ gIth $ truncateB x
+    | otherwise = R1 $ gIth $ truncateB $ e x - n
+   where
+    n = natToNum @(GElementCount a)
+    e = extend @Index @(GElementCount (a :+: b))
+      @(If (GElementCount (a :+: b) <=? GElementCount a) 1 0)
+
+  gIndex = \case
+    L1 x -> extend (gIndex x)
+    R1 x -> extend (gIndex x) + natToNum @(GElementCount a)
+
+instance Finite Void
+instance Finite ()
+instance Finite Bool
+instance Finite Ordering
+instance Finite (Proxy a)
+
+instance KnownNat n => Finite (SNat n) where
+  type ElementCount (SNat n) = 1
+  elements     = SNat :> Nil
+  lowest       = SNat
+  lowestMaybe  = Just SNat
+  highest      = SNat
+  highestMaybe = Just SNat
+  predMaybe    = const Nothing
+  succMaybe    = const Nothing
+  ith          = const SNat
+  index        = const 0
+
+instance KnownSymbol s => Finite (SSymbol s) where
+  type ElementCount (SSymbol s) = 1
+  elements     = SSymbol :> Nil
+  lowest       = SSymbol
+  lowestMaybe  = Just SSymbol
+  highest      = SSymbol
+  highestMaybe = Just SSymbol
+  predMaybe    = const Nothing
+  succMaybe    = const Nothing
+  ith          = const SSymbol
+  index        = const 0
+
+instance c => Finite (Dict c) where
+  type ElementCount (Dict c) = 1
+  elements     = Dict :> Nil
+  lowest       = Dict
+  lowestMaybe  = Just Dict
+  highest      = Dict
+  highestMaybe = Just Dict
+  predMaybe    = const Nothing
+  succMaybe    = const Nothing
+  ith          = const Dict
+  index        = const 0
+
+deriving via BoundedEnumEq 0x110000         Char   instance Finite Char
+deriving via BoundedEnumEq (2^BitSize Int)  Int    instance Finite Int
+deriving via BoundedEnumEq (2^8)            Int8   instance Finite Int8
+deriving via BoundedEnumEq (2^16)           Int16  instance Finite Int16
+deriving via BoundedEnumEq (2^32)           Int32  instance Finite Int32
+deriving via BoundedEnumEq (2^64)           Int64  instance Finite Int64
+deriving via BoundedEnumEq (2^BitSize Word) Word   instance Finite Word
+deriving via BoundedEnumEq (2^8)            Word8  instance Finite Word8
+deriving via BoundedEnumEq (2^16)           Word16 instance Finite Word16
+deriving via BoundedEnumEq (2^32)           Word32 instance Finite Word32
+deriving via BoundedEnumEq (2^64)           Word64 instance Finite Word64
+
+deriving newtype instance Finite a         => Finite (Const a b)
+deriving newtype instance Finite a         => Finite (Identity a)
+deriving newtype instance Finite (f (g a)) => Finite (Compose f g a)
+
+instance  Finite a                    => Finite (Maybe a)
+instance (Finite a,     Finite b    ) => Finite (Either a b)
+instance (Finite (f a), Finite (g a)) => Finite (Product f g a)
+instance (Finite (f a), Finite (g a)) => Finite (Sum f g a)
+
+instance KnownNat n => Finite (Index n) where
+  type ElementCount (Index n) = n
+  elements     = indicesI
+  lowest       = minBound
+  lowestMaybe  = case toUNat (SNat @n) of
+    UZero -> Nothing
+    _     -> Just minBound
+  highest      = maxBound
+  highestMaybe = case toUNat (SNat @n) of
+    UZero -> Nothing
+    _     -> Just maxBound
+  predMaybe    = case toUNat (SNat @n) of
+    UZero -> const Nothing
+    _     -> \n -> if n == minBound then Nothing else Just $ n - 1
+  succMaybe    = case toUNat (SNat @n)of
+    UZero -> const Nothing
+    _     -> \n -> if n == maxBound then Nothing else Just $ n + 1
+  ith          = id
+  index        = id
+
+instance KnownNat n => Finite (Signed n) where
+  type ElementCount (Signed n) = 2^n
+  elements     = iterateI (+1) minBound
+  lowest       = minBound
+  lowestMaybe  = Just minBound
+  highest      = maxBound
+  highestMaybe = Just maxBound
+  predMaybe n  = if n == minBound then Nothing else Just $ n - 1
+  succMaybe n  = if n == maxBound then Nothing else Just $ n + 1
+  ith          = unpack . xor (complement (complement 0 `shiftR` 1)) . pack
+  index        = unpack . xor (complement (complement 0 `shiftR` 1)) . pack
+
+instance KnownNat n => Finite (Unsigned n) where
+  type ElementCount (Unsigned n) = 2^n
+  elements     = iterateI (+1) minBound
+  lowest       = minBound
+  lowestMaybe  = Just minBound
+  highest      = maxBound
+  highestMaybe = Just maxBound
+  predMaybe n  = if n == minBound then Nothing else Just $ n - 1
+  succMaybe n  = if n == maxBound then Nothing else Just $ n + 1
+  ith          = bitCoerce
+  index        = bitCoerce
+
+instance Finite Bit where
+  type ElementCount Bit = 2
+  elements     = low :> high :> Nil
+  lowest       = low
+  lowestMaybe  = Just low
+  highest      = high
+  highestMaybe = Just high
+  predMaybe b  = if b == low then Nothing else Just low
+  succMaybe b  = if b == high then Nothing else Just high
+  ith          = \case { 0 -> low; _ -> high }
+  index b      = if b == low then 0 else 1
+
+instance KnownNat n => Finite (BitVector n) where
+  type ElementCount (BitVector n) = 2^n
+  elements     = iterateI (+1) 0
+  lowest       = minBound
+  lowestMaybe  = Just minBound
+  highest      = maxBound
+  highestMaybe = Just maxBound
+  predMaybe bv = if bv == minBound then Nothing else Just $ bv - 1
+  succMaybe bv = if bv == maxBound then Nothing else Just $ bv + 1
+  ith          = pack
+  index        = unpack
+
+data PowV (k :: Nat) (a :: Type) (f :: TyFun Nat Type) :: Type
+type instance Apply (PowV k a) n = Vec (k^n) (Vec n a)
+
+instance (KnownNat n, Finite a) => Finite (Vec n a) where
+  type ElementCount (Vec n a) = ElementCount a^n
+
+  elements = dfold
+    (Proxy @(PowV (ElementCount a) a))
+    (\_ _ -> concatMap ((<$> elements) . (:<)))
+    (Nil :> Nil)
+    (repeat @n ())
+
+  lowest = case toUNat (SNat @n) of
+    UZero -> Nil
+    _ -> case compareSNat (SNat @1) (SNat @(ElementCount a)) of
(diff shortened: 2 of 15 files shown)


clash-prelude/src/Clash/Explicit/Prelude.hs
```diff
@@ -242,11 +244,11 @@ window
...
   where
-    res  = x :> prev
-    prev = case natVal (asNatProxy prev) of
+    res  = x :> prev'
+    prev' = case natVal (asNatProxy prev') of
              0 -> repeat def
-             _ -> let next = x +>> prev
-                  in  registerB clk rst en (repeat def) next
+             _ -> let next' = x +>> prev'
+                  in  registerB clk rst en (repeat def) next'
```
martijnbastiaan, line 251: The fact that you have to do this is a pretty strong hint these names shouldn't be in `Clash.Prelude`. Grepping my projects for `prev` and `next` also reveals a number of uses.

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,751 @@
+{-|
+Copyright  :  (C) 2024, QBayLogic B.V.
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+-}
+
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
martijnbastiaan, line 7: Is there a great reason to use `AllowAmbiguousTypes`? It makes error messages really bad. (I know you said very new GHCs do this better, but we're not even close to dropping most of them messing up..)

clash-prelude/src/Clash/Num/Zeroing.hs
```diff
@@ -1,11 +1,12 @@
...
-Copyright   : (C) 2021, QBayLogic B.V.
+{-|
+Copyright   : (C) 2021-2024, QBayLogic B.V.
 License     : BSD2 (see the file LICENSE)
 Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
```
martijnbastiaan, line 9: Why does importing a module cause an additional language pragma?

clash-prelude/src/Clash/Num/Wrapping.hs
```diff
@@ -1,11 +1,12 @@
...
-Copyright   : (C) 2021, QBayLogic B.V.
+{-|
+Copyright   : (C) 2021-2024, QBayLogic B.V.
 License     : BSD2 (see the file LICENSE)
 Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
```
martijnbastiaan, line 9: Same as other comment.

clash-prelude/src/Clash/Num/Saturating.hs
```diff
@@ -1,11 +1,12 @@
...
-Copyright   : (C) 2021, QBayLogic B.V.
+{-|
+Copyright   : (C) 2021-2024, QBayLogic B.V.
 License     : BSD2 (see the file LICENSE)
 Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeFamilies #-}
```
martijnbastiaan, line 9: Same as other comment.

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,751 @@
...
+--   @ 'iterateI' ('>>=' prev) (highest \@a) = 'Just' '<$>' (elements \@a) @
+-- [Index Isomorphism]
+--   @ith (index x) = x@
+-- [Minimum Predecessor]
+--   @ lowest '>>=' prev = 'Nothing' @
+-- [Maximum Successor]
+--   @ highest '>>=' next = 'Nothing' @
+-- [No Uninhabited Extremes]
+--   @ lowest \@a = 'Nothing' /and/ highest \@a = 'Nothing'
+--     /if and only if/ ElementCount a = 0 @
```
martijnbastiaan, line 139: Would it be sensible to add hedgehog laws for this? Could these be added to the test suite for the types that derive / define an instance?

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,751 @@
...
+-- | The class of types holding only a finite number of elements. The
+-- class offers type level access to the number of elements @n@ and
+-- defines a total order on the elements via indexing them from @0@ to
+-- @n-1@. The class gives access to the vector of all elements of the
+-- type, allows to iterate over the elements in order or to map back
+-- and forth between their associated indices.
+--
+-- In general, the class can be considered as a more hardware-friendly
+-- alternative to 'Bounded' and 'Enum', utilizing 'Index' instead of
+-- 'Int' and vectors instead of lists.
```
martijnbastiaan, line 97: Is there a reason that we don't provide `FiniteEnum` (or `CEnum`) directly? This would give developers an immediate intuition for what the classes are.

Related: I've wondered many times whether it would be sensible for `Clash.Prelude` to export a Clash-friendly version of all these `Int`/`Integer` based APIs. I.e., replace `base`'s `Enum` (and friends) in favor of Clash ones.

martijnbastiaan (commented):
Couple of questions! Mostly wondering whether it would make sense to provide an `Index`-based `Enum` instead.

clash-prelude/src/Clash/Explicit/Prelude.hs
kleinreact: That's kind of unavoidable. Every new name we choose might already be picked up in existing user code and I doubt that it's in the interest of the user, if we choose artificially unintuitive names just because of that.

So yes, the user might experience some new warnings about "clashing" names after a Clash update, which are always easy to fix via renaming or hiding the new stuff from `Clash.Prelude`, in case they are not desired.

Please note that finding a good (and intuitive) naming scheme that doesn't clash with basic existing libraries (in particular `base`) is a challenge already, especially if the offered primitives are quire fundamental. I also tried to improve the situation here based on some prior experience I had with the [finite](https://hackage.haskell.org/package/finite-1.4.1.2/docs/Finite.html) library, which comes with a similar interface.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: Newer GHCs do much better on this! Maybe you should consider to update :stuck_out_tongue_winking_eye:.

kleinreact: The alternative would be to use proxies, which I used in my [finite](https://hackage.haskell.org/package/finite-1.4.1.2/docs/Finite.html) library in the past, but the usage experience turns out to be really unsatisfactory. Using ambiguous types instead definitely improves the situation here.

IMO users that like to stick with older GHCs and don't like the bad error messages just shouldn't use the new feature then. As it's just the error messages tough, there is no problem in developing with a recent GHC and supporting the old ones with the error-free code in the end. 

I developed this interface with GHC 9.10 and I definitely experience several ambiguous type errors, which all where easy to identify and fix, but I never had to inspect any of the error messages of older GHCs.

gergoerdi: I was unable to find an example/test module that uses these new definitions in the commit, but if the issue here is using `foo @3` vs `foo (Proxy @3)` etc., if you are aiming for GHC 9.10 you should consider using a required type argument (i.e. using `forall n -> ...` instead of `forall n. ...`) so you can write it as `foo 3`.

clash-prelude/src/Clash/Num/Zeroing.hs
kleinreact: It's because `Finite` also has been added to the `deriving newtype` list and the default implementation of `Finite` requires `TypeFamilies`. The `deriving newtype` strategy does not use the default implementation tough, which is why the dependency technically is not necessary here. This seems to be fixed in GHC 9.4 and later, but GHC 9.2 and older still required the language extension to be enabled.

We could add a preprocessor define to only enable the extension for the old GHCs, but then we also need to enable `CCP` for any compiler version, so it's a matter of taste in the end. A CPP define only makes sense if we plan to clean up the code as soon as we drop support for GHC 9.2 and older, because it makes it easier identify the legacy parts.

clash-prelude/src/Clash/Num/Wrapping.hs
kleinreact: Same as my previous answer :upside_down_face:.

clash-prelude/src/Clash/Num/Saturating.hs
kleinreact: Same as my previous answer :upside_down_face:.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: @gergoerdi I haven't looked into how the interface would particularly change with required type arguments yet, but my experience is that you don't need to pass the type explicitly in most of the cases. 

Thus in terms of writing code, I'd say that ambiguous types are the most elegant solution here :sparkles:.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: Definitely! It's just not necessary at the current code reviewing stage IMO. Hence, I considered it to be a nice addition for a later stage. I also tested all of the features in REPL already :tm:.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: I would not change the semantics of the classes that are already available in `base`, because this only leads to trouble and confusion. The right strategy should be to provide hardware-friendy alternatives, like `Signed n` is to be preferred over `Int` in hardware, which then can coexist with the software-friendly types.

Most of the design choices for this extension have been inspired by my [finite](https://hackage.haskell.org/package/finite-1.4.1.2/docs/Finite.html) library for sure, where I considered it more useful to have a single class covering all of the properties of a "finite" type, instead of breaking things up into multiple classes like `Enum` or `Bounded`, which definitely makes sense for types with infinitely many inhabitants.

I'd also like to avoid to make the interface here too similar to `Enum` and friends, because there still are some fundamental differences, which could be mixed up easily otherwise.

clash-prelude/src/Clash/Explicit/Prelude.hs
DigitalBrains1: I think the point is we write many state machines in Clash and there `prev` and `next` are very intuitive local names. Of course you can't avoid all name clashes, but there are some words that are very common in Clash code and `prev` and `next` are two examples of that specific category. So I agree with Martijn that trying to come up with an alternative is worthwhile in this specific case.

DigitalBrains1: before/after?

prior/later?

back/forward?

Did not think them through much, just throwing it out there.

gergoerdi: `fpred` / `fsucc` for finite versions of Enum functions?

clash-prelude/src/Clash/Num/Zeroing.hs
martijnbastiaan: Interesting. Not worth the CPP IMO, so let's leave it like it is.

clash-prelude/src/Clash/Explicit/Prelude.hs
martijnbastiaan: > Please note that finding a good (and intuitive) naming scheme that doesn't clash with basic existing libraries (in particular base) is a challenge already, especially if the offered primitives are quire fundamental. 

IMO this again reinforces my point: `base`'s Prelude (+Haskell itself) already squats an uncomfortable number of identifiers. Another point is that I don't really like synonyms. I can basically never remember whether to use `min` or `minimum` for example, I'd have the same problems with `succ` and `next`. (Though TBF I've used `succ` so many times now that I'd recognize `next` as the newer identifier, but I think my point is clear.)

`fpred` / `fsucc` makes sense to me.

clash-prelude/src/Clash/Class/Finite/Internal.hs
martijnbastiaan: > IMO users that like to stick with older GHCs and don't like the bad error messages just shouldn't use the new feature then.

You're introducing this in the code base though, it's not like maintainers will have much of a choice there..

> but my experience is that you don't need to pass the type explicitly in most of the cases.

This statement doesn't make much sense to me, they're ambiguous for a reason. If anything, it's almost always required to fully type apply ambiguous functions. Which also shows in this PR: most (if not all?) ambiguous functions you define in the PR you also use in a fully type applied manner.

IMO until we can do required type arguments (`forall n ->`) we should use `Proxy`. Once we drop support the last GHC not supporting them, we can instruct our users (+ourselves) to regex the `Proxy`s out.

clash-prelude/src/Clash/Class/Finite/Internal.hs
martijnbastiaan: > Most of the design choices for this extension have been inspired by my [finite](https://hackage.haskell.org/package/finite-1.4.1.2/docs/Finite.html) library for sure, where I considered it more useful to have a single class covering all of the properties of a "finite" type, instead of breaking things up into multiple classes like Enum or Bounded, which definitely makes sense for types with infinitely many inhabitants.
>
> I'd also like to avoid to make the interface here too similar to Enum and friends, because there still are some fundamental differences, which could be mixed up easily otherwise.

Could you please elaborate? Both `Enum` and `Bounded` are mentioned in the documentation, but their relation (or rather differences) aren't clear to me. What makes them good classes for "infinite" types, but not for "finite" types?

clash-prelude/src/Clash/Explicit/Prelude.hs
kleinreact: Thanks for the feedback and all the suggestions. I like to change it to `nextElement` and `previousElement` then, which I still consider intuitive and recognizable. 

While `fpred` and `fsucc` are nice and short, they still might be mixed up with their `Enum` counterparts, which I'd prefer to avoid.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: > You're introducing this in the code base though, it's not like maintainers will have much of a choice there.

Sry, but I don't get your point. As long as the users don't use the new class they also won't face any ambiguous type errors, right?

> This statement doesn't make much sense to me, they're ambiguous for a reason. If anything, it's almost always required to fully type apply ambiguous functions. Which also shows in this PR: most (if not all?) ambiguous functions you define in the PR you also use in a fully type applied manner.

I also can't see why either `Proxy` or required type arguments are desired here. For example, if you like to create the result column of the truth table of conjunction, then with ambiguous types you can easily do so via
```haskell
uncurry (&&) <$> elements
```
where type derivation does it's job and deduces that `elements` needs to be of type `Vec 4 (Bool, Bool)`. With `Proxy` or required type arguments you always would need to pass `Proxy` or a type, which is just overhead here.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: **Enum**
* Already the type of `succ, pred :: Enum a => a -> a` assumes that every element has successor and predecessor, which makes perfect sense for infinite types, but requires finite types to error on certain inputs. Note that wrapping behavior is disallowed according to the documentation of `base`.
* As a consequence, `Enum` instances must ship partial functions for most finite types.
* The same way, using `Int` as the index type creates a natural mismatch between the number of inhabitants of the index type with the ones of the indexed type. For infinite types, that's "kind of ok", because here the practical assumption  is that `Int` ~ `Integer`, i.e., we never enumerate stuff that won't fit into into an `Int` anyway, but that's just an efficiency trade off in the end.
* Generic deriving for `Enum` is only supported for sum data types, but they cannot contain any products (because that doesn't work out with infinite types), which is kind of a strange conceptual mix between the two worlds.

**Bounded**
* The class is often used to identify the bounds of finite types, but just having a `minBound` and a `maxBound` does not automatically make your type finite, e.g. the interval [0,1] over the reals has a `minBound` and `maxBound` too, but is infinite.
* The `Bounded` class doesn't support uninhabited types like `Void` or `Index 0`, which makes it kind of incomplete for me. Though, I think it depends on how you look at it. Nevertheless, having to exclude the `Index 0` case just because of that turns out to be non-elegant in then end.

clash-prelude/src/Clash/Explicit/Prelude.hs
DigitalBrains1: I agree with `fpred`/`fsucc`.

clash-prelude/src/Clash/Class/Finite/Internal.hs
DigitalBrains1: It seems to compile fine on GHC 9.8.2 without the `AllowAmbiguousTypes` pragma.

clash-prelude/src/Clash/Explicit/Prelude.hs
kleinreact: I changed the function names to `before` and `after` now, as I just didn't feel comfortable with the options discussed so far.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: @DigitalBrains1 Thanks for pointing that out. I initially had the dictionary definitions also in `Internal.hs` and just moved them to a separate file after cleaning up. I just didn't catch in the end that this indeed is the only part requiring `AllowAmbiguousTypes`. 

I cleaned it up now, which should finally resolve this discussion. I'm sorry for all the back and forth in the first place then.

kleinreact: And as it turns out, everything prior to GHC 9.2 requires `AllowAmbiguousTypes` here :disappointed:.

https://github.com/clash-lang/clash-compiler/actions/runs/12573389973/job/35046345649?pr=2858

clash-prelude/src/Clash/Explicit/Prelude.hs
martijnbastiaan: > Thanks for the feedback and all the suggestions. I like to change it to nextElement and previousElement then, which I still consider intuitive and recognizable.

This suffers from the same synonym (`min` vs `minimum`) issue I mentioned.

clash-prelude/src/Clash/Class/Finite/Internal.hs
martijnbastiaan: Gotcha, I think some of this should make it into the documentation. I'm not opposed to "large" classes btw.

martijnbastiaan:
Another thing to consider: there is at least some overlap with [Counter](https://github.com/clash-lang/clash-compiler/blob/master/clash-prelude/src/Clash/Class/Counter/Internal.hs) I believe. Perhaps `Counter`'s superclass should be `Finite`?

kleinreact:
> Another thing to consider: there is at least some overlap with [Counter](https://github.com/clash-lang/clash-compiler/blob/master/clash-prelude/src/Clash/Class/Counter/Internal.hs) I believe. Perhaps `Counter`'s superclass should be `Finite`?

Good point. Making `Finite` a superclass would be a recognizable API change tough. Also, you technically can have a `Counter` instance for a type with infinitely many inhabitants, so adding the superclass requirement here would introduce a limitation.

 How about adding another deriving strategy for `Counter` via `FiniteDerive` instead, like already present for `Enum`? Then the users can decide on their own, whether counting via the implicit index order of the `Finite` instance is exactly what they want.

martijnbastiaan:
> Making Finite a superclass would be a recognizable API change tough.

I think however we slice it this PR will be an API change, unless we decide to not export it from `Prelude`. So that's fine with me (provided that there wouldn't be runtime changes).

> How about adding another deriving strategy for Counter via FiniteDerive instead, like already present for Enum? Then the users can decide on their own, whether counting via the implicit index order of the Finite instance is exactly what they want.

I think `Finite` would count the same as `Enum` for sum-types right? In that case I don't see the benefit of offering both, but I'm also not opposed to it.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: I added the demanded tests. 

Note that #2871 is required for the tests to run in a reasonable amount of time.

clash-prelude/src/Clash/Class/Finite/Internal/TH.hs
```diff
@@ -0,0 +1,171 @@
...
+                (z:zs) -> (z:zs)
+                _ -> error "maxTupleSize < 3"
+        tuple xs = foldl' AppT (TupleT $ length xs) xs
+
+        -- Instance declaration
+        context = AppT finite . VarT <$> names
+        instTy = AppT finite $ tuple vs
+
+    return $ InstanceD Nothing context instTy []
+#else
```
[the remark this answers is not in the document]
kleinreact: Thanks for the hint. I didn't test compilation with `large-tuples` enabled, neither does CI. 

I removed the short version then.

clash-prelude/src/Clash/Class/Finite/Internal/Dictionaries.hs
```diff
@@ -0,0 +1,58 @@
...
+#if !MIN_VERSION_base(4,16,0)
+import GHC.TypeLits.Extra (CLog)
+#endif
+import Unsafe.Coerce (unsafeCoerce)
+
+-- | Evidence that exponentiation can never return a zero result.
+powGTZero ::
+  forall (n :: Nat) (m :: Nat).
+  Dict (1 <= n^m)
+powGTZero = unsafeCoerce (Dict :: Dict (0 <= 0))
```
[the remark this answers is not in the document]
kleinreact: Nice catch! Fixed it by adding a `1 <= n` constraint.

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,760 @@
...
+  -- | The elements of the type.
+  elements :: Vec (ElementCount a) a
+  default elements ::
+    ( Generic a, GFinite (Rep a)
+    , ElementCount a ~ GElementCount (Rep a)
+    ) => Vec (ElementCount a) a
+  elements = to <$> gElements
+
+  -- | Just the @0@ indexed element. Nothing if @ElementCount a = 0@.
+  lowest :: Maybe a
```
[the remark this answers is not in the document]
kleinreact: We could change the type of `lowest` and `highest` to
```haskell
lowest, highest :: 1 <= ElementCount a => a
```
and add some additional methods `lowestMaybe` and `highestMaybe` for user convenience. Only requires `ConstrainedClassMethods`, which shouldn't cause any issues.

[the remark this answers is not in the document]
martijnbastiaan: Hmmyeah, I'm also conflicted. I think we've got a couple of options:

1. `Maybe`
2. Constraint (`1 <= ElementCount a`)
3. Producing bottoms (`errorX "...."`)
4. Define two classes: one that allows `Void`-like types, one that doesn't.
5. Expose both functions

If precedent is worth anything, we've typically opted for (3). In cases where we did pick (2), we later moved to (3) because the constraints were annoying in practice. Producing bottoms isn't ideal either: I wish we had a non-partial `unpack`, for example.

So what about 5: having a `lowestMightBeMaybe :: Maybe a` (name to be bikeshedded) _and_ a `lowestMightProduceErrorX :: a`?

clash-prelude/src/Clash/Explicit/Prelude.hs
martijnbastiaan: I was thinking, there is precedent for this:

* `succ`
* `countSucc`
* `satSucc`

I that list `finSucc` makes sense.

If I can poke a little bit in your thoughts: I think what you're rubbing against is the fact that you think names in `Clash.Class.Finite` shouldn't be influenced by names in other classes because it is its own namespace. This is typically true, but we're making one God-module that contains everything (`Clash.Prelude`), which makes the namespace mangled. The friction becomes especially annoying when you're trying to the RightThing(tm) and use qualified imports and then end up with:

```haskell
Finite.finSucc
```

which succs!

clash-prelude/src/Clash/Class/Finite/Internal.hs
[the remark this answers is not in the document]
kleinreact: I don't see the difference between `lowestMightProduceErrorX` and `fromJust . lowestMightBeMaybe`. Hence, if the user prefers to introduce partial functions in his codebase, then the `Maybe` version gives all freedom to do so.

This is why I prefer having `lowest` and `lowestMaybe` as suggested above. With that, the user has all options at hand:
1. user prefers maybe: use `lowestMaybe`
2. user prefers errors: use `fromJust lowestMaybe` (or `fromMaybe (error "🙀") lowestMaybe` if an individual error message is desired)
3. user prefers constraints:  use `lowest`

Also note that the constraint only appears for polymorphic instances. Thus, you really need to write code that needs to work for all `Finite` instances (using a `Finite a =>` constraint) to encounter that constraint. In practice that's usually only the case when creating a new container type (like `Vec`) with a `Finite` instance derivation rule.

clash-prelude/src/Clash/Explicit/Prelude.hs
kleinreact: > This suffers from the same synonym (min vs minimum) issue I mentioned.

@martijnbastiaan Sry, I didn't saw your edit till now. I still don't get what the synonym issue is about? For me `min` is a shorthand for `minimum` and `minimum` is a proper English word you can find in a dictionary. So the name characterization looks quite clear.

Can you elaborate on your thoughts here a bit?

It's not that I don't like the names, but I think it's just not necessary to introduce new terminology, if we already have words that very well describe the functionality of these methods here. Just consider you'd live in universe that neither knows Clash nor Haskell _(which certainly applies to a bunch of people in the world)_? How would you interpret `fpred` and `fsucc`, just given the names?

clash-prelude/tests/Clash/Tests/Laws/Finite.hs
```diff
@@ -0,0 +1,181 @@
...
+  , testCase "No Uninhabited Extremes" $ noUninhabitedExtremes proxy
+  ]
+
+testFiniteLaws ::
+  (NFData a, Show a, Eq a, Finite a) =>
+  String ->
+  Proxy a ->
+  TestTree
+testFiniteLaws typeName =
+  testGroup typeName . finiteLaws
```
martijnbastiaan, line 106: Micro-optimization but:

```suggestion
testFiniteLaws ::
  (NFData a, Show a, Eq a, Finite a, Typeable a) =>
  Proxy a ->
  TestTree
testFiniteLaws t =
  testGroup (show (typeRep t)) (finiteLaws t)
```

which would allow you to remove duplication:

```haskell
  [ testFiniteLaws (Proxy @Void)
  , testFiniteLaws (Proxy @())
  , testFiniteLaws (Proxy @Bit)
  , testFiniteLaws (Proxy @Bool)
  ..
```

clash-prelude/src/Clash/Explicit/Prelude.hs
martijnbastiaan: The problem with synonyms is that they are words that refer to the same thing (modulo some culturally defined emphasis perhaps). So it makes it impossible to tell them apart -- you have to learn it by hard. As an example I used `min` vs `minimum`: is it:

```
min :: Ord a => a -> a -> a
minimum :: (Foldable t, Ord a) => t a -> a
```

or the other way around? You can't tell by the names! Similarly, if we were to do this for the various `succ`s around `Clash.Prelude` you'd get (for example): `succ`, `successor`, `next`, and `after`. Sure, the identifiers are different, but you'd have to learn by hard which identifiers refer to which concept. That's why I think `satSucc`, `finSucc`, `countSucc`, and `succ` are much much better.

(To be clear: I think the _actual_ solution is proper name spacing, but that ship has more or less sailed. Something we could consider is using `succ` for the classes and then renaming then prefixing them in `Clash.Prelude`.)

clash-prelude/tests/Clash/Tests/Laws/Finite.hs
kleinreact: Thanks for the tip. Updated it accordingly. I primarily stole the layout from the `Enum` law tests, which could be improved in that regard as well.

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,760 @@
...
+--      = 'Just' '<$>' 'reverse' (elements \@a) @
+-- [Index Isomorphism]
+--   @ith (index x) = x@
+-- [Minimum Predecessor]
+--   @ lowest '>>=' before = 'Nothing' @
+-- [Maximum Successor]
+--   @ highest '>>=' after = 'Nothing' @
+-- [No Uninhabited Extremes]
+--   @ lowest \@a = 'Nothing' /and/ highest \@a = 'Nothing'
+--     /if and only if/ ElementCount a = 0 @
```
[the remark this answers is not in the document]
kleinreact: You are right, the index order of `Down` should be reversed. I updated it accordingly.

`ReversedIndexOrder` originally was intended to reverse the index order of new data types without the need of wrapping them into a newtype. However, I missed to make it dependent on the generic class instead, which is fixed now. 

To make that more clear, I renamed the newtype to `GenericReverse` and added some documentation that explains why we wanna have both: `Down` and `GenericReverse`.

clash-prelude/src/Clash/Class/Finite/Internal.hs
martijnbastiaan: Looks good!

clash-prelude/src/Clash/Explicit/Prelude.hs
kleinreact: Gotcha. Thanks for the elaboration. I wasn't aware of how many `succ` variants we already provide.

One important mismatch that I still see here is the underlying ordering relation utilized. There is the common misconception that types with a `Num` instance and an `Enum` instance must satisfy `x + 1 = succ x` and `x - 1 = pred x`, but that's just not given. `Enum` just enumerates the elements according to **some** (_freely chosen_) index order, which does not have to be compatible with the notion of addition and subtraction at all. The same principles apply for `Finite`.

`SaturatingNum` is based on `Num`, so `satSucc` should be compatible with (+ 1), but can be incompatible with `succ`, `Coutner` is based on no other class, hence, `countSucc` again can behave completely different.

 The point that I don't like with using similar names here, is that they suggest that all of the different variants have to behave similarly for the same type / are compatible with each other, but there is actually no requirement that they do.

martijnbastiaan: > The point that I don't like with using similar names here

Synonyms _are_ similar names, just without any clear connection to a specific type class.

> The point that I don't like with using similar names here, is that they suggest that all of the different variants have to behave similarly for the same type / are compatible with each other, but there is actually no requirement that they do.

I'm honestly not sure why you would even think that. The point of different functions is that they have different behavior.. With prefixes you clearly link them to specific classes and (hopefully) remember what they do.

kleinreact: > > SaturatingNum is based on Num, so satSucc should be compatible with (+ 1)
>
> I really don't know why you would think that.
```haskell
-- | 'Num' operators in which overflow and underflow behavior can be specified
-- using 'SaturationMode'.
class (Bounded a, Num a) => SaturatingNum a where

...

-- | Get successor of (or in other words, add 1 to) given number
satSucc :: SaturationMode -> a -> a
```
Do I miss something fundamental here? I read that like `satSucc` is equivalent to `(+ 1)`, except for the boundary cases, where the behavior might differ. That's what I consider by "to compatible with each other".

> Synonyms are similar names, just without any clear connection to a specific type class.

Mhm, if you phrase it like that it suddenly makes a lot of sense. :thinking: 

> I'm honestly not sure why you would even think that. The point of different functions is that they have different behavior.. With prefixes you clearly link them to specific classes and (hopefully) remember what they do.

I just wanted to say that they behave similarly, not they behave equal. Not the best formulation of mine, I agree.

kleinreact: I don't like to loose the focus of our main discussion here though. I agree, that the situation is not ideal, as we somehow ended up with a bunch of concepts (reflected by the different classes), that all kind of do similar things, but all of them with a different kind of flavor.

My original intention was that `Finite` should offer at least the features of `Enum` and `Bounded`, but with an interface that is better suited finite types. In that regard, as everything in hardware is finite, I'd say that you should be able to use `Finite` as a drop-in replacement of `Enum` and `Bounded` in Clash.

So maybe a more fundamental question at this point. How do you think about removing `base`'s `Enum` and `Bounded` from `Clash.Prelude` and adding `Finite` instead, with the names just chosen the same as for `Enum` and `Bounded`? If a user still needs the classes from `base`, he can `qualified import` them from `Prelude`, similar as it needs to be done for the operations on lists. This would make it clear more that `Finite` is the desired alternative for hardware development in this case.

It's just an idea, and I know you won't like it because it's a breaking API change, but that should't hold me back to put it as a proposal at least :wink:.

kleinreact: Alright. I figured out a clean solution in the end. We just name them `succMaybe` and `predMaybe` which makes it just perfectly clear that these are just safe variants of `succ` and `pred`. I updated it already.

I also added some conditional laws that there must be compatibility between the operations if `Bounded` and `Enum` if those instances exist. That makes `Finite` indeed a proper extension of `Bounded` and `Enum` without any sacrifices in the end.

I kept `lowest`, `lowestMaybe`, `highest` and `highestMaybe` to be considered as safe variants of `minBound` and `maxBound`, respectively.

clash-prelude/src/Clash/Class/Finite/Internal.hs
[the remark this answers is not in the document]
kleinreact: The different variants are available now.

clash-prelude/src/Clash/Class/Finite/Internal.hs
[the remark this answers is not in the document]
kleinreact: Conditional laws on existing `Bounded` and `Enum` instances have been added.

_(also see my comment above)_

clash-prelude/src/Clash/Explicit/Prelude.hs
DigitalBrains1: Sorry for straying a bit off course, but I still feel I should point it out:

> There is the common misconception that types with a `Num` instance and an `Enum` instance must satisfy `x + 1 = succ x` and `x - 1 = pred x`, but that's just not given. `Enum` just enumerates the elements according to **some** (_freely chosen_) index order, which does not have to be compatible with the notion of addition and subtraction at all.

https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#t:Enum says:

> `succ :: a -> a`
>
> Successor of a value. For numeric types, [succ](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:succ) adds 1.

> `pred :: a -> a`
>
> Predecessor of a value. For numeric types, [pred](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:pred) subtracts 1.

While the Haskell Report might be silent on what a type with both `Enum` and `Num` should do, the Haskell Prelude documentation does specify it.

kleinreact: > > `succ :: a -> a`
> >
> > Successor of a value. For numeric types, [succ](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:succ) adds 1.
>
> > `pred :: a -> a`
> >
> > Predecessor of a value. For numeric types, [pred](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:pred) subtracts 1.

I think that if they wanted to enforce that, then they would have added a law. But yeah, those two sentences are kind of misleading indeed.

DigitalBrains1: If the Haddock of a class definition is not normative, what is? They usually use the word _law_ in these cases, but there's no wiggle room in the doc that I see. If they wanted to document something that is common but not normative, I believe they would at the least add the word _usually_.

kleinreact: > My point is that these sentences are not misleading. Can you point me to an authoritative text that says otherwise? They are laws. They just didn't use the word law.

I'm not aware of any authoritative text, but the style of class requirement descriptions in `base` and `ghc-prim` varies a lot, so it doesn't look to me like they have a strict enforcement on wording at this point. I gave some examples below. According to those you need to take the `Applicative` laws very seriously for example, while the equality laws are more kind of a nice-to-have :upside_down_face:.

#### [Eq](https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#t:Eq)
_The Haskell Report defines no laws for [Eq](https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#t:Eq). However, [==](https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#v:-61--61-) is customarily expected to implement an equivalence relationship where two values comparing equal are indistinguishable by "public" functions, with a "public" function being one not allowing to see implementation details. For example, for a type representing non-normalised natural numbers modulo 100, a "public" function doesn't make the difference between 1 and 201. It is **expected to have** the following properties:_

#### [Ord](https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#t:Ord)
_The Haskell Report defines no laws for [Ord](https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#t:Ord). However, [<=](https://hackage.haskell.org/package/ghc-prim-0.6.1/docs/GHC-Classes.html#v:-60--61-) is **customarily expected to** implement a non-strict partial order and **have** the following properties:_

#### [Applicative](https://hackage.haskell.org/package/base-4.20.0.1/docs/Control-Applicative.html#t:Applicative)
_Further, any definition **must satisfy** the following:_

#### [Monad](https://hackage.haskell.org/package/base-4.20.0.1/docs/Control-Monad.html#t:Monad)
_Instances of [Monad](https://hackage.haskell.org/package/base-4.20.0.1/docs/Control-Monad.html#t:Monad) **should satisfy** the following:_

kleinreact: > `satSucc` and `satPred` are based on `Enum`.

Sure `satSucc` has a default implementation, which is `satSucc s n = satAdd s n 1`, but there is no requirement that I cannot chose a different implementation for my own type instances at this point. 

It's also a questionable, whether just because of that every class instance should depend on a `Num`. It would be much more user-friendly to have a default signature that requires `Num`, so the constraint is only required if the user wants to use the default in the first, e.g.,
```haskell
satSucc :: SaturationMode -> a -> a
-- Default method suitable for types that can represent the number 1
default satSucc :: Num a => SaturationMode -> a -> a
satSucc s n = satAdd s n 1
{-# INLINE satSucc #-}
```
This would elide the requirement on `Num` for every class instance.

DigitalBrains1: Regarding `satSucc` and `satPred` I have primarily not been talking about documentation or implementation, I'm talking about invisible things like design and intent. And also how they changed once we noticed deficiencies.

> Sure `satSucc` has a default implementation, which is `satSucc s n = satAdd s n 1`, but there is no requirement that I cannot chose a different implementation for my own type instances at this point.

We mean our documentation to be normative. So yes, I definitely interpret it as a requirement that they should add/subtract 1.¹ You yourself argued this above:

> Do I miss something fundamental here? I read that like `satSucc` is equivalent to `(+ 1)`, except for the boundary cases, where the behavior might differ. That's what I consider by "to compatible with each other".

I'd strongly argue that even in the boundary cases the behaviour matches adding/subtracting 1, it's just that `(+ 1)` does not actually add 1 in those cases.

(continuing with quote from your latest message)

> It's also a questionable, whether just because of that every class instance should depend on having a `Num` instance as well.

I think discussing creating another class for `satSucc` and `satPred` is out-of-scope here.

¹ Unless you're talking about boundary cases where the default implementation does not actually satisfy the laws, but then I don't understand why you bring it up.

DigitalBrains1:
Could you hide the three internal modules from the Haddock?

Also, I think we should have instances for `Clash.Sized.Fixed`. I don't know if there are any more obvious instances that were overlooked, but that one occurred to me.

kleinreact:
> Could you hide the three internal modules from the Haddock?

:+1:

DigitalBrains1:
Okay, when this is merged, could you create an issue that we should add the instance?

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,967 @@
...
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise #-}
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.KnownNat.Solver #-}
+{-# OPTIONS_GHC -fplugin GHC.TypeLits.Extra.Solver #-}
+
+{-# OPTIONS_HADDOCK not-home #-}
```
DigitalBrains1, line 24: Could you make this `hide`? I don't think internal modules should generally be in the generated Haddock (the fact that functions _have_ Haddock is definitely good, it just doesn't need to have an entry in the contents of the module).

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,967 @@
...
+
+  gIndex = \case
+    L1 x -> extend (gIndex x)
+    R1 x -> extend (gIndex x) + natToNum @(GElementCount a)
+
+instance Finite Void
+instance Finite ()
+instance Finite Bool
+instance Finite Ordering
+instance Finite (Proxy a)
```
DigitalBrains1, line 396: `SNat` is really just a `Proxy` with benefits, so just like we have the `Proxy` instance I feel we should have an `SNat` instance.

Same for `SSymbol`.

clash-prelude/src/Clash/Class/Finite/Internal.hs
kleinreact: No, as haddock does not find the missing references then:
https://gitlab.com/clash-lang/clash-compiler/-/pipelines/1620169101

DigitalBrains1: \*shakes fist at Haddock\* 😠

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,1006 @@
...
+
+  -- | The elements of the type.
+  elements :: Vec (ElementCount a) a
+  default elements ::
+    ( Generic a, GFinite (Rep a)
+    , ElementCount a ~ GElementCount (Rep a)
+    ) => Vec (ElementCount a) a
+  elements = to <$> gElements
+
+  -- | The @0@ indexed element.
```
DigitalBrains1, line 183:
```suggestion
  -- | The element at index 0.
```

Just a suggestion... it's just that to me it feels a zero-indexed something is something that has indexes starting at 0, like most arrays and stuff in computer science.

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,1006 @@
...
+  lowest = to gLowest
+
+  -- | Just the @0@ indexed element. Nothing if @ElementCount a = 0@.
+  lowestMaybe :: Maybe a
+  default lowestMaybe ::
+    ( Generic a, GFinite (Rep a)
+    ) => Maybe a
+  lowestMaybe = to <$> gLowestMaybe
+
+  -- | The @(ElementCount a - 1)@ indexed element
```
DigitalBrains1, line 197:
```suggestion
  -- | The element at index @(ElementCount a - 1)@
```

clash-prelude/src/Clash/Class/Finite/Internal.hs
```diff
@@ -0,0 +1,1006 @@
...
+  predMaybe    = fmap Down . succMaybe . getDown
+  succMaybe    = fmap Down . predMaybe . getDown
+  ith          = Down . ith . (maxBound -)
+  index        = (maxBound -) . index . getDown
+#endif
+
+-- | A newtype wrapper that reverses the index order, which normally
+-- would be used by the 'Finite' instance of the inner type via
+-- 'Generic' deriving. The newtype is only intended be used with the
+#if MIN_VERSION_base(4,15,0)
```
DigitalBrains1, line 744: This does not appear to work. CPP and Haddock do not play nice. I think it works better with the `{-` construction, but still, you'll have to jump through some hoops and you're probably better off with a little duplication. Otherwise you'll get CPP-generated markers as part of your Haddock.

Please test with GCC versions both matching the `#if` and the `#else`.

Note how your documentation ends at "used with the" in the Haddock generated in CI:
https://clash-lang.gitlab.io/-/clash-compiler/-/jobs/8839034866/artifacts/hadocs/clash-prelude/clash-prelude/Clash-Class-Finite.html#t:GenericReverse

DigitalBrains1 (changes requested):
I had another pass over the Haddock and feel there's still some needed changes.

clash-prelude/src/Clash/Class/Finite/Internal.hs
DigitalBrains1: At 0 I said "just a suggestion" but I changed my mind. I think

> The `(ElementCount a - 1)` indexed element

is an odd sentence, and I would like it changed. That also means it makes complete sense to change the 0-indexed phrasing.

clash-prelude/src/Clash/Class/Finite.hs
```diff
@@ -0,0 +1,41 @@
...
+infinitely many inhabitants. The type of @'succ', 'pred' :: 'Enum' a
+=> a -> a@ clearly reflects this design choice, as it assumes that
+every element has a successor and predecessor, which makes perfect
+sense for infinite types, but requires finite types to error on
+certain inputs. Wrapping behavior is forbidden according to the
+documentation of 'Enum' (assuming that finite types usually have a
+'Bounded' instance) such that 'Enum' instances must ship partial
+functions for most finite types. Likewise, the number of inhabitants
+does not align with the number of indices offered by 'Int' for most
+types, which 'Finite' resolves by using 'Index n' instead.
```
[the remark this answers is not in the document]
kleinreact: Yes, it does.

clash-prelude/src/Clash/Class/Finite.hs
```diff
@@ -0,0 +1,41 @@
...
+-}
+module Clash.Class.Finite
+  ( -- * Finite Class
+    Finite(..)
+    -- * Extensions
+  , GenericReverse(..)
+  , WithUndefined(..)
```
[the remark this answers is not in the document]
kleinreact: I made `BoundedEnumEq` also available via the non-internal class. The name of the type is quite artificial, but I don't think it will hurt in any sense at this point.

clash-prelude/src/Clash/Class/Finite/Internal/Dictionaries.hs
```diff
@@ -0,0 +1,83 @@
...
+powLawsRewrite ::
+  forall (a :: Nat) (n :: Nat).
+  Dict ((a^(2^(n + 1))) ~ ((a^(2^n)) * (a^(2^n))))
+powLawsRewrite = unsafeCoerce (Dict :: Dict (0 ~ 0))
+
+#if !MIN_VERSION_base(4,16,0)
+-- | Evidence that exponentiation and clog are dual to each other.
+powCLogDual ::
+  forall (a :: Nat) (n :: Nat).
+  Dict (CLog a (a^n) ~ n)
```
leonschoorl, line 74: This isn't correct for `a < 2`.
As `1 ^ n = 1` for any `n`, so out can't get it back out with (c)log.
And similar for 0.

DigitalBrains1: I suspect maybe the following law is enough for `Finite`?

```suggestion
powCLogDual ::
  forall (n :: Nat).
  Dict (CLog 2 (2^n) ~ n)
```

DigitalBrains1: Yes, the following compiles under 8.10.7, so I think that works out

```diff
--- a/clash-prelude/src/Clash/Class/Finite/Internal.hs
+++ b/clash-prelude/src/Clash/Class/Finite/Internal.hs
@@ -886,7 +886,7 @@ instance KnownNat n => Finite (WithUndefined (BitVector n)) where
       , Dict <- powPositiveIfPositiveBase @2 @n
       , Dict <- powPositiveIfPositiveBase @3 @n
 #if !MIN_VERSION_base(4,16,0)
-      , Dict <- powCLogDual @2 @n
+      , Dict <- powCLogDual @n
       , Dict <- leqOnePlusMinus @(2^n) @(3^n)
 #endif
       = second (pack . complement . truncateB @Index @(2^n) @(3^n - 2^n))
diff --git a/clash-prelude/src/Clash/Class/Finite/Internal/Dictionaries.hs b/clash-prelude/src/Clash/Class/Finite/Internal/Dictionaries.hs
index 631430d61..1769d6512 100644
--- a/clash-prelude/src/Clash/Class/Finite/Internal/Dictionaries.hs
+++ b/clash-prelude/src/Clash/Class/Finite/Internal/Dictionaries.hs
@@ -70,8 +70,8 @@ powLawsRewrite = unsafeCoerce (Dict :: Dict (0 ~ 0))
 #if !MIN_VERSION_base(4,16,0)
 -- | Evidence that exponentiation and clog are dual to each other.
 powCLogDual ::
-  forall (a :: Nat) (n :: Nat).
-  Dict (CLog a (a^n) ~ n)
+  forall (n :: Nat).
+  Dict (CLog 2 (2^n) ~ n)
 powCLogDual = unsafeCoerce (Dict :: Dict (0 ~ 0))
 
 -- | Evidence that substraction and addition of the same nat cancels
```

kleinreact: Thanks for checking and the suggestion. I changed it accordingly.

DigitalBrains1 (approved):
I strongly feel the functions should have been called `fpred` and `fsucc` (or `finSucc`), but since you've made it clear that this is not up for discussion, I suppose I'll yield and approve.

martijnbastiaan (commented):
I'm going to neither approve nor deny this PR on the grounds that I  strongly feel squatting the names `succMaybe` in the Prelude is _no bueano_. The functions `Counter` exposes could just as well have been called this with slightly different design decisions. They should be called `finSucc` (in line with `satSucc` and `countSucc`) or _maybe_ `fsucc`.

kleinreact:
> The functions Counter exposes could just as well have been called this with slightly different design decisions. They should be called finSucc (in line with satSucc and countSucc) or maybe fsucc.

I propose to switch your point of view, as `satSucc` and `countSucc` can have other semantics than `succ`, while `succMaybe` is required, by the `Enum Compatibility` law, to expose the same semantics as `Just . succ` for all the preimage of `succ`. 

Thus, using the name `succMaybe` makes this semantic connection explicit, while names like `finSucc` of `fSucc` do not. The same applies for `pred` and `predMaybe`.

martijnbastiaan:
I don't think I agree with your logic, but more importantly: you are basing your arguments on very subtle details that will be lost on the majority of Clash users. As a result, the API will be less memorable.

kleinreact:
> you are basing your arguments on very subtle details

Why do you consider the semantics of a function to be subtle details?

> that will be lost on the majority of Clash users. As a result, the API will be less memorable.

How do you know that?

DigitalBrains1:
Also note that when we add the instance for `Fixed`, the semantics of the `pred` and `succ` of this class will no longer match the semantics of `Enum` for `Fixed`.

I really don't feel like participating in this discussion otherwise, but you raise an important point: your laws prohibit compatibility with our existing `Enum` instance for `Fixed`. You say both that `Finite` should enumerate all elements and that it should behave as `succ` and `pred` from `Enum` which don't do that, they add and subtract one.

DigitalBrains1 (changes requested):
The laws are contradictory. They state `predMaybe` and `succMaybe` should enumerate all values, and simultaneously that they need to be compatible with `Enum`. Our `Fixed` type will break this as it can not satisfy both. It also points out how they are overconstrained. The `Enum` compatibility law probably needs to be removed; the existing laws should already fix the behaviour of `predMaybe` and `succMaybe`.

kleinreact:
> Also note that when we add the instance for Fixed, the semantics of the pred and succ of this class will no longer match the semantics of Enum for Fixed.

It looks to me like the `Enum` instance for `Fixed` is wrong then. I consider the successor of `1.5 :: UFixed 2 1` to be `2`and not `2.5` according to the Haskell 2010 Language Report.

The current `Enum (UFixed 2 1)` instance implements a similar behavior as the `Enum` instance for `Float`, but the report makes an explicit exception for that particular type. It defines:

_For all **four** numeric types_ **(which are only `Int`, `Integer`, `Float`, and `Double`)** _, succ adds 1, and pred subtracts 1_.

For **any** other type the semantics is:

_Class `Enum` deﬁnes operations on sequentially ordered types. The functions succ and pred return the
successor and predecessor, respectively, of a value._

The sequential order of `UFixed 2 1` is
```
0.0, 0.5, 1.0, 1.5, 2.0, 2.5, 3.0, 3.5
```
and, thus, the successor of `1.5` must be `2.0`.

As `UFixed 2 1` also has a `Bounded` instance, there even is proof that the current implementation is wrong, because in that case it is also required that

```haskell
enumFromThen x y = enumFromThenTo x y bound
  where
    bound | fromEnum y >= fromEnum x = maxBound
          | otherwise = minBound
```
which does not hold for `UFixed 2 1`, `x = 2.5` and `y = 2.0`, resulting in
```
[2.5,2.0,1.5,1.0,0.5,0.0] = []
```

DigitalBrains1:
Yeah, we're not going to change `Fixed` again. Maybe changing it to conform to the class's documentation in the first place was a mistake, but we're not changing it back. We already had the fallout of at least one Clash user losing a lot of time trying to find the regression when we changed it the first time. Plus, being able to get the intuitive behaviour of `(+ 1)` for degenerate instances of `Fixed` is also nice.

Also, you're missing the bigger point, it doesn't matter what `Fixed` does. The `Enum` laws do not require the type to enumerate each and every element, but your `Finite` laws do. So somebody out there might already have a type for which they've written an `Enum` and now want to add a `Finite`. You can't at the same time fully fix the behaviour by requiring it to enumerate each and every possible value and require it to conform to `Enum` as well, which does not have that requirement.

kleinreact:
> Yeah, we're not going to change Fixed again.

That's not an issue of this PR. It's not conform with `Haskell2010` as it is implemented right now. I opened #2888 to keep that separate .

> The `Enum` laws do not require the type to enumerate each and every element, but your Finite laws do.

This is only true for `Int`, `Integer`, `Float`, `Double` and infinite types. Every finite type that has some  sequential order is able to traverse every element according to that order. Otherwise it should not be an instance of `Enum` in the first place.

> You can't at the same time fully fix the behaviour by requiring it to enumerate each and every possible value and require it to conform to Enum as well, which does not have that requirement.

Why not? Every type that doesn't satisfy this requirement simply is not eligible to be an instance of `Finite`. I don't require anything. I define the class to capture only those types that have that property.

> Finally, if they really wanted succ and pred to be the successor and predecessor of all possible values a type can take, why didn't they do this for `Float` and `Double`?

`Float` and `Double` are infinite types, where all the considerations made by `Finite` simply don't apply. Sure, on the finite state machines on our desks they are represented by finitely many bits, but on an analog computation architecture, for example, that might not be a limitation.

> Why did you re-request a review?

Because you requested changes with the assumption that the laws are contradictory. However, I don't see anything contradicting. If a type cannot satisfy the laws, it simply has no `Finite` instance.

> What exactly do you foresee your additional requirement add?

The user shall be able to expect that `succ` and `succMaybe` don't behave contradictory to each other. Hence, any implementation of a `Finite` instance needs to take this into account.

DigitalBrains1 (changes requested):
> If a type cannot satisfy the laws, it simply has no `Finite` instance.

I don't agree and I am _way_ beyond done with all the discussion in this PR.

Unless you remove the `Enum` compatibility law, I do not want to approve this PR. I'm also still quite conflicted about the naming.

kleinreact:
> I don't agree and I am way beyond done with all the discussion in this PR.
>
>Unless you remove the Enum compatibility law, I do not want to approve this PR. I'm also still quite conflicted about the naming.

I understand. The Enum compatibility law is an essential part of the intended design.

Thank you all for the feedback.

kleinreact:
For all those who are still interested in using this. The package is now available on Hackage ([clash-finite](https://hackage.haskell.org/package/clash-finite-1.0.0.0)) and the sources are available [here](https://github.com/kleinreact/clash-finite).

PR #2871, opened 2025-01-06
Linear Clash.Sized.Vector.reverse

kleinreact:
The current implementation of `Clash.Sized.Vector.reverse` has a simulation runtime that is quadratic in the input. This PR improves this to a linear version.

## Still TODO:

  - [x] ~Write a changelog entry (see changelog/README.md)~


martijnbastiaan (approved):
Wow, nice find.

PR #2873, opened 2025-01-13
Avoid warnings instead of hiding them

kleinreact:
Certain GHC warnings are currently disabled in in `Clash.Prelude` to hide the fact some imports and bindings are not used, and that some of the patterns in `Clash.Sized.Vector` are incomplete. As this is avoidable though, and having those flags enabled may also hide problems that are easy to catch otherwise. Therefore, this PR just avoids production of the respective warnings in the first place.

The only current exception is `Clash.Tutorial`, where the unused imports are used to have shorter Haddock references. We also can avoid that one by using fully qualified imports for the references instead. I'm in favor of introducing those changes as part of this PR too, but please let me know first, whether that's of common interest, as it takes a few minutes to apply all the changes along the file.

## Requires (parts of)
* #2843 (3c1953f in particular)

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Sized/Index.hs
+++ b/clash-prelude/src/Clash/Sized/Index.hs
@@ -1,5 +1,6 @@
 {-|
 Copyright  :  (C) 2013-2016, University of Twente
+                  2025     , QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
 -}
@@ -18,7 +19,6 @@ where
 import GHC.TypeLits (KnownNat, type (^))
 import GHC.TypeLits.Extra (CLog) -- documentation only
 
-import Clash.Promoted.Nat (SNat (..), pow2SNat)
 import Clash.Sized.Internal.BitVector (BitVector)
 import Clash.Sized.Internal.Index
 
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -8,6 +8,7 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
 
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE EmptyCase #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
@@ -22,8 +23,6 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.KnownNat.Solver #-}
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise       #-}
-
-{-# OPTIONS_GHC -fno-warn-incomplete-patterns -fno-warn-redundant-constraints #-}
 
 {-# OPTIONS_HADDOCK show-extensions #-}
 
@@ -184,6 +183,9 @@ data Vec :: Nat -> Type -> Type where
 
 {-# COMPLETE Nil, (:>) #-}
 
+-- | Tag for K1: @n ~ 0@ proof
+data N
+
 -- | In many cases, this Generic instance only allows generic
 -- functions/instances over vectors of at least size 1, due to the
 -- /n-1/ in the /Rep (Vec n a)/ definition.
@@ -194,7 +196,7 @@ data Vec :: Nat -> Type -> Type where
 instance KnownNat n => Generic (Vec n a) where
   type Rep (Vec n a) =
     D1 ('MetaData "Vec" "Clash.Data.Vector" "clash-prelude" 'False)
-      (C1 ('MetaCons "Nil" 'PrefixI 'False) U1 :+:
+      (C1 ('MetaCons "Nil" 'PrefixI 'False) (K1 N (Dict (n ~ 0))) :+:
        C1 ('MetaCons "Cons" 'PrefixI 'False)
         (S1 ('MetaSel 'Nothing
                 'NoSourceUnpackedness
@@ -206,13 +208,14 @@ instance KnownNat n => Generic (Vec n a) where
                 'NoSourceStrictness
                 'DecidedLazy)
             (Rec0 (Vec (n-1) a))))
-  from Nil         = M1 (L1 (M1 U1))
+  from Nil         = M1 (L1 (M1 (K1 Dict)))
   from (Cons x xs) = M1 (R1 (M1 (M1 (K1 x) :*: M1 (K1 xs))))
   to (M1 g) = case compareSNat (SNat @n) (SNat @0) of
     SNatLE -> case leZero @n of
       Sub Dict -> Nil
     SNatGT -> case g of
       R1 (M1 (M1 (K1 p) :*: M1 (K1 q))) -> Cons p q
+      L1 (M1 (K1 eqZero)) -> case eqZero of {}
 
 instance (KnownNat n, Typeable a, Data a) => Data (Vec n a) where
   gunfold k z _ = case compareSNat (SNat @n) (SNat @0) of
@@ -449,6 +452,12 @@ singleton = (`Cons` Nil)
 -}
 head :: Vec (n + 1) a -> a
 head (x `Cons` _) = x
+#if !MIN_VERSION_base(4,16,0) || MIN_VERSION_base(4,17,0)
+head xs = unreachable xs
+ where
+  unreachable :: forall n a. 1 <= n => Vec n a -> a
+  unreachable (x `Cons` _) = x
+#endif
 
 -- See: https://github.com/clash-lang/clash-compiler/pull/2511
 {-# CLASH_OPAQUE tail #-}
@@ -494,7 +503,13 @@ head (x `Cons` _) = x
 #endif
 -}
 tail :: Vec (n + 1) a -> Vec n a
-tail (_ `Cons` xs) = xs
+tail (_ `Cons` xr) = xr
+#if !MIN_VERSION_base(4,16,0) || MIN_VERSION_base(4,17,0)
+tail xs = unreachable xs
+ where
+  unreachable :: forall n a. 1 <= n => Vec n a -> Vec (n - 1) a
+  unreachable (_ `Cons` xr) = xr
+#endif
 
 -- See: https://github.com/clash-lang/clash-compiler/pull/2511
 {-# CLASH_OPAQUE last #-}
@@ -541,7 +556,13 @@ tail (_ `Cons` xs) = xs
 -}
 last :: Vec (n + 1) a -> a
 last (x `Cons` Nil)         = x
-last (_ `Cons` y `Cons` ys) = last (y `Cons` ys)
+last (_ `Cons` y `Cons` xr) = last (y `Cons` xr)
+#if !MIN_VERSION_base(4,16,0) || MIN_VERSION_base(4,17,0)
+last xs = unreachable xs
+ where
+  unreachable :: 1 <= n => Vec n a -> a
+  unreachable ys@(Cons _ _) = last ys
+#endif
 
 -- See: https://github.com/clash-lang/clash-compiler/pull/2511
 {-# CLASH_OPAQUE init #-}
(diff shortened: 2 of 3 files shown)


clash-prelude/src/Clash/Sized/Index.hs
```diff
@@ -29,7 +27,7 @@ import Clash.Sized.Internal.Index
 -- That is, the type of 'Clash.Class.BitPack.unpack' is:
 --
 -- @
--- __unpack__ :: 'BitVector' ('CLog' 2 n) -> 'Index' n
+-- __unpack__ :: 'BitVector' ('GHC.TypeLits.Extra.CLog' 2 n) -> 'Index' n
```
DigitalBrains1, line 30: There are two reasons we do it with an import:
*  Because we get a CI build failure if we make a typo. If you link to an identifier that is not in scope, Haddock does not check whether the identifier exists.
* It also doesn't check whether it's a type or a value; in this case, the link changes from a link to the type to a link to the value, which doesn't exist. We can fix that with an annotation, except for that we'd first need to update our Haddock generation to a more recent version so it understands the annotation.

In other words, under current circumstances, this link is broken.

More in general, I feel you should _always_ manually check Haddock produced the result you expected, and you have to pay a lot of attention. In my experience, the only way to not make mistakes is to spot them in a careful check afterwards.

clash-prelude/src/Clash/Examples/Internal.hs
```diff
@@ -13,16 +14,10 @@ Licence   : Creative Commons 4.0 (CC BY 4.0) (https://creativecommons.org/licens
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
 
-{-# OPTIONS_GHC -fno-warn-unused-imports #-}
-{-# OPTIONS_GHC -fno-warn-unused-binds #-}
-{-# OPTIONS_GHC -fno-warn-type-defaults #-}
-{-# OPTIONS_GHC -fno-warn-missing-signatures #-}
-{-# OPTIONS_GHC -fno-warn-incomplete-patterns #-}
-
```
DigitalBrains1, line 21: I think there is one reason and one reason only for the existence of this module: to, in a very clunky way, check that the code we have in `Clash.Examples` actually compiles.

So if you then make the code in `Clash.Examples.Internal` something different than what is in `Clash.Examples` is really a bad idea.

And it is debatable, but I feel some warnings in code in `Clash.Examples` is fine if that makes it more accessible to the beginner. That especially goes for warnings that are not turned on by default.

So it is my opinion we should leave `Clash.Examples` and `Clash.Examples.Internal` as they are.

DigitalBrains1 (changes requested):
Woops, my comment directly above was meant to be part of a review.

Speaking in general, I'm all in favour of this effort and I thank you for fiddling with it and trying to get it all working.

But I do have some comments regarding documentation and documentation-related subjects.

clash-prelude/src/Clash/Sized/Index.hs
DigitalBrains1: > you should always manually check Haddock produced the result you expected

And please take care to use the exact same setup as we use for publishing to Hackage. Stuff changes a lot between versions of Haddock.

DigitalBrains1: I guess that also answers your suggestion in the cover letter:

> The only current exception is `Clash.Tutorial`, where the unused imports are used to have shorter Haddock references. We also can avoid that one by using fully qualified imports for the references instead.

This is unfortunately incorrect. The two are not comparable.

kleinreact: > Because we get a CI build failure if we make a typo. If you link to an identifier that is not in scope, Haddock does not check whether the identifier exists.

What exactly do you mean by that? Do you refer to a typo in the unused import or the documentation? Haddock won't complain about a typo in the documentation in neither case, so I don't see any benefit of using unused imports here.

> It also doesn't check whether it's a type or a value; in this case, the link changes from a link to the type to a link to the value, which doesn't exist. We can fix that with an annotation, except for that we'd first need to update our Haddock generation to a more recent version so it understands the annotation.

If there is the potential for an ambiguous reference, then you have to add the `t` or `v` marker, as described in [the docs](https://haskell-haddock.readthedocs.io/latest/markup.html#markup). 

> More in general, I feel you should always manually check Haddock produced the result you expected, and you have to pay a lot of attention. In my experience, the only way to not make mistakes is to spot them in a careful check afterwards.

The purpose of haddock is to generate documentation, not to check anything. It's the **full** responsibility of the user to produce correct documentation, who even might want that reference is broken. Haddock does not take any responsibility for judging about correct documentation, because that is not what it is made for.

> We run .ci/build_sdist.sh under GHC 8.10.7 to build the documentation

In which sense is the compiler version relevant for the invocation of haddock? If at all, the haddock version should be of any relevance, but I cannot find any requirements on that one in the `cabal.project` or elsewhere. Accordingly, our documentation must be correct for any version of haddock the user might wanna use, right? 

Users can only be expected to clone the repo and run `cabal haddock`, but I doubt that we should require them to use a particular (old) version of GHC for building their docs.

> This is unfortunately incorrect. The two are not comparable. Well, if you mean "if we fully qualify all identifiers", I don't think you can change -Wunused-imports by fully qualifying imports, but I don't think you meant that.

I proposed to change identifiers like `Num` to `GHC.Enum.Num` in the documentation, which is equivalent from a haddock perspective to using `Num` and importing `GHC.Enum` for bringing the name into scope. 

I don't see why those are incomparable. Can you maybe a bit more elaborate your thoughts on that?

clash-prelude/src/Clash/Examples/Internal.hs
kleinreact: > I think there is one reason and one reason only for the existence of this module: to, in a very clunky way, check that the code we have in Clash.Examples actually compiles.

Ah, I see. But isn't it kind of bad practice then to suggest beginners to use incomplete patterns or not to add type signatures. I mean, there is a good reason why GHC produces warnings in those cases.

In that sense, I'd say we have to update `Clash.Examples` as well, to also avoid those warnings.

clash-prelude/src/Clash/Sized/Index.hs
DigitalBrains1: > What exactly do you mean by that? Do you refer to a typo in the unused import or the documentation? 

If you make a typo in `'CLog'`, Haddock will emit a

```
Warning: 'CLog' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
```

which we [explicitly check for](https://github.com/clash-lang/clash-compiler/blob/d9e2a53fe81eb23616d9fbd2e7944845c5c6f6c4/.ci/build_docs.sh#L42) in CI.

whereas if you make a typo in `'GHC.TypeLits.Extra.CLog'` anywhere, you'll never get any warning, just a not-working link.

> If there is the potential for an ambiguous reference, then you have to add the `t` or `v` marker, as described in [the docs](https://haskell-haddock.readthedocs.io/latest/markup.html#markup).

... which doesn't work under GHC 8.10.7. Hence why I said _except for that we'd first need to update our Haddock generation to a more recent version so it understands the annotation._

> The purpose of haddock is to generate documentation, not to check anything.

Obviously I'm talking about checking whether the result is what you intended. Haddock could do way more to warn when something looks fishy, and its parsing could be way less ambiguous. Also, I find this whole avenue of discussion unproductive, this will be the last I will say about it.

> It's a classical [WYSIWYM](https://en.wikipedia.org/wiki/WYSIWYM) application

Oh man, thanks, I needed a laugh. The following:

```
What you see/what I see? We may
never know/be certain.
```

renders as

> What you see/what I see? We may never know/be certain.

However, those lines are a bit short, we have 80 columns, let's reflow.

```
What you see/what I see? We may never know/be certain.
```

which renders as

> What you see*what I see? We may never know*be certain.

If I change documentation, I might reflow the paragraph to fit within the 80-char limit. Slashes that were previously on separate lines might now come together, and *I didn't even change the content of that part of the paragraph*. This is just one of the many ways in which you can accidentally introduce wrong Haddock.

For this specific case, I feel that a single slash on a line should be a hard parsing error. If you mean a literal slash, you should escape it. If you mean italics, there should be two.

There are much more insidious cases, this is just my go-to explanation of Haddock pitfalls.

> In which sense is the compiler version relevant for the invocation of haddock?

They're tightly coupled. You cannot get a different version of Haddock unless you recompile GHC (and I doubt it would work).

> Accordingly, our documentation must be correct for any version of haddock the user might wanna use, right?

Yes, but I am talking about the version we use to upload to Hackage. We can't help that different versions of Haddock unfortunately sometimes completely differently interpret markup and there is no way to write something that is compatible with both unless you `#ifdef` the whole lot. Luckily, it's mostly pretty okay.

> I proposed to change identifiers like `Num` to `GHC.Enum.Num` in the documentation, which is equivalent from a haddock perspective to using `Num` and importing `GHC.Enum` for bringing the name into scope.

It is not. With unqualified identifiers, it checks whether they exist, with qualified identifiers, you get no checks at all, not even if the module is in scope. And it will also check to see if something does not live in the term namespace so it will generate a type link instead. Of course this falls apart when an identifier lives in both spaces.

> I don't see why those are incomparable. Can you maybe a bit more elaborate your thoughts on that?

I just did :-D.

clash-prelude/src/Clash/Examples/Internal.hs
DigitalBrains1: > I mean, there is a good reason why GHC produces warnings in those cases.

It doesn't by default.

```haskell
module Warnings where

f True = 1
```
```
$ ghc-9.10 Warnings.hs
[1 of 1] Compiling Warnings         ( Warnings.hs, Warnings.o )
$
```

DigitalBrains1: > But isn't it kind of bad practice then to suggest beginners to use incomplete patterns

That's why I said it's debatable. My point of view is: No, it's more important to focus on the basics first, and only once the reader understands your code snippet tell them: also, there's another issue with this, and it's incomplete patterns.

If we change our examples to no longer have incomplete patterns or missing type signatures (which were probably conscious decisions made when writing it), we need to discuss that in a larger forum, with more people working on Clash. I don't think we should do that in this PR, but rather, for instance, on Discourse.

clash-prelude/src/Clash/Sized/Index.hs
kleinreact: Thanks for the in-depth feedback. I wasn't aware of the fact that Haddock handles qualified imports differently than non-qualified ones. I never stumbled across that.

To take the discussion to an end: I will update the PR then in the sense that disabling unused import warnings is not touched. I agree on your argument that that one probably never is going to cause any harm.

clash-prelude/src/Clash/Examples/Internal.hs
kleinreact: Alright, I revert the examples to be untouched then. Thanks.

clash-prelude/src/Clash/Sized/Index.hs
kleinreact: It's updated.

clash-prelude/src/Clash/Examples/Internal.hs
kleinreact: It's reverted.

DigitalBrains1 (approved):

DigitalBrains1:
Why did you include part of PR #2843 here? Is there something blocking the merge of PR #2843?

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -450,6 +452,12 @@ singleton = (`Cons` Nil)
 -}
 head :: Vec (n + 1) a -> a
 head (x `Cons` _) = x
+#if !MIN_VERSION_base(4,16,0) || MIN_VERSION_base(4,17,0)
```
DigitalBrains1, line 455: So why are all these bits unneeded on GHC 9.2 (base 4.16)? Is it possible to briefly explain that? Is it because of some `CmpNat` thing that got reverted in GHC 9.4?

kleinreact:
> Why did you include part of PR #2843 here?

@DigitalBrains1 The version of `select` from #2843 is just much better suited for getting rid of the incomplete patterns, which  is why I just cherry picked that version to avoid unnecessary extra work.

> Is there something blocking the merge of #2843?

Not from my side. Shall I merge it?

clash-prelude/src/Clash/Sized/Vector.hs
kleinreact: I haven't checked the reason, but 9.2 is the only version being able to detect the missing patterns to be unreachable. Hence, you have to exclude them there explicitly, because you get a warning about them being non-reachable otherwise :upside_down_face:.

DigitalBrains1: Hahaha! 🙃 indeed!

DigitalBrains1:
> Shall I merge it?

Point 1: _definitely_ not without rebasing first, after this long a time you really can't trust that that old CI success is still valid!

Point 2: I think it needs approving reviews first. Rowan and I both raised issues, you resolved the issues, but I don't see any approving reviews yet. I think the only thing that's keeping me personally from an approving review, after I've now re-studied it, is that I didn't check yet whether all the changed code is well covered in unit tests and/or HDL tests. It looks fine, but does it work fine? :-D

So perhaps I need to go hunt for existing tests and then either approve the PR or ask you to produce tests if they are not well covered.

PR #2874, opened 2025-01-15
Combinatorial => combinational

martijnbastiaan:
See https://en.wikipedia.org/wiki/Combinational_logic. ~~Combinatorial is a
mathematical term related to counting.~~

Edit: Seems like it's not necessarily wrong, though "combinational" seems to be more popular. We mix both currently.


DigitalBrains1 (approved):
Yeah, as the hatnote at [wp:Combinatorics](https://en.wikipedia.org/wiki/Combinatorics) implies, _combinatorial_ is not wrong. Not that I'd trust Wikipedia to always get this right...

If we said _combinatory_, we'd be wrong (that's what your Wikipedia article shows).

I think consistency is good, and I'm fine with this change, but the inconsistency never really bothered me. It's just written by different people with different customs. As long as we don't switch inside a single piece of Haddock it's fiinnnnnne.

kleinreact (approved):
Yes, both are correct, but I also prefer `combinational`, as it is more common in electrical engineering, and having consistent wording in general.

Fun fact, for me:
* [Combinational MAC](https://duckduckgo.com/?q=combinational+MAC&t=ffab&ia=web) lists the wiki article first, while
* [Combinatorial MAC](https://duckduckgo.com/?q=combinatorial+MAC&t=ffab&ia=web) first lists some medical pages, then the wiki and then our `clash-lang.readthedocs.io`, 

which kind of hints that `combinational` indeed is the better choice.

PR #2876, opened 2025-01-16
New: ddrForwardClock

DigitalBrains1:
This PR was inspired by the `clash-cores` [PR #7 Add RGMII core](https://github.com/clash-lang/clash-cores/pull/7) which uses this construction but keeps the output clock as a `Signal domDDR Bit`. With the new `ddrOutClock`, we can correctly declare the thing to be a clock.

There's a qualitative difference between `Clock` and `Signal _ Bit`. In a nutshell, a signal is relative to a clock, it has setup and hold constraints and carries data. All of these don't apply to a clock. You could say a clock leads and a signal follows. The resulting output on the pin is no different, it's just about correctly declaring what something is in the Haskell world.

The new `ddrForwardClock` function gets a DDR output primitive as an argument. If you look at `clash-prelude`, the only directly fitting one is `Clash.Explicit.DDR.ddrOut` (if you apply the reset value first). However, in a PR that I will create soon, the API for `Clash.Intel.DDR` and `Clash.Xilinx.DDR` is changed to accomodate the signature introduced in this PR. Even with the current code, it's possible to adapt the vendor primitives to fit `ddrForwardClock`, but soon it will be a natural fit.

## Still TODO:

### DIFF
--- a/changelog/2025-01-16T14_20_35+01_00_ddroutclock
+++ b/changelog/2025-01-16T14_20_35+01_00_ddroutclock
@@ -1,0 +1,1 @@
+NEW: The function `Clash.Explicit.DDR.ddrForwardClock`, which uses a DDR output primitive to forward a clock signal to a DDR-capable output pin.
--- a/clash-prelude/src/Clash/Explicit/DDR.hs
+++ b/clash-prelude/src/Clash/Explicit/DDR.hs
@@ -1,8 +1,9 @@
 {-|
 Copyright  :  (C) 2017, Google Inc
                   2019, Myrtle Software Ltd
+                  2025, QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
-Maintainer :  Christiaan Baaij <christiaan.baaij@gmail.com>
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
 We simulate DDR signal by using 'Signal's which have exactly half the period
 (or double the speed) of our normal 'Signal's.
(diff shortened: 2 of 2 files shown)


DigitalBrains1:
GHC 8.10.7 should get its priorities straight:

```
    • Expected kind ‘Nat’,
        but ‘DomainPeriod domIn ~ 2’ has kind ‘Constraint’
    • In the first argument of ‘(*)’, namely ‘DomainPeriod domIn ~ 2’
      In the type signature:
[...]
                        DomainPeriod domIn ~ 2 * DomainPeriod domDdr) =>
[...]
```

Why is this same expression no longer a problem in `clash-cores` [PR #7 Add RGMII core](https://github.com/clash-lang/clash-cores/pull/7)? I couldn't get it to work here. I did notice that code had fewer `KnownDomain`s than I expected; if you look at our docs, a use of `DomainPeriod` is documented as needing `KnownDomain` in its example, but apparently you don't?

I'll remove some `KnownDomain`s and add parentheses to the expression.

rowanG077:
Yes this is very nice! Thank you!

rowanG077 (approved):

rowanG077:
Perhaps `ddrForwardClock` is a more appropriate name as this technique is generally referred to a clock forwarding, or at least that's what inspired me.

DigitalBrains1:
I've changed the name from `ddrOutClock` to `ddrForwardClock`.

PR #2891, opened 2025-02-15
Switch away from data-default-class

DigitalBrains1:
data-default used to be split into several smaller packages, with data-default itself just depending on them all and re-exporting the class. These days, data-default contains everything and data-default-class just exists for easier upgrades, it just depends on data-default itself.

Note that the split stll carries a risk of confusion downstream; we cannot help this. Suppose one package uses the new data-default 0.8 and another uses an old data-default-class <0.2. The packages could both be dependencies, or one is a (dependency of a) dependency and the other is the primary package. You can get a bewildering error message "No instance for (Default T)" because there exist two different classes both called `Default` and the error message doesn't qualify the class name.

I suppose the worst situation you can have is where one function `f` has a `Default` constraint, and you want to apply `f` in your program both to:
1.  Types with a `Data.Default.Default` instance from `data-default` 0.8
2. Types with a `Data.Default.Class.Default` instance from `data-default-class` <0.2.

The only solution that occurs to me is adding an orphan instance for the type missing the needed instance, which of course is grimy.

No, wait. There's another bad solution that is nonetheless less bad. You could `allow-newer` the package depending on the old `data-default-class`, so it will pick the new 0.2 that _is_ compatible with the new `data-default`. I see no way how that could go wrong. Of course, this doesn't help other people in turn adding your package as a dependency, this solution will always have to be applied at the project level.

I suppose we should be glad not that much uses `Default`. And we could seriously consider stopping doing so ourselves, but that is for later.

## Still TODO:


martijnbastiaan (approved):
> And we could seriously consider stopping doing so ourselves, but that is for later.

Yes, please.

I see you didn't add a label, but just to be sure: we can't backport this even though the change is tiny.

DigitalBrains1:
> I see you didn't add a label, but just to be sure: we can't backport this even though the change is tiny.

Yes, I am aware!

PR #2899, opened 2025-03-05
CI: Downgrade Cabal for GHC 8.10.7

DigitalBrains1:
Cabal 3.14 silently ignores `haddock-options` in `cabal.project`. The change is deliberate (though not warning about it probably is not). The change is documented in some places:

https://github.com/haskell/cabal/blob/master/release-notes/cabal-install-3.14.0.0.md

https://github.com/haskell/cabal/blob/master/release-notes/WIP-Cabal-3.12.x.0.md 

https://github.com/haskell/cabal/pull/9177

(Note that the deprecation message apparently never made it to the release notes)

We should probably think about how to fix this in the future, but for now rolling back to Cabal 3.12 will allow us to build the documentation as we want to again. The mentioned `haddock-version-cpp` flag in the linked cabal PR 9177 above  to get a CPP define for `__HADDOCK_VERSION__` did not work at all for me.

PR #2201 was supposed to update our Haddock generation to 9.0.2 but accidentally picked 8.10.7 for the uploads to Hackage, which also fixed issue #2200. I'm careful about upgrading Haddock now as I've experienced several times now that there were undocumented changes going from one version to the next that broke some rendering, and carefully rereading our complete Haddock to scan for such breakage is an impossible task. One day we'll have to upgrade, when we drop GHC 8.10 from Clash.

## Still TODO:

### DIFF
--- a/.ci/docker/Dockerfile
+++ b/.ci/docker/Dockerfile
@@ -111,7 +111,7 @@ RUN git clone https://github.com/verilator/verilator verilator \
 
 FROM builder AS build-ghc
 
-ARG ghcup_version="0.1.30.0"
+ARG ghcup_version="0.1.40.0"
 
 # Must be explicitly set
 ARG ghc_version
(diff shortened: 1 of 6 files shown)


martijnbastiaan (approved):
LGTM

PR #2910, opened 2025-03-19
Clarify ShowX examples

DigitalBrains1:
The fact that we used `errorX "undefined"` has the potential of misleading people into thinking `ShowX` will render the message of their `XException` rather than the literal string `undefined`. Adjust examples so this misconception is no longer possible.

## Still TODO:


martijnbastiaan (approved):
I guess it could though?

PR #2911, opened 2025-03-26
Fix DDR primitives for 1.8 (version without API changes)

DigitalBrains1:
This is a partial backport of #2833 but with an addition. PR #2833 prevents people from instantiating the vendor IP cores in a domain where the falling edge is the active edge, but for 1.8 we need a solution that does not change the API. So instead, this PR uses `clashCompileError` to at least prevent people from generating HDL in this defective scenario, to hold the fort until 1.10 comes along to fix this properly.

I ran all the Intel tests locally (with GHC 9.10.1) and had CI run all the DDR tests, including the ones that would only run in nightly and thus wouldn't show up as a CI failure on the PR itself. I also ran all ModelSim DDR tests locally as we do not run these on CI at all. I'm pretty confident this PR is good to go :-D.

Note that there is one somewhat-issue with the Changelog files. If we release another 1.8, we should document the `clashCompileError` case, but if we don't and the next version is 1.10.0, there will never have been a stable release with the `clashCompileError` solution, so it should not appear in the Changelog. What this boils down to, is that `changelog/2024-12-14T13_35_37+01_00_fix_ddr_prims` differs between the `master` branch and the `1.8` branch after this PR is merged. If we release another 1.8, this will show up when we backport the release changes to the `master` branch as part of the release process, as Git will warn that the deleted file has been modified. The correct solution is to accept the deletion of the file despite the difference in contents. If the next version we release is 1.10.0, this situation will never occur and everything will be smooth sailing all the way through.

## Still TODO:

### DIFF
--- a/changelog/2024-12-14T13_35_37+01_00_fix_ddr_prims
+++ b/changelog/2024-12-14T13_35_37+01_00_fix_ddr_prims
@@ -2,8 +2,10 @@ FIXED: `Clash.Explicit.DDR`:
   - `ddrIn`: VHDL: Remove data input from sensitivity list of `ddrIn_neg_latch` register as it is superfluous. This should not affect functionality.
   - `ddrOut`: VHDL: Fix incorrect usage of `Enable` input when the domain is set to asynchronous resets. Deasserting the `Enable` exhibited wrong behavior before this fix.
 FIXED: `Clash.Xilinx.DDR`:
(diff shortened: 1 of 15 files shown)


DigitalBrains1:
It turns out GHC 8.6 needs some parentheses in this expression:
```haskell
type Fast = "Fast" :: Domain
```
GHC 8.8 and later are fine with the expression as-is, butI've added the parentheses to make 8.6 happy.

martijnbastiaan (approved):

Issue #2913, opened 2025-03-30
`(:>)` vs. `Cons`


kleinreact:
I recognized that
```haskell
mymap :: (a -> b) -> Vec n a -> Vec n b
mymap _ Nil           = Nil
mymap f (x `Cons` xr) = f x `Cons` mymap f xr
```
type-checks, but
```haskell
mymap' :: (a -> b) -> Vec n a -> Vec n b
mymap' _ Nil       = Nil
mymap' f (x :> xr) = f x :> mymap' f xr
```
does not (GHC 9.10):
```
<interactive>:4:11: error: [GHC-25897]
    • Couldn't match type ‘n’ with ‘n0 + 1’
      Expected: Vec n a
        Actual: Vec (n0 + 1) a
      ‘n’ is a rigid type variable bound by
        the type signature for:
          mymap' :: forall a b (n :: Nat). (a -> b) -> Vec n a -> Vec n b
        at <interactive>:2:1-40
    • In the pattern: x :> xr
      In an equation for ‘mymap'’:
          mymap' f (x :> xr) = f x :> mymap' f xr
    • Relevant bindings include
        mymap' :: (a -> b) -> Vec n a -> Vec n b
          (bound at <interactive>:3:1)
```
This is kind of annoying, as up to my knowledge most people use `(:>)` and are not much aware about `Cons`. A little research revealed that there already was a discussion about this in the past (#966, #867, #943), but that discussion was within the scope of some old GHC (we don't even support any more). Thus, it might be that these past considerations need some re-evaluation first.

My primary question would be: do we still need both: `Cons` and `(:>)` these days?

---

A possible solution would be to update the `(:>)` synonym to
```haskell
pattern (:>) ::
  forall n a. () =>
  forall m. n ~ m + 1 =>
  a -> Vec m a -> Vec n a
pattern (:>) x xs = Cons x xs
```
which elides the aforementioned type-mismatch (and yes, the explicit quantifier split is necessary here). However, this change also introduces a different quantifier alternation than `Cons`, which might be non-desirable.

Another solution (my favorite) would be to just rename `Cons` to `(:>)` and to drop `Cons`, as I could not find any counterexample, where this solution would cause any trouble, i.e., where using the constructor directly causes any issues.

DigitalBrains1:
Relevant: #2716

martijnbastiaan:
So after reading the history I think there are three open questions:

[...]

PR #2914, opened 2025-03-30
(:>) pattern simplification

kleinreact:
Let's see what it breaks :-).

### DIFF
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -272,10 +272,11 @@ instance NFData a => NFData (Vec n a) where
 -- g :: Num a => Vec ((((n + 1) + 1) + 1) + 1) a -> a
 -- >>> g (1:>2:>3:>4:>5:>Nil)
 -- 12
-pattern (:>) :: a -> Vec n a -> Vec (n + 1) a
-pattern (:>) x xs <- ((\ys -> (head ys,tail ys)) -> (x,xs))
-  where
-    (:>) x xs = Cons x xs
+pattern (:>) ::
+  forall {n} a. () =>
+  forall m. n ~ m + 1 =>
+  a -> Vec m a -> Vec n a
+pattern (:>) x xs = Cons x xs
 
 infixr CONS_PREC :>
 


martijnbastiaan:
One of the reasons we couldn't remove the pattern synonym last time is that lazy pattern matches didn't work on GADTs. Is that still the case?

kleinreact:
Lazy pattern still can cause problems (cf https://github.com/clash-lang/clash-compiler/issues/2913#issuecomment-2772472500), which why I for example also needed to change the type signature in `T1340` [here](https://github.com/clash-lang/clash-compiler/pull/2914/files#diff-19a1ca06a8c9529c1681deb87d9e8d884382c8eafee1d14632284264f3cc6b4d). There's usually a a proper fix for these kind of issues, but it might cause some new type errors for code that previously just compiled fine.

The problem here more are the plugins, which don't work equally well with the constraints being introduced. My hope was that we could improve the plugins in that regard first, such that our users won't observe any different, but that's bit more work, unfortunately.

It's quite hard to estimate how much effect this particular change could have for our users. I'm fine with just trying it out :wink:.

kleinreact:
So just to summarize my current understanding:
* Pattern matching on the currently used constructor `Cons` brings the constructor and the associated `m ~ n + 1` constraint into scope (where `m` is the length of the vector and `n` is the length of the tail).
* Pattern matching via the `(:>)` pattern instead only brings some (new) `m ~ n + 1` constraint into scope that, however, is not associated with the constructor.
* Lazy pattern matching won't bring the constructor into scope, which is why you get some `m ~ n + 1` constraint with `~(x :> xs)` but not with `~(Cons x xs)`.
* The `m ~ n + 1` constraint associated with the constructor and the `m ~ n + 1` constraint introduced by `(:>)` are handled differently. Either by the plugins or even GHC, as it was observed in https://github.com/clash-lang/clash-compiler/issues/2913#issuecomment-2772472500.

This PR associates the constructor `m ~ n + 1` constraint with `(:>)`, which is more desirable IMO. Whether a pattern for creating some new `m ~ n + 1` constraint just for the lazy pattern match case is desirable or not may be a different question. I don't see any need for lazily matching, but that does not have to mean anything.

kleinreact:
> I was under the impression that you wanted to completely remove the pattern synonym, which would make it impossible to use lazy pattern matches. Is that not the case?

Yes, the idea is to swap the names of `(:>)` and `Cons` and to discard `Cons` (the pattern synonym then) eventually. 

Would missing support for lazy pattern matches be a big deal though? You still can just use `head` and `tail` instead, which are also evaluated lazily, while their usage is more explicit than the pattern synonym in terms of not introducing any new constraints.

My problem here is that the pattern does two things: (1) it rewrites the match into a view pattern over the head and tail and (2) it introduces a new constraint. (1) is fine, but (2) can cause trouble. Most use cases won't need a lazy match in the first place, so just matching on the constructor instead should be fine. In some cases you might need a lazy match, i.e., (1), but then you probably don't need (2). Hence, my recommendation is to explicitly use a view pattern with `head` and `tail` in these cases instead, to avoid the confusion caused by (2).

I think that the user needs to explicitly take care of these difference here, and really only use lazy matching, if required. Currently we introduce a potential problem with every use of `:>`, which, however, is not necessary in 99% of the cases.

martijnbastiaan:
I don't see how `head` and `tail` would provide a clean transition, but we could introduce `uncons`. Then we would be able to write a changelog telling users to rewrite:

```haskell
f ~(x :> xs)
```

into

```haskell
f (uncons -> (x, xs))
```

It's still a bit more verbose, but:

1. I don't think it happens that often
2. The patch should always work?
3. We could even provide an upgrade tool through [retrie](https://hackage.haskell.org/package/retrie)

To be clear I just care about a clean, simple upgrade path for our users. There are big code bases out there. Minor changes like these just another thing on the pile of things they need to do in order to bump Clash..

kleinreact:
I'm fine with the introduction of `uncons`, as it makes it explicit that the user want's to use a lazy match, like `~(x :> xs)` is syntactically visible in the code, but does not suffer from the lazy pattern GADT error.

christiaanb:
The fact that we have to tell users to use lazy patterns for certain situations is already annoying enough, adding an extra step for `Vec` seems like going in the wrong direction when it comes to learnability and useability for the majority of our (potential) users.

I.e. my assumption is that more users will have to use a lazy pattern match for `Vec` than there are users who will need to write recursive functions over the structure of `Vec`; most of our users will get by with the existing higher-order functions.

So the status quo, where it is easy to have a lazy match on `:>`, is the global optimum in the design space.

martijnbastiaan:
I can totally see that @christiaanb. While I do think it's a wart, it's not truly a user-facing wart. Most people can go years without realizing `:>` is a pattern synonym..

PR #2915, opened 2025-03-30
Add `Clash.Class.Convert`

martijnbastiaan:
Utilities for safely converting between various Clash number types.

## To discuss / investigate (but maybe not for this PR?)

- [ ] The instances add `1 <= n` constraints to every _result_ where the argument is `Index n`
  - This makes sense from a perspective of: `Index 0` has no inhabitants, so we cannot create it. On the other hand, we can of course just call `errorX "foo"`. I'm not sure whether I would say this violates the "guaranteed to succeed" rule.
- [ ] The instances add `1 <= n` constraints to every _argument_ where the argument is `Index n`
  - This is similar to the first point, but differs in that we shove the responsibility for creating these values to the caller. 
- [ ] `convert` and `maybeConvert` will _always_ return `XException` if given an `XException`.
  - Does this make sense for extending a `BitVector 0`? The HDL will happily do it, why wouldn't Clash simulation?

I'd push for kicking these questions to after this PR, because:

* We can always relax constraints
* We can always relax `XException` behavior (but not introduce it).
* Having a safe way of converting between Clash types is _way_ more important than dropping a minor constraint.
* It's a fairly delicate thing I don't think we have a proper/consistent view on yet across `clash-prelude`.

## Still TODO:

  - [ ] Decide whether to call it `Clash.Class.Convert` or `Clash.Class.Convertible`. The latter has some buy-in (?) from the community, given that `covertible` already exists. On the other hand, this might make users think it _is_ `convertible`, but it isn't.


martijnbastiaan:
I'll amend the point! But if not super controversial I'd like to postpone, yes.

DigitalBrains1:
I'd vote against naming it `Convertible`, since I think GHC will by default not qualify the class name in an error message of the form `No instance for ‘Convertible T’`, which might be very misleading.

kleinreact (approved):
LGTM overall. 

I'm going to accept this, because I cannot see anything really controversial in the sense that it would be hard to fix it later on.

Regarding the `Convert` vs `Convertible` question: I don't like either name, as they both suggest much freedom in terms of which types could be converted with the classes. Your documentation clearly restricts the setting to number types, which is in contrast to `Convertible` for example, which is also open for conversion between custom non-number types. So maybe we should keep the users of that package happy by not producing any name clash at this point.

A name like `changeNumType` would make it more clear to me what the  purpose of this function is. `Convert`, `Convertible`, and `convert` are just super generic names!

I still find it strange to consider `BitVector` a number type. If we are going to have a discussion about this at some point, then I'd suggest to keep it out here for the time being. Adding it later is easy and the user still can define the instance locally, if necessary.

I don't like the fact that we need to have two classes in the end, but I get the point that the need for individual selection of constraints per instance introduces that requirement.

martijnbastiaan:
Thanks for the comments @kleinreact, I'll get to them, though I'm fairly time constrained at the moment :(.

I think your comment about the name `convert` makes a lot of sense. Maybe `convertNum`? You're right that the class really constrains it to just number types. (Maybe _that_ is an issue?)

DigitalBrains1:
Somehow I feel `NumConvert` feels more natural than `ConvertNum` for the class name... But not so for the function name.

Also, we could rewrite the laws such that the class can also be used for non-numbers. I think that would work?

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
```diff
@@ -0,0 +1,164 @@
...
+>>> import Clash.Class.Convert
+-}
+
+{- | Conversions that are, based on their types, guaranteed to succeed.
+
+== __Laws__
+A conversion is safe and total if a round trip conversion is guaranteed to be
+lossless. I.e.,
+
+> Just x == maybeConvert (convert @a @b x)
```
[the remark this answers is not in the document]
martijnbastiaan: Hmm, well it isn't imposed by anything, but it seems natural to me that everything that defines a `Convert` also defines a `MaybeConvert`. I get that that makes laws like these a bit iffy -- I'm happy to add a superclass constraint if that makes more sense to you. I'm not sure how to rephrase this law properly to _just_ talk about `convert`.

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
```diff
@@ -0,0 +1,164 @@
...
+> toInteger x == toInteger (convert @a @b x)
+
+Instances should make sure their constraints are as \"tight\" as possible. I.e.,
+if an instance exist, but the constraints cannot be satisfied, then
+'Clash.Class.Convert.convertMaybe' should return 'Nothing' for one or more values in
+the domain of the source type @a@:
+
+> L.any isNothing (L.map (maybeConvert @a @b) [minBound ..])
+
+Additionally, any implementation should be translatable to synthesizable RTL.
```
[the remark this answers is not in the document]
martijnbastiaan: Agreed

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
```diff
@@ -0,0 +1,164 @@
...
+instance (Convert (Unsigned 64) a) => Convert Word64 a where
+  convert = convert . bitCoerce @_ @(Unsigned 64)
+instance (Convert (Unsigned 32) a) => Convert Word32 a where
+  convert = convert . bitCoerce @_ @(Unsigned 32)
+instance (Convert (Unsigned 16) a) => Convert Word16 a where
+  convert = convert . bitCoerce @_ @(Unsigned 16)
+instance (Convert (Unsigned 8) a) => Convert Word8 a where
+  convert = convert . bitCoerce @_ @(Unsigned 8)
+
+instance (Convert (Signed 64) a) => Convert Int a where
```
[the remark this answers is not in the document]
martijnbastiaan: Oh woops. I think this just fails to compile on any platform that is not 32-bit, as the `bitCoerce` would type error. This did cross my mind, but I never followed up on it -- I'll add some CPP foo.

clash-prelude/src/Clash/Class/Convert/Internal/MaybeConvert.hs
```diff
@@ -0,0 +1,160 @@
...
+
+  >>> maybeConvert (1 :: Index 8) :: Maybe (Unsigned 2)
+  Just 1
+  >>> maybeConvert (7 :: Index 8) :: Maybe (Unsigned 2)
+  Nothing
+
+  For the time being, if the input is an @XException@, then the output is too.
+  This property might be relaxed in the future.
+  -}
+  maybeConvert :: a -> Maybe b
```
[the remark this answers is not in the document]
martijnbastiaan: I'm fine with either as far as naming schemes go. I do think we've already set a precedent in `clash-prelude`, i.e.:

```
martijn@volthe:~/code/clash-compiler$ egrep -r '^maybe[A-Z].*::'  clash-prelude/src | wc -l
7
martijn@volthe:~/code/clash-compiler$ egrep -r '^.*Maybe ::'  clash-prelude/src | wc -l
1
```

so I think we should continue that. (Or change that, but that's for another PR.)

clash-prelude/tests/Clash/Tests/Convert.hs
```diff
@@ -0,0 +1,307 @@
...
+{- | Tightness law: this law is tested for if there is _no_ instance of
+'Convert'. If this is the case, 'MaybeConvert' should at least return a 'Nothing'
+once when converting the domain of @a@ to @b@. If all conversions are possible,
+the constraints of the instances should be relaxed. If the domain of @a@ is
+empty, this law is considered satisfied too.
+-}
+convertLaw3 :: forall a b. (MaybeConvert a b, Bounded a, Enum a) => Proxy a -> Proxy b -> Bool
+convertLaw3 _ _ = L.any isNothing results || not (L.null results)
+ where
+  results = L.map (maybeConvert @a @b) [minBound ..]
```
DigitalBrains1, line 92: So this test completes in two cases:
1. There's a `Nothing` in the `results`
2. `results` is not empty

I feel this would capture the stated behaviour:
```haskell
convertLaw3 _ _ = L.any isNothing results || L.null results
```

but that `L.null results` never actually has an effect, and it still makes the tests fail for:
```haskell
convert @(Index 0) @(Signed 0)
convert @(Unsigned 0) @(Signed 0)
convert @(Signed 0) @(Signed 0)
```
but for two distinct reasons.

For `Index 0`, the problem is:
```haskell
>>> convertLaw3 (Proxy @(Index 0)) (Proxy @(Signed 0))
<interactive>:6:1: error: [GHC-64725]
    • Cannot satisfy: 1 <= 0
    • In the expression:
        convertLaw3 (Proxy @(Index 0)) (Proxy @(Signed 0))
      In an equation for ‘it’:
          it = convertLaw3 (Proxy @(Index 0)) (Proxy @(Signed 0))
```
(because `instance (KnownNat n, KnownNat m, 1 <= n) => MaybeConvert (Index n) (Signed m)`)

For `Unsigned 0` and `Signed 0`, the problem is that 
```haskell
>>> L.map (maybeConvert @(Unsigned 0) @(Signed 0)) [minBound ..]
[Just 0]
>>> L.map (maybeConvert @(Signed 0) @(Signed 0)) [minBound ..]
[Just 0]
````

File: clash-prelude/tests/Clash/Tests/MaybeConvert.hs [hunk 6]
```diff
@@ -0,0 +1,323 @@
...
+  Bool
+convertMaybeLaw4 Proxy x =
+  isJust (maybeConvert @_ @b x) == (i x >= i (minBound @b) && i x <= i (maxBound @b))
+ where
+  i :: (Integral c) => c -> Integer
+  i = toInteger
+
+-- | Checks whether an @XException@ in, means an @XException@ out
+convertXException :: forall a b. (MaybeConvert a b) => Proxy a -> Proxy b -> Bool
+convertXException _ _ = isLeft $ isX $ maybeConvert @a @b (errorX "" :: a)
```
DigitalBrains1, line 109: If you make it so that the original `XException` is always preserved, you could make this test stricter by testing that the message of the original `XException` is also the message of the ouput `XException` (same for the `Convert` tests).

DigitalBrains1 (changes requested):
I also feel `Convert` is too generic a name and squats too much, except if it is only to be used qualified. I'm personally leaning towards renaming it `NumConvert`, `numConvert`, `MaybeNumConvert` and `maybeNumConvert` as the names used, and exporting them from `Prelude`. Note that you currently to not export them from `Prelude`, I'd like that to be added.

Something I noticed is that for `Convert`, you seem to put a bang pattern only on the instances that need one to preserve the `XException`, yet on `MaybeConvert` you just bang almost everything. Was this intended? I'm not against it, I'm just wondering.

clash-prelude/tests/Clash/Tests/Convert.hs
DigitalBrains1: Note though that for `Index 0` we'd get:

```haskell
ghci> [minBound @(Index 0) ..]
*** Exception: X: maxBound of 'Index 0' is undefined
  errorX, called at src/Clash/Sized/Internal/Index.hs:300:10 in clash-prelude-1.9.0-inplace:Clash.Sized.Internal.Index
```
so this is also not `L.null`, it bottoms.

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
```diff
@@ -0,0 +1,164 @@
...
+instance (KnownNat n, KnownNat m, 1 <= n, n <= 2 ^ m) => Convert (Index n) (BitVector m) where
+  convert !a = resize $ pack a
+
+instance (KnownNat n, KnownNat m, 1 <= m, 2 ^ n <= m) => Convert (Unsigned n) (Index m) where
+  convert !a = bitCoerce $ resize a
+
+instance (KnownNat n, KnownNat m, n <= m) => Convert (Unsigned n) (Unsigned m) where
+  convert = resize
+
+{- | Note: Conversion from @Unsigned 0@ to @Signed 0@ is total, but not within the
```
DigitalBrains1, line 95: Ah skipped one. As per earlier comment: let's get rid of the word _total_ here.

clash-prelude/tests/Clash/Tests/Convert.hs
DigitalBrains1: Let me be a bit more explicit:

The `L.null results` branch of the test never helps you. I think this case simply never occurs, but it is clear you can remove it and not affect the outcome: `convertLaw3 _ _ = L.any isNothing results` works just as well.

The three problematic tests are precisely those of which you already note:
> Note: Conversion from `Index 1` to `Signed 0` is total, but not within the constraints of the instance.

So your test correctly tests the law, it just needs some exceptions for the cases where you document a deviation from the laws. I think in your Python script you actually already flag these cases with `constraints[(a, b)][0]`, but you never use this boolean, you only use the second element of the tuple. I suspect that you should use that flag to generate
```haskell
assertBool (show (n, m))
  (convertLaw3 (Proxy @(Index (n + 1))) (Proxy @(Signed m)) || (n == 0 && m == 0))
```
instead of the current
```haskell
assertBool (show (n, m)) (convertLaw3 (Proxy @(Index (n + 1))) (Proxy @(Signed m)))
```

martijnbastiaan:
@DigitalBrains1 For some reason GitHub doesn't let me reply to your comment on the word total, so I'll reply here: totally fair criticism! I think _lossless_ is the word I want to use. (After rephrasing the sentence a bit.)

DigitalBrains1:
> For some reason GitHub doesn't let me reply to your comment on the word total

Yeah, that's a confusing formatting issue. As part of a review, you can reply to an earlier review comment. It then shows it [in the original place](https://github.com/clash-lang/clash-compiler/pull/2915#discussion_r2058705668) where you _can_ reply, and also as part of the new review. If you click on the date in the header of the latter (_3 weeks ago_ it currently says), you can go to the former and comment... why they did this, I don't know.

I think _lossless_ is fair, although you could argue that `bitCoerce @(Unsigned 8) @(Signed 8) 200` is in fact lossless even though the result is `-56`. But when put into context, _lossless_ is in fact a proper description that people should be able to understand.

martijnbastiaan:
Me? Puns? Never.

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
```diff
@@ -0,0 +1,164 @@
...
+
+for all values @x@ of type @a@. It should also preserve the numerical value
+interpretation of the bits. For types that have an "Integral" instance, this
+intuition is captured by:
+
+> toInteger x == toInteger (convert @a @b x)
+
+Instances should make sure their constraints are as \"tight\" as possible. I.e.,
+if an instance exist, but the constraints cannot be satisfied, then
+'Clash.Class.Convert.convertMaybe' should return 'Nothing' for one or more values in
```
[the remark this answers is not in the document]
martijnbastiaan: What do you think of:

```
Instances should make sure their constraints are as \"tight\" as possible. I.e.,
if an instance's constraints cannot be satisfied, then for the same types
'Clash.Class.NumConvert.maybeConvert' should return 'Nothing' for one or more
values in the domain of the source type @a@:
```

I find it quite hard to move away from my mental model of "instance resolver has found a match == instance exists", so I'm having trouble really rewriting this.

clash-prelude/src/Clash/Class/Convert/Internal/MaybeConvert.hs
```diff
@@ -0,0 +1,160 @@
...
+instance (KnownNat n, KnownNat m) => MaybeConvert (BitVector n) (Signed m) where
+  maybeConvert !a = maybeResize $ unpack @(Signed (n + 1)) $ extend a
+
+instance (KnownNat n, KnownNat m) => MaybeConvert (BitVector n) (BitVector m) where
+  maybeConvert !a = maybeResize a
+
+instance (MaybeConvert (Unsigned 64) a) => MaybeConvert Word a where
+  maybeConvert = maybeConvert . bitCoerce @_ @(Unsigned 64)
+instance (MaybeConvert (Unsigned 64) a) => MaybeConvert Word64 a where
+  maybeConvert = maybeConvert . bitCoerce @_ @(Unsigned 64)
```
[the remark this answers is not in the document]
martijnbastiaan: I've added a test

clash-prelude/tests/Clash/Tests/MaybeConvert.hs
martijnbastiaan: Test amended!

clash-prelude/src/Clash/Class/Convert/Internal/MaybeConvert.hs
```diff
@@ -0,0 +1,160 @@
...
+instance (MaybeConvert (Signed 32) a) => MaybeConvert Int32 a where
+  maybeConvert = maybeConvert . bitCoerce @_ @(Signed 32)
+instance (MaybeConvert (Signed 16) a) => MaybeConvert Int16 a where
+  maybeConvert = maybeConvert . bitCoerce @_ @(Signed 16)
+instance (MaybeConvert (Signed 8) a) => MaybeConvert Int8 a where
+  maybeConvert = maybeConvert . bitCoerce @_ @(Signed 8)
+
+instance (MaybeConvert a (Unsigned 64)) => MaybeConvert a Word where
+  maybeConvert = fmap (bitCoerce @(Unsigned 64)) . maybeConvert
+instance (MaybeConvert a (Unsigned 64)) => MaybeConvert a Word64 where
```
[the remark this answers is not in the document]
martijnbastiaan: yes

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
```diff
@@ -0,0 +1,164 @@
...
+the domain of the source type @a@:
+
+> L.any isNothing (L.map (maybeConvert @a @b) [minBound ..])
+
+Additionally, any implementation should be translatable to synthesizable RTL.
+-}
+class Convert a b where
+  {- | Convert a supplied value of type @a@ to a value of type @b@. The conversion
+  is guaranteed to succeed.
```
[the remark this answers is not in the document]
martijnbastiaan: That would make sense to me if this was documentation purely for the implementer of the class, but 9/10 times it is for users looking to use it.

martijnbastiaan:
I've done a couple of things (now hidden in resolved conversations):

* I've renamed the class
* I've touched up the documentation pointed out by the reviews
* I've fixed a few issues with the test bench
* I've added instances for `Bit`

I _think_ I got to all the comments.

martijnbastiaan:
> Something I noticed is that for Convert, you seem to put a bang pattern only on the instances that need one to preserve the XException, yet on MaybeConvert you just bang almost everything. Was this intended? I'm not against it, I'm just wondering.

I mostly followed the error messages brought up by the tests. TBF, it is a bit too much for me to do precisely so I might have been over eager..

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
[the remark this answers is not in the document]
kleinreact: Good point! I still consider _"is guaranteed to succeed"_ to be too strong though, as it sounds like even the implementer cannot do wrong. Maybe something like _"is intended to always succeed"_ or _"is required to always succeed"_ to make it clear that this is a design constraint and not a design feature.

Consider that users can also be implementers of the class. Hence, I don't see why there should be no documentation for the implementer's use case as well.

It is useful to take into account that wording in that regard is standardized, e.g., in https://www.rfc-editor.org/rfc/rfc2119

clash-prelude/src/Clash/Class/Convert/Internal/Convert.hs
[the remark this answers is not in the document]
DigitalBrains1: > What do you think of:

Yes, excellent work.

> The qualification is not necessary; the function is documented in this module.

(my own comment) Ah, I made a mistake. I was probably looking at the Haddock for `Clash.Class.NumConvert` when I concluded `maybeConvert` was defined locally, because it's clearly not the case.

clash-prelude/src/Clash/Class/NumConvert.hs
```diff
@@ -0,0 +1,33 @@
+{- |
+Copyright  :  (C) 2025     , Martijn Bastiaan
+License    :  BSD2 (see the file LICENSE)
+Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
+
+Utilities for converting between Clash number types in a non-erroring way. Its
+existence is motivated by the observation that Clash users often need to convert
+between different number types (e.g., 'Clash.Sized.Unsigned.Unsigned' to
+'Clash.Sized.Signed.Signed') and that it is not always clear how to do so
+properly. Two classes are exported:
```
DigitalBrains1:
```suggestion
between different number types (e.g., @Unsigned@ to @Signed@) and that it is not
always clear how to do so properly. Two classes are exported:
```

With GHC 8.10, you cannot link to qualified type names :-(. Only terms. If you want to link to a type, you need to import it and use it unqualified in the Haddock.

clash-prelude/src/Clash/Class/NumConvert.hs
```diff
@@ -0,0 +1,33 @@
...
+synthesizable HDL.
+
+== __Relation to @convertible@__
+Type classes exported here are similar to the @convertible@ package in that it
+aims to facilitate conversions between different types. It is different in three
+ways:
+
+ 1. It offers no partial functions.
+ 2. All its conversions are translatable to synthesizable HDL.
+ 3. It is focused on (Clash's) number types
```
DigitalBrains1: Stupid little stylistic detail: either all these sentence fragments need to omit the period, or they all should have a period. I do believe I omit the period when I add documentation.

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
```diff
@@ -0,0 +1,181 @@
...
+failure is expressed by returning 'Nothing', never by an 'Clash.XException.XException'.
+
+== __Laws__
+A conversion is either successful or it fails gracefully. I.e., it does not
+produces produce errors (also see "Clash.XException"). I.e.,
+
+> x == fromMaybe x (maybeNumConvert @a @b x >>= maybeNumConvert @b @a)
+
+for all values @x@ of type @a@. It should also preserve the numerical value
+interpretation of the bits. For types that have an "Integral" instance, this
```
DigitalBrains1:
```suggestion
interpretation of the bits. For types that have an @Integral@ instance, this
```

Double quotes link to modules, not declarations. I don't think the type is in scope, so let's just use typewriter font.

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
```diff
@@ -0,0 +1,181 @@
...
+intuition is captured by:
+
+> toInteger x == fromMaybe (toInteger x) (toInteger (numConvert @a @b x))
+
+If a conversion succeeds one way, it should also succeed the other way. I.e.,
+
+> isJust (maybeNumConvert @a @b x) `implies` isJust (maybeNumConvert @a @b x >>= maybeNumConvert @b @a)
+
+A conversion should succeed if and only if the value is representable in the
+target type. For types that have a "Bounded" and "Integral" instance, this
```
DigitalBrains1:
```suggestion
target type. For types that have a @Bounded@ and @Integral@ instance, this
```

clash-prelude/src/Clash/Class/NumConvert/Internal/NumConvert.hs
```diff
@@ -0,0 +1,178 @@
...
+successful conversion retains the numerical value interpretation of the source
+type in the target type and does not produce errors.
+
+== __Laws__
+A conversion is successful if a round trip conversion is lossless. I.e.,
+
+> Just x == maybeNumConvert (numConvert @a @b x)
+
+for all values @x@ of type @a@. It should also preserve the numerical value
+interpretation of the bits. For types that have an "Integral" instance, this
```
DigitalBrains1:
```suggestion
interpretation of the bits. For types that have an @Integral@ instance, this
```

clash-prelude/src/Clash/Class/NumConvert/Internal/NumConvert.hs
```diff
@@ -0,0 +1,178 @@
...
+intuition is captured by:
+
+> toInteger x == toInteger (numConvert @a @b x)
+
+Instances should make sure their constraints are as \"tight\" as possible. I.e.,
+if an instance's constraints cannot be satisfied, then for the same types
+'Clash.Class.NumConvert.maybeConvert' should return 'Nothing' for one or more
```
DigitalBrains1:
```suggestion
'Clash.Class.NumConvert.maybeNumConvert' should return 'Nothing' for one or more
```

clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
```diff
@@ -0,0 +1,181 @@
...
+> isJust (maybeNumConvert @a @b x) == (i x >= i (minBound @b) && i x <= i (maxBound @b))
+
+where @i = toInteger@.
+
+All implementations should be total, i.e., they should not produce \"bottoms\".
+
+Additionally, any implementation should be translatable to synthesizable HDL.
+-}
+class MaybeNumConvert a b where
+  {- | NumConvert a supplied value of type @a@ to a value of type @b@. If the value
```
DigitalBrains1:
```suggestion
  {- | Convert a supplied value of type @a@ to a value of type @b@. If the value
```

As discussed

clash-prelude/src/Clash/Class/NumConvert/Internal/NumConvert.hs
```diff
@@ -0,0 +1,178 @@
...
+values in the domain of the source type @a@:
+
+> L.any isNothing (L.map (maybeNumConvert @a @b) [minBound ..])
+
+All implementations should be total, i.e., they should not produce \"bottoms\".
+
+Additionally, any implementation should be translatable to synthesizable HDL.
+-}
+class NumConvert a b where
+  {- | NumConvert a supplied value of type @a@ to a value of type @b@. The conversion
```
DigitalBrains1:
```suggestion
  {- | Convert a supplied value of type @a@ to a value of type @b@. The conversion
```

As discussed

clash-prelude/tests/Clash/Tests/MaybeNumConvert.hs
```diff
@@ -0,0 +1,326 @@
...
+  Bool
+maybeConvertLaw4 Proxy x =
+  isJust (maybeNumConvert @_ @b x) == (i x >= i (minBound @b) && i x <= i (maxBound @b))
+ where
+  i :: (Integral c) => c -> Integer
+  i = toInteger
+
+-- | Checks whether an 'XException' in, means an 'XException' out
+convertXException :: forall a b. (MaybeNumConvert a b) => Proxy a -> Proxy b -> Bool
+convertXException _ _ = isLeft $ isX $ maybeNumConvert @a @b (errorX "" :: a)
```
DigitalBrains1:
```suggestion
-- | Checks whether an 'XException' in, means an 'XException' out
convertXException :: forall a b. (MaybeNumConvert a b) => Proxy a -> Proxy b -> Bool
convertXException _ _ = case isX $ maybeNumConvert @a @b (errorX "BOO" :: a) of
  Left s -> "BOO" `L.isInfixOf` s
  Right _ -> False
```

Needs some adjustments to imports too. But the test passes!

DigitalBrains1 (changes requested):
Nice work on the improvements! It did introduce a few more issues, but I think after that, that should be it. As discussed, I will add the improvements as a commit to save you the trouble. But the commit follows tomorrow, I'm already past out of time right now.

clash-prelude/tests/Clash/Tests/MaybeNumConvert.hs
DigitalBrains1: > `errorX "BOO"`

👻

(since Haskell supports Unicode, you could use that as the error message)

.
.
.

(please don't use `errorX "👻"`)

DigitalBrains1 (approved):
As discussed, I've added a commit implementing my change requests. When you squash the merge, please remove the _Co-authored by Peter_, a bit of review is not co-authorship.

PR #2921, opened 2025-04-02
Inline global binders in `termToData`

martijnbastiaan:
**I've marked it as draft to prevent merges (see TODOs), but feel free to review.**

GHC sometimes decides to pull out data structures into their own global variables. For example:

    Clash.Cores.Xilinx.Xpm.Cdc.Internal.InstConfig[8214565720323807388]
      Clash.Cores.Xilinx.Unisim.DnaPortE2.Internal.dnaPortE32[8214565720323807390][GlobalId]
      Clash.Cores.Xilinx.Unisim.DnaPortE2.Internal.dnaPortE29[8214565720323807391][GlobalId]
      Clash.Cores.Xilinx.Unisim.DnaPortE2.Internal.dnaPortE26[8214565720323807392][GlobalId]
      (GHC.Maybe.Nothing[3891110078048108565]
        @Clash.Cores.Xilinx.Xpm.Cdc.Internal.XilinxWizard[8214565720323807393])

This makes it impossible for `termToData` to properly resolve it. We now pass a `BindingMap` to it, so it can inline the definition and continue working.

## Related
https://github.com/clash-lang/clash-cores/pull/28

## Still TODO:

  - [ ] Wait for https://github.com/clash-lang/clash-cores/pull/28 to be based on this and check that _it_ passes CI

### DIFF
--- a/clash-lib/src/Clash/Core/TermLiteral.hs
+++ b/clash-lib/src/Clash/Core/TermLiteral.hs
@@ -1,7 +1,7 @@
 {-|
-Copyright   :  (C) 2019, Myrtle Software Ltd,
-                   2021, QBayLogic B.V.
-                   2022, Google Inc.
+Copyright   :  (C) 2019,      Myrtle Software Ltd,
+                   2021,      QBayLogic B.V.
+                   2022-2025, Google Inc.
 License     :  BSD2 (see the file LICENSE)
 Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Core/TermLiteral.hs
```diff
@@ -62,11 +74,51 @@ import           Clash.Core.TermLiteral.TH
...
+  pure (termToData (netlistState ^. bindings) term)
+
+-- | Replace any global binders with their corresponding term
+inlineGlobalBinders :: BindingMap -> Term -> Term
+inlineGlobalBinders binders = go
+ where
+  go = \case
+    Var v
+      | isGlobalId v
+      , Just (bindingTerm -> t) <- lookupVarEnv v binders -> go t
```
martijnbastiaan, line 103: Perhaps `go` should take an integer to stop infinite loops here. (Though TBF `TermLiteral` isn't supposed to "parse" infinitely self recursive data types.)

martijnbastiaan:
Unfortunately, the binders I'm after aren't even in the binder map. This would mean even more changes the Clash, which I don't think is a good idea -- I've changed the PR at `clash-cores` to just not trigger this behavior anymore.

PR #2925, opened 2025-04-05
Strip ticks for literal parsing in `TermLiteral`

martijnbastiaan:
Note that the other instances are fine, because they use `collectArgs`.

## Still TODO:

### DIFF
--- a/changelog/2025-04-05T17_48_30+02_00_ignore_ticks
+++ b/changelog/2025-04-05T17_48_30+02_00_ignore_ticks
@@ -1,0 +1,1 @@
+INTERNAL FIX: Ignore `Tick`s in `TermLiteral Integer`, `TermLiteral Char`, `TermLiteral Natural`, and `TermLiteral (SNat n)`
--- a/clash-lib/src/Clash/Core/TermLiteral.hs
+++ b/clash-lib/src/Clash/Core/TermLiteral.hs
@@ -48,7 +48,7 @@ import           Clash.Core.DataCon              (DataCon(..))
 import           Clash.Core.Literal
 import           Clash.Core.Name                 (Name(..))
 import           Clash.Core.Pretty               (showPpr)
-import           Clash.Core.Term                 (Term(Literal, Data), collectArgs)
+import           Clash.Core.Term                 (Term(Literal, Data, Tick), collectArgs)
 import           Clash.Promoted.Nat
 import           Clash.Promoted.Nat.Unsafe
 import           Clash.Sized.Index               (Index)
@@ -127,6 +127,7 @@ instance TermLiteral Word where
   termToData t = Left t
 
 instance TermLiteral Integer where
+  termToData (Tick _ e) = termToData e
   termToData (Literal (IntegerLiteral n)) = Right n
   termToData (collectArgs -> (_, [Left (Literal (IntegerLiteral n))])) = Right n
   termToData t = Left t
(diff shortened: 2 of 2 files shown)


DigitalBrains1:
> Note that the other instances are fine, because they use `collectArgs`.

Why was the `collectArgs` use in `Integer`, `Char` and `Natural` not sufficient? And why couldn't `SNat` be solved the same as the ones that use `collectArgs`?

martijnbastiaan:
Because it can either be directly a literal (potentially wrapped in `Tick`), or an application to some primitive builder (`fromIntegral# (Literal ...)`. The latter will be parsed using `collectArgs`, the former has a direct pattern match.

TBF, I'm somewhat suspicious of this pattern. I'm not sure why there are multiple ways of representing literals in our `Term`..

martijnbastiaan:
> And why couldn't SNat be solved the same as the ones that use collectArgs?

I think you're asking why there isn't a pattern match with `collectArgs` for `SNat`? Probably because we haven't encountered it in the wild. (Yes, it's ad-hoc like that 😅.)

DigitalBrains1:
Right, so the direct pattern match is a more-primitive case than the `collectArgs` view pattern match, if I understand correctly. In that case I'm wondering about the difference here:

```haskell
instance TermLiteral Integer where
  termToData (Tick _ e) = termToData e
  termToData (Literal (IntegerLiteral n)) = Right n    -- <- "more-primitive" case
  termToData (collectArgs -> (_, [Left (Literal (IntegerLiteral n))])) = Right n    -- <- "less-primitive" case
  termToData t = Left t

instance TermLiteral Char where
  termToData (Tick _ e) = termToData e
  termToData (collectArgs -> (_, [Left (Literal (CharLiteral c))])) = Right c
  termToData t = Left t
```
All the ones that now have an explicit match on `Tick` also have a direct match on `FooLiteral` (for some `Foo`) except `Char`, that one breaks that pattern. It feels like if `collectArgs` does not handle the occurrence of this "more-primitive case", and `Char` needed the `Tick` here, then it should also have a pattern match for the "more-primitive case", something like
```haskell
instance TermLiteral Char where
  termToData (Tick _ e) = termToData e
  termToData (Literal (CharLiteral c)) = Right c
  termToData (collectArgs -> (_, [Left (Literal (CharLiteral c))])) = Right c
  termToData t = Left t
```

DigitalBrains1:
In fact, I think for a fact that the `Char` one doesn't make sense, as it is already handled by `collectArgs`:

```haskell
-- | Split a (Type)Application in the applied term and it arguments
collectArgs :: Term -> (Term, [Either Term Type])
collectArgs = go []
  where
    go args (App e1 e2) = go (Left e2:args) e1
    go args (TyApp e t) = go (Right t:args) e
    go args (Tick _ e)  = go args e    -- <- HERE
    go args e           = (e, args)
```

martijnbastiaan:
You're right the `Char` one is overzealous. Though..... `collectArgs` doesn't seem to look through ticks in arguments, so perhaps the correct way of doing it is the following (don't mind pseudo code, I'm on my phone):

```haskell
go = \case
  Tick e -> go e
  Primitive e -> e
  (collectArgs -> (_, e)) -> go e
  _ -> .......
```

martijnbastiaan:
I think this is still an improvement over the status-quo, even if it isn't as thorough as https://github.com/clash-lang/clash-compiler/issues/2926. Do we want to merge @DigitalBrains1?

DigitalBrains1:
Yeah, I'm fine with that if you drop the change to `Char` as it is, in your words, overzealous. In my words, meaningless and redundant 😂.

Issue #2932, opened 2025-04-11
Problematic SystemVerilog as top entity is inlined in test bench


DigitalBrains1:
While adding SystemVerilog to the Vivado tests, one of the tests in `clash-testsuite`, `shouldwork.Vector.DFold`, errored out ([source](https://github.com/clash-lang/clash-compiler/blob/f166a204da010e267b8653680c193209c5c1fec6/tests/shouldwork/Vector/DFold.hs)). On GHC 9.0 and 9.2, it causes an error while on everything else (8.10 and 9.4 through 9.10) it succeeds. Note that Questa Intel Starter FPGA Edition 2023.3 is actually fine with the SystemVerilog in question.

The error message is:
```
ERROR: [VRFC 10-805] illegal context for assignment pattern [/tmp/clash-test_DFold-5364f0c46dd5157e/vivado-testBench/DFold.testBench/testBench.sv:36]
```

Looking at the produced HDL reveals that on GHC 9.0, the test bench seems to inline the top entity despite a `NOINLINE` annotation. With 9.4 and up, we use `OPAQUE`, suggesting the bug is fixed by that. Why 8.10 does not inline the top entity I don't know.

The error message refers to this expression:
```systemverilog
    if (('{0: 64'sd7
   ,1: 64'sd8
   ,2: 64'sd9
   ,3: 64'sd0
   ,4: 64'sd1
   ,5: 64'sd2
   ,6: 64'sd3
   ,7: 64'sd4
   ,8: 64'sd5
   ,9: 64'sd6}) !== (DFold_testBench_types::array_of_10_signed_64_from_lv(c$vec[(64'sd0)]))) begin
```

This expression is supposed to be an `assert` on the output of the top entity, but is in fact the constant value of `topEntity` inlined in the assertion. I think the bigger issue here is: _why_ do we generate HDL that trips up Vivado? The fact that the top entity is inlined despite a `NOINLINE` might be something we'll just have to accept.

This same expression looks a lot better under GHC 9.10. `i0` is the output of the top entity:

```systemverilog
    if ((i0) !== (DFold_testBench_types::array_of_10_signed_64_from_lv(c$vec[(64'sd0)]))) begin
```
This is the full SystemVerilog for the test bench under GHC 9.0:

 <details><summary><code>testBench.sv</code> (GHC 9.0)</summary>

```systemverilog
/* AUTOMATICALLY GENERATED SYSTEMVERILOG-2005 SOURCE CODE.
[...]

PR #2933, opened 2025-04-11
Add Vivado SystemVerilog to CI

DigitalBrains1:
Something prevented us from adding this before, but I don't fully recall.

It works fine now, so let's add it.

One test is affected by #2932. The same test has always had Verilator excluded, but it turns out Verilator is also affected by #2932, which is SystemVerilog-specific. Verilator runs fine on the Verilog, so that test is added.

Note that ModelSim is not affected by #2932 and runs succesfully.

## Still TODO:

### DIFF
--- a/.ci/gitlab/test.yml
+++ b/.ci/gitlab/test.yml
@@ -204,6 +204,15 @@ suite:vivado:verilog:
     - local
     - vivado-2022.1-standard
 
+suite:vivado:systemverilog:
+  extends: .test-cache-local-nightly
+  script:
+    - source /opt/tools/Xilinx/Vivado/2022.1/settings64.sh
+    - bin/clash-testsuite:clash-testsuite -j$THREADS -p .SystemVerilog --hide-successes --no-modelsim --no-ghdl --no-iverilog --no-verilator --no-symbiyosys
+  tags:
+    - local
+    - vivado-2022.1-standard
+
 # XXX: Vivado randomly fails with the following error message:
 #
 #          FATAL_ERROR: Vivado Simulator kernel has discovered an exceptional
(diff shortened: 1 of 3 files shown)


DigitalBrains1:
All the GitHub CI failures are an issue with Stackage. I'll wait for a bit to give them time to properly update their repo and then restart them.

.ci/gitlab/test.yml
```diff
@@ -204,6 +204,15 @@ suite:vivado:verilog:
     - local
     - vivado-2022.1-standard
 
+suite:vivado:systemverilog:
```
martijnbastiaan, line 207: Should `suite:systemverilog` be renamed to `suite:verilator:systemverilog`?

martijnbastiaan: Could you add a comment mentioning that we run Vivado/SV as a separate job due to performance (I'm assuming that's why)?

martijnbastiaan (approved):

.ci/gitlab/test.yml
DigitalBrains1: We already have that:
https://github.com/clash-lang/clash-compiler/blob/f166a204da010e267b8653680c193209c5c1fec6/.ci/gitlab/test.yml#L188-L189
Is this insufficient?

DigitalBrains1: > Should suite:systemverilog be renamed to suite:verilator:systemverilog?

I think it's more of a coincidence that we only run Verilator on SystemVerilog; it's not a design decision, it's simply the case because we never made the effort to install Questa on our runners.

Conversely, running Vivado in a separate job is truly a design decision.

As far as I'm concerned, some day soon we install Questa Intel FPGA Starter Edition on all our runners and run that as part of `suite:systemverilog`. I think it's fast enough to run both Verilator and Questa as part of every CI run. It's just Vivado that's such a 🐢.

martijnbastiaan: Ah sorry, nothing to be done here then.

Issue #2808, opened 2024-09-08
Spine-lazy vector functions


gergoerdi:
Functions that construct vectors with a known length should be able to produce the spine fully lazily. For example, I don't think there's a good reason why `map (const 0) (undefined :: Vec 5 ())`  should bottom out.

So I propose rewriting the vector functions a la the following examples:

```
map :: forall n a b. (KnownNat n) => (a -> b) -> Vec n a -> Vec n b
map f = go (toUNat SNat)
  where
    go :: forall n. UNat n -> Vec n a -> Vec n b
    go UZero     ~Nil       = Nil
    go (USucc n) ~(x :> xs) = f x :> go n xs

zipWith :: forall n a b c. (KnownNat n) => (a -> b -> c) -> Vec n a -> Vec n b -> Vec n c
zipWith f = go (toUNat SNat)
  where
    go :: forall n. UNat n -> Vec n a -> Vec n b -> Vec n c
    go UZero     ~Nil       ~Nil       = Nil
    go (USucc n) ~(x :> xs) ~(y :> ys) = f x y :> go n xs ys
```

(I'm including `zipWith` here to demonstrate how nicely unbiased it becomes with this technique).

gergoerdi:
As @kleinreact can attest to, I have real-life code where, due to `zipWith` being unnecessarily lazy in its first argument's spine, `f <$> x <*> y` loops, whereas `pure f <*> x <*> y` works, which can be hella confusing unless one is aware of `zipWith`'s details.

gergoerdi:
Related: If `KnownNat n`, then `Vec n` can be made a [distributive functor](https://hackage.haskell.org/package/distributive-0.6.2.1/docs/Data-Distributive.html), which is basically the generalization of the observation that we can go from `Maybe (Vec n a)` to `Vec n (Maybe a)` in the same build-the-spine-from-the-type way.

gergoerdi:
Hmm, I wanted to start working on this but very quickly realized that `map` has no `KnownNat n` constraint, and neither does the `Functor (Vec n)` instance, so this would actually be quite a big & nasty API change. *Sigh*.

DigitalBrains1:
[...]

PR #2936, opened 2025-04-15
Rewrite `(+>>)` to no longer require `KnownNat`

kleinreact:
Drops the `KnownNat` constraint from `(+>>)`, as it is not needed.

## Still TODO:

    ~_(I don't see why this update should require a changelog entry, as dropping constraints is backward compatible by default.)_~
    (edited by @DigitalBrains1)

### DIFF
--- a/changelog/2025-04-18T15_27_46+02_00_less_knownnat
+++ b/changelog/2025-04-18T15_27_46+02_00_less_knownnat
@@ -1,0 +1,1 @@
+CHANGED: Changed implementation of `(+>>)` to no longer require `KnownNat`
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
I think this change is an improvement.

But I think a better description of this PR would be 
> Rewrite (+>>) to no longer require `KnownNat`

I would add a changelog nonetheless; people might be interested in the removal of constraints, since this might simplify their code.

I was wondering a bit about whether the change is desired. #2808 ended prematurely because it would require adding `KnownNat` constraints to existing functions. We would have had better vector functions now if they had always carried a `KnownNat` constaint. In fact, I think it might well be justified to add new `KnownNat` constraints to the functions discussed in that issue just to get the improved versions. In practice, you really quickly require such a `KnownNat` constraint anyway in Clash code, so it doesn't seem to help much in practice to have a function that doesn't require such a constraint.

But I do think that getting rid of unnecessary constraints is good, so...

DigitalBrains1:
FYI, I don't know what went wrong with the status reporting, but GitLab CI _is_ running for the current tip of the branch. I expect the pending check to resolve into a completion status when the completion is finally reported by GitLab.

christiaanb (approved):
LGTM

PR #2939, opened 2025-04-18
Test 9.10 on macOS, add to README

DigitalBrains1:
On Linux, Clash has been working on GHC 9.10 for a while, both `master` and 1.8. This PR verifies that the same is the case on macOS, and adds these facts to our README.

Windows is not supported at this time, due to a dependency conflict regarding the `Win32` package.

All versions of Stackage run in CI are bumped to their latest minors.

For Stackage LTS 22 (GHC 9.6) the version bump made a workaround obsolete.

We can be a bit more selective with `allow-newer` in Stackage LTS 23 (GHC 9.8).

## Still TODO:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["macOS-13", "windows-latest"]
-        ghc: ["8.10", "9.0", "9.2", "9.4", "9.6", "9.8"]
+        ghc: ["8.10", "9.0", "9.2", "9.4", "9.6", "9.8", "9.10"]
         exclude:
           # Some tests fail with a mysterious -11 error code.
           - os: "macOS-13"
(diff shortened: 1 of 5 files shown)


martijnbastiaan (approved):

.github/workflows/ci.yml
```diff
@@ -37,6 +37,11 @@ jobs:
           - os: "windows-latest"
             ghc: "9.8"
 
+          # Cannot build on Windows due to problem with dependency on package
+          # "Win32"
```
martijnbastiaan, line 41: I think if you'd run `cabal outdated` on Windows you'd see the out-of-bounds message for `Win32`.

martijnbastiaan:
@DigitalBrains1 I think we can just remove the `Win32` dependency bounds. They're only used in in the `clash-ghc/src-bin-*`, which we inherit from GHC. Same story for `unix`.

DigitalBrains1:
> @DigitalBrains1 I think we can just remove the `Win32` dependency bounds. They're only used in in the `clash-ghc/src-bin-*`, which we inherit from GHC. Same story for `unix`.

As discussed, we copy from the [`ghc-bin` Cabal package](https://gitlab.haskell.org/ghc/ghc/-/blob/6111c5e4c4aad2f4e97af62ae304e96bed6ed1e3/ghc/ghc-bin.cabal.in), and this is not a transitive dependency of us. So we deemed it safer to not remove the bounds and just bump them.

PR #2945, opened 2025-04-27
Support 9.10 on Windows, add to README

DigitalBrains1:
This is primarily a backport of #2939 _Support 9.10 on Windows, add to README_, but some changes to CI were necessary as well, so these changes have also been backported. Specifically, we needed the new `haskell-actions/setup@v2` GitHub CI Action to support GHC 9.10, but for that, we also needed to change the architecture of our macOS runners from `aarch64` to `x86-64`.

The other changes to the GitHub CI workflow file that were made to `master` but not yet backported to `1.8` were also added, such that the workflow file is now once again up to date with `master`. These changes were not necessary, but are beneficial, and make future backports easier.

PR #2944 (adding a changelog file) is also included.

All GitHub CI actions, including the ones that are explicitly disabled or only run in certain scenarios, were run once on this PR to verify they still _could_ work.

## Still TODO:


martijnbastiaan (approved):
Surprising amount of work as always...

PR #2964, opened 2025-06-06
Fix Nix and test it in CI

DigitalBrains1:
Our dependencies for the Nix flake were not correctly declared. This could happen because our CI job for Nix actually only tries to build `clash-ghc`, but the problems are in packages that `clash-ghc` does not depend on. So change CI to actually try to build all our packages, and run any test suites they have. (Just Cabal `test-suite`s, not the executable `clash-testsuite`).

## Still TODO:

### DIFF
--- a/.ci/build_nix.sh
+++ b/.ci/build_nix.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+set -u
+
+nix \
+  --extra-experimental-features nix-command \
+  --extra-experimental-features flakes \
+  build -j$THREADS --log-format raw --max-silent-time 3600 "$@"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -87,8 +87,21 @@ nix-build:
   stage: test
   before_script:
    - nix-env -i gawk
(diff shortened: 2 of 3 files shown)


DigitalBrains1:
The dependency on Tasty 1.5 has been fixed, but the dependency on `tasty-flaky` still needs to be resolved, so this PR is unfinished. (I'd also like to do a few more manual tests: temporarily break stuff so I can verify CI actually fails.)

DigitalBrains1:
My thanks to @jaschutte for helping me with this.

.ci/build_nix.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
```
martijnbastiaan, line 2: Could you add `set -u` in case `$THREADS` isn't set? That would make it fail loudly instead of fail silently.

martijnbastiaan (approved):

.ci/build_nix.sh
DigitalBrains1: Yes, good point!

Issue #2972, opened 2025-06-16
No blackbox found for: `GHC.Internal.Base.eqString`


martijnbastiaan:
Looks like things got renamed. This is slightly different from https://github.com/clash-lang/clash-compiler/issues/2966, because there are no black boxes for it, only evaluator rules.

martijnbastiaan:
Oh, yeah, I like that! I'll report an `enhancement` issue after closing this.

DigitalBrains1:
We've released [v1.8.3](https://github.com/clash-lang/clash-compiler/releases/tag/v1.8.3), which includes a fix for this issue.

PR #2973, opened 2025-06-16
Use TH names in evaluator

martijnbastiaan:
Fixes https://github.com/clash-lang/clash-compiler/issues/2972

I've left the names for GHC 8.10 untouched, because I didn't feel like going through the pain for a GHC that we'll (hopefully) soon remove..

## Still TODO:

### DIFF
--- a/changelog/2025-06-16T11_46_57+02_00_fix2966
+++ b/changelog/2025-06-16T11_46_57+02_00_fix2966
@@ -1,0 +1,1 @@
+FIXED: No blackbox found for: `GHC.Internal.Control.Exception.Base.recSelError` on GHC 9.10 [#2966](https://github.com/clash-lang/clash-compiler/issues/2966)


clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -57,6 +57,7 @@ module Clash.Sized.Vector
   , merge
     -- * Modifying vectors
   , replace
+  , replace_int
     -- ** Permutations
```
christiaanb: Ah, too bad this PR required the export of said function. Understandable though.

christiaanb (approved):

clash-prelude/src/Clash/Sized/Vector.hs
```diff
@@ -33,7 +33,7 @@ module Clash.Sized.Vector
     -- ** Length information
   , length, lengthS
     -- ** Indexing
-  , (!!), head, last, at
+  , (!!), index_int, head, last, at
```
christiaanb: How come this export is needed, but the one for `replace_int` is not?

martijnbastiaan: Ah right, I had intentionally exported them, but it caused test failures (due to a test defining `replace_int`). I then changed my mind and just used a string again.. I'll do that for `index_int` too. I think the real(tm) solution is to create an `Internal` module.

martijnbastiaan: I've reported an issue here https://github.com/clash-lang/clash-compiler/issues/2974 to pick up later

clash-prelude/src/Clash/Sized/Vector.hs
martijnbastiaan: As discussed in the other comment, I've removed this export.

PR #2985, opened 2025-07-28
Patch tasty and time-compat dependencies on ghc98 and ghc96 shells.

jaschutte:
Tasty 1.5 was not available in these sets of packages, which broke those devshells. This includes the right version of tasty (by directly grabbing it from hackage).

Time-compat broke due to that (it's tests required tasty <1.5), solved by simply not running those tests.

## Still TODO:


DigitalBrains1 (approved):
Thanks!

PR #2993, opened 2025-08-21
Add `PolyKinds` to `Clash.Class.BitPack.Internal`

lmbollen:
We now also have `KnownNat n => BitPack (Proxy n)` . This should have been added in #2990


DigitalBrains1 (changes requested):
I don't think a commit titled _Add regression doctests for BitPack (Proxy a)_ should also change the definition of the instance without explaining why. Likewise, the whole PR and PR cover letter need to mention this change.

Your commit message for the other commit more clearly describes the change than the PR cover letter. How about changing the PR cover letter to:

> This gives us `BitPack` instances for `Proxy a` where `a` is a Kind from our derived `BitPack` instance for `Proxy a`. Examples are `Proxy 42`, `Proxy (n :: Nat)` or `KnownNat n => Proxy n`.
>
> This should have been added in https://github.com/clash-lang/clash-compiler/pull/2990
>
> This PR also corrects a mistake in the `BitPack` instance definition, which was not intended to have any constraints.
>
> This should also be backported to 1.8, see https://github.com/clash-lang/clash-compiler/pull/2991

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -12,7 +12,7 @@ Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
-
+{-# LANGUAGE PolyKinds #-} -- Required for BitPack (KnownNat n) => BitPack (Proxy n) instances
```
DigitalBrains1, line 15: Sorry I missed this on initial review, but what does `BitPack (KnownNat n)` mean? I initially mentally autocorrected this to `KnownNat n => BitPack (Proxy n)` which makes sense although it is overly specific: it forces `n` to be a `Nat` which needs `PolyKinds`. How about:

```suggestion
{-# LANGUAGE PolyKinds #-} -- Required for BitPack (Proxy (n :: Nat)) instances
```

It's still overly specific, but at least it makes sense. I don't think a `BitPack (KnownNat n)` constraint means anything?

DigitalBrains1 (changes requested):
As I indiicated in PR #2994, I'm confused by the structure you're going for. Why make a PR on your topic branch instead of just including the default extension here or closing this and making a new PR with a new title?

PR #2994, opened 2025-08-25
Add `PolyKinds` as default extension to `clash-prelude`

lmbollen:
`PolyKinds` is now a `default-extension` of `clash-prelude`. `NoPolyKinds` prevented the derivation of
typeclass instances of types that contained a Kind, for example `BitPack (Proxy 52)`.

* This PR also corrects a mistake in the BitPack instance definition, which was not intended to have any constraints.
* ~~This should also be backported to 1.8~~ We can't backport this to 1.8, as it is a change that makes some (shoddy) code not compile anymore.
 
## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Class/BitPack/Internal.hs
+++ b/clash-prelude/src/Clash/Class/BitPack/Internal.hs
@@ -58,6 +58,7 @@ import Clash.Sized.Internal.BitVector
 >>> :m -Prelude
 >>> :set -XDataKinds
 >>> import Clash.Prelude
+>>> import Data.Proxy
 -}
 
 -- | Convert data to/from a 'BitVector'. This allows functions to be defined
@@ -92,7 +93,7 @@ import Clash.Sized.Internal.BitVector
 -- places.
 --
 -- Clash provides some generic functions on packable types in the prelude, such
--- as indexing into packable stuctures (see "Clash.Class.BitPack.BitIndex") and
+-- as indexing into packable structures (see "Clash.Class.BitPack.BitIndex") and
 -- bitwise reduction of packable data (see "Clash.Class.BitPack.BitReduction").
 --
 class KnownNat (BitSize a) => BitPack a where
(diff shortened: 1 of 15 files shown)


DigitalBrains1:
Why did you create a PR on a PR instead of just including it into PR #2993? This feels confusing. If you first merge this into #2993 and then merge that into `master`, what will the history even look like?

I took the liberty of copy-editing the title.

lmbollen:
> Why did you create a PR on a PR instead of just including it into PR #2993? This feels confusing. If you first merge this into #2993 and then merge that into `master`, what will the history even look like?
> 
> I took the liberty of copy-editing the title.

Because I intend this to be a separate PR and now the commit history in the GUI reflects that.

DigitalBrains1:
Ah, I never considered merging PR #2933 as-is, adding `PolyKinds` to a module for a really really brief time and then reverting it again.

My preference would be to close the other PR and include all the changes here. But if you don't want to do that, I am not against merging first #2933 and then this. However, there's [one last thing to address](https://github.com/clash-lang/clash-compiler/pull/2993#discussion_r2301209921) in #2933. Again, I'm sorry I missed that one on initial review, I mentally corrected it to what I expected it to say.

DigitalBrains1:
I removed a superfluous change to `clash-prelude/src/Clash/Class/BitPack/Internal.hs` in the first commit, and added the removal of `PolyKinds` in `clash-prelude/src/Clash/Annotations/SynthesisAttributes.hs` as I think you simply overlooked that one.

DigitalBrains1:
Corrected a typo in the first commit message (and formatted the `changelog/` file to be a single line, which makes creating the `CHANGELOG.md` easier).

DigitalBrains1:
> I can make a more precise argument, but I hope this vague description already convinces you.

It does :-)

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -478,7 +479,15 @@ instance ( BitPack a
...
-instance BitPack (Proxy a) => BitPack (Proxy a)
+
+-- |
+-- >>> pack (Proxy @())
+-- 0
+-- >>> pack (Proxy @5)
+-- 0
+-- >>> pack (Proxy @Bool)
+-- 0
+instance BitPack (Proxy a)
```
leonschoorl, line 491: Unless the reader is very familiar with the details of the `Show BitVector` instance, it might not be clear from the example above that `BitSize (Proxy a) ~ 0`.
(I missed this point myself, reading it the first time.)

Maybe we should explicit like document that.

DigitalBrains1: Yes, the 0 is pretty confusing at first! I completely agree we should document it explicitly as a 0-sized construct.

martijnbastiaan: Added more documentation

martijnbastiaan (approved):

DigitalBrains1 (approved):
Here, have another aproval

Nice work untangling everything!

PR #3010, opened 2025-09-19
Add specialisation for powImpl1 over Ints

Rowan-Mather:
Adds an evaluator rule to handle another GHC specialization of `(^)` for `Int` arguments, namely the worker `$w$spowImpl1`.

## Still TODO:

  - [x] Add regression test


martijnbastiaan (approved):
Thanks! I'll add a changelog. After that I think we can merge :-).

martijnbastiaan:
I've added a regression test and found a few more missing. I've also added the change log, so this PR LGTM now.

martijnbastiaan:
@kloonbot run_ci ba17259510d7edeaf210386e5888dda3327d466f

PR #3029, opened 2025-10-03
Add convenience primitives

kleinreact:
The PR adds some more high level primitives that are convenient to use during Clash circuit design. In particular, the PR adds:

* ~~the static left and right shift operators `(<<%)` and `(%>>)`~~
* the ~~`apWhen`~~ `apEn` primitive, which is just a convenience extension of `mux`
* the `regEnN` primitive, which is the probably most primitive FIFO just chaining `regEn`s.

## Still TODO:

### DIFF
--- a/changelog/2025-10-03T10_14_19+02_00_add_convenience_primitives
+++ b/changelog/2025-10-03T10_14_19+02_00_add_convenience_primitives
@@ -1,0 +1,1 @@
+ADDED: `apEn`, and `regEnN`.


gergoerdi:
As a small comment, to me the name `apWhen` conjures the idea of `whenA :: (Applicative f) => f Bool -> f () -> f ()`. I think your `apWhen` is more closely related to `regEn`, so maybe `apEn` or `updateEn` or something would be a more intuitive name?

kleinreact:
> As a small comment, to me the name `apWhen` conjures the idea of `whenA :: (Applicative f) => f Bool -> f () -> f ()`. I think your `apWhen` is more closely related to `regEn`, so maybe `apEn` or `updateEn` or something would be a more intuitive name?

I totally agree. I'm happy with `apEn` as well. It's nice, short, free and **ap**pealing.

clash-prelude/src/Clash/Class/BitPack/BitShift.hs
```diff
@@ -0,0 +1,50 @@
...
+-- 1
+-- >>> (1 :: Unsigned 3) <<% d2
+-- 4
+-- >>> (1 :: Unsigned 3) <<% d4
+-- 0
+-- >>> pack (-1 :: Signed 6)
+-- 0b11_1111
+-- >>> reduceAnd (-1 :: Signed 6)
+infixl 7 <<%
+(<<%) :: forall a n. BitPack a => a -> SNat n -> a
```
christiaanb: I think this is too general and should be limited to `Bits a`. To me it makes no sense to add operators that can shift an e.g. `Either Bool (Maybe Int)` value.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -759,6 +762,35 @@ regEn = \clk rst gen initial en i ->
...
+  Signal dom Bool ->
+  -- ^ "push next input" indicator
+  Signal dom a ->
+  -- ^ input stream
+  Signal dom a
+  -- ^ final FIFO element
+regEnN clk rst gen n@SNat initial en = case compareSNat n (SNat @0) of
+  SNatLE -> id
+  SNatGT -> regEn clk rst gen initial en
+          . leToPlus @1 @n (regEnN clk rst gen (predSNat n) initial en)
```
christiaanb: Using `Vec` here would be probably be faster for both synthesis and simulation.

clash-prelude/src/Clash/Explicit/Signal.hs
```diff
@@ -759,6 +762,35 @@ regEn = \clk rst gen initial en i ->
   register# clk rst (andEnable gen en) initial initial i
 {-# INLINE regEn #-}
 
+-- | A simple queuing FIFO that pushes data through whenever the
```
christiaanb: This really is the most degenerate of FIFOs.... with many annoying properties;
1. Cannot distinguish new values from the initial values
2. Elements put into the FIFO cannot be gotten out until the FIFO has been completely filled up.

I would rather it be classified as a delay line rather than a FIFO.

clash-prelude/src/Clash/Class/BitPack/BitShift.hs
martijnbastiaan: I agree with this perspective @christiaanb. Also, I don't think it makes sense to go through a `BitVector` conversion for things that already have an efficient simulation path through `Bits`.

clash-prelude/src/Clash/Explicit/Signal.hs
martijnbastiaan: Yeah, something like:

```haskell
last . iterate (succSNat n) (register ....) 
```

would have an efficient HDL generation path and equivalentish simulation path.

clash-prelude/src/Clash/Signal/Internal.hs
```diff
@@ -1653,6 +1654,13 @@ mux :: Applicative f => f Bool -> f a -> f a -> f a
 mux = liftA3 (\b t f -> if b then t else f)
 {-# INLINE mux #-}
 
+-- | Updates a value inside an 'Applicative' context if and only if
+-- the given Boolean condition evaluates to 'True' in the same
+-- context.
+apEn :: Applicative f => f Bool -> (a -> a) -> f a -> f a
+apEn cond upd x = mux cond (upd <$> x) x
+{-# INLINE apEn #-}
```
martijnbastiaan: This documentation doesn't help hardware designers from traditional backgrounds / newcomers and is very distinct from the style of explanation above it. I think it's best to rely on the intuition built by `mux` and introduce it like you do in your PR cover letter:

> which is just a convenience extension of mux

I.e., something like:

```suggestion
-- | Like 'mux', but muxes between a given argument and an updated version.
-- Given "@'muxUpdate' b f s@", output an updated version of @s@ (i.e., @f s@)
-- when @b@ is 'True', but return it unchanged (i.e., @s@) when @b@ is 'False'.
muxUpdate :: Applicative f => f Bool -> (a -> a) -> f a -> f a
muxUpdate cond upd x = mux cond (upd <$> x) x
{-# INLINE muxUpdate #-}
```

I feel like I have written this construction inline many times, so I do think it makes sense to add it to the prelude.

clash-prelude/src/Clash/Class/BitPack/BitShift.hs
```diff
@@ -0,0 +1,47 @@
...
+{-# INLINE (<<%) #-}
+-- | Left shift the bits by the given amount.
+--
+-- >>> (1 :: Unsigned 3) <<% d0
+-- 1
+-- >>> (1 :: Unsigned 3) <<% d2
+-- 4
+-- >>> (1 :: Unsigned 3) <<% d4
+-- 0
+infixl 7 <<%
```
martijnbastiaan: Where does the `%` come from? Is this a pattern we establish anywhere else?

I get the impulse to have these short symbols for things, but if we do it to much Clash code will end up looking like a ~~Donald Duck cursing~~lens, see the discussion [over here](https://github.com/clash-lang/clash-compiler/pull/2545#issuecomment-2662314676). I'd therefore much rather just call them and use them like:

```haskell
d5 `snatBitShiftR` y
d5 `staticBitShiftR` y
-- (or whatever name seems sensible)
```

and restrain the punctuation overloading in `clash-prelude` to `.` and `#` (and `_` and ....) . Yes it's more verbose, but you don't have to rely on memorization to read Clash code.

clash-prelude/src/Clash/Class/BitPack/BitShift.hs
kleinreact: As discussed offline, it would be desirable to limit this operator to data structures that are product types, but no sum types. Basically every structure that is no sum type has proper semantics with respect to the structural operations of `Bits`. 

We can use `BitPack` to classify the instances accordingly. I added two PRs #3030 and #3031 both implementing the idea. One uses type `Bool`s for classification, the other one is purely constraint based.

With either of these extensions in place, we could add a `IsSumType ~ False` (#3030) or a `NoSumType` (#3031) constraint, which then gives the desired restriction making sure the user cannot mess up with this operator accidentally being applied to sum types in the first place.

clash-prelude/src/Clash/Class/BitPack/BitShift.hs
kleinreact: I found the `%` to be an appealing symbol for marking `SNat` arguments. We havn't used in that regard so far, but I also don't see any reasons speaking against it.

I'm usually in favor of using more descriptive names, but everybody immediately grasps the semantics when seeing `<<` and `>>` to read them as shifts, while using `snatBitShiftR` and `staticBitShiftR` will blow up your code significantly instead, e.g., compare
```haskell
(m %>> x & (1 <<% d3)) <<% m
```
with

```haskell
m `staticBitShiftL` (m `staticBitShiftR` x & (d3 `staticBitShiftL` 1)) `
```

It just hurts my eyes to find the `L` and `R` letters within all those characters and to not mix up the argument order every time.

martijnbastiaan: That's fair enough, though shouldn't it be something like `.<<.%`? Which is also very ugly, but at least consistent with existing operators from `Data.Bits`. Not sure about this.

> not mix up the argument order every time.

Well.. :-). The number of positions to shift by should always be the second argument. This is a precedent set by virtually all other programming languages, but also by Haskell itself (see `.<<.` and `.>>.`, but also `shiftL` and `shiftR`).

kleinreact: > That's fair enough, though shouldn't it be something like `.<<.%`?

That looks familiar, right! Wait ... give me sec ...

... yes! That's the guy I've seen at the train station yesterday ...
```
 %%%
.<<.%
 \_/
/---\
```
... very ugly face ... indeed  :sweat_smile:.

---

Anyway, I don't insist on any particular notion (nor do I care much about consistency with respect to the choice of characters). I only want to be able to remember the operator somehow and prefer to not break my fingers while typing it.

> Well.. :-). The number of positions to shift by should always be the second argument.

Right, but it breaks the syntactic hint of the `%` always pointing to the `SNat` argument. Then, you'd need `>>%` and `<<%` instead, which ... I don't know ... I just don't like as much :stuck_out_tongue_winking_eye:.

clash-prelude/src/Clash/Explicit/Signal.hs
kleinreact: Thanks for the suggestions. I updated it.

kleinreact:
As far as I can see all of the previous concerns already have been resolved a while ago. So is anybody still interested in these additions or can we close the PR?

PR #3031, opened 2025-10-05
Add `NoProductType` and `NoSumType` constraints

kleinreact:
The PR adds constraints to `BitPack`, which allow to check the instance type for being **no** sum- or **no** product type. Checking negated properties results from the limitation that constraints can only be composed conjunctively in this case, which is required for recursive detection over the generic structure.

An alternative `Bool`-based implementation is provided by #3030.

## Still TODO:


martijnbastiaan:
This won't be necessary anymore because https://github.com/clash-lang/clash-compiler/pull/3029#issuecomment-3405823286.

PR #3037, opened 2025-10-09
Only run `collapseRHSNoops` after constant propagation

### LINKED ISSUE #3036: Clash fails to constant fold (due to `collapseRHSNoops`)
I've tried my best to come up with a reproducer, but I can't seem to make Clash fail in a small example. Over at [bittide-hardware](https://github.com/bittide/bittide-hardware/), I've managed to get it down to:

```haskell
vexRiscvTest ::
  Clock System ->
  Reset System ->
  Enable System ->
  ( ( ()
    , Signal "System" (WishboneM2S 16 4 (BitVector 32))
    )
  , Signal "System" ()
  ) ->
  ( ( SimOnly MM.MemoryMap
    , Signal "System" (WishboneS2M (BitVector 32))
    )
  , Signal "System" (Unsigned 64)
  )
vexRiscvTest clk rst ena =
  withBittideByteOrder $
    withClockResetEnable clk rst ena $
      toSignals $
        timeWb @System @16
{-# OPAQUE vexRiscvTest #-}
```

where `timeWb` is:

```haskell
timeWb ::
  forall dom addrW.
  ( HiddenClockResetEnable dom
  , ?busByteOrder :: ByteOrder
  , ?regByteOrder :: ByteOrder
  , HasCallStack
  , KnownNat addrW
  , 1 <= DomainPeriod dom
  ) =>
  Circuit
    (MM.ConstBwd MM.MM, Wishbone dom 'Standard addrW (Bytes 4))
    (CSignal dom (Unsigned 64))
timeWb = circuit $ \mmWb -> do
  [scratchWb] <- MM.deviceWb "Timer" -< mmWb
  MM.registerWbI_ (MM.registerConfig "scratchpad") (0 :: Unsigned 64) -< (scratchWb, Fwd (pure Nothing))
  idC -< Fwd 0
```

Compiling this yields a constant folding error:

```
<no location info>: error:
    
    Clash.Netlist.BlackBox(218): Couldn't instantiate blackbox for Clash.Signal.Internal.register#. Verification procedure reported:
    
    Argument 5 should be literal, as blackbox used ~CONST[5], but was:
    
    DataCon (Vector 2 (BitVector 32)) VecAppend [Identifier (UniqueIdentifier {i_baseName = "result", i_baseNameCaseFold = "result", i_extensionsRev = [1,6], i_idType = Basic, i_hdl = Verilog, i_provenance = []}) (Just (Indexed (Vector 2 (BitVector 32),1,0))),DataCon (Vector 1 (BitVector 32)) VecAppend [Identifier (UniqueIdentifier {i_baseName = "result", i_baseNameCaseFold = "result", i_extensionsRev = [4,6], i_idType = Basic, i_hdl = Verilog, i_provenance = []}) (Just (Nested (Indexed (Vector 2 (BitVector 32),1,1)) (Indexed (Vector 1 (BitVector 32),1,0))))]]
```

As far as I can tell (but again, this doesn't trigger the bug), what we're doing boils down to:

```haskell
topEntity ::
  Clock System ->
  Reset System ->
  Enable System ->
  Signal System (BitVector 64) ->
  Signal System (BitVector 64)
topEntity clk rst ena = register clk rst ena initial
 where
  initial =
    pack
      . reverse
      . unpack @(Vec 8 (BitVector 8))
      . (resize :: BitVector 64 -> BitVector 64)
      . pack
      $ (0 :: Unsigned 64)
```

During the compilation process, Clash also emits a whole bunch of warnings (with `-fclash-debug DebugSilent`):

```
Unmatchable constant as case subject: reverse (unsafeCoerce# (unconcatBitVector# removedArg removedArg (fromInteger# 64 0 0)))
WHNF is: reverse
  (unsafeCoerce#
     (Cons _CO_
        (case (split# 56 (fromInteger# 64 0 0)) of
           (,) x bv' ->
             x[LocalId])
        (unconcatBitVector# 7 removedArg
           (case (split# 56 (fromInteger# 64 0 0)) of
              (,) x bv' ->
                bv'[LocalId]))))
```

After a little bit of digging it turns out that `collapseRHSNoops` introduces a `unsafeCoerce#`, which makes Clash unable to constant fold. Disabling that transformation also gets rid of the error and Clash happily generates (HITL-proof) HDL.

Going through the history of `collapseRHSNoops` it seems that the only reason we want this is because it can produce cleaner HDL. The downside is, apparently, that Clash can fail to generate HDL with it. I'll propose a patch that disables the transformation by default soon.

# Symptom management 
The "solution" I propose does sound a bit like symptom management, not an actual cure. The fact that Clash relies critically on the _cleanup_ phase to constant fold seems like an issue.

# `DebugApplied` log
[log.zip](https://github.com/user-attachments/files/22803564/log.zip)

martijnbastiaan:
Fixes #3036

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations/Inline.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Inline.hs
@@ -60,6 +60,7 @@ import Clash.Core.HasType
 import Clash.Core.Name (Name(..), NameSort(..))
 import Clash.Core.Pretty (PrettyOptions(..), showPpr, showPpr')
 import Clash.Core.Subst
+import qualified Clash.Core.Term as Term
 import Clash.Core.Term
   ( CoreContext(..), Pat(..), PrimInfo(..), Term(..), WorkInfo(..), collectArgs
   , collectArgsTicks, mkApps , mkTicks, stripTicks)
@@ -397,13 +398,20 @@ simply a variable reference. See issue #779 -}
 -- synthesis boundaries (NOINLINE/OPAQUE functions) to avoid running too early
 -- on functions that might be inlined later. See #3036.
 collapseRHSNoops :: HasCallStack => NormRewrite
-collapseRHSNoops _ letrec@(Letrec binds body) = do
+collapseRHSNoops _ letrec@(Let letBind body) = do
   (curFunId, _) <- Lens.use curFun
   curBinding <- lookupVarEnv curFunId <$> Lens.use bindings
   case curBinding of
     Just binding | isNoInline (bindingSpec binding) -> do
-      binds1 <- mapM runCollapseNoop binds
-      pure (Letrec binds1 body)
+      -- Explicitly match on Let instead of using LetRec, because we need to
+      -- preserve the structure. See https://github.com/clash-lang/clash-compiler/issues/3044.
+      case letBind of
+        Term.Rec binds -> do
+          binds1 <- mapM runCollapseNoop binds
+          pure (Let (Term.Rec binds1) body)
+        Term.NonRec b0 e0 -> do
+          (b1, e1) <- runCollapseNoop (b0, e0)
+          pure (Let (Term.NonRec b1 e1) body)
     _ -> pure letrec
   where
     runCollapseNoop orig =


clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -121,7 +121,7 @@ removeUnusedExpr _ e@(collectArgsTicks -> (p@(Prim pInfo),args,ticks)) = do
       tcm <- Lens.view tcCache
       (args1, Monoid.getAny -> hasChanged) <- listen (go tcm 0 usedArgs1 args)
       if hasChanged then
-        return (mkApps (mkTicks p ticks) args1)
+        changed (mkApps (mkTicks p ticks) args1)
```
christiaanb: I don’t understand the need for this change. All 
`changed` does is set the `Writer` to `Any True`: https://github.com/clash-lang/clash-compiler/blob/7db1405a3bb6c9d6574a472664c9cc1a0737948c/clash-lib/src/Clash/Rewrite/Util.hs#L323

But it has already been set to `Any True` otherwise we wouldn’t have taken this branch. That is, `listen` does __not__ reset the Writer back to `mempty`: https://hackage-content.haskell.org/package/transformers-0.6.2.0/docs/src/Control.Monad.Trans.RWS.Strict.html#listen

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -395,12 +395,15 @@ simply a variable reference. See issue #779 -}
...
 collapseRHSNoops :: HasCallStack => NormRewrite
-collapseRHSNoops _ (Letrec binds body) = do
-  binds1 <- mapM runCollapseNoop binds
-  return $ Letrec binds1 body
+collapseRHSNoops _ letrec@(Letrec binds body) = do
+  (changes, binds1) <- unzip <$> mapM runCollapseNoop binds
+  if or changes then changed (Letrec binds1 body) else pure letrec
   where
-    runCollapseNoop orig =
-      runMaybeT (collapseNoop orig) >>= Maybe.maybe (return orig) changed
```
christiaanb, line 403: I don’t understand this change. The `Any` Writer that we use already does the or-ing: https://hackage.haskell.org/package/ghc-internal-9.1201.0/docs/src/GHC.Internal.Data.Semigroup.Internal.html#line-239

There shouldn’t be a need to do it explicitly.

clash-lib/src/Clash/Normalize.hs
```diff
@@ -381,7 +381,12 @@ flattenCallTree (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
                  apply "reduceNonRepPrim" reduceNonRepPrim >->
                  apply "removeUnusedExpr" removeUnusedExpr) >->
                bottomupR (apply "flattenLet" flattenLet)) !->
-      topdownSucR (apply "topLet" topLet)
+      topdownSucR (apply "topLet" topLet) >->
+      -- See [Note] relation `collapseRHSNoops` and `inlineCleanup`
+      -- Note that we do this as the very last step, after all constant propagation
+      -- has been done to avoid #3036.
+      topdownSucR (apply "collapseRHSNoops" collapseRHSNoops) >->
```
christiaanb: This has the potential to still run collapseRHSNoops to early if the function in which we are currently flattening might itself be inlined and further flattened later. You need to take care that you only run the transformation when the function we are currently flattening is marked NOINLINE or OPAQUE.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
martijnbastiaan: That's what I thought too, but without this change compilation fails with an error that an expression has changed without a claim that it has changed. 

I'll see whether I can find the culprit.

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: So I think the property we're looking for is whether it is a "synthesis boundary". So, special names, noinline, and opaque? Or do we secretly add these annotations to special names?

Also, perhaps a stupid question but doesn't noinline/opaque get ignored when the function is non-representable? 

Anyway, just asking as a shortcut. I'll see if I can find the answers to these questions when I'm back at a computer.

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
martijnbastiaan: I agree, but I didn't see this behavior in practice. See other comment.

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: > So I think the property we're looking for is whether it is a "synthesis boundary". So, special names, noinline, and opaque? Or do we secretly add these annotations to special names?

They get an annotation using `setNoInlineTopEntities`

> Also, perhaps a stupid question but doesn't noinline/opaque get ignored when the function is non-representable?

Yes, but, these functions also don't get normalized on their own.

tests/Main.hs
```diff
@@ -724,7 +724,12 @@ runClashTest = defaultMain
           -- Vivado's mod misbehaves on negative dividend
           runTest "IntegralTB" def{hdlSim=hdlSim def \\ [Vivado]}
 
-        , runTest "NumConstantFoldingTB_1" def{clashFlags=["-itests/shouldwork/Numbers"]}
+        , runTest "NumConstantFoldingTB_1" def
+            { clashFlags = ["-itests/shouldwork/Numbers"]
+              -- IVerilog core dumps with:
+              -- ivl: draw_vpi.c:315: draw_vpi_taskfunc_args: Assertion `(unsigned)(dp - buffer) <= sizeof buffer' failed.
```
christiaanb: Because you are only allowed to print bit literals of 4096 characters in iverilog...

https://github.com/steveicarus/iverilog/blob/884349caabc81da7fb0780db24939a71c8aeb563/tgt-vvp/draw_vpi.c#L281

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
martijnbastiaan: So this turns out to be `Let` vs `Letrec`.

martijnbastiaan:
So after a bunch of back and forth, @christiaanb and I decided to go the original route:

* Add a flag enabling `collapseRHSNoops`. I.e., the transformation will be opt-in.
* Report an issue saying that a bunch of tests can't live without the flag
  * One of them pointing to _another_ Clash bug
  * One of them pointing to an IVerilog bug

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
martijnbastiaan: https://github.com/clash-lang/clash-compiler/issues/3044

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
martijnbastiaan: https://github.com/clash-lang/clash-compiler/issues/3044

tests/Main.hs
martijnbastiaan: https://github.com/clash-lang/clash-compiler/issues/3046

martijnbastiaan:
@christiaanb managed to fix all the issues mentioned in this thread, so we can do the neat solution of keeping the optimization in _and_ having the testsuite work as it did. Much rejoicing.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -393,11 +393,18 @@ in HDL, but not in Haskell, by `unsafeCoerce`.
...
+-- on functions that might be inlined later. See #3036.
 collapseRHSNoops :: HasCallStack => NormRewrite
-collapseRHSNoops _ (Letrec binds body) = do
-  binds1 <- mapM runCollapseNoop binds
-  return $ Letrec binds1 body
+collapseRHSNoops _ letrec@(Letrec binds body) = do
+  (curFunId, _) <- Lens.use curFun
+  curBinding <- lookupVarEnv curFunId <$> Lens.use bindings
+  case curBinding of
+    Just binding | isOpaque (bindingSpec binding) -> do
```
christiaanb: I think `isNoInline` is preferred https://github.com/clash-lang/clash-compiler/blob/a6f3230a26ccff44b076f10c0c74cfe396cc36e8/clash-lib/src/GHC/BasicTypes/Extra.hs#L52-L58

because that is what we use to determine how much of the hierarchy we flatten: https://github.com/clash-lang/clash-compiler/blob/a6f3230a26ccff44b076f10c0c74cfe396cc36e8/clash-lib/src/Clash/Normalize.hs#L298

clash-lib/src/Clash/Normalize.hs
christiaanb: I think we want to use `!->` instead of `>->`, so that we only run `inlineCleanup` when `collapseRHSNoops` succeeds.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
martijnbastiaan: That makes sense

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: Very well

PR #3043, opened 2025-10-15
Define `iterateI` `replicate`-like, preventing space leaks

### LINKED ISSUE #3042: Each element of `iterateI` retains pointer to whole list, causing space leaks
See https://github.com/bittide/bittide-hardware/issues/784#issuecomment-3401927342

martijnbastiaan:
Fixes #3042

## Still TODO:


christiaanb:
For every recursively defined function in `Vector` we have a blackbox, reduceNonRepPrim, and prim eval rule, so that compile-time evaluation is fast. This matters especially for “large” vectors. Perhaps we want to do the  same here?

PR #3060, opened 2025-10-24
Backport flake.nix changes to 1.8

jaschutte:
Cherry picked the commits from the main branch regarding the flakes.nix to the 1.8 branch.

### DIFF
--- a/nix/overlay-ghc910.nix
+++ b/nix/overlay-ghc910.nix
@@ -59,7 +59,7 @@ in
 
   singletons-th = prev.singletons-th_3_4;
 
-  # nixplgs doesn't include revision 1, changing dependency on template-haskell.
+  # nixpkgs doesn't include revision 1, changing dependency on template-haskell.
   string-interpolate = pkgs.haskell.lib.compose.overrideCabal (drv: {
     revision = "1";
     editedCabalFile = "sha256-oh0tR+LDFcVzQnm4kSrmhAU+P7wdai536d72Cvhzipg=";
--- a/nix/overlay-ghc96.nix
+++ b/nix/overlay-ghc96.nix
@@ -8,9 +8,10 @@ in
   # is basically abandonware it catches fire with brick 1.0+.
   brick = doJailbreak prev.brick_0_70_1;
 
+  # brick 0.70.1 requires vty < 6.0.
   vty = prev.callHackage "vty" "5.39" { };
 
-  # Marked as broken in nixpkgs, since it needs on a newer hashable than the
+  # Marked as broken in nixpkgs, since it needs a newer hashable than the
   # .cabal file currently uploaded to hackage.
   concurrent-supply = doJailbreak (markUnbroken prev.concurrent-supply);
 
--- a/nix/overlay-ghc98.nix
+++ b/nix/overlay-ghc98.nix
@@ -30,7 +30,7 @@ in
     sha256 = "sha256-irLM3aVMxpBgsM72ArulMXcoLY2glalVkG//Lrj2JBI=";
   }) {};
 
-  # This version of tasty isn't available in the nix ghc96 package set
+  # This version of tasty isn't available in the nix ghc98 package set
   tasty = prev.callHackageDirect {
     pkg = "tasty";
     ver = "1.5.3";
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -78,9 +78,30 @@ let
         });
 
       clash-cores =
-        hprev.callCabal2nixWithOptions "clash-cores" ../clash-cores "--flag nix" {
-          inherit (hfinal) clash-prelude;
-        };
+        let
+          # Remove the -fplugin and Setup.hs settings in the .cabal
+          # For ghc9101+ these options don't matter, but for ghc964 and ghc982 this breaks installation
+          # When entering the installPhase something (I'm not entirely sure what) goes wrong
+          # between Nix and GHC, causing Setup.hs to get invoked with the wrong set of packages
+          # (I think?). Removing the specific flags during installation fixes the issue for Nix,
+          # whilst not breaking regular compilation.
+          #
+          # Do note that this patch only gets applied during *installation* and not *compilation*
+          # That means these flags are still in place during compilation
+          override-attrs = if compilerVersion == "ghc964" || compilerVersion == "ghc982" then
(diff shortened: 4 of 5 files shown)


jaschutte:
@DigitalBrains1 would it be possible to re-run the failing task? I think it's a fluke. Aside from that I think this provides a solid base for the mergify PRs to *not* fail.

nix/overlay-ghc96.nix
```diff
@@ -1,13 +1,18 @@
 { pkgs }:
 final: prev:
 let
-  inherit (pkgs.haskell.lib) doJailbreak dontCheck markUnbroken;
+  inherit (pkgs.haskell.lib) doJailbreak markUnbroken dontCheck;
 in
 {
   # Use an older version than the default in nixpkgs. Since rewrite-inspector
   # is basically abandonware it catches fire with brick 1.0+.
   brick = doJailbreak prev.brick_0_70_1;
```
DigitalBrains1, line 10:
```suggestion

  # brick 0.70.1 requires vty < 6.0.
```

nix/overlay-ghc96.nix
```diff
@@ -1,13 +1,18 @@
...
 
+  vty = prev.callHackage "vty" "5.39" { };
```
DigitalBrains1, line 12: Why is this
```nix
  vty = prev.callHackage "vty" "5.39" { };
```
in `ghc96` but
```nix
  vty = doJailbreak (prev.callHackage "vty" "5.39" { });
```
in `ghc98`?

nix/overlay-ghc96.nix
```diff
@@ -1,13 +1,18 @@
...
+
+  # Marked as broken in nixpkgs, since it needs on a newer hashable than the
```
DigitalBrains1:
```suggestion
  # Marked as broken in nixpkgs, since it needs a newer hashable than the
```

DigitalBrains1: What does it mean when you say

> Needs a newer _x_ than the .cabal file currently uploaded to hackage.

That the package needs a newer version of _x_ than is currently uploaded to Hackage? So it needs an unreleased version? (Usually, we view the latest that is on Hackage as the latest release of a Haskell package.)

While the `.cabal` file specifies the version number for a specific version of a package, it's the uploaded package itself that is at that version. So you can just shorten it to

> Needs a newer _x_ than the latest version on Hackage.

But it feels like you mean something else entirely, as it's rather surprising that a released version of a package strictly requires an unreleased version of another package, so I'm probably misinterpreting.

nix/overlay-ghc98.nix
```diff
@@ -0,0 +1,43 @@
...
+  # We need a new tasty-flaky. The one from Hackage doesn't build for some weird
+  # reason..
+  tasty-flaky = prev.callCabal2nix "tasty-flaky" (pkgs.fetchFromGitHub {
+    owner = "LaurentRDC";
+    repo  = "tasty-flaky";
+    rev = "fc31a9d622c1eb60030a50152258a9bef785e365";
+    sha256 = "sha256-irLM3aVMxpBgsM72ArulMXcoLY2glalVkG//Lrj2JBI=";
+  }) {};
+
+  # This version of tasty isn't available in the nix ghc96 package set
```
DigitalBrains1:
```suggestion
  # This version of tasty isn't available in the nix ghc98 package set
```

nix/overlay-ghc910.nix
```diff
@@ -0,0 +1,71 @@
...
+    repo  = "tasty-flaky";
+    rev = "fc31a9d622c1eb60030a50152258a9bef785e365";
+    sha256 = "sha256-irLM3aVMxpBgsM72ArulMXcoLY2glalVkG//Lrj2JBI=";
+  }) {};
+
+  tasty = prev.tasty_1_5;
+
+  singletons-th = prev.singletons-th_3_4;
+
+  # nixplgs doesn't include revision 1, changing dependency on template-haskell.
```
DigitalBrains1:
```suggestion
  # nixpkgs doesn't include revision 1, changing dependency on template-haskell.
```

Pronounce the old one for fun 😂

nix/overlay.nix
```diff
@@ -192,7 +194,7 @@ let
               "clash-testsuite"
               ../tests
               "--flag workaround-ghc-mmap-crash" {
-              inherit (hfinal) clash-cores clash-ghc clash-lib clash-prelude;
+              inherit (hfinal) clash-ghc clash-lib clash-prelude;
```
DigitalBrains1: You'll find that [tests/shouldwork/Issues/T2781.hs](https://github.com/clash-lang/clash-compiler/blob/1328c34ac33de0c3214e370ebb9bb191d56d40c6/tests/shouldwork/Issues/T2781.hs) requires `clash-cores`, so please don't remove it from the `clash-testsuite` dependencies in this PR. I think it makes sense to properly remove this old `clash-cores` package from the 1.8 branch, but not here and now.

DigitalBrains1 (changes requested):
Ah, thanks! I have just a few points, I think we're almost there!

nix/overlay.nix
jaschutte: Oops! I noticed that one of the commits removed clash-cores so I was manually adding them back, thanks for catching it!

nix/overlay-ghc96.nix
jaschutte: I didn't write these comments haha, I cherry picked them from the main branch. However I think the comments refer to the version of Hackage pinned by the flake. It could be that actual Hackage has a newer version however the pinned set of packages is older, so we jailbreak it.

nix/overlay-ghc96.nix
jaschutte: Ghc98 has a different set of packages (since the compiler is newer) and in this set of packages for ghc98 certain packages (deepseq) has the wrong version, however it still compiles fine. That is why the jailbreak is required in ghc98 and not in ghc62 (where deepseq has the right version).

DigitalBrains1: Okay!

nix/overlay-ghc96.nix
DigitalBrains1: Oh, sorry, I should have checked where it came from before raising it here.

Well, you know what, if they just come from `master` let's use them as they are.

PR #3061, opened 2025-10-24
Bump typelit plugins

### LINKED ISSUE #3078: `CLog` type family reduction
Clash's normalization time for the following reproducer still grows exponentially in the argument to `Run`, e.g., on my machine `6s` for `Run 17`, `12s` for `Run 18`, `24s` for `Run 19`, and so forth ...

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE TypeFamilies #-}

{-# OPTIONS_GHC -fplugin GHC.TypeLits.Extra.Solver #-}
{-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise #-}
{-# OPTIONS_GHC -fplugin GHC.TypeLits.KnownNat.Solver #-}

import Clash.Prelude
import GHC.TypeLits.KnownNat

type family Calc (n :: Nat) (t :: Nat) :: Nat where
  Calc 0 _ = 1
  Calc n t = Calc (n - 1) (t + t)

instance (KnownNat n, KnownNat t) => KnownNat2 $(nameToSymbol ''Calc) n t
 where
  natSing2 = SNatKn $ calc (natToNum @n) (natToNum @t)
  {-# INLINE natSing2 #-}

calc :: Natural -> Natural -> Natural
calc 0 _ = 1
calc n t = calc (n - 1) (t + t)

type Run n = Calc (n - CLog 2 2) 0

topEntity :: Int
topEntity = natToNum @(Run 17)
```

The recipe is simple: we use a custom type family `Calc` recursing `n`-times over the first argument and creating an exponentially growing expression via `t + t` in the second argument. Unfolding the expression, but not evaluating the `+`-type family then causes the exponential blowup.

Now this problem should have been fixed with #3071, which takes care of reducing the type expression before handing it over to Clash. However, the problem here is the `CLog 2 2` being part of the expression to be reduced as well. Looking at the implementation of [CLog](https://hackage-content.haskell.org/package/ghc-typelits-extra-0.5.1/docs/GHC-TypeLits-Extra.html#t:CLog) in `ghc-typelits-extra` reveals that the equations of the type-family are given by the plugin, but are not part of the Haskell definition at first. Hence, the solution of #3071 won't be able to reduce `CLog 2 2`, as there is no reduction defined.

This being the actual issue is quite easy to check, as just replacing `CLog 2 2` by `1` in the above reproducer fixes the exponential blowup. The behavior can also be checked via loading the example with `ghci` and executing

```
ghci> :k! Calc (11 - CLog 2 2) 0
Calc (11 - CLog 2 2) 0 :: Nat
= Calc (11 - CLog 2 2) 0
ghci> :k! Calc (11 - 1) 0
Calc (11 - 1) 0 :: Natural
= 1
```

It may be that the problem is better to be fixed in the plugin, however, the exponential blowup still is a problem of Clash, which is why I'm reporting the error here at first.

kleinreact:
The PR bumps `ghc-typelits-natnormalise`, `ghc-typelits-knownnat`, and `ghc-typelits-extra` to the latest versions and introduces some required code changes in `clash-prelude` to make it compile.

## Still TODO:
  - [x] Blocked by https://github.com/clash-lang/ghc-typelits-natnormalise/issues/108
  - [x] ~~Requires #3062~~ obsolete due to https://github.com/clash-lang/ghc-typelits-natnormalise/pull/110
  - [x] Blocked by https://github.com/clash-lang/ghc-typelits-natnormalise/issues/109
  - [x] Requires https://github.com/clash-lang/ghc-typelits-natnormalise/pull/110

## Nice to have:
  - [x] https://github.com/clash-lang/ghc-typelits-extra/pull/64
  - [x] Check whether it fixes #3078


clash-prelude/clash-prelude.cabal
```diff
@@ -335,9 +335,11 @@ Library
...
                       extra                     >= 1.6.17  && < 1.9,
                       ghc-prim                  >= 0.5.1.0 && < 0.13,
-                      ghc-typelits-extra        >= 0.4     && < 0.5,
-                      ghc-typelits-knownnat     >= 0.7.2   && < 0.8,
-                      ghc-typelits-natnormalise >= 0.7.2   && < 0.8,
+  -- >>> lower bounds temporarily increased for testing purposes
+                      ghc-typelits-extra        >= 0.5.1   && < 0.6,
+                      ghc-typelits-knownnat     >= 0.8.2   && < 0.9,
+                      ghc-typelits-natnormalise >= 0.9.1   && < 0.10,
+  -- <<<
```
leonschoorl: If the new code in Vector.hs doesn't typecheck with the old version of the plugin(s), the lower bound increases should be permanant.
But possibly conditional on the GHC version.

christiaanb:
I’m adding some features and fixes in https://github.com/clash-lang/ghc-typelits-natnormalise/pull/110 which will hopefully only require haddock/doctest changes w.r.t. error messages for `clash-prelude`.

DigitalBrains1:
> [...] all that's needed to compile clash-prelude warning-free and pass the unit-tests and doctests: [9b4ca8a](https://github.com/clash-lang/clash-compiler/commit/9b4ca8a8ceb9604ab0054fd3b8481784f94641b6)


I note you don't increase the lower bounds on the dependencies. So does it work with the whole range of versions in the interval then? I'm having trouble understanding how you fix a compilation error by merely allowing _more_ versions of a dependency to be used...

christiaanb:
I have released version [0.9.3](https://hackage.haskell.org/package/ghc-typelits-natnormalise-0.9.3/changelog) of the natnormalise plugin, which requires no changes to any of our `clash-*` packages. See also https://github.com/clash-lang/clash-compiler/commit/c6f2ed3e4f19157d2e7bb01ab04973a426f1c11e

DigitalBrains1:
There have been [three](https://github.com/clash-lang/clash-compiler/pull/3061#issuecomment-3442760021) [separate](https://github.com/clash-lang/clash-compiler/pull/3061#discussion_r2569553925) [comments](https://github.com/clash-lang/clash-compiler/pull/3061#issuecomment-3601478934) discussing the need to exclude certain versions to not produce build plans that don't compile. These have not been addressed in any way. Could you please explain why, e.g., a compilation with `ghc-typelits-natnormalise` 0.9.1, is now suddenly okay when it wasn't before? Or if it isn't, what is preventing a user from constructing such a build plan? I have to admit I'm a bit peeved by the total lack of a response to these issues.

christiaanb:
Well, as you can see, that build had an updated cabal.project file with an allow newer on all the plugins: https://github.com/clash-lang/clash-compiler/commit/18934e9a8f1dcd37a1c670a011491f74b12899eb#diff-b8ed06757045fac949c89f2139a862498ad2b6d1f82c61a31e7c91d6cf0eaa70R62-R67

Which yeah, does show that everything builds and passes tests, but did hide that we forgot to up the upper bound with `clash-prelude-hedgehog`

DigitalBrains1:
Since that means that the CI for this PR didn't actually test the changes in the PR, I'd say that needs to be addressed.

clash-prelude/clash-prelude.cabal
DigitalBrains1: The code changes have been reverted as no longer needed with the updated plugins.

Actually skipping the use of the problematic plugin version that necessitated these code changes will be addressed by deprecating plugins on Hackage that lead to build plans that do not compile (per https://github.com/clash-lang/clash-compiler/pull/3061#issuecomment-3607642495)

clash-lang-builder (approved):

christiaanb (approved):

PR #3062, opened 2025-10-24
Bump typelit plugins (preparations)

kleinreact:
The PR introduces some preparations for #3061 to make `Clash.Sized.Vector` compile with `ghc-typelits-natnormalise-0.9.1`, `ghc-typelits-knownnat-0.8.2`, and `ghc-typelits-extra-0.5.1`. 

The update avoids the erroneous, but currently required unification under type families (cf. https://github.com/clash-lang/ghc-typelits-natnormalise/issues/108), which has been fixed by https://github.com/clash-lang/ghc-typelits-natnormalise/commit/c994419b8ad3f6c3c8bb21150ae4e11c5183957a.

 The modifications are only introduced for GHCs between 9.4 and 9.10, as the introduced fix won't work for GHCs prior to 9.4 and is not needed any more with GHC 9.12 due to https://github.com/ghc/ghc/commit/a63ee33a116599c88e703d9793e025e17c59f7f9.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Sized/Vector.hs
+++ b/clash-prelude/src/Clash/Sized/Vector.hs
@@ -2654,16 +2654,42 @@ dtfold :: forall p k a . KnownNat k
 dtfold _ f g = go (SNat :: SNat k)
   where
     go :: forall n . SNat n -> Vec (2^n) a -> (p @@ n)
+#if __GLASGOW_HASKELL__ >= 904 && __GLASGOW_HASKELL__ < 912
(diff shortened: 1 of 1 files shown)


kleinreact:
I added the strictness requirement in the argument again, which fixes the failing doctest. However, the pattern matches require some additional inaccessibility proofs for the non-reachable cases then, which have been added as well. 

This version will also work with the updated plugins avoiding the issue reported in https://github.com/clash-lang/ghc-typelits-natnormalise/issues/108.

PR #2941, opened 2025-04-19
Add support for GHC 9.12

martijnbastiaan:
## Still TODO:

  - [ ] Write a changelog entry (see changelog/README.md)
  - [X] ~~Check copyright notices are up to date in edited files~~
  - [ ] Fix issues exposed with `cabal run -- clash-testsuite -j24 -p VACC -p clash --hide-successes`

[comment]: # (Depending on the change, some of these points may not be necessary. If so, leave the boxes unchecked so it is easier for a reviewer to see what has been omitted.)

PR #3070, opened 2025-11-06
1.8.4 release

DigitalBrains1:


martijnbastiaan:
@DigitalBrains1 Do you want me to merge https://github.com/clash-lang/clash-compiler/pull/2941, so we can just do a constraint bump once 9.12.3 is out? We should be _really_ close to a release.

Issue #3082, opened 2025-11-17
`dumpVCD` can only dump 93 signals


marijn-qbaylogic:
The `dumpVCD` function only supports up to 93 traces, stating that this is the maximum allowed by the standard, when it is in fact not. More signals can be assigned by assigning multi-character labels to signals.

DigitalBrains1:
We've released [v1.8.5](https://github.com/clash-lang/clash-compiler/releases/tag/v1.8.5), which includes a fix for this issue.

PR #3083, opened 2025-11-17
Removed the limit on the number of traces in `dumpVCD`

marijn-qbaylogic:
`dumpVCD` was limited to 93 traces, because it could only generate single character trace identifier codes. The codes are now generated ad infinitum, and the restriction has been removed.

Additionally, single bit traces were logged with an extra newline, causing empty lines in the VCD. This has been removed as well.

Fixes: #3082

### DIFF
--- a/changelog/2025-11-17T21_35_22+01_00_fixed_dumpvcd_trace_limit
+++ b/changelog/2025-11-17T21_35_22+01_00_fixed_dumpvcd_trace_limit
@@ -1,1 +1,1 @@
-FIXED: Removed the limit on the number of traces in `dumpVCD`. Also removed unnecessary newlines after 1-bit signals. [#3082](https://github.com/clash-lang/clash-compiler/issues/3082)
+FIXED: Removed the limit on the number of traces in `dumpVCD`. Also removed unnecessary double newlines after 1-bit signals. [#3082](https://github.com/clash-lang/clash-compiler/issues/3082)
--- a/clash-prelude/src/Clash/Signal/Trace.hs
+++ b/clash-prelude/src/Clash/Signal/Trace.hs
@@ -372,7 +372,7 @@ dumpVCD## (offset, cycles) traceMap now
  where
   offensiveNames = filter (any (not . printable)) traceNames
 
-  -- Generate labels like reversed digits: 0:[1,2,01,11,21,02,12,22,001,...]
+  -- Generate labels in the pattern a,b,c,aa,ab,ac,ba,bb,bc,ca,cb,cc,aaa,...
   labels = concatMap (\s -> map (snoc s) alphabet) ([]: labels)
    where
     alphabet = map chr [33..126]


clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -373,7 +371,12 @@ dumpVCD## (offset, cycles) traceMap now
  where
   offensiveNames = filter (any (not . printable)) traceNames
 
-  labels = map chr [33..126]
+  -- Generate labels like reversed digits: 0:[1,2,01,11,21,02,12,22,001,...]
+  labels = "!" : map go [1..]
+   where
+    go 0 = ""
+    go n = chr ( 33 + n `mod` k) : go (n `div` k)
+    k = 126-33+1
```
DigitalBrains1: How about this instead:
```suggestion
  labels = concatMap (\s -> map (snoc s) alphabet) ([]: labels)
   where
    alphabet = map chr [33..126]
```

You'll need to import
```haskell
import Data.List.Extra (snoc)
```

clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -413,22 +416,22 @@ dumpVCD## (offset, cycles) traceMap now
...
 
   -- | Format single value according to VCD spec
-  format :: Width -> Char -> Value -> String
-  format 1 label (0,0)   = ['0', label, '\n']
-  format 1 label (0,1)   = ['1', label, '\n']
-  format 1 label (1,_)   = ['x', label, '\n']
+  format :: Width -> String -> Value -> String
+  format 1 label (0,0)   = '0': label <> "\n"
+  format 1 label (0,1)   = '1': label <> "\n"
+  format 1 label (1,_)   = 'x': label <> "\n"
```
DigitalBrains1: This file seems to use `++` for concatenating strings, in the interest of consistency I'd like to see that maintained here.

marijn-qbaylogic: Looking at these lines, why do we even add a newline? It isn't there a few lines later:
```hs
format n label (mask,val) =
  "b" ++ map digit (reverse [0..n-1]) ++ " " ++ label
```

clash-prelude/src/Clash/Signal/Trace.hs
marijn-qbaylogic: This is one fancy way of doing things.
Are there specific things that make it better? I do agree it's better, but there might be stuff I am missing.

DigitalBrains1: If the alphabet were `"abc"`, this is what it would generate:

```haskell
>>> import Text.Show.Pretty 
>>> pPrint $ P.take 20 labels
[ "a"
, "b"
, "c"
, "aa"
, "ab"
, "ac"
, "ba"
, "bb"
, "bc"
, "ca"
, "cb"
, "cc"
, "aaa"
, "aab"
, "aac"
, "aba"
, "abb"
, "abc"
, "aca"
, "acb"
]
```

It is almost how a base-_n_ number system would work, except that 0 is also significant in leading positions. It just seems like a clean numbering to me.

It's also more Haskelly if you ask me :-).

clash-prelude/src/Clash/Signal/Trace.hs
DigitalBrains1: You're completely right, they don't need a newline! It's already handled in two `Text.intercalate "\n"`s. Let's remove it.

clash-prelude/src/Clash/Signal/Trace.hs
DigitalBrains1: The recursion step of the algorithm in words is roughly this:

To form all labels of length up to _n_+1 we take all labels up to a length _n_ and for all of those labels and all letters of the alphabet, we append the letter of the alphabet to the label. Also, for each letter of the alphabet, include just that letter as a label (because by increasing the length of all labels we lost the labels of length 1 plus we clearly need to start somewhere).

changelog/2025-11-17T21_35_22+01_00_fixed_dumpvcd_trace_limit
```diff
@@ -0,0 +1 @@
+FIXED: Removed the limit on the number of traces in `dumpVCD`. Also removed unnecessary newlines after 1-bit signals. [#3082](https://github.com/clash-lang/clash-compiler/issues/3082)
```
DigitalBrains1:
```suggestion
FIXED: Removed the limit on the number of traces in `dumpVCD`. Also removed unnecessary double newlines after 1-bit signals. [#3082](https://github.com/clash-lang/clash-compiler/issues/3082)
```

clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -373,7 +372,10 @@ dumpVCD## (offset, cycles) traceMap now
  where
   offensiveNames = filter (any (not . printable)) traceNames
 
-  labels = map chr [33..126]
+  -- Generate labels like reversed digits: 0:[1,2,01,11,21,02,12,22,001,...]
```
DigitalBrains1:
```suggestion
  -- Generate labels almost like digits: if the alphabet were "012", generate
  -- [0,1,2,00,01,02,10,11,12 ..]
```

Although frankly it can do without the comment; it's just that the current comment is wrong. Feel free to just omit it altogether.

DigitalBrains1 (changes requested):
I'd like to stress that newlines are necessary after one-bit signals; it's just that we emitted two of them.

Please also add something similar to the PR cover letter (the first message in the PR).

changelog/2025-11-17T21_35_22+01_00_fixed_dumpvcd_trace_limit
DigitalBrains1: A different way to phrase it would be

```suggestion
FIXED: Removed the limit on the number of traces in `dumpVCD`. Also removed unnecessary blank lines after 1-bit signals. [#3082](https://github.com/clash-lang/clash-compiler/issues/3082)
```

I think that's even more clear.

PR #3085, opened 2025-11-19
GitHub CI: New MacOS + update actions versions

DigitalBrains1:
MacOS 13 is phased out: GitHub maintains the latest two versions of a runner OS.

I have also run the _Build and Test_ CI job to be absolutely sure the updated action didn't break stuff.

## Still TODO:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,12 +20,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


DigitalBrains1:
I belatedly discovered we switched architectures. The default architecture for the old macOS-13 runners was x86_64, but the default architecture for macOS-15 is aarch64. Since we actually expect more people to be interested in aarch64, this change is not a problem.

PR #3087, opened 2025-11-19
Add Nix caching

jaschutte:
Adds caching as part of CI. This caches *all* packages Clash exposes and all developer shells. Cachix gets populated on main and the local cache on every commit. CI will benifit from the local cache while all 'stable' commits are available on Cachix.

I split the building of the Nix packages up in the 'common' packages (ghc, lib and prelude) and the 'other' packages. The common packages are build and uploaded sequentially, while the other packages all get their own job. By splitting the tasks up like this it speeds up the final compilation & upload time.

## Still TODO:


jaschutte:
This will also package it to the local cache. The GitLab only caches the default GHC version of Clash while this caches all supported versions (at least, supported by the flake).

This is mostly the code from the rewrite from Gitlab -> Github actions and I thought it may be useful to have this be split off as a separate PR.

martijnbastiaan (approved):
Got it!

.ci/attic-auth.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -uo pipefail
+
+mkdir -p ~/.config/attic
+cat <<EOF > ~/.config/attic/config.toml
+default-server = "public"
+
+[servers.public]
+endpoint = "http://192.168.102.136:9200"
```
martijnbastiaan: AFAIK, `diepenheim.local:9200` should just work.

PR #3108, opened 2025-12-11
Add fail-fast and cancel old jobs to Nix caching

jaschutte:
Cancels older caching jobs on the same branch when pushed to. Also sets fail-fast to false. Despite one caching job failing, others may still be useful when cached (for re-runs for example).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -6,7 +6,7 @@ on:
       - '*'
 
 concurrency:
-  group: ${{ github.head_ref || github.run_id }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:


DigitalBrains1:
> on the same commit (for force-pushes). 

Not commit, but branch. And for any push, not just force pushes. And that is precisely what we want.

We use the `github.head_ref` context property in `ci.yml` but note that `ci.yml` only runs for PR's (and `master`). `head_ref` is only set for PR's. But the `nix.yml` workflow runs for any push, not just PR's. I'm looking at the documentation and wondering why we don't use `github.ref` throughout. Also note that this concurrency is a global property, by which I mean that if both `ci.yml` and `nix.yml` specify a group key of `github.ref`, they will cancel each other as well. This is currently not a problem, but we could solve it by using `group: ${{ github.workflow }}-${{ github.ref }}`. That feels cleaner, that way the concurrency decision is up to each workflow instead of them interacting with each other.

I'd like to ask @martijnbastiaan if he knows why we use `github.head_ref` instead of `github.ref`, and if there is no important reason, I'd like to suggest we use `group: ${{ github.workflow }}-${{ github.ref }}` in both `ci.yml` and `nix.yml`. 

The change to `ci.yml` is not strictly necessary, but may be good for our own sanity and the clarity of the mechanism. No need to think about what happens if someone names a PR branch `Nix Packaging-refs/heads/main` and their next push cancelling the `nix.yml` job on `main` that was running. Oh man, I am really reaching now, what would be the odds of that 😂.

DigitalBrains1:
Okay, so let's use `group: ${{ github.workflow }}-${{ github.ref }}` in both `ci.yml` and `nix.yml`!

[edit]
> provided I was the one who did it in the first place

[You were not](https://github.com/clash-lang/clash-compiler/commit/aa9335e39ea601e951ef004652bb2665b5b3b252)
[/edit]

PR #3111, opened 2025-12-19
Improve deriveAutoReg constraint calculations

leonschoorl:
This fixes the constraint calculation issues in `deriveAutoReg` seen in https://clash-lang.discourse.group/t/deriveautoreg-with-superclass-constraints/100

In addition to looking for constraints on AutoReg instances fields,
`deriveAutoReg` now also looks for constraints on the NFDataX instance of the whole type.

## Still TODO:

  - [x] Add a test?
  - [x] Update changelog file to qlog format?


clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -435,6 +441,21 @@ findTyVarSubsts = go
...
+
+When looking up typeclass instances for tuples we get a slight mismatch
+in the representation of the tuple types.
+
+Internally in GHC a 3-tuple is represented as `TupleT 3`.
+But our API is Name based, so we look them up as `ConT "(,,)"`, see 'deriveAutoRegTuple'.
+This confused 'findTyVarSubsts', as it gets both forms and tries to unify them.
+To allow it to continue I just allowed it to see them as equal.
+This may seem dangerous, but isn't any more dangerous then seeing arbitratry ConT's
+or LitT's as equal.
```
martijnbastiaan: > This may seem dangerous, but isn't any more dangerous then seeing arbitratry ConT's or LitT's as equal.

But `findTyVarSubsts` doesn't do that either, so that doesn't seem like a valid excuse.

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -435,6 +441,21 @@ findTyVarSubsts = go
...
     _ -> error $ unlines [ "findTyVarSubsts: Unexpected types"
                          , "ty1:", pprint ty1,"ty2:", pprint ty2]
+{-
+NOTE [TupleT ConT equality]
+
+When looking up typeclass instances for tuples we get a slight mismatch
+in the representation of the tuple types.
+
+Internally in GHC a 3-tuple is represented as `TupleT 3`.
+But our API is Name based, so we look them up as `ConT "(,,)"`, see 'deriveAutoRegTuple'.
```
martijnbastiaan, line 474: Local patch ups like these are asking for trouble IMO as they allow bugs (in GHC this time) to silently pass. So two questions:

* Why couldn't we emit `TupleT`?
* In this case, could we compare tuple constructor names? That way we only really fire when we want.

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -421,6 +426,7 @@ findTyVarSubsts = go
     (ImplicitParamT _ x1, ImplicitParamT _ x2) -> go x1 x2
     (PromotedT _          , PromotedT _          ) -> []
     (TupleT _             , TupleT _             ) -> []
+    (TupleT _             , ConT _               ) -> [] -- see NOTE [TupleT ConT equality]
```
martijnbastiaan, line 452: `TupleT` / `ConT` is handled, but `ConT` / `TupleT` isn't. That might be fine in the current context, but it would be very surprising if the context ever changes.

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -326,7 +326,10 @@ deriveAutoRegProduct tyInfo conInfo = go (constructorName conInfo) fieldInfos
             [] -> [| $tyConE |]
 
     autoRegDec <- funD 'autoReg [clause argsP (normalB body) decls]
-    ctx <- calculateRequiredContext conInfo
+    ctxAutoReg <- calculateRequiredContext conInfo
+    -- look up if the NFDataX superclass has any (extra) constraints
+    ctxNFDataX <- constraintsWantedFor ''NFDataX [ty]
+    let ctx = nub (ctxAutoReg ++ ctxNFDataX)
```
martijnbastiaan: `nub` has very poor computational complexity, is `nubOrd` an option?

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -326,14 +326,36 @@ deriveAutoRegProduct tyInfo conInfo = go (constructorName conInfo) fieldInfos
...
-    ctxNFDataX <- constraintsWantedFor ''NFDataX [ty]
-    let ctx = nub (ctxAutoReg ++ ctxNFDataX)
+    ctxNFDataX <- fmap nub $ constraintsWantedFor ''NFDataX [ty]
+    let ctxNFDataXfiltered = removedImpliedNFXby ctxAutoReg ctxNFDataX
+    let ctx = nub (ctxAutoReg ++ ctxNFDataXfiltered)
     return [InstanceD Nothing ctx (AppT (ConT ''AutoReg) ty)
               [ autoRegDec
               , PragmaD (InlineP 'autoReg Inline FunLike AllPhases) ]]
 
+-- | Looks through the constraits in the 2nd argument and
```
martijnbastiaan: constraints

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
martijnbastiaan: Also: "arbitrary" and "more dangerous than".

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
```diff
@@ -326,14 +326,36 @@ deriveAutoRegProduct tyInfo conInfo = go (constructorName conInfo) fieldInfos
...
+-- drops any `NFDataX a`, when there is a corresponding `AutoReg a` in the first argument
+removedImpliedNFXby :: Cxt -> Cxt -> Cxt
```
martijnbastiaan: `removedImpliedNFDataXby`

martijnbastiaan (changes requested):

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
leonschoorl: I was referring to this:
```
  go ty1 ty2 = case (ty1,ty2) of
    ...
    (ConT _               , ConT _)                -> []
    (LitT _               , LitT _               ) -> []
```

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
leonschoorl: 1. We don't emit `ConT "GHC.Tuple.Tuple2"` ourselves, we get it from [DatatypeInfo.datatypeName](https://hackage-content.haskell.org/package/th-abstraction-0.7.2.0/docs/Language-Haskell-TH-Datatype.html#t:DatatypeInfo), which we get from [reifyDatatype](https://hackage-content.haskell.org/package/th-abstraction-0.7.2.0/docs/Language-Haskell-TH-Datatype.html#v:reifyDatatype).
2. As for checking the tuple constructor names, we could.
   But is seems somewhat arbitrary to check that for example `TupleT 3`, matches  `ConT "GHC.Tuple.Tuple3"`, but not guard against `ConT "Unsigned"` not matching `ConT "Signed"`, or even `TupleT 3` and `TupleT 4`.

   (And also such a check would be somewhat complicated code to get it working with multiple GHC versions, because the Tuple type names have changed a couple of times between GHC versions.)

leonschoorl:
@rowanG077 We don't have to recursively keep doing instance lookups

```haskell
module Constraints where
import Prelude

-- simplified versions of relevant type classes:
class NFDataX a where
  rnfX :: a -> ()
class NFDataX a => AutoReg a where
  autoReg :: a -> [a] -> [a]

data Foo a = Foo a deriving Show

-- Lets say our NFDataX instance has some extra constraint:
instance Eq (Foo a) => NFDataX (Foo a) where
  -- ...

{-
Before this PR, deriveAutoReg would generate an instance that looks like this:
> instance AutoReg a => AutoReg (Foo a) where ...

But GHC would refuse that with:
    • Could not deduce ‘Eq (Foo a)’
        arising from the superclasses of an instance declaration
      from the context: AutoReg a
        bound by the instance declaration at Constraints.hs:21:10-39
    • In the instance declaration for ‘AutoReg (Foo a)’
-}

-- This PR looks up the instance of NFDataX (Foo a) and copies its constraints to generate:
instance (AutoReg a, Eq (Foo a)) => AutoReg (Foo a) where
  -- ... 


-- A question Rowan had was what happens if the Eq (Foo a) instance also has some extra constraint,
-- don't we have to keep recursively looking up instances?
-- 
-- Lets say our Eq instance also has some weird extra constraint:
instance Show a => Eq (Foo a) where
  x == y = show x == show y

-- The answer is we don't have to, we can push that constraint onto the user of our instance.
-- For example:
f i = autoReg (Foo i)
-- Whose type gets inferred as:
--  f :: (AutoReg a,Show a) => ..
```

clash-prelude/src/Clash/Class/AutoReg/Internal.hs
leonschoorl: I didn't mean to imply that the code sees a ConT as equal to a LitT.

I've simplified the comment to avoid this confusion.

rowanG077:
Stuff like this does not yet work:

```
data Extra = Extra

data Message = Message Int
  deriving (Generic)

instance NFDataX Extra => NFDataX Message
deriveAutoReg ''Message
```

leonschoorl:
@rowanG077 it is intentional that that gives an error.

`deriveAutoReg ''Message` generates:
```
instance AutoReg Message where ...
```
which fails to compile with:
```
error: 
    • No instance for ‘NFDataX Extra’
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘AutoReg Message’
```

`deriveAutoReg` could be changed to generate the instance like this:
```
instance NFDataX Message => AutoReg Message where ...
```
Which seems like it would silence this error. (and it does)
But it just moves the error to the point where the `AutoReg Message` instance is used.

The error is a valid, the definition `instance NFDataX Extra => NFDataX Message` requires an `instance NFDataX Extra`, which isn't provided.
The user needs to provide that instance, and the sooner they learn that (via an error message) the better.

rowanG077 (approved):
LGTM

PR #3112, opened 2025-12-23
Update flake

rowanG077:
This updates the nix flake to the newest nixos-unstable. Fully updates to the most recent GHC minor version as well as bump the typelit plugins.

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -236,7 +236,7 @@ let
                 prev.yosys
               ]} \
               --set LIBRARY_PATH ${prev.lib.makeLibraryPath [
-                prev.ghdl-llvmf
+                prev.ghdl-llvm
                 prev.zlib.static
               ]}
           '';


clash-lib/clash-lib.cabal
```diff
@@ -373,7 +373,7 @@ test-suite doctests
     build-depends:
       base,
       clash-lib,
-      doctest-parallel >= 0.2 && < 0.4,
+      doctest-parallel >= 0.2 && < 0.5,
```
rowanG077, line 376: @martijnbastiaan Was there a reason this (and clash-prelude) was still locked to `<0.4` searching the issues brought nothing up. Quickly looking over the changes from `doctest-parallel` 0.3 to 0.4 also didn't bring anything controversial for me. Doctests still seem to work.

.github/workflows/nix.yml
```diff
@@ -18,9 +18,9 @@ jobs:
       fail-fast: false
       matrix:
         ghc: [
-          "ghc9101",
-          "ghc982",
-          "ghc964"
+          ghc9102,
```
kleinreact: Why not directly switching to 9.10.3? We are successfully using it with Clash in our project flakes.

rowanG077: Good catch! I checked which version was in nixpkgs and somehow overlooked 9.10.3

DigitalBrains1: Well, we don't actually test 9.10.3 in CI, so by the old adage "if we don't test it, it doesn't work", actually we don't support 9.10.3 ;-).

If you want to use 9.10.3, I suggest you add it to CI as well so we actually know it passes.

clash-lib/clash-lib.cabal
martijnbastiaan: No reason. IMO we can also just make this upper-bound less, I'm not planning on big changes.

martijnbastiaan:
I have no idea @DigitalBrains1, sorry. It wouldn't surprise me if it was deliberate though, every other GHC release seems to be broken in some way.

DigitalBrains1:
Maybe I'm just mixing stuff up with GHC 9.12, though.

GHC 9.12 is broken for us, and that is why we are not in Stackage nightly. But I'm not sure what the situation is for 9.10.

I think we can switch to 9.10.3, provided we add it to GitLab CI, because we want to know when we break something.

@christiaanb Do you agree? Or do you remember why we're still on 9.10.2?

martijnbastiaan:
IMO just do it. If it didn't work, we should have left a note..

DigitalBrains1:
Right, I wasn't very explicit, but I also meant, let's upgrade to 9.10.3.

DigitalBrains1 (approved):
My requested changes have been taken care of, approved.

However, I think you will still need to do something about 9.8.4 on Nix. I suspect `ffi-interface-tests` is your culprit https://github.com/clash-lang/clash-compiler/pull/3112#issuecomment-3689988120.

rowanG077:
The reason the clash-ffi build succeeds on GHC 9.10.3 and not on GHC 9.8.4 seems to be that cabal errors on empty package definition with the cabal that is used with GHC 9.8.4 but no longer with GHC 9.10.3.

What I have done now is marked `clash-ffi` broken on the GHC version that it doesn't function on in nix as well and removed the builds from the matrix in CI.

@jaschutte To do this I had to change the order of the overlays. With `ghcOverlay` first there was no way to mark `clash-ffi` broken only on some specific GHC version. I don't think this has further impact for us but I'd like you to confirm that!

I think if @jaschutte is oke with this and CI passes it can be merged.

clash-ffi/clash-ffi.cabal
```diff
@@ -78,13 +78,14 @@ library
     Clash.FFI.VPI.Reg
   -- TODO: remove this once https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
   -- is fixed
-  if impl(ghc < 9.8.3) || impl(ghc == 9.10.1)
+  if impl(ghc < 9.8.3) || impl(ghc == 9.10.1) || impl(ghc > 9.10.2)
```
christiaanb, line 80: Why not:
```
if impl(ghc < 9.8.3) || impl(ghc >= 9.10.1)
```

clash-ffi/clash-ffi.cabal
```diff
@@ -115,7 +116,7 @@ executable ffi-interface-tests
     , tasty-smallcheck
   -- TODO: remove this once https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
   -- is fixed
-  if impl(ghc < 9.8.3) || impl(ghc == 9.10.1)
+  if impl(ghc < 9.8.3) || impl(ghc == 9.10.1) || impl(ghc > 9.10.2)
```
christiaanb, line 117: Why not
```
if impl(ghc < 9.8.3) || impl(ghc >= 9.10.1)
```

clash-ffi/clash-ffi.cabal
DigitalBrains1: Because GHC 9.10.2 has the issue.

DigitalBrains1: Of course we could do

```
if impl(ghc < 9.8.3) || impl(ghc > 9.10.2)
```

and just ignore the fact that 9.10.1 would actually work.

christiaanb: Ah, perhaps we should reverse the logic and do:
```
-- GHC 9.8.4 and 9.10.2 have issues, see https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
if impl(ghc == 9.8.4) || impl(ghc == 9.10.2)
  buildable: False
else
  buildable: True
```

DigitalBrains1: You're thinking nobody will be trying 9.8.3 anymore? In that case, it would indeed match everything in practice.

What happened to 9.8.3? :-) Do you know why it was pulled from GHCup? I couldn't find anything at a really quick glance, and my dinner is cooling.

DigitalBrains1: The GHCup default distribution channel decided not to ship 9.8.3 because of issues with the `filepath` package: https://discourse.haskell.org/t/ghc-9-8-3-is-now-available/10576/26?u=hasufell

However, 9.8.3 is still in the `vanilla` channel, and I think this might have some users? In that case, we're not helping them by `impl(ghc == 9.8.4) || impl(ghc == 9.10.2)`. I'd feel happier doing `impl(ghc < 9.8.3) || impl(ghc > 9.10.2)`; it's not more verbose and switches the sense. Instead of allowing 9.8.3 to do broken stuff, we disallow 9.10.1 even though it works fine. And let's be honest, how many people use `clash-ffi` in the first place? :-)

[edit]
> Do you know why it was pulled from GHCup?

Not pulled; never released. Made a wrong assumption.
[/edit]

christiaanb: Ah, sorry, I missed that 9.8.3 also has the bug... so yeah, then it would be:
```
-- GHC 9.8.3, 9.8.4 and 9.10.2 have issues, see https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
if impl(ghc == 9.8.3) || impl(ghc == 9.8.4) || impl(ghc == 9.10.2)
  buildable: False
else
  buildable: True

Comment (christiaanb) [reply clash-ffi/clash-ffi.cabal:80]: Perhaps we can just leave the logic as is, but update the comment to specify which GHCs are affected

Comment (DigitalBrains1) [reply clash-ffi/clash-ffi.cabal:80]: Fine by me! What do we do with
```haskell
-- TODO: remove this once https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
-- is fixed
```

I'm not sure what it suggests we do, but I think it's saying: once GHC 9.8 and GHC 9.10 both have received a minor version update that fixes the bug, remove this stanza. I wonder whether that would even be a good idea.

Also, after reading the discussion about 9.8.3 I linked to in my previous message, I doubt there will ever be another 9.8 release.

You wrote the comment; what did you intend?

christiaanb: Ah, I meant that when a fix was released, we should properly fix the bounds. The original version didn’t allow building on e.g. GHC 9.12

DigitalBrains1: Okay I've updated the comments from
```
-- TODO: remove this once https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
-- is fixed
```
to
```
-- GHC 9.8.3, 9.8.4 and 9.10.2 have issues, see https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12264#note_602406
```

I note that in `.ci/build.sh` the phrasing was a bit different:
https://github.com/clash-lang/clash-compiler/blob/d0f65c47fe946699cb7818d24ae16dd8e7aff286/.ci/build.sh#L42-L44

I think it'll be quite  a while before we can actually _put it back into tests_; I suppose that will be once we drop 9.8 support from Clash. In the interest of clarity and maintainability, I've rewritten it a bit so it doesn't duplicate shell code and is easier to grok.

Are you satisfied with how it is now, @christiaanb ?

rowanG077: @christiaanb Just pinging you again on this PR so it doesn't slip through the cracks. If you are not satisfied yet please let us know.

christiaanb: Looks good!

Issue #3115, opened 2026-01-08
`holdReset` output is not registered


JvWesterveld:
The output of [`holdReset`](https://hackage-content.haskell.org/package/clash-prelude-1.8.4/docs/Clash-Explicit-Reset.html#v:holdReset) is not registered: it is a comparison with a counter value. I believe it should be preferable to have flipflops drive reset signals rather than LUTs given their typical high-fanout nature.

The implementation cannot be changed if wishing to retain the current timing behaviour. Perhaps a note or warning could be added to the Haddock documentation where the user is recommended to register the output themselves through something like:
```
rstReg :: forall dom. (KnownDomain dom) => Clock dom -> Reset dom -> Reset dom
rstReg clk = unsafeToReset . E.register clk noReset enableGen initVal . unsafeFromReset
 where
  initVal = case resetPolarity @dom of
    SActiveHigh -> True
    SActiveLow -> False
```

DigitalBrains1:
Erm you would think the glitchy nature of the comparison would also be a problem with asynchronous resets.

I think the current function is unsafe for that reason, and that is a major issue since it is not called `unsafeHoldReset`. I'm leaning towards making it safe for asynchronous resets and making the output registered in one go, and _still_ backporting it to 1.8.

[edit]
Hah! Replied at the exact same time!
[/edit]

christiaanb:
How can it glitch as a result of its internal state? So there are four situations:

1. The counter is not at max value, the reset is unasserted
2. The counter is not at max value, the reset is asserted
3. The counter is at max value, the reset is unasserted
4. The counter is at max value, the reset is asserted

I don't see in what situations the output goes `asserted -> unsasserted -> asserted` or `unasserted -> asserted -> unasserted` between clock cycles. How would the comparitor be implemented at the gate-level to cause this glitching?

rowanG077:
A concrete example: Imagine `maxBound` is `110`, current state is `011`. `011` is incremented to `100`. You'd expect no assertions. However since all bit paths don't change at the same time there could be this ordering:

1.  bit 1 is flipped `011` -> `111`
2.  Bit 3 is flipped `111` -> `110` -> ASSERTION
3.  Bit 2 is flipped `110` -> `100`

DigitalBrains1:
We've released [v1.8.5](https://github.com/clash-lang/clash-compiler/releases/tag/v1.8.5), which includes a fix for this issue.

PR #3116, opened 2026-01-08
Fix `holdReset`  glitches for async resets and add `registerSyncReset`

rowanG077:
Aimed to address https://github.com/clash-lang/clash-compiler/issues/3115

This updates the documentation of `holdReset` to indicate the problems it has. It also gives a new function `stretchReset` that should have the correct behaviour but has slightly different timing characteristics so it's not a drop in replacement.

I have marked the `holdReset` function with a warning, I don't like it since now we introduce warning into our own project when exposing it and using it in tests. I also thought a deprecation pragma. Both have some implication as that requires a version bump according to haskell PVP specification.

## Still TODO:

  - [x] Write test
  - [x] Adjust doctest

### DIFF
--- a/clash-prelude/src/Clash/Explicit/Reset.hs
+++ b/clash-prelude/src/Clash/Explicit/Reset.hs
@@ -1,5 +1,5 @@
 {-|
-Copyright  :  (C) 2020-2023, QBayLogic B.V.,
+Copyright  :  (C) 2020-2026, QBayLogic B.V.,
                   2022-2023, Google LLC
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
@@ -23,6 +23,7 @@ module Clash.Explicit.Reset
   , resetGlitchFilter
   , resetGlitchFilterWithReset
   , unsafeResetGlitchFilter
+  , registerSyncReset
   , holdReset
   , convertReset
   , noReset
@@ -399,12 +400,54 @@ resetGlitchFilter# SNat reg dffSync rstIn0 =
       SActiveHigh -> True
       SActiveLow -> False
 
--- | Hold reset for a number of cycles relative to an incoming reset signal.
+-- | Register a synchronous reset signal.
+--
+--   `registerSyncReset` delays an incoming reset by one clock cycle using a
+--   register. This can be useful to break combinational paths involving reset
+--   logic.
+--
+--   __NB__: This is not a synchronizer. Use `resetSynchronizer` to synchronize
+--   a reset.
+--
+--   Example:
+--
+-- >>> let sampleReset = sampleN 7 . unsafeToActiveHigh
+-- >>> let rst = unsafeFromActiveHigh (fromList [False, True, False, False, True, False])
+-- >>> sampleReset (registerSyncReset @XilinxSystem clockGen rst enableGen True)
+-- [True,False,True,False,False,True,False]
+--
+registerSyncReset
+  :: forall dom
+   . KnownDomain dom
+  => HasSynchronousReset dom
+  => Clock dom
+  -> Reset dom
+  -> Enable dom
+  -> Bool
+  -- ^ Initial assert value of the register if supported by the domain.
+  --   If True the initial reset value is asserted.
+  --   If False the initial reset value is de-asserted.
+  -> Reset dom
+registerSyncReset clk (unsafeFromReset -> rst) en initialValue = unsafeToReset outRst
+  where
+    intialRst :: Bool
+    intialRst =
+      case resetPolarity @dom of
+        SActiveHigh -> initialValue
+        SActiveLow -> not initialValue
+    outRst = delay clk en intialRst rst
+
+-- | Hold reset for a number of cycles relative to an incoming reset
+-- signal.
+--
+-- __NB__: The output of this function is combinational for @n > 0@ on domains
+-- with a synchronous reset. Use `registerSyncReset` to add an output register if
+-- desired.
 --
 -- Example:
 --
--- >>> let sampleWithReset = sampleN 8 . unsafeToActiveHigh
--- >>> sampleWithReset (holdReset @System clockGen enableGen (SNat @2) (resetGenN (SNat @3)))
+-- >>> let sampleReset = sampleN 8 . unsafeToActiveHigh
+-- >>> sampleReset (holdReset @System clockGen enableGen (SNat @2) (resetGenN (SNat @3)))
 -- [True,True,True,True,True,False,False,False]
 --
 -- 'holdReset' holds the reset for an additional 2 clock cycles for a total
@@ -412,7 +455,7 @@ resetGlitchFilter# SNat reg dffSync rstIn0 =
 -- intermediate assertions of the reset signal:
 --
 -- >>> let rst = fromList [True, False, False, False, True, False, False, False]
--- >>> sampleWithReset (holdReset @System clockGen enableGen (SNat @2) (unsafeFromActiveHigh rst))
+-- >>> sampleReset (holdReset @System clockGen enableGen (SNat @2) (unsafeFromActiveHigh rst))
 -- [True,True,True,False,True,True,True,False]
 --
 holdReset
@@ -427,11 +470,66 @@ holdReset
   -> Reset dom
   -- ^ Reset to extend
   -> Reset dom
-holdReset clk en SNat rst =
-  unsafeFromActiveHigh ((/=maxBound) <$> counter)
+holdReset clk en n rst = case resetKind @dom of
+  SSynchronous -> holdResetSync clk en n rst
+  SAsynchronous -> holdResetAsync clk en n rst
+
+holdResetSync
+  :: forall dom n
+   . KnownDomain dom
+  => DomainResetKind dom ~ 'Synchronous
+  => Clock dom
+  -> Enable dom
+  -- ^ Global enable
+  -> SNat n
+  -- ^ Hold for /n/ cycles
+  -> Reset dom
+  -- ^ Reset to extend
+  -> Reset dom
+holdResetSync clk en sn@SNat rst = rstOut
  where
-  counter :: Signal dom (Index (n+1))
-  counter = register clk rst en 0 (satSucc SatBound <$> counter)
+  rstOut = case snatToInteger sn of
+    0 -> rst
+    1 -> orReset rst (unsafeToReset regReset)
+      where
+        isActiveHigh = case resetPolarity @dom of { SActiveHigh -> True; _ -> False }
+        regReset = register clk rst en isActiveHigh (pure (not isActiveHigh))
+    _ -> orReset rst (unsafeToReset rawRst)
+      where
+        counter :: Signal dom (Index (n + 1))
+        counter = register clk rst en 0 (satSucc SatBound <$> counter)
+        rawRst :: Signal dom Bool
+        rawRst = case resetPolarity @dom of
+          SActiveHigh -> (/=maxBound) <$> counter
+          SActiveLow -> (==maxBound) <$> counter
+
+holdResetAsync
+  :: forall dom n
+   . KnownDomain dom
+  => DomainResetKind dom ~ 'Asynchronous
+  => Clock dom
+  -> Enable dom
+  -- ^ Global enable
+  -> SNat n
+  -- ^ Hold for /n/ cycles
+  -> Reset dom
+  -- ^ Reset to extend
+  -> Reset dom
+holdResetAsync clk en sn@SNat rst = rstOut
+ where
+  isActiveHigh = case resetPolarity @dom of { SActiveHigh -> True; _ -> False }
+  rstOut = case toUNat sn of
+    UZero -> rst
+    USucc UZero -> unsafeToReset (register clk rst en isActiveHigh (pure (not isActiveHigh)))
+    USucc (USucc _) -> unsafeToReset (register clk rst en isActiveHigh rawRst)
+      where
+        counter :: Signal dom (Index n)
+        counter = register clk rst en 0 (satSucc SatBound <$> counter)
+        rawRst :: Signal dom Bool
+        rawRst = case resetPolarity @dom of
+          SActiveHigh -> (/=maxBound) <$> counter
+          SActiveLow -> (==maxBound) <$> counter
+
 
 -- | Convert between different types of reset, adding a synchronizer when
 -- the domains are not the same. See 'resetSynchronizer' for further details
--- a/changelog/2026-01-08T16_58_32+01_00_add_registerSyncReset
+++ b/changelog/2026-01-08T16_58_32+01_00_add_registerSyncReset
@@ -1,0 +1,1 @@
+ADDED: `registerSyncReset` [#3115](https://github.com/clash-lang/clash-compiler/issues/3115)
--- a/changelog/2026-01-08T16_58_32+01_00_fix_holdreset
+++ b/changelog/2026-01-08T16_58_32+01_00_fix_holdreset
@@ -1,0 +1,1 @@
+FIXED: Fix `holdReset` glitch behaviour for asynchronous resets and wrong hold cycles for sync resets. [#3115](https://github.com/clash-lang/clash-compiler/issues/3115)
--- a/clash-prelude/src/Clash/Signal.hs
+++ b/clash-prelude/src/Clash/Signal.hs
@@ -2,7 +2,7 @@
 Copyright  :  (C) 2013-2016, University of Twente,
                   2016-2019, Myrtle Software Ltd,
                   2017     , Google Inc.,
-                  2021-2024, QBayLogic B.V.
+                  2021-2026, QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -156,6 +156,7 @@ module Clash.Signal
   , unsafeFromActiveLow
   , resetSynchronizer
   , resetGlitchFilter
+  , registerSyncReset
   , holdReset
     -- * Enabling
   , Enable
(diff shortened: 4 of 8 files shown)


clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -433,6 +450,110 @@ holdReset clk en SNat rst =
...
+  -> SNat n
+  -- ^ Hold for /n/ cycles, counting from the moment the incoming reset
+  -- signal becomes deasserted.
+  -> Reset dom
+  -- ^ Reset to extend
+  -> Reset dom
+stretchResetAsync clk en SNat rst = rstOut
+ where
+  -- This ensures the index at least has one cycle for counting.
+  counter :: Signal dom (Index ((Max 2 ((Max 1 n) - 1))))
```
rowanG077: I dislike this a lot but imo it's better than forcing `3 <= n` constraint on the user.

christiaanb: Why not make it so that you create different circuits for different values of `n`? i.e.
```haskell
stretchResetAsync clk en sn@SNat rst = rstOut
 where
  rstOut = case snatToInteger sn of
    0 -> rst
    1 -> unsafeToReset (delay clk en isActiveHigh (unsafeFromReset rst))
    _ -> unsafeToReset (if isActiveHigh then fmap (/=maxBound) counter else fmap (==maxBound) counter)

  counter :: Signal dom (Index n)
  counter = register clk rst en 0 (fmap (satSucc SatBound) counter)

  isActiveHigh = case resetPolarity @dom of { SActiveHigh -> True; _ -> False }
```
I explicitly left out the resetSynchronizer because the incoming reset should already be synchronously deasserted for the counter to work properly.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -433,6 +447,135 @@ holdReset clk en SNat rst =
...
+-- count it as a full cycle.
+--
+stretchReset
+  :: forall dom n
+   . KnownDomain dom
+  => Clock dom
+  -> Enable dom
+  -- ^ Global enable
+  -> SNat n
+  -- ^ Hold for /n/ cycles
```
christiaanb: I think we should create circuits that respect this instruction, and we should just create different circuits for different values of _n_:
* Identify for 0
* A simple dflipflop for 1
* The counter for 2 or more

That also simplifies the whole documentation

clash-prelude/src/Clash/Explicit/Reset.hs
rowanG077: You need the reset synchronizer to stop the glitches as indicated by the comment in https://github.com/clash-lang/clash-compiler/issues/3115#issuecomment-3723921420

I _thought_ only one FF is enough though since there should be Recovery and Removal timing checks from the counter output FFs to the reset input. But I'm not 100% sure. 

My reasoning for not being 100% sure is that the recovery and removal check only apply to the de-assertion path so timing closure will not ensure any glitch assertion will be safe with respect to recovery and removal checks.

clash-prelude/src/Clash/Explicit/Reset.hs
rowanG077: This is possible for sync resets. Not for async resets as you need the reset synchronizer.

clash-prelude/src/Clash/Explicit/Reset.hs
DigitalBrains1: In the linked comment you write _assertion_, but what you are describing is a glitch _deassertion_ of the outgoing reset. Just FYI; I wondered if you got them mixed up.

christiaanb: You just need a simple register on the output to stop the glitching. So in that case `rstOut` would be something along the lines of:
```haskell
case snatToInteger sn of
  0 -> rst
  1 -> unsafeToReset (register clk rst en isActiveHigh (pure (not isActiveHigh))
  _ -> unsafeToReset (register clk rst en isActiveHigh (if isActiveHigh then fmap (/=maxBound) counter else fmap (==maxBound) counter)
```

rowanG077: @DigitalBrains1 I think I described both? 2. is a assertion glitch, 3. is a de-assertion glitch. Am I missing something here? 

@christiaanb 

Yes, you can do that!  However now the async reset is no longer asserted in the same cycle as before. So it introduces a cycle delay. I thought same cycle reset is desirable. One extra FF is really a tiny cost imo.

DigitalBrains1: > @DigitalBrains1 I think I described both? 2. is a assertion glitch, 3. is a de-assertion glitch. Am I missing something here?

The reset output is `unsafeFromActiveHigh ((/=maxBound) <$> counter)` and `maxBound` is `0b110`. We're counting from `0b011` to `0b100`, so we were expecting to stay in reset, we still have 2 cycles to go after that. So the reset outputs for the given sequence of events is:

1.  bit 1 is flipped `011` -> `111` -> reset asserted
2.  Bit 3 is flipped `111` -> `110` -> RESET DEASSERTED
3.  Bit 2 is flipped `110` -> `100` -> reset asserted

This is a single glitch, where we wanted the reset to stay asserted but instead it briefly glitched to deassertion. I only see a single deassertion glitch?

rowanG077: Hmmm thinking more about it:

is one standard FF really safe? On an incoming asynchronous assertion the counter is asynchronous reset which which might hit in the setup/hold window of the register input causing potential metastability on the output of the FF. Async resets inputs are designed to handle that. I honestly don't know if this is safe for the standard register input.

Edit:

The key thing that might save us there is that the register is asynchronously reset using the original reset. So if the counter based reset on the FF input is asserted then at the very least you know for certain the register is in reset, or very soon will be in reset.

christiaanb: @rowanG077

For:
```haskell
rstOut = case snatToInteger sn of
  0 -> rst
  1 -> unsafeToReset (register clk rst en isActiveHigh (pure (not isActiveHigh))
  _ -> unsafeToReset (register clk rst en isActiveHigh (if isActiveHigh then fmap (/=maxBound) counter else fmap (==maxBound) counter)
```
The `rstOut` follows the assertion of `rst` almost immediately because the registers are level sensitive, there is no full clock cycle delay on the assertion.

The implementation of my first comment, using `delay`, was wrong. The implementation using `register` has the behavior I intended.

christiaanb: @rowanG077 as I already indicated, the reset that we want to stretch already has to synchronized (synchronous deassertion) otherwise the `counter` goes meta-stable.

rowanG077: Yes I'm talking about asynchronous assertion case. An asynchronous assertion on the D input can cause meta stability, but since the FF is also reset with the same asynchronous assertion it is _probably safe_. Personally I'd at least like a canonical source of that. Especially if the goal is that it just saves us a single FF/allows a single cycle less stretch.

Perhaps you missed my edit.

rowanG077: > > @DigitalBrains1 I think I described both? 2. is a assertion glitch, 3. is a de-assertion glitch. Am I missing something here?
> 
> The reset output is `unsafeFromActiveHigh ((/=maxBound) <$> counter)` and `maxBound` is `0b110`. We're counting from `0b011` to `0b100`, so we were expecting to stay in reset, we still have 2 cycles to go after that. So the reset outputs for the given sequence of events is:
> 
>     1. bit 1 is flipped `011` -> `111` -> reset asserted
> 
>     2. Bit 3 is flipped `111` -> `110` -> RESET DEASSERTED
> 
>     3. Bit 2 is flipped `110` -> `100` -> reset asserted
> 
> 
> This is a single glitch, where we wanted the reset to stay asserted but instead it briefly glitched to deassertion. I only see a single deassertion glitch?

Yes you are correct! I mixed them up.

christiaanb: @rowanG077 With the edit I can now longer follow whether you agree or disagree that my implementation is (glitch-)safe or not?

rowanG077: @christiaanb I'm not sure myself :) I agree with the reasoning but it's finicky. We are doing something that is not really allowed (Potential assertion of the D input during setup/hold window) and explain it away with, well yes we do that, but if that happens the FF will go into reset so it's fine. 

Unless there is a canonical example or this I'd err on the side of just including the tried and true standard reset synchronizer.

christiaanb: But the asynchronous change on `D` only happens when the asynchronous CLR/SET ports are asserted, which have priority, meaning the register will not go meta-stable.

DigitalBrains1: Yeah, isn't this what usually happens on an asynchronous reset? Flip-flops will transition their outputs immediately, leading to timing violations, but this is normal and okay because the flip-flops with the inputs are also reset. It only becomes a problem when you go out of the reset domain.

[edit]
The same can be said for the second FF in our asynchronous reset synchronizer. When the incoming reset asserts, the Q of the first FF will transition immediately, but that's fine because the second FF is also reset. The output of the second FF is a reset signal; the situation appears nearly identical except that there are LUTs between the flip-flops where in the asynchronous reset synchronizer the connection is direct, so the transition on D is even closer to the reset assertion.
[/edit]

christiaanb: And even if the outer flop does go meta-stable because CLR/SET port is asserted later on the outer flop than the CLR/SET ports of the `counter` flops, the circuit will recover, because eventually the assertion will arrive at the CLR/SET port of the outer flop.

christiaanb: Also, if you want to use the standard dual FF reset synchronizer, that's fine by me. The only thing I care about is that `stretchReset` has a simple semantics. If I want the to stretch the deassertion of the reset by 1 cycle, it should just stretch the deassertion by 1 cycle, regardless of whether I have synchronous or asynchronous resets.

DigitalBrains1: I think it's important that you can configure it precisely, to control the order in which stuff comes out of reset, but the ability to stretch by very small amounts seems less important (not sure, just can't think of a case where that'd matter much). So if it can't stretch by just 1 or 2 cycles, I think that'd be fine, but if you say `n` it should really mean `n`.

rowanG077: Yes indeed! I now am convinced that this is safe. I was thinking of a possibility where the rst source -> sync FF would be much slower than rst source -> counter FF -> comb logic -> sync FF could induce meta stability. But that's just a standard case with async resets and does not introduce problems since eventually everything will reset. I realized it when I drew it out.

I will update the code and doc. Thank you for the discussion!

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -433,6 +447,135 @@ holdReset clk en SNat rst =
...
+-- Synchronous resets
+--
+-- If used with a synchronous reset the assertion of the reset is delayed by 1
+-- cycle. If your domain can support initial values the first clock cycle
+-- after boot the reset will be asserted. Afterwards the reset keeps being
+-- asserted for (n + 1) cycles.
+--
+-- Examples on a domain with synchronous reset and supporting initial values:
+--
+-- >>> let sampleWithReset = sampleN 8 . unsafeToActiveHigh
```
JvWesterveld: Nit: wouldn't `sampleReset` be more apt than `sampleWithReset` since you're sampling a reset itself?

DigitalBrains1: Especially since `sampleWithReset` is a function in our Prelude you're re-defining! :-)

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -433,6 +447,135 @@ holdReset clk en SNat rst =
...
+--
+-- The first 3 cycles the input reset is asserted and `stretchReset` should
+-- stretch it by 2 cycles. That results in a reset that is asserted for 6
+-- cycles. This additional cycle is due to the first cycle always being reset if
+-- initial values are supported.  'stretchReset' also works on intermediate
+-- assertions of the reset signal:
+--
+-- >>> let rst = fromList [False, False, False, True, False, False, False, False]
+-- >>> sampleWithReset (stretchResetBy2 (unsafeFromActiveHigh rst))
+-- [True,True,True,False,False,True,True,False]
```
leonschoorl: I feel this example is confusing.

The function is called `stretchResetBy2`, so I would expect it to "stretch the incoming reset pulse by 2".
Here it looks like it's given a one cycle reset pulse, and it stretched it to a 2 cycle long pulse.
Where as going by the name, one might expect a 3 cycle long pulse on the output.

leonschoorl: Or maybe the synchronous side of stretchReset is just wrong.
As having to provide a `(SNat @3)` to implement `stretchResetBy2` is also a bit weird.

rowanG077: Yes, this is fixed.

clash-prelude/src/Clash/Explicit/Reset.hs
rowanG077: I agree, this is fixed now. It was an artifact when copying over the documentation form `holdReset`. I have fixed it there as well.

clash-prelude/src/Clash/Explicit/Reset.hs
rowanG077: This is fixed. For a single clock cycle stretch you'd still need the counter though. So identity for 0 and counter for 1 or more.  For the async case as well as we don't need a full reset synchronizer as is discussed in the other thread.

clash-prelude/src/Clash/Explicit/Reset.hs
rowanG077: This is solved now. Pleas re-review.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -427,11 +472,66 @@ holdReset
...
+  -> Reset dom
+holdResetSync clk en sn@SNat rst = rstOut
  where
-  counter :: Signal dom (Index (n+1))
-  counter = register clk rst en 0 (satSucc SatBound <$> counter)
+  isActiveHigh = case resetPolarity @dom of { SActiveHigh -> True; _ -> False }
+
+  rstOut = case snatToInteger sn of
+    0 -> rst
+    1 -> unsafeToReset (register clk rst en isActiveHigh (unsafeFromReset rst))
```
leonschoorl: You could do the same here as in the Async version:
```suggestion
    1 -> unsafeToReset (register clk rst en isActiveHigh (pure (not isActiveHigh)))
```
And have slightly less demand on the rst signal.

leonschoorl:
This now again has the weird quirk that for synchronous resets is it holds it for 1 cycle less then one might expect.

```
>>> let sampleReset = sampleN 8 . unsafeToActiveHigh
>>> let rst = fromList [False, False, False, False, True, False, False, False]
>>> sampleReset $ holdReset @System clockGen enableGen d2 $ unsafeFromActiveHigh rst
[True,True,False,False,True,True,True,False]
>>> sampleReset $ holdReset @XilinxSystem clockGen enableGen d2 $ unsafeFromActiveHigh rst
[True,True,False,False,False,True,True,False]
```

Although that is something that `holdReset` always did.
And hasn't changed in this PR.

So maybe this should be a different issue/PR, but the synchronous behavior is something we should document (with a doctest example).
And possibly change, I'm not sure.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -427,11 +472,63 @@ holdReset
...
+  -> Reset dom
+  -- ^ Reset to extend
+  -> Reset dom
+holdResetSync clk en sn@SNat rst = rstOut
+ where
+  rstOut = case snatToInteger sn of
+    0 -> rst
+    _ -> unsafeToReset rawRst
+      where
+        counter :: Signal dom (Index (n + 2))
```
JvWesterveld: This should probably stay as before to keep the same behaviour with synchronous resets if the plan is to backport.
```suggestion
        counter :: Signal dom (Index (n + 1))
```

JvWesterveld (commented):
@rowanG077 Yes, approach LGTM. Thank you for picking it up.

clash-prelude/src/Clash/Explicit/Reset.hs
rowanG077: Will do

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -427,11 +472,63 @@ holdReset
...
+  -> SNat n
+  -- ^ Hold for /n/ cycles
+  -> Reset dom
+  -- ^ Reset to extend
+  -> Reset dom
+holdResetSync clk en sn@SNat rst = rstOut
+ where
+  rstOut = case snatToInteger sn of
+    0 -> rst
+    _ -> unsafeToReset rawRst
```
christiaanb: Don’t we want to use `orReset rst rawRst` here so that assertion is not delayed by one cycle?

DigitalBrains1: I think [my message above](https://github.com/clash-lang/clash-compiler/pull/3116#issuecomment-3762976756) is relevant

DigitalBrains1: But the fact the reset is delayed isn't documented further at all, so you could say this delay has been a bug all along and fixing it is just a bugfix, which can go into 1.8. It's a bit worrying we would change the timing of people's reset manager, although the reset _asserting_ a cycle sooner is much more benign than deasserting the reset sooner or later.

I think this is actually a fair assessment; it wasn't documented at all, let's just fix it.

> the weird quirk that for synchronous resets is it holds it for 1 cycle less then one might expect.

I initially misinterpreted this quirk, this is exactly that: if we `orReset` it will hold the reset for the expected amount of time. The quirk was really that it asserted too late, not that it deasserted too soon.

rowanG077: This is implemented now!

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -427,11 +472,67 @@ holdReset
...
+    (1, _) -> unsafeToReset (register clk rst en isActiveHigh (pure (not isActiveHigh)))
+    (_, SNatLE) -> unsafeToReset (register clk rst en isActiveHigh rawRst)
+      where
+        counter :: Signal dom (Index n)
+        counter = register clk rst en 0 (satSucc SatBound <$> counter)
+        rawRst :: Signal dom Bool
+        rawRst = case resetPolarity @dom of
+          SActiveHigh -> (/=maxBound) <$> counter
+          SActiveLow -> (==maxBound) <$> counter
+    (_, SNatGT) -> clashCompileError "holdResetAsync: Impossbile! n =! 0 && n != 1 && 2 > n, Please report this at https://github.com/clash-lang/clash-compiler/issues."
```
DigitalBrains1: Perhaps you could do
```haskell
case toUNat sn of
  UZero ->
  USucc UZero ->
  USucc (USucc _) ->
```

For large _n_ this makes no sense but for 0, 1 and >=2 it neatly avoids the absurd case. Without resorting to the weird

```haskell
case (snatToInteger sn, compareSNat d2 sn) of
  (0, _) ->
  (_, SNatGT) ->
  (_, SNatLE) ->
```
which should also work but is quite twisty :-P

rowanG077: Indeed, that is a way nicer way to write that and it works.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -399,20 +400,62 @@ resetGlitchFilter# SNat reg dffSync rstIn0 =
...
 -- [True,True,True,True,True,False,False,False]
 --
 -- 'holdReset' holds the reset for an additional 2 clock cycles for a total
 -- of 5 clock cycles where the reset is asserted. 'holdReset' also works on
 -- intermediate assertions of the reset signal:
 --
 -- >>> let rst = fromList [True, False, False, False, True, False, False, False]
--- >>> sampleWithReset (holdReset @System clockGen enableGen (SNat @2) (unsafeFromActiveHigh rst))
+-- >>> sampleReset (holdReset @System clockGen enableGen (SNat @2) (unsafeFromActiveHigh rst))
 -- [True,True,True,False,True,True,True,False]
```
leonschoorl, line 459: Now that we're changing the behaviour for synchronous resets (XilinxSystem).
Maybe we should add a doctest with the XilinxSystem domain, both as a way of documenting the synchronous behaviour, which we didn't (explicitly) did before.
And as a was to catch accidental changes in the future.

rowanG077: There is a test for it in the test suite. I didn't add a doctest because the behavior is the same for either reset types.

rowanG077:
@leonschoorl We talked about this in the office and in chat. Since the old behavior is considered broken we think it should be backported.

clash-prelude/src/Clash/Explicit/Reset.hs
```diff
@@ -399,20 +400,62 @@ resetGlitchFilter# SNat reg dffSync rstIn0 =
...
+  where
+    intialRst :: Bool
+    intialRst =
+      case resetPolarity @dom of
+        SActiveHigh -> initialValue
+        SActiveLow -> not initialValue
+    outRst = delay clk en intialRst rst
+
+-- | Hold/stretch reset for a number of cycles relative to an incoming reset
+-- signal.
```
leonschoorl: I wonder if maybe this is a better explanation of what it's doing:
```suggestion
-- | Holds the reset asserted for an extra @n@ cycles after the incoming reset is deasserted
```

leonschoorl: Or if we think "stretch" is important terminology that somebody might search for, we could keep it with something like:
```
-- | Holds/stretches the reset by keeping it asserted for an extra @n@ cycles after the incoming reset is deasserted
```

PR #3117, opened 2026-01-15
Fix ambiguous data files on duplicate idir

### LINKED ISSUE #3141: Clash errors if it finds the same data file twice
```haskell
-- In `src/Memory.hs`
module Memory where

import Clash.Explicit.Prelude

topEntity ::
  Clock System ->
  Enable System ->
  Signal System (Unsigned 4) ->
  Signal System (Maybe (Unsigned 4, BitVector 8)) ->
  Signal System (BitVector 8)
topEntity clk ena = blockRamFile clk ena d16 "memory.bin"
```

Reproducer:

```
$ cabal run clash -- -isrc -isrc Memory --vhdl
<no location info>: error:
    Clash error call:
    Multiple data files for "memory.bin" found. The following candidates were found:
      ["src/memory.bin","src/memory.bin"]
    Please disambiguate data files.
      error, called at src/Clash/Netlist/BlackBox/Util.hs:343:20 in clash-lib-1.9.0-inplace:Clash.Netlist.BlackBox.Util
```

This of course sounds silly, but passing down the same directory twice can happen more easily in build systems / generic environments.

TristanCacqueray:
When library and simulator code share the same hs-source-dirs, clash fails to resolve data file with the following error:
```
  Multiple data files for "font.bin" found. The following candidates were found:
    ["./font.bin","./font.bin"]
```
This change fix that bug by removing duplicated source dirs.

Fixes #3141

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Util.hs
@@ -46,6 +46,7 @@ import           Control.Monad.State             (State, StateT (..), lift, gets
 import           Data.Bitraversable              (bitraverse)
 import           Data.Bool                       (bool)
 import           Data.Coerce                     (coerce)
+import           Data.Containers.ListUtils       (nubOrd)
 import           Data.Foldable                   (foldrM)
 import           Data.Hashable                   (Hashable (..))
 import qualified Data.HashMap.Strict             as HashMap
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Netlist/BlackBox/Util.hs
```diff
@@ -332,7 +332,7 @@ canonicalizeDataFilePath idirs toCanonicalize
       (canonicalizePath toCanonicalize)
       (error [I.i|Could not find data file #{show toCanonicalize}. Does it exist?|])
   | otherwise = unsafePerformIO $ do
-      let candidates = map (</> toCanonicalize) idirs
+      let candidates = map (</> toCanonicalize) (nub idirs)
```
martijnbastiaan:
```suggestion
      let candidates = map (</> toCanonicalize) (nubOrd idirs)
```

`nub` is `O(n^2)`

martijnbastiaan (commented):
Thanks for your PR! I'm not sure I understand:

> When library and simulator code share the same hs-source-dirs

You mean when you have a Cabal file with a `library` and `executable` (simulator) that source from the same source dir, you get this error?

clash-lib/src/Clash/Netlist/BlackBox/Util.hs
TristanCacqueray: Thanks!

PR #3123, opened 2026-01-20
Add evaluator rule for `bigNatEq#`

### LINKED ISSUE #3084: Issue when using `testEquality`/`decideNat` in some contexts involving large `Nat`s
I faced an issue using `testEquality` with large (256 bits) `Nat`s. I created a reproducer:
```haskell
import Clash.Prelude hiding (SNat, Mod)
import Clash.Annotations.TH (makeTopEntity)

import Data.Type.Equality
import GHC.TypeLits (SNat)

type P = 2 ^ 256 - 2 ^ 224 + 2 ^ 192 + 2 ^ 96 - 1 -- any small-ish number would work (e.g. 300)

topEntity :: "hey" ::: Unsigned (CLog 2 P) -> "ho" ::: Index P
topEntity = f

f :: forall m. (KnownNat m, 1 <= m) => Unsigned (CLog 2 m) -> Index m
f =
 case testEquality (natSing :: SNat m) (natSing :: SNat P) of
  Just Refl -> g -- g @P would work
  Nothing   -> bitCoerce

g :: forall m. (KnownNat m, 1 <= m) => Unsigned (CLog 2 m) -> Index m
g =
  case testEquality (natSing :: SNat m) (natSing :: SNat P) of
    Just Refl -> h
    Nothing   -> bitCoerce

h :: forall m . (KnownNat m, 1 <= m) => Unsigned (CLog 2 m) -> Index m
h i = 1 + bitCoerce i

makeTopEntity 'topEntity
```

This snippet fails to synthesize with ```Clash.Netlist.BlackBox(343): No blackbox found for: GHC.Num.BigNat.bigNatEq#. Did you forget to include directories containing primitives? You can use '-i/my/prim/dir' to achieve this.```

However, either having `type P = 300` (e.g.) or replacing the call to `g` by `g @P` makes it synthesize - I assume that aggressively monomorphizing the code at call sites makes somehow `testEquality` (or its Core equivalent) disappear in the generated Core. Otherwise, if the `Nat` is small enough, it's treated as an `Int` which Clash can handle.

DigitalBrains1:
Fixes #3084

## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
@@ -2,7 +2,7 @@
   Copyright   :  (C) 2013-2016, University of Twente,
                      2016-2017, Myrtle Software Ltd,
                      2017-2022, Google Inc.,
-                     2017-2024, QBayLogic B.V.
+                     2017-2026, QBayLogic B.V.
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
@@ -52,6 +52,7 @@ import           GHC.Int
 import           GHC.Integer
   (decodeDoubleInteger,encodeDoubleInteger,compareInteger,orInteger,andInteger,
    xorInteger,complementInteger,absInteger,signumInteger)
+import           GHC.Num.BigNat      (bigNatEq#)
 import           GHC.Num.Integer (Integer (..), integerEncodeFloat#)
 import           GHC.Num.Natural     (naturalSubUnsafe)
 import           GHC.Natural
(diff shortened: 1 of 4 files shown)


clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -1896,6 +1897,11 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
     | [i] <- naturalLiterals' args
     -> reduce (Literal (WordLiteral (signum i)))
 
+  $(namePat 'GHC.Num.BigNat.bigNatEq#)
+    | [Lit (ByteArrayLiteral i), Lit (ByteArrayLiteral j)] <- args
+    -> let boolPrim = if i == j then 1 else 0
```
christiaanb: I would have implemented this using `bigNatEq#`, but this also looks fine.

christiaanb (approved):

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
DigitalBrains1: You mean like this?

```haskell
  $(namePat 'GHC.Num.BigNat.bigNatEq#)  
    | [ Lit (ByteArrayLiteral (BA.ByteArray i))
      , Lit (ByteArrayLiteral (BA.ByteArray j))] <- args
    -> reduce (Literal (IntLiteral (IS (bigNatEq# i j))))
```

I can change it to do that. It appears to work, and the effort has already been invested (you made me curious for the solution).

It also definitely looks more regular, instead of re-inventing `bigNatEq#` on lifted arguments.

christiaanb: Yeah, like that. Now you can never get a mismatch in Haskell runtime behavior and Clash symbolic compile time behavior.

DigitalBrains1: Done! Thanks for the education!

PR #3127, opened 2026-01-24
GitLab CI: Fix cache fallback

DigitalBrains1:
Since GitLab 18.6.1, 18.5.3, 18.4.5, CI runs triggered by a user with at least the Maintainer role are protected:
https://docs.gitlab.com/ci/caching/#cache-key-names

This probably applies to all our CI. Our nightlies get the -protected suffix at least, so our cache fallback key was no longer correct.

Testing indicates `$CI_JOB_IMAGE` is now properly expanded by GitLab, so we can simplify the fallback key as well.

## Still TODO:

### DIFF
--- a/.ci/gitlab/common.yml
+++ b/.ci/gitlab/common.yml
@@ -14,9 +14,13 @@ default:
   variables:
     CLASH_DOCKER_TAG: 20251007
     CACHE_BUST_TOKEN: 3
-    # Note that we copy+paste the image name into CACHE_FALLBACK_KEY. If we don't,
-    # $GHC_VERSION gets inserted at verbatim, instead of resolving to some ghc version.
-    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-ghcr.io/clash-lang/clash-ci:$GHC_VERSION-$CLASH_DOCKER_TAG-$CACHE_BUST_TOKEN-3-non_protected
+    # Note that the fallback key has two cache bust tokens. `$CACHE_BUST_TOKEN`
+    # is our own, defined right above, and the `-3` at the end in the fallback
+    # key below is GitLab's cache bust token that gets incremented when the user
+    # clicks "Clear runner caches" in the GitLab web interface. We need to
+    # specify both in the fallback key specifically, but *only* our own in the
+    # `cache.key` keyword of a job.
+    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-$CI_JOB_IMAGE-$CACHE_BUST_TOKEN-3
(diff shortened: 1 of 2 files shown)


DigitalBrains1:
I have verified it works correctly.

There is an unrelated issue with caching though: only the machines `diepenheim` and `lonneker` are able to access our cache, the machines `glanerbrug` and `gronau` get this error:

> WARNING: received: 403 Forbidden. Request failed with code: SignatureDoesNotMatch, message: The request signature we calculated does not match the signature you provided. Check your key and signing method.

(Only the four mentioned machines were involved in the test run of this PR I did today, that's why I can only say something about those). I currently don't have a user account on `glanerbrug`, but I do on `gronau`, so that machine was a lot easier to check. The thing is: the keys in `/etc/gitlab/config.toml` are 100% certain the same as on `diepenheim`! I don't know why it gets an authentication failure.

We also should investigate this error:

> WARNING: Container based cache volumes creation is disabled. Will not create volume for "/cache"

.ci/gitlab/common.yml
```diff
@@ -14,9 +14,7 @@ default:
   variables:
     CLASH_DOCKER_TAG: 20251007
     CACHE_BUST_TOKEN: 3
-    # Note that we copy+paste the image name into CACHE_FALLBACK_KEY. If we don't,
-    # $GHC_VERSION gets inserted at verbatim, instead of resolving to some ghc version.
-    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-ghcr.io/clash-lang/clash-ci:$GHC_VERSION-$CLASH_DOCKER_TAG-$CACHE_BUST_TOKEN-3-non_protected
+    CACHE_FALLBACK_KEY: $CI_JOB_NAME-master-$CI_JOB_IMAGE-$CACHE_BUST_TOKEN-3
```
leonschoorl, line 23: Every time I see this, I think that extra `-3` seems wrong.
Because `CACHE_BUST_TOKEN=3` and the cache key is just `-$CACHE_BUST_TOKEN`.

But it is correct, that extra `-3` gets added automatically by gitlab to the cache key, but not to the fallback cache key.
There is a button in gitlab somewhere that increments this number, see [the note here](https://docs.gitlab.com/ci/caching/#clear-the-cache-manually) for more info
Which we shouldn't push, because then we'd have to manually update this `-3` in the CACHE_FALLBACK_KEY here.

Maybe we should add a small comment here that explains this whole situation.

DigitalBrains1: I will add a comment since it is indeed very confusing.

PR #3131, opened 2026-01-30
Various small fixes - part 14

leonschoorl:
No user facing changes expect the [Polish docs](https://github.com/clash-lang/clash-compiler/commit/cdd96ef058b715105a5c376826c1f3e165a48053)

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Signal/Trace.hs
+++ b/clash-prelude/src/Clash/Signal/Trace.hs
@@ -515,7 +515,7 @@ dumpReplayable n oSignal traceName = do
 -- | Take a serialized signal (dumped with @dumpReplayable@) and convert it
 -- back into a signal. Will error if dumped type does not match requested
 -- type. The first value in the signal that fails to decode will stop the
--- decoding process and yield an error. Not that this always happens if you
+-- decoding process and yield an error. Note that this always happens if you
 -- evaluate more values than were originally dumped.
 replay
   :: forall a dom n
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -314,7 +314,7 @@ Added:
 Added:
   * `Clash.Class.Counter`: a class that defines a odometer-style supercounter. [#1763](https://github.com/clash-lang/clash-compiler/pull/1763)
   * `isLike` function for BitPack types. [#1774](https://github.com/clash-lang/clash-compiler/pull/1774)
-  * 'seqErrorX' for catching both `XException` and `ErrorCall`. [#1774](https://github.com/clash-lang/clash-compiler/pull/1774)
+  * 'seqErrorX' for catching both `XException` and `ErrorCall`. [#1839](https://github.com/clash-lang/clash-compiler/pull/1839)
   * `Clash.Explicit.BlockRam.File.memFile`, a function for creating the contents of the data files this blockRAM uses. Can also be imported from `Clash.Prelude.BlockRam.File`, `Clash.Prelude.ROM.File` and `Clash.Explicit.ROM.File`. [#1840](https://github.com/clash-lang/clash-compiler/pull/1840)
   * Support for Yosys compatible SVA to `Clash.Verification`. This enables formal verification using SymbiYosis for Verilog and SystemVerilog. [#1798](https://github.com/clash-lang/clash-compiler/pull/1798)
   * `Clash.Explicit.Signal.Delayed.forward`, a function that can be used to retime a `DSignal` into the future without applying any logic. [#1882](https://github.com/clash-lang/clash-compiler/pull/1882)
--- a/clash-lib/src/Clash/Core/EqSolver.hs
+++ b/clash-lib/src/Clash/Core/EqSolver.hs
@@ -88,7 +88,7 @@ solveEq tcm solveSet (coreView tcm -> left, coreView tcm -> right) =
             []
 
 -- | Solve equations supported by @normalizeAdd@. See documentation of
--- @TypeEqSolution@ to understand the return value.
+-- 'TypeEqSolution' to understand the return value.
 solveAdd
   :: VarSet
   -> (Type, Type)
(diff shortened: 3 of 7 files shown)


tests/src/Test/Tasty/Clash/CoreTest.hs
```diff
@@ -68,9 +67,9 @@ runToCoreStage _target f src = do
 
 findBinding
   :: OccName
-  -> (BindingMap, TyConMap, Supply)
+  -> (ClashEnv, ClashDesign, Supply)
```
DigitalBrains1: Should we even bother, isn't this unnecessary code churn? The tests were [taken out](https://github.com/clash-lang/clash-compiler/commit/e7f7c522499cb7db94ce51ad21fb0a79737d1492) four and a half years ago and that particular partial evaluator effort abandoned. To me, I'd say either just remove it or leave it in as-is.

clash-prelude/src/Clash/Signal/Trace.hs
```diff
@@ -74,7 +74,6 @@ module Clash.Signal.Trace
   -- * Internal
   -- ** Types
   , Period
-  , Changed
```
DigitalBrains1, line 77: Why would we restrict what we export in the _Internal_ section? I understand this particular type synonym doesn't occur in any of the other definitions, but why does that matter? If we apply that reasoning to exported definitions recursively, we end up without an _Internal_ section. Case in point: maybe someone wants to copy a definition from here and tweak it; the copied code might still refer to the `Changed` type. But more broadly, I'd just say: either export the stable API or export everything. And we generally want to expose internals in some way. I don't see what the purpose is of tweaking what you do and do not export as internal definitions.

Note that for 1.10, we want to overhaul this module.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
```diff
@@ -30,7 +30,7 @@ module Clash.Explicit.Synchronizer
 where
 
 import Data.Bits                   (complement, shiftR, xor)
-import Data.Constraint             ((:-)(..), Dict (..))
+import Data.Constraint             ((:-)(Sub), Dict (Dict))
```
DigitalBrains1: While I understand that using `T(..)` basically makes us sensitive to additions to the API, it seems we regularly use it instead of naming only the constructors we want to use. I'm not sure why this is, but what makes `Data.Constraint` different to give it special treatment? Why are the other instances of `T(..)` left alone? I'd think a commit/PR that removes all uses of `T(..)` for imports from external dependencies would be fine, but picking just two specific ones out of the list of occurrences of this pattern is too much of a halfway solution.

(You also forgot an instance of `Data.Constraint` in `Clash.Promoted.Nat`)

tests/src/Test/Tasty/Clash/CoreTest.hs
DigitalBrains1: Hmmm I'm not so sure anymore. We do use the partial evaluator code. I retract my objection.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
DigitalBrains1: I'm fine with making a PR that addresses all cases of `T(..)`, I didn't mean to put extra work on you :-). You're more than welcome to, but I can do it as well.

clash-prelude/src/Clash/Signal/Trace.hs
leonschoorl: > I understand this particular type synonym doesn't occur in any of the other definitions
That was basically my reasoning why it was pointless to export this.

It was my understanding at the time that all the exported "internal" functions were primitives, and they needed to be exported because they were primitives.

And assumed that the "internal" type were exported because they were used in the type  signatures of the internal functions.

`Changed` seem to be the odd one out.
It wasn't used in any of the types signatures of the exported functions.
That led me to conclude that there was no good reason for it be exported, and was exported by mistake.
And this commit was an attempt to cleanup/correct that mistake.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
leonschoorl: I think was looking at some code that used that `Sub` constructor and got confused where it came from.
It's not a very meaningful name to me and not easily connected to the `:-` type.
For example trying to find it with [hoogle](https://hoogle.haskell.org/?hoogle=Sub&scope=set%3Astackage), does show it, but only half way down the page, buried on a sea of other things with the same name.

I don't think we need to do this everywhere, I'm fine with for example `SNat (..)`, because:
 1. the data constructor is easily linked to the type, because they have the same name
 2. I'm very familiar with SNat, and I suspect anyone else working on the clash-lib source would also be familiar with SNat

But I wasn't very familiar with `:-` and `Sub`, this caused me some confusion when trying to read some code used it.
And I thought I might spare somebody else that confusion, by making the import of `Sub` explicit.

As for the `Dict(..)` vs `Dict(Dict)`, I don't have strong preference. It fits `1.`, but not `2.`

leonschoorl: I've update the PR, reverted back to `Dict(..)`, and changed to `(:-)(Sub)` everywhere

clash-prelude/src/Clash/Signal/Trace.hs
leonschoorl: Looking further it turns out I was mistaken and these "internal" functions are just internal functions and not primitives.

So I don't see a good reason for exporting them.

> And we generally want to expose internals in some way.
Do we?

We also aren't exporting `Clash.Signal.Trace.iso8601Format`.
And I feel we shouldn't.

clash-prelude/src/Clash/Explicit/Synchronizer.hs
DigitalBrains1: Ah, so I completely misguessed your reasons. Yeah, I'm fine with the clarification.

clash-prelude/src/Clash/Signal/Trace.hs
leonschoorl: I've taken out this commit

Issue #3132, opened 2026-02-02
Source annotations are output even without `-g`


pieter-bos:
To my best knowledge Clash should output source annotations if and only if the relevant module was compiled with [debug information](https://downloads.haskell.org/ghc/latest/docs/users_guide/debug-info.html) (`-g`). In the case that a file is compiled from source using Clash there is no choice: the source annotations seem to always appear. It would be nice if Clash respects the absence of `-g`.

For us this is not entirely a cosmetic complaint: in the case that we are compiling sources from the nix store this can lead to very coarse-grained cache invalidation, due to the source annotation listing the full absolute path. The path includes a hash of all input involved in making the HDL.

```haskell
module Repro where

import Clash.Prelude

topEntity :: Maybe Int -> Int
topEntity x = case x of Nothing -> 0; Just x -> x
```

```
$ clash Repro.hs --verilog
$ cat verilog/Repro.topEntity/topEntity.v 
/* AUTOMATICALLY GENERATED VERILOG-2001 SOURCE CODE.
** GENERATED BY CLASH 1.9.0. DO NOT MODIFY.
*/
`default_nettype none
`timescale 100fs/100fs
module topEntity
    ( // Inputs
      input wire [64:0] x


      // Outputs
    , output wire signed [63:0] result
    );
  // Repro.hs:6:1-9
  wire signed [63:0] x1;

  assign x1 = $signed(x[63:0]);

  assign result = x[64:64] ? x1 : (64'sd0);


endmodule
`default_nettype wire
```

DigitalBrains1:
We've released [v1.8.5](https://github.com/clash-lang/clash-compiler/releases/tag/v1.8.5), which includes a fix for this issue.

PR #3133, opened 2026-02-02
Suppress source location notes if `-g` is not passed to clash

rowanG077:
Fixes [#3132](https://github.com/clash-lang/clash-compiler/issues/3132).

First time working with this part of clash. I'm iffy about storing this in `GHC2CoreEnv` but it seems to fit the best.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver/Types.hs
+++ b/clash-lib/src/Clash/Driver/Types.hs
@@ -2,8 +2,8 @@
   Copyright  :  (C) 2013-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017     , QBayLogic, Google Inc.,
-                    2020-2022, QBayLogic,
                     2022     , Google Inc.,
+                    2020-2026, QBayLogic,
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -315,6 +315,11 @@ data ClashOpts = ClashOpts
   -- Command line flag: -fclash-evaluator-fuel-limit
   , opt_debug :: DebugOpts
   -- ^ Options which control debugging. See 'DebugOpts'.
+  , opt_ghcDebugLevel :: Int
+  -- ^ What the GHC debug level is (i.e. @-g<N>@).
+  --
+  -- Clash uses this to decide whether to emit source-location information in
+  -- generated HDL.
   , opt_cachehdl :: Bool
   -- ^ Reuse previously generated output from Clash. Only caches topentities.
   --
@@ -407,6 +412,7 @@ defClashOpts
   , opt_inlineConstantLimit = 0
   , opt_evaluatorFuelLimit  = 20
   , opt_debug               = debugNone
+  , opt_ghcDebugLevel       = 0
   , opt_cachehdl            = True
   , opt_clear               = False
   , opt_primWarn            = True
--- a/tests/shouldwork/Netlist/HDLNotContainsLoc.hs
+++ b/tests/shouldwork/Netlist/HDLNotContainsLoc.hs
@@ -1,0 +1,39 @@
+{-# LANGUAGE DataKinds #-}
+
+module HDLNotContainsLoc where
+
+import Clash.Prelude
+
+import qualified Prelude as P
+import Data.List (isInfixOf)
+import System.Environment (getArgs)
+import System.FilePath ((</>), takeDirectory)
+
+topEntity :: Maybe Int -> Int
+topEntity x = case x of Nothing -> 0; Just x -> x
+
+assertNotIn :: String -> String -> IO ()
+assertNotIn needle haystack
+  | not (needle `isInfixOf` haystack)
+  = return ()
+  | otherwise
+  = P.error $ P.concat [ "Did not expect:\n\n  ", needle
+                       , "\n\nIn:\n\n", haystack ]
+
+mainVHDL :: IO ()
+mainVHDL = do
+  [topDir] <- getArgs
+  content <- readFile (topDir </> show 'topEntity </> "topEntity.vhdl")
+  assertNotIn "HDLNotContainsLoc.hs:" content
+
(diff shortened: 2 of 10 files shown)


clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -3,7 +3,7 @@
                     2016-2017, Myrtle Software Ltd,
                     2017     , QBayLogic, Google Inc.,
                     2020-2022, QBayLogic,
-                    2022     , Google Inc.,
+                    2026     , Google Inc.,
```
DigitalBrains1: I think you meant to add a copyright year for QBayLogic, not postdate Google's copyright :-D

DigitalBrains1: You can just change the line above from `2020-2022` to `2020-2026`. We don't record which years had development, just the overal span. But maybe move the line down so it's the last line in the list.

tests/src/Test/Tasty/SourceLoc.hs
```diff
@@ -0,0 +1,69 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Test.Tasty.SourceLoc (sourceLocTests) where
```
martijnbastiaan: Could you comment on why this isn't an `outputTest`?

clash-lib/src/Clash/Driver/Types.hs
DigitalBrains1: Ah, sorry, I should have mentioned you did this twice! Also in `clash-lib/src/Clash/Netlist.hs`. I shouldn't have assumed you'd notice.

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -315,6 +315,11 @@ data ClashOpts = ClashOpts
   -- Command line flag: -fclash-evaluator-fuel-limit
   , opt_debug :: DebugOpts
   -- ^ Options which control debugging. See 'DebugOpts'.
+  , opt_ghcDebugInfo :: Bool
+  -- ^ Whether GHC debug info is enabled (e.g. @-g@).
```
DigitalBrains1: It's a minor detail but...

Did you really mean e.g. (for example) or did you mean i.e. (that is)? In other words, are there indeed other things than the command line option `-g` that affect the `GHC.Driver.DynFlags.debugLevel` flag?

I grepped the command line option documentation for the word "debug" and did not think there were more things, but I could be mistaken.

clash-lib/src/Clash/Driver/Types.hs
```diff
@@ -315,6 +315,11 @@ data ClashOpts = ClashOpts
   -- Command line flag: -fclash-evaluator-fuel-limit
   , opt_debug :: DebugOpts
   -- ^ Options which control debugging. See 'DebugOpts'.
+  , opt_ghcDebugInfo :: Bool
```
christiaanb: Why not use the same type as GHC uses to track verbosity? Instead of the binary choice of "not verbose"/"verbose" and move the decision whether to generate/copy ticks or not based on the set verbosity level.

clash-lib/src/Clash/Driver/Types.hs
rowanG077: Fixed.

tests/src/Test/Tasty/SourceLoc.hs
rowanG077: Yes, I have added a comment.

clash-lib/src/Clash/Driver/Types.hs
rowanG077: I checked the GHC sources. Only `-g` sets it. So I have amended this comment.

clash-lib/src/Clash/Driver/Types.hs
rowanG077: Sure possible, since we don't do anything else with this flag I found it cleaner to make it a bool. But I'm not hung up on it so I have implemented your proposal.

tests/src/Test/Tasty/SourceLoc.hs
martijnbastiaan: I'm confused, I think your comments claims that an `outputTest` does simulation, but it doesn't. Unless I'm mistaken it does:

```
clash --hdl
run mainHdl
```

Which is AFAIK what you're doing here.

rowanG077: That's what I get for looking to quickly early in the morning. I assumed running the build executable would be a simulation. But it isn't. Yes I can replace this with an `outputTest`.

rowanG077: I have made them `outputTest` now. This requires having almost duplicate sources.

tests/shouldwork/Netlist/HDLNotContainsLoc.hs
```diff
@@ -0,0 +1,39 @@
...
+
+topEntity :: Maybe Int -> Int
+topEntity x = case x of Nothing -> 0; Just x -> x
+
+assertNotIn :: String -> String -> IO ()
+assertNotIn needle haystack
+  | not (needle `isInfixOf` haystack)
+  = return ()
+  | otherwise
+  = P.error $ P.concat [ "Expected:\n\n  ", needle
```
DigitalBrains1:
```suggestion
  = P.error $ P.concat [ "Did not expect:\n\n  ", needle
```

tests/src/Test/Tasty/SourceLoc.hs
martijnbastiaan: They can import each other, if that helps.

rowanG077: The source is so small that I don't think it's a win having 3 files vs 2.

kleinreact (approved):

DigitalBrains1 (approved):
All the things I commented about are now fine, but note that I have no experience with this piece of our code.

PR #3136, opened 2026-02-07
Fix overlapping instances in `NumConvert` with `Canonical`

### LINKED ISSUE #2986: Overlapping instances make some `NumConvert` usages impossible
```haskell
import Clash.Prelude
import Data.Word

f :: Word32 -> Word64
f = numConvert
```

```
martijn@deskei:~/code/clash-compiler$ cabal run clash -- Test.hs
Configuration is affected by the following files:
- cabal.project
Loaded package environment from /home/martijn/code/clash-compiler/.ghc.environment.x86_64-linux-9.10.1
[1 of 2] Compiling Main             ( Test.hs, Test.o )
Test.hs:6:5: error: [GHC-43085]
    • Overlapping instances for NumConvert Word32 Word64
        arising from a use of ‘numConvert’
      Matching instances:
        instance NumConvert a (Unsigned 64) => NumConvert a Word64
          -- Defined in ‘Clash.Class.NumConvert.Internal.NumConvert’
        instance NumConvert (Unsigned 32) a => NumConvert Word32 a
          -- Defined in ‘Clash.Class.NumConvert.Internal.NumConvert’
    • In the expression: numConvert
      In an equation for ‘f’: f = numConvert
  |
6 | f = numConvert
  |     ^^^^^^^^^^
```

martijnbastiaan:
Previously, conversions such as:

    f :: Word32 -> Word64
    f = numConvert

would fail with an "overlapping instances" error, because it could match on both `NumConvert a Word64` and `NumConvert Word32 a`. This can't be solved through `OVERLAPPING` pragmas (and friends), because neither is more specific than the other.

To fix this we make the intuition that some types are "equal" to others explicit using a `Through` type class. This can then be used in `numConvertThrough` and `maybeNumConvertThrough`.

Fixes #2986

## Dear reviewer (1)
After this patch we ask our users to differentiate between `numConvert` and `numConvertThrough`. This is a bit meh, though the API is clear and understandable . An alternative would be to make `numConvertThrough` _the_ function to use. I.e., do something like:

```haskell
type NumConvert a b = 
  ( NumConvertDirect a (Through a)
  , NumConvertDirect (Through a) (Through b)
  , NumConvertDirect (Through b) b
  )

numConvert :: NumConvert a b => a -> b
numConvert = 
    numConvertDirect @(Through b) @b
  . numConvertDirect @(Through a) @(Through b)
  . numConvertDirect @a @(Through a)
```

The user would be none the wiser, until error messages show up. What do you think?

## Dear reviewer (2)
I didn't go through the effort of building hedgehog/exhaustive tests. IMO there is very little that can go wrong given the abstract definitions...

## AI disclaimer
I used AI to generate a bunch of fixed test cases. I used AI to shuffle around instances.

## Still TODO:

  - [X] ~~Write a changelog entry (see changelog/README.md)~~ I don't think `NumConvert` has been part of a release yet.

### DIFF
--- a/clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
+++ b/clash-prelude/src/Clash/Class/NumConvert/Internal/MaybeNumConvert.hs
@@ -48,7 +48,7 @@ failure is expressed by returning 'Nothing', never by an 'Clash.XException.XExce
 
 == __Laws__
 A conversion is either successful or it fails gracefully. I.e., it does not
-produces produce errors (also see "Clash.XException"). I.e.,
+produce errors (also see "Clash.XException"). I.e.,
 
 > x == fromMaybe x (maybeNumConvert @a @b x >>= maybeNumConvert @b @a)
 
@@ -56,7 +56,7 @@ for all values @x@ of type @a@. It should also preserve the numerical value
 interpretation of the bits. For types that have an @Integral@ instance, this
 intuition is captured by:
 
-> toInteger x == fromMaybe (toInteger x) (toInteger (numConvert @a @b x))
+> toInteger x == fromMaybe (toInteger x) (toInteger <$> maybeNumConvert @a @b x)
 
 If a conversion succeeds one way, it should also succeed the other way. I.e.,
 


clash-prelude/src/Clash/Class/NumConvert/Internal/Through.hs
```diff
@@ -0,0 +1,48 @@
...
+import Clash.Sized.Signed
+import Clash.Sized.Unsigned
+
+import Data.Int (Int16, Int32, Int64, Int8)
+import Data.Word (Word16, Word32, Word64, Word8)
+
+-- | Type family mapping types to their canonical \"unwrapped\" Clash form.
+-- This is used by 'Clash.Class.NumConvert.numConvertThrough' and
+-- 'Clash.Class.NumConvert.maybeNumConvertThrough' to determine the intermediate type.
+type family Through a
```
rowanG077: I'd like to word `Canonical` much better. As `Through` doesn't fit what it actually means imo.

martijnbastiaan: Would you still do `numConvertThrough`? Or would it be `numConvertCanonical`? (Not asking to be funny, I also like `Canonical` better for the type family. Just wondering what it would entail.)

rowanG077: Yeah `numConvertThrough` should be fine, more descriptive would be `numConvertThroughCanonical` but that length gets a bit ridiculous. Maybe `numConvViaCanon`? Also not ideal. So I think `numConvertThrough` is fine. The doc makes clear that the `Through` part is through the canonical representation.

kleinreact: To also throw in my naming considerations: I would use `numConvertVia` at this point, simply because
 * it aligns with already known Haskell naming such as `DerivingVia`, and
 * it is easier to speak out loud and type :slightly_smiling_face:.

I also don't see any particular requirements on having the type family and the conversion function to use similar names.

martijnbastiaan: Via works for me too

martijnbastiaan:
```
ghci> numConvertVia (0 :: Word64) :: Word32
<interactive>:40:1: error: [GHC-64725]
    • Cannot satisfy: 64 <= 32
    • In the expression: numConvertVia (0 :: Word64) :: Word32
      In an equation for ‘it’: it = numConvertVia (0 :: Word64) :: Word32
```
```
ghci> numConvertVia (0 :: Int32) :: Word32
<interactive>:41:1: error: [GHC-39999]
    • No instance for ‘Clash.Class.NumConvert.Internal.NumConvert.NumConvert
                         (Signed 32) (Unsigned 32)’
        arising from a use of ‘numConvertVia’
    • In the expression: numConvertVia (0 :: Int32) :: Word32
      In an equation for ‘it’: it = numConvertVia (0 :: Int32) :: Word32
```
```
ghci> numConvertVia (0 :: Unsigned 64) :: Unsigned 32
<interactive>:43:1: error: [GHC-64725]
    • Cannot satisfy: 64 <= 32
    • In the expression:
          numConvertVia (0 :: Unsigned 64) :: Unsigned 32
      In an equation for ‘it’:
          it = numConvertVia (0 :: Unsigned 64) :: Unsigned 32
```

I don't think these are bad at all. For `Int32` -> `Word32` you don't get the original types in the error message, but IMO the correspondence should be immediately obvious. Let's go this way?

clash-prelude/src/Clash/Class/NumConvert/Internal/Through.hs
martijnbastiaan: I've amended the names to `Canonical` and `*Via` (but also see my new main-thread comment).

rowanG077:
In that case I would say make `numConvertVia` THE way to do it as the error message are good and it will work more like people expect. 

Edit: You just posted a hair faster than me. Yes this looks great!  It has my stamp of approval

martijnbastiaan:
I don't get it. How are these concepts related?

martijnbastiaan:
@rowanG077 Something to think about w.r.t. `Erroring`/`Saturating`: it might just be that users expect `numConvert` to respect these "settings". I.e., maybe they expect `numConvert` to not impose any constraints at all for these instances, but to act error/saturate on conversion failure.

For now I've done the dumbest thing and simply strip/add the wrappers. We could relax this in the future. Alternatively, we might not include it at all for now.

DigitalBrains1 (approved):
Nice, approved! But with the following comments:

1. I think this is a good moment to fix two mistakes in the Haddock which I wrote to you about on Slack.
2. I'm not convinced adding the `Erroring`, `Saturating` and stuff here. I agree that it is rather open to interpretation what the behaviour should be, and people might make incorrect assumptions.

Issue #3147, opened 2026-02-10
Verilog blackbox for `iterateI` uses `wire`  for sequential assignments


lmbollen:
The blackbox for verilog uses `wire` instead of `reg` which causes Vivado to throw the following tantrum:
```
ERROR: [Synth 8-2577] procedural assignment to a non-register iterateI_ho1_0_res_21 is not permitted, left-hand side should be reg/integer/time/genvar
```
on the folllowing piece of verilog code:
```verilog
  wire [63:0] iterateI_ho1_0_res_21;
  wire [63:0] iterateI_ho1_1_res_1;
  wire [63:0] iterateI_ho1_2_res_1;
  // delay begin
  always @(posedge bFwd_66[1240:1240]) begin : c$app_arg_361_delay_2
    iterateI_ho1_0_res_21 <= cnt;
  end
  // delay end

  // delay begin
  always @(posedge bFwd_66[1240:1240]) begin : c$app_arg_361_delay_0_0
    iterateI_ho1_1_res_1 <= iterateI_ho1_0_res_21;
  end
  // delay end

  // delay begin
  always @(posedge bFwd_66[1240:1240]) begin : c$app_arg_361_delay_1_1
    iterateI_ho1_2_res_1 <= iterateI_ho1_1_res_1;
  end
  // delay end
```

Reproducer:
```hs
module Example where

import Clash.Prelude

delayedCounter ::
  forall dom d n.
  (HiddenClockResetEnable dom, KnownNat d, KnownNat n) =>
  SNat d ->
  Signal dom (Unsigned n)
delayedCounter d = last $ iterate (succSNat d) dflipflop cnt
 where
  cnt = register 0 (cnt + 1)

topEntity :: Clock System -> Reset System -> Enable System -> Signal System (Unsigned 8)
topEntity = exposeClockResetEnable $ delayedCounter d3
```

I tried to quickly reproduce failure with yosys but "unfortunately" yosys did not care and just synthesized.

leonschoorl:
Just to avoid confusion: the [reproducer doesn't reproduce](https://github.com/clash-lang/clash-compiler/pull/3148#issuecomment-3884632117), but needs explicit clocks

DigitalBrains1:
We've released [v1.10.0](https://github.com/clash-lang/clash-compiler/releases/tag/v1.10.0), which includes a fix for this issue.

PR #3148, opened 2026-02-10
Fix higher order blackboxes not propagating usage metadata

rowanG077:

### DIFF
--- a/changelog/2026-02-10T18_38_12+01_00_fix_iteratei_verilog_reg
+++ b/changelog/2026-02-10T18_38_12+01_00_fix_iteratei_verilog_reg
@@ -1,0 +1,1 @@
+FIXED: Higher order blackboxes not propagating usage metadata for generated results [#3147](https://github.com/clash-lang/clash-compiler/issues/3147)
--- a/clash-lib/src/Clash/Primitives/DSL.hs
+++ b/clash-lib/src/Clash/Primitives/DSL.hs
@@ -1,6 +1,6 @@
 {-|
   Copyright   :  (C) 2019,      Myrtle Software Ltd.
-                     2020-2024, QBayLogic B.V.
+                     2020-2026, QBayLogic B.V.
                      2021,      Myrtle.ai
                      2022-2023, Google Inc
   License     :  BSD2 (see the file LICENSE)
@@ -95,6 +95,7 @@ module Clash.Primitives.DSL
   ) where
 
 import           Control.Lens                    hiding (Indexed, assign)
+import           Control.Exception               (throw)
 import           Control.Monad                   (forM, forM_, zipWithM)
 import           Control.Monad.State
 import           Data.Default                    (Default(def))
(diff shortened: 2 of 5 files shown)


changelog/2026-02-10T18_38_12+01_00_fix_iteratei_verilog_reg
```diff
@@ -0,0 +1 @@
+FIXED: HO blackboxes not propagating usage metadata for generated results [#3147](https://github.com/clash-lang/clash-compiler/issues/3141)
```
DigitalBrains1: I suspect the link to #3147 is a mistake, could you please check?

Also, maybe write out higher order in full for clarity.

DigitalBrains1: I suspect the link to #3147 is a mistake, could you please check?

Also, maybe write out higher order in full for clarity.

DigitalBrains1: I tried to post this comment an hour ago but pressed the wrong button.

Now I do post it and see that the URL points to #3141 but the rendered text still says #3147.

I now also see what wrong button I pressed. Ctrl-Enter queues for review instead of posting.

DigitalBrains1: In 3147:

image

:eyes:

If I'm not misremembering, this is buggy output and this won't actually happen. Merely mentioning an issue doesn't close it but it says that it does.

rowanG077: Afaik that should work. The link is indeed a typo...

rowanG077: Fixed.

clash-lib/src/Clash/Primitives/DSL.hs
```diff
@@ -835,6 +835,10 @@ instHO bbCtx fPos (resTy, bbResTy) argsWithTypes = do
 
   resName <- declare' (ctxName <> "_" <> "ho" <> showt fPos <> "_"
                                <> showt fSubPos <> "_res") resTy
+  -- Pick the fSubPos-th HO instantiation (if present) and reuse its usage.
+  case IntMap.lookup fPos (bbFunctions bbCtx) >>= listToMaybe . drop fSubPos of
+    Just (_, usage, _, _, _, _) -> declareUseOnce usage resName
+    Nothing -> pure ()
```
christiaanb: Is this what we want? Should we be allowed to instantiate a function argument more often that its function plurality? Shouldn't we error out if that happens? @martijnbastiaan

martijnbastiaan: Yes, agreed, we should error out. Using a function more times than the function plurality will lead to broken HDL, as Clash uses the plurality to determine how often it should instantiate the function.

I'm not 100% sure on the other way around, I think Clash will just generate superfluous but not necessarily broken HDL.

rowanG077: Later on in `renderElem` this check is already performed:

https://github.com/clash-lang/clash-compiler/blob/fb1fedf46e5b266eeac638143960aa0fa53ea693/clash-lib/src/Clash/Netlist/BlackBox/Util.hs#L485-L489

Do you want to duplicate that check here? Or would a comment suffice that makes clear that this invariant is upheld later on?

martijnbastiaan: Factor the check out, call it twice, as just discussed in our call :)

rowanG077: done. Actually found the same check duplicated one more time.

PR #3163, opened 2026-03-15
Simplify `KnownDomain`

kleinreact:
The PR implements the suggestions of #3178, except for the clock periods, where it instead keeps the exiting integer period for backward compatibility, while just adding the period fraction that may result from conversion imprecision. This fraction then can be used to recover the precise frequency, if needed.

## Still TODO:

---

Approved for public release. This work has been supported by funding from the _Agentur für Innovation in der Cybersicherheit GmbH ([Cyberagentur](https://www.cyberagentur.de))_ as part of the [Clash Formal](https://clash-formal.org) project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ cabal.config
 *.bin
 *.log
 *.tar.gz
+*.agdai
 
 *~
 *.DS_Store
--- a/Clash.hs
+++ b/Clash.hs
@@ -58,7 +58,8 @@ doHDL Proxy opts src = do
   putStrLn $ "Loading dependencies took " ++ prepStartDiff
 
   generateHDL clashEnv clashDesign (Just backend)
-    (ghcTypeToHWType (opt_intWidth opts)) ghcEvaluator evaluator Nothing startTime
+    (ghcTypeToHWType (envDomainTCUs clashEnv) (opt_intWidth opts))
+    ghcEvaluator evaluator Nothing startTime
 
 main :: IO ()
 main =
--- a/benchmark/benchmark-concurrency.hs
+++ b/benchmark/benchmark-concurrency.hs
@@ -4,19 +4,19 @@ module Main (main) where
 
 import Criterion.Main
 import Data.List (isPrefixOf, partition)
-import Data.Time (getCurrentTime)
+--import Data.Time (getCurrentTime)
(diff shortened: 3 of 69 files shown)


christiaanb:
Looking at the failing tests, it seems the existing encoding is ingrained in the Clash compiler to figure out things about domain configurations during HDL generation:

1. https://github.com/clash-lang/clash-compiler/blob/a20806dd0dba0cf8d27b64d35149e048a76284c7/clash-ghc/src-ghc/Clash/GHC/LoadModules.hs#L564-L638
2. https://github.com/clash-lang/clash-compiler/blob/a20806dd0dba0cf8d27b64d35149e048a76284c7/clash-lib/src/Clash/Netlist/BlackBox/Util.hs#L674-L679
3. https://github.com/clash-lang/clash-compiler/blob/a20806dd0dba0cf8d27b64d35149e048a76284c7/clash-lib/src/Clash/Netlist/BlackBox/Util.hs#L724-L747

rowanG077:
I would like to do this. In the past I tried something like this as well but I never got it working without breaking type inference on existing code. If you want @kleinreact I can take up this PR and see if I can get it to pass all existing tests.

kleinreact:
Thanks for the offer @rowanG077. I am currently evaluating some more improvements that should be applicable here, where I'd prefer to ship them all in a single PR in the end, as this is a quite sensible part of the prelude. I hope to be able to finish my evaluation soon.

kleinreact:
> I am currently evaluating some more improvements that should be applicable here, where I'd prefer to ship them all in a single PR in the end, as this is a quite sensible part of the prelude. I hope to be able to finish my evaluation soon.

You can find the results of my evaluation for further discussion in https://github.com/clash-lang/clash-compiler/discussions/3178.

kleinreact:
The PR now includes a full implementation of the changes proposed in #3178, except for the clock periods, which I solved sightly differently than originally proposed.

A few tests are still failing, but I could neither find any good solution for fixing them yet nor do I understand the origin of their failure. Nevertheless, I'd like to focus and discuss the particular design decisions made, basically sharing this as a proposal with the goal to eventually reach some consensus of what we want. 

My proposal is to continue any technical discussions here in the PR and keep general design considerations for #3178.

PR #3169, opened 2026-03-26
Support new Tasty in `clash-testsuite`

DigitalBrains1:
Tasty 1.5.4, released to Hackage on 2026-03-25, deprecated `sequentialTestGroup`, with `dependentTestGroup` offering a drop-in replacement. Use that, so we don't error out because of our `-Werror`.

Bump the Stack LTS's to the latest minor.

Nix: Disable checks for criterion

## Still TODO:


martijnbastiaan (approved):
Makes sense, not sure why the nix builds are failing: the change should be drop-in.

DigitalBrains1:
I noted I copy-pasted a bit too much, all the comments in the Nix files say "This version of tasty isn't available in the nix ghc910 package set", for GHC 9.8 and 9.6 too. That's why I flipped it to Draft; I'll fix the comments once I know the rest is okay. But I'd like to look at these weird Nix failures too.

DigitalBrains1:
Thanks Rowan for helping with Nix.

PR #3172, opened 2026-03-31
Don't create netlist cases on Integer or Natural constructors

leonschoorl:
Fixes #3157

## Still TODO:

### DIFF
--- a/changelog/2026-04-16T14_23_19+02_00_dont_create_netlist_cases_on_Integer_or_Natural_constructors
+++ b/changelog/2026-04-16T14_23_19+02_00_dont_create_netlist_cases_on_Integer_or_Natural_constructors
@@ -1,0 +1,1 @@
+FIXED: Clash won't create (invalid) netlist cases on Integer and Natural constructors anymore [#3157](https://github.com/clash-lang/clash-compiler/issues/3157)
--- a/clash-lib/src/Clash/Netlist.hs
+++ b/clash-lib/src/Clash/Netlist.hs
@@ -66,6 +66,7 @@ import           Clash.Core.TermInfo              (multiPrimInfo', splitMultiPri
 import           Clash.Core.Type
   (Type (..), coreView1, splitFunForallTy, splitCoreFunForallTy)
 import           Clash.Core.TyCon                 (TyConMap)
+import           Clash.Core.TysPrim               (integerPrimTy, naturalPrimTy)
 import           Clash.Core.Util                  (splitShouldSplit)
 import           Clash.Core.Var                   (Id, Var (..), isGlobalId)
 import           Clash.Core.VarEnv
@@ -498,6 +499,7 @@ mkSelection declType bndr scrut altTy alts0 tickDecls = do
   (_,sp) <- Lens.use curCompNm
   ite <- Lens.use backEndITE
   altHTy <- unsafeCoreTypeToHWTypeM' $(curLoc) altTy
+  let e = Case scrut altTy (NE.toList alts0)
   case iteAlts scrutHTy alts0 of
     Just (altT,altF)
       | ite
@@ -525,6 +527,10 @@ mkSelection declType bndr scrut altTy alts0 tickDecls = do
          | otherwise
           -> do dstAssign <- contAssign dstId (IfThenElse scrutExpr altTExpr altFExpr)
                 return $! scrutDecls ++ altTDecls ++ altFDecls ++ tickDecls ++ [dstAssign]
+    _ | scrutTy == integerPrimTy
+      -> throw (ClashException sp ($(curLoc) ++ "Can't create netlist for:  case (_ :: Integer) of ...") (Just $ showPpr e))
+    _ | scrutTy == naturalPrimTy
+      -> throw (ClashException sp ($(curLoc) ++ "Can't create netlist for:  case (_ :: Natural) of ...") (Just $ showPpr e))
     _ -> do
       reprs <- Lens.view customReprs
       let alts1 = (reorderDefault . reorderCustom tcm reprs scrutTy) alts0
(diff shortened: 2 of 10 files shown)


clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -702,3 +704,37 @@ elimExistentials (TransformContext is0 _) (Case scrut altsTy alts0) = do
...
+-- @
+-- data Natural = NS Word# | NB ByteArray#
+-- data Integer = IS Int#  | IP ByteArray#  | IN ByteArray#
+-- @
+--
+-- Because the NB/IP/IN don't/can't exist in HDL, this transformation looks for case with
+-- patterns for NS/IS and just always picks those alternatives.
+--
+-- This is as "safe" as the rest of the Natural/Integer handling that clash does in HDL,
+-- because numbers bigger then Word/Int can't exist there anyway.
```
rowanG077, line 725: Perhaps it's good to document here that GHC has the invariant that if a number fits into `Word#`/`Int#` it will always use the `NS/IS` constructor.

clash-lib/src/Clash/Netlist.hs
```diff
@@ -525,6 +527,10 @@ mkSelection declType bndr scrut altTy alts0 tickDecls = do
          | otherwise
           -> do dstAssign <- contAssign dstId (IfThenElse scrutExpr altTExpr altFExpr)
                 return $! scrutDecls ++ altTDecls ++ altFDecls ++ tickDecls ++ [dstAssign]
+    _ | scrutTy == integerPrimTy
+      -> throw (ClashException sp ($(curLoc) ++ "Can't create netlist for:  case (_ :: Integer) of ...") (Just $ showPpr e))
+    _ | scrutTy == naturalPrimTy
+      -> throw (ClashException sp ($(curLoc) ++ "Can't create netlist for:  case (_ :: Integer) of ...") (Just $ showPpr e))
```
rowanG077: `Integer` -> `Natural`

rowanG077 (approved):
Small nits but in general I think this looks good. Regarding where to put this. I can't really give meaningful input there.  

I would like some tests some tests that it indeed eliminates these cases.

martijnbastiaan:
Should this be behind a flag like we discussed @leonschoorl? Or is that for something else?

leonschoorl:
@martijnbastiaan This is just fixing a bug in the way clash works with GHC >= 9.4.
And without it clash can generate broken HDL code in some cases.
So it strictly improving the current situation, and therefor shouldn't be behind a flag.

I do feel something for your idea to create a flag that would stop clash generating any HDL that contains Naturals or Integers.
But that could be its own issue/PR.

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -36,6 +36,7 @@ normalization =
   bindConst >-> letTL
   >-> evalConst
   >-!-> cse >-!-> cleanup >->
+  elimCaseBigNum >->
```
christiaanb: Why so late in the chain? Why not run it as the first transformation, eliminating normalization on all the ultimately unused bignum branch.

clash-lib/src/Clash/Normalize/Transformations/Case.hs
rowanG077: please don't forget this :)

clash-lib/src/Clash/Normalize/Strategy.hs
leonschoorl: As the evaluator is able to work with big numbers, my instinct was to put this new transformation fairly late.
Out of fear of doing incorrect transformations on a Natural hidden inside a SNat/Nat/KnownNat, that actually was actually big, and could have been constant folded away.

christiaanb: Right, that makes sense! It's worth a comment/note in the source!

leonschoorl: I've added a note.
And I've managed to construct an example were clash is actually having to do a caseCon transformation on a big natural inside an SNat, that fails when `elimCaseBigNumInternals` is done too early.
That's been added to the test

changelog/2026-04-16T14_23_19+02_00_dont_create_netlist_cases_on_Integer_or_Natural_constructors
```diff
@@ -0,0 +1 @@
+FIXED Clash won't create (invalid) netlist cases on Integer and Natural constructors anymore [#3157](https://github.com/clash-lang/clash-compiler/issues/3157)
```
rowanG077, line 1: We always use `FIXED:` see the missing `:`

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -136,6 +138,17 @@ constantPropagation =
         -- See Note [zeroWidthSpec enabling transformations]
       ]
 
+{-
+[Note] late elimCaseBigNum
+
+elimCaseBigNum is placed fairly late in the pipeline, after any constant folding and after caseCon.
+Because Naturals also hide inside SNat/Nat/KnownNats, and those can be bigger then 64bits.
+But they're ultimatily monomorphic and thus static and will be evaluated.
```
rowanG077, line 146: `ultimatily` -> `ultimately`

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -136,6 +138,17 @@ constantPropagation =
...
+We just have to make sure that evaluation happens before we do elimCaseBigNum.
+
+See foo in the test T3157_IntegerNaturalInternals.hs for an example of where this can happen.
```
rowanG077, line 149: Please add the full path since it might not immediately obvious for people where `T3157_IntegerNaturalInternals.hs` resides.

rowanG077 (changes requested):
small doc things, the rest looks good.

rowanG077 (approved):
Make sure to squash the commits since this is really one atomic change.

leonschoorl:
@rowanG077 I explicitly did not squash it, as the first 5 commits are kind of auxiliary to the last 2, which implement the core of the PR.
And it seemed clearer to me to not have them all lumped together.

PR #3177, opened 2026-04-05
Fallible unpacking

rslawson:

As per #3150, this is an implementation of fallible unpacking (`BitVector (BitSize a) -> Maybe a`). This is largely separate from the existing `unpack` implementations, except where those can be relied upon for easy implementations (`maybeUnpack = Just . unpack`).

## Still TODO:

  - [x] Ensure that the new feature is tested properly

### DIFF
--- a/changelog/2026-07-23T11_50_16+02_00_add_maybeUnpack
+++ b/changelog/2026-07-23T11_50_16+02_00_add_maybeUnpack
@@ -1,0 +1,5 @@
+ADDED: `maybeUnpack` to `BitPack`, providing a synthesizable way to reject bit
+patterns that do not represent a value. This is a source-breaking addition for
+manually written `BitPack` instances: they must either define `maybeUnpack` or
+satisfy its generic default implementation. See
+[#3177](https://github.com/clash-lang/clash-compiler/pull/3177).
--- a/changelog/2026-07-23T11_50_16+02_00_add_maybeUnpack
+++ b/changelog/2026-07-23T11_50_16+02_00_add_maybeUnpack
@@ -1,4 +1,4 @@
-ADDED: `maybeUnpack` to `BitPack`, providing a synthesizable way to reject bit
+ADDED: Added `maybeUnpack` to `BitPack`, providing a synthesizable way to reject bit
 patterns that do not represent a value. This is a source-breaking addition for
 manually written `BitPack` instances: they must either define `maybeUnpack` or
 satisfy its generic default implementation. See
--- a/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
+++ b/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
@@ -59,7 +59,7 @@ import qualified Clash.Annotations.BitRepresentation.Util
 
 import           Clash.Annotations.Primitive  (hasBlackBox)
 import           Clash.Class.BitPack
-  (BitPack, BitSize, pack, packXWith, unpack)
+  (BitPack, BitSize, pack, packXWith, unpack, maybeUnpack)
 import           Clash.Class.Resize           (resize)
 import           Language.Haskell.TH.Compat   (mkTySynInstD)
 import           Clash.Sized.BitVector        (BitVector, low, (++#))
@@ -948,6 +948,21 @@ buildUnpack (DataReprAnn _name _size constrs) = do
   let func          = FunD 'unpack [Clause [VarP argNameIn] (NormalB unpackApplied) []]
   return [func]
 
+buildMaybeUnpack
+  :: DataReprAnn
+  -> Q [Dec]
+buildMaybeUnpack (DataReprAnn _name _size constrs) = do
+  argNameIn   <- newName "toBeUnpackedIn"
+  argName     <- newName "toBeUnpacked"
+  matches     <- mapM (buildUnpackIfE argName) constrs
+  let justMatches      = map (\(g, e) -> (g, AppE (ConE 'Just) e)) matches
+      otherwiseNothing = [(NormalG (ConE 'True), ConE 'Nothing)]
+      unpackBody       = MultiIfE (justMatches ++ otherwiseNothing)
+      unpackLambda     = LamE [VarP argName] unpackBody
+      unpackApplied    = (VarE 'dontApplyInHDL) `AppE` unpackLambda `AppE` (VarE (argNameIn))
+      func             = FunD 'maybeUnpack [Clause [VarP argNameIn] (NormalB unpackApplied) []]
+  return [func]
+
 -- | Derives BitPack instances for given type. Will account for custom bit
 -- representation annotations in the module where the splice is ran. Note that
 -- the generated instance might conflict with existing implementations (for
@@ -987,8 +1002,9 @@ deriveBitPack typQ = do
               []  -> fail "No custom bit annotation found."
               _   -> fail "Overlapping bit annotations found."
 
-  packFunc   <- buildPack ann
-  unpackFunc <- buildUnpack ann
+  packFunc        <- buildPack ann
+  unpackFunc      <- buildUnpack ann
+  maybeUnpackFunc <- buildMaybeUnpack ann
 
   let (DataReprAnn _name dataSize _constrs) = ann
 
@@ -1001,7 +1017,7 @@ deriveBitPack typQ = do
                    -- Context
                    (AppT (ConT ''BitPack) typ)
                    -- Type
-                   (bitSizeInst : packFunc ++ unpackFunc)
+                   (bitSizeInst : packFunc ++ unpackFunc ++ maybeUnpackFunc)
                    -- Declarations
                ]
   alreadyIsInstance <- isInstance ''BitPack [typ]
--- a/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
+++ b/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
@@ -913,22 +913,43 @@ dontApplyInHDL f a = f a
 
 buildUnpackField
   :: Name
+  -> Name
   -> Integer
   -> Q Exp
-buildUnpackField valueName mask =
-  let ranges = bitRanges mask in
-  let vec = select' (VarE valueName) ranges in
-  [| unpack $vec |]
+buildUnpackField decoderName valueName mask = do
+  let ranges = bitRanges mask
+  vec <- select' (VarE valueName) ranges
+  pure (AppE (VarE decoderName) vec)
+
+buildConstr
+  :: Name
+  -> [Exp]
+  -> Exp
+buildConstr name = foldl AppE (ConE name)
+
+buildMaybeConstr
+  :: Name
+  -> [Exp]
+  -> Exp
+buildMaybeConstr name [] =
+  AppE (ConE 'Just) (ConE name)
+buildMaybeConstr name (field:rest) =
+  foldl
+    (\acc arg -> InfixE (Just acc) (VarE '(<*>)) (Just arg))
+    (AppE (AppE (VarE 'fmap) (ConE name)) field)
+    rest
 
 buildUnpackIfE
-  :: Name
+  :: (Name -> [Exp] -> Exp)
+  -> Name
+  -> Name
   -> ConstrRepr
   -> Q (Guard, Exp)
-buildUnpackIfE valueName (ConstrRepr name mask value fieldanns) = do
-  let valueName' = return $ VarE valueName
+buildUnpackIfE buildResult decoderName valueName (ConstrRepr name mask value fieldanns) = do
+  let valueName' = pure (VarE valueName)
   guard  <- NormalG <$> [| ((.&.) $valueName' mask) == value |]
-  fields <- mapM (buildUnpackField valueName) fieldanns
-  return (guard, foldl AppE (ConE name) fields)
+  fields <- mapM (buildUnpackField decoderName valueName) fieldanns
+  pure (guard, buildResult name fields)
 
 -- | Build an /unpack/ function corresponding to given DataRepr
 buildUnpack
@@ -937,7 +958,7 @@ buildUnpack
 buildUnpack (DataReprAnn _name _size constrs) = do
   argNameIn   <- newName "toBeUnpackedIn"
   argName     <- newName "toBeUnpacked"
-  matches     <- mapM (buildUnpackIfE argName) constrs
+  matches     <- mapM (buildUnpackIfE buildConstr 'unpack argName) constrs
   let fallThroughLast []      = []
       fallThroughLast [(_,e)] = [(NormalG (ConE 'True), e)]
       fallThroughLast (x:xs)  = x:fallThroughLast xs
@@ -952,15 +973,11 @@ buildMaybeUnpack
   :: DataReprAnn
   -> Q [Dec]
 buildMaybeUnpack (DataReprAnn _name _size constrs) = do
-  argNameIn   <- newName "toBeUnpackedIn"
   argName     <- newName "toBeUnpacked"
-  matches     <- mapM (buildUnpackIfE argName) constrs
-  let justMatches      = map (\(g, e) -> (g, AppE (ConE 'Just) e)) matches
-      otherwiseNothing = [(NormalG (ConE 'True), ConE 'Nothing)]
-      unpackBody       = MultiIfE (justMatches ++ otherwiseNothing)
-      unpackLambda     = LamE [VarP argName] unpackBody
-      unpackApplied    = (VarE 'dontApplyInHDL) `AppE` unpackLambda `AppE` (VarE (argNameIn))
-      func             = FunD 'maybeUnpack [Clause [VarP argNameIn] (NormalB unpackApplied) []]
+  matches     <- mapM (buildUnpackIfE buildMaybeConstr 'maybeUnpack argName) constrs
+  let otherwiseNothing = [(NormalG (ConE 'True), ConE 'Nothing)]
+      unpackBody       = MultiIfE (matches ++ otherwiseNothing)
+      func             = FunD 'maybeUnpack [Clause [VarP argName] (NormalB unpackBody) []]
   return [func]
 
 -- | Derives BitPack instances for given type. Will account for custom bit
--- a/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
+++ b/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
@@ -59,7 +59,7 @@ import qualified Clash.Annotations.BitRepresentation.Util
 
 import           Clash.Annotations.Primitive  (hasBlackBox)
 import           Clash.Class.BitPack
-  (BitPack, BitSize, pack, packXWith, unpack)
+  (BitPack, BitSize, pack, packXWith, unpack, maybeUnpack)
 import           Clash.Class.Resize           (resize)
 import           Language.Haskell.TH.Compat   (mkTySynInstD)
 import           Clash.Sized.BitVector        (BitVector, low, (++#))
@@ -953,6 +953,21 @@ buildUnpack (DataReprAnn _name _size constrs) = do
   let func          = FunD 'unpack [Clause [VarP argNameIn] (NormalB unpackApplied) []]
   return [func]
 
+buildMaybeUnpack
+  :: DataReprAnn
+  -> Q [Dec]
+buildMaybeUnpack (DataReprAnn _name _size constrs) = do
+  argNameIn   <- newName "toBeUnpackedIn"
+  argName     <- newName "toBeUnpacked"
+  matches     <- mapM (buildUnpackIfE argName) constrs
+  let justMatches      = map (\(g, e) -> (g, AppE (ConE 'Just) e)) matches
+      otherwiseNothing = [(NormalG (ConE 'True), ConE 'Nothing)]
+      unpackBody       = MultiIfE (justMatches ++ otherwiseNothing)
+      unpackLambda     = LamE [VarP argName] unpackBody
+      unpackApplied    = (VarE 'dontApplyInHDL) `AppE` unpackLambda `AppE` (VarE (argNameIn))
+      func             = FunD 'maybeUnpack [Clause [VarP argNameIn] (NormalB unpackApplied) []]
+  return [func]
+
 -- | Derives BitPack instances for given type. Will account for custom bit
 -- representation annotations in the module where the splice is ran. Note that
 -- the generated instance might conflict with existing implementations (for
@@ -992,8 +1007,9 @@ deriveBitPack typQ = do
               []  -> fail "No custom bit annotation found."
               _   -> fail "Overlapping bit annotations found."
 
-  packFunc   <- buildPack ann
-  unpackFunc <- buildUnpack ann
+  packFunc        <- buildPack ann
+  unpackFunc      <- buildUnpack ann
+  maybeUnpackFunc <- buildMaybeUnpack ann
 
   let (DataReprAnn _name dataSize _constrs) = ann
 
@@ -1006,7 +1022,7 @@ deriveBitPack typQ = do
                    -- Context
                    (AppT (ConT ''BitPack) typ)
                    -- Type
-                   (bitSizeInst : packFunc ++ unpackFunc)
+                   (bitSizeInst : packFunc ++ unpackFunc ++ maybeUnpackFunc)
                    -- Declarations
                ]
   alreadyIsInstance <- isInstance ''BitPack [typ]
--- a/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
+++ b/clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
@@ -918,22 +918,43 @@ dontApplyInHDL f a = f a
 
 buildUnpackField
   :: Name
+  -> Name
   -> Integer
   -> Q Exp
-buildUnpackField valueName mask =
-  let ranges = bitRanges mask in
-  let vec = select' (VarE valueName) ranges in
-  [| unpack $vec |]
+buildUnpackField decoderName valueName mask = do
+  let ranges = bitRanges mask
+  vec <- select' (VarE valueName) ranges
+  pure (AppE (VarE decoderName) vec)
+
+buildConstr
+  :: Name
+  -> [Exp]
+  -> Exp
+buildConstr name = foldl AppE (ConE name)
+
+buildMaybeConstr
+  :: Name
+  -> [Exp]
+  -> Exp
+buildMaybeConstr name [] =
+  AppE (ConE 'Just) (ConE name)
+buildMaybeConstr name (field:rest) =
+  foldl
+    (\acc arg -> InfixE (Just acc) (VarE '(<*>)) (Just arg))
+    (AppE (AppE (VarE 'fmap) (ConE name)) field)
+    rest
 
 buildUnpackIfE
-  :: Name
+  :: (Name -> [Exp] -> Exp)
+  -> Name
+  -> Name
   -> ConstrRepr
   -> Q (Guard, Exp)
-buildUnpackIfE valueName (ConstrRepr name mask value fieldanns) = do
-  let valueName' = return $ VarE valueName
+buildUnpackIfE buildResult decoderName valueName (ConstrRepr name mask value fieldanns) = do
+  let valueName' = pure (VarE valueName)
   guard  <- NormalG <$> [| ((.&.) $valueName' mask) == value |]
-  fields <- mapM (buildUnpackField valueName) fieldanns
-  return (guard, foldl AppE (ConE name) fields)
+  fields <- mapM (buildUnpackField decoderName valueName) fieldanns
+  pure (guard, buildResult name fields)
 
 -- | Build an /unpack/ function corresponding to given DataRepr
 buildUnpack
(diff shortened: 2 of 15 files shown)


clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -190,6 +191,18 @@ unpack# :: KnownNat n => BitVector (CLogWZ 2 n 0) -> Index n
 unpack# (BV 0 i) = fromInteger_INLINE (naturalToInteger i)
 unpack# bv = undefError "Index.unpack" [bv]
 
+{-# OPAQUE maybeUnpack# #-}
+{-# ANN unpack# hasBlackBox #-}
```
martijnbastiaan:
```suggestion
{-# ANN maybeUnpack# hasBlackBox #-}
```

Though I think you should just implement this without black boxes. Just use compare and the like without deconstructing the `Index` :). It save you trouble, it reduces the "attack surface" for bugs, and it would eliminate the need to do HDL testing.

clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -190,6 +191,18 @@ unpack# :: KnownNat n => BitVector (CLogWZ 2 n 0) -> Index n
...
+maybeUnpack# :: KnownNat n => BitVector (CLogWZ 2 n 0) -> Maybe (Index n)
+maybeUnpack# (BV 0 i) = maybeFromInteger_INLINE (naturalToInteger i)
+maybeUnpack# bv = undefError "Index.maybeUnpack" [bv]
+
+{-# INLINE maybeFromInteger_INLINE #-}
+maybeFromInteger_INLINE :: forall n . (HasCallStack, KnownNat n) => Integer -> Maybe (Index n)
+maybeFromInteger_INLINE i = bound `seq` if i > (-1) && i < bound then Just (I i) else Nothing
```
martijnbastiaan: Huh, why the `seq`?

clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
```diff
@@ -948,6 +948,21 @@ buildUnpack (DataReprAnn _name _size constrs) = do
...
+  -> Q [Dec]
+buildMaybeUnpack (DataReprAnn _name _size constrs) = do
+  argNameIn   <- newName "toBeUnpackedIn"
+  argName     <- newName "toBeUnpacked"
+  matches     <- mapM (buildUnpackIfE argName) constrs
+  let justMatches      = map (\(g, e) -> (g, AppE (ConE 'Just) e)) matches
+      otherwiseNothing = [(NormalG (ConE 'True), ConE 'Nothing)]
+      unpackBody       = MultiIfE (justMatches ++ otherwiseNothing)
+      unpackLambda     = LamE [VarP argName] unpackBody
+      unpackApplied    = (VarE 'dontApplyInHDL) `AppE` unpackLambda `AppE` (VarE (argNameIn))
```
martijnbastiaan: Any reason you're using magic like `dontApplyInHDL`? Just naively I would expect the functionality of this to be entirely implementable without it.

martijnbastiaan (commented):
@rslawson Not a real review yet, just wanted to drop by to say that it's awesome that you're working on this :). It seems very sensible to me to add this as a separate `maybeUnpack` function that's a member of `BitPack`.

clash-prelude/src/Clash/Sized/Internal/Index.hs
rslawson: Entirely because I copied `fromInteger_INLINE` and just changed a couple things. @rowanG077 suggested when I talked with him in the office that I should probably rewrite that function in terms of this one, but I forgot to do that before committing this.

clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
rslawson: It's there from copying over parts of the code in `buildUnpack`. I wasn't sure which parts I should or shouldn't keep, so I just went with the strategy of take what I can if it doesn't break things.

clash-prelude/src/Clash/Sized/Internal/Index.hs
rowanG077: The `seq` is there in `fromInteger_INLINE` and was copied here I guess. I'm not sure why it's there either. It was introduced in https://github.com/clash-lang/clash-compiler/commit/8ca93af9a30c2cfb08677738117ee9d33555e658 but the only thing the commit says is `A more readable implementation of Index.fromInteger`

clash-prelude/src/Clash/Sized/Internal/Index.hs
rslawson: Ah, right, I think I just forgot to remove this annotation. Added it early on and just forgot it was there, oops.

clash-prelude/src/Clash/Annotations/BitRepresentation/Deriving.hs
DigitalBrains1: The source of that `dontApplyInHdl` is [this commit](https://github.com/clash-lang/clash-prelude/commit/085cadb0b1776f0bb56e7384310a109f3265d93c) (part of [this PR](https://github.com/clash-lang/clash-prelude/pull/168), the introduction of custom bit-representations).

While it does clean up the generated HDL, it is also the only occurrence I know of so far that _requires_ `BitPack` to have the same bit representation as the type itself, [causing errors](https://github.com/clash-lang/clash-compiler/discussions/3150#discussioncomment-15861926) if it is not.

rowanG077:
To give an update on this:

I want to do some tests with yosys locally to see what kind of resource usage we can expect from `maybeUnpack`. I'm also curious if now  `fromJustX . maybeUnpack` will actually make it free.

rowanG077:
I tested bittide by replacing `unpack` with `fromJustX . maybeUnpack` and these are the results when synthesizing it for an ECP5 target. The proprietary blackboxes are stubbed out.

```
  ┌─────────────────┬──────────┬─────────┬───────┐
  │ Resource        │ Baseline │ Patched │ Delta │
  ├─────────────────┼──────────┼─────────┼───────┤
  │ LUT4            │    64801 │   65014 │  +213 │
  │ TRELLIS_FF      │    45452 │   45452 │     0 │
  │ CCU2C           │     5939 │    5926 │   -13 │
  │ L6MUX21         │      305 │     372 │   +67 │
  │ PFUMX           │     5869 │    5857 │   -12 │
  │ DP16KD          │      826 │     826 │     0 │
  │ TRELLIS_DPR16X4 │      132 │     132 │     0 │
  │ MULT18X18D      │       12 │      12 │     0 │
  └─────────────────┴──────────┴─────────┴───────┘
```

So it is a regression. Testing separately does show it does optimize things out correctly but it's probably not always possible in larger structures. For me this PR can now be merged. I will retag the reviewers.

martijnbastiaan:
@rowanG077 Do you have a top entity you can share? I'll push it through Vivado.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -441,16 +509,19 @@ instance BitPack Half where
...
+  unpack  _   = ()
+  maybeUnpack = Just . unpack
 
 instance BitPack Char where
   type BitSize Char = 21
-  pack   = packXWith packChar#
-  unpack = checkUnpackUndef unpackChar#
+  pack        = packXWith packChar#
+  unpack      = checkUnpackUndef unpackChar#
+  maybeUnpack = Just . unpack
```
DigitalBrains1:
```haskell
>>> unpack 0x110000 :: Char
*** Exception: Prelude.chr: bad argument: 1114112
```

Unicode code points lie in the range U+0000 to U+D7FF and U+E000 to U+10FFFF. It seems Haskell isn't too concerned with the gap, it doesn't error for values in that range that I tried. But it does bottom with an `ErrorCall` for numbers larger than 0x10ffff. Not an XException, mind you.

I think it'd be better if `maybeUnpack` returned `Nothing` for out-of-range numbers.

DigitalBrains1:
Ah, this is a very welcome addition, thanks! I've only quickly scrolled through so far, I'll give it a bit more time later.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
DigitalBrains1: Values in the gap round-trip neatly in GHC 9.10.3:
```haskell
>>> all (\x -> x == (ord $ chr x)) [0xd800 .. 0xdfff]
True
```

martijnbastiaan: Agree Peter! This is also one of the rules of NumConvert: succes -> not bottom. It makes a lot of sense to make conversion functions do this across the prelude.

rowanG077:
@martijnbastiaan Yes 

The `maybeUnpack.zip` contains the clash generated files for `unpack = fromJustX . maybeUnpack`. `unpack.zip` is the baseline.

[maybeUnpack.zip](https://github.com/user-attachments/files/28349455/maybeUnpack.zip)
[unpack.zip](https://github.com/user-attachments/files/28349457/unpack.zip)

clash-prelude/src/Clash/Class/BitPack/Internal/TH.hs
```diff
@@ -107,4 +116,39 @@ deriveBitPackTuples bitPackName bitSizeName packName unpackName = do
                 []
             ]
 
-    in InstanceD Nothing context instTy [bitSizeType, pack, unpack]
+        maybeUnpack =
```
DigitalBrains1, line 126: Hah, you go in thinking it'll just be very similar to `unpack` but with a `Functor (Maybe _)` like so many other instances, and it turns out to be a lot more complicated! Well, in TH. In Haskell it's still very basic, but still.

This definition looks fine, I was just wondering why it was more complicated than others and understand it now. The TH is still somewhat readable, but I asked GHC to read it for me so it becomes more readable. Just for reference, `-ddump-splices` tells us the three-tuple is:

```haskell
instance (BitPack a,
          KnownNat (BitSize a),
          BitPack (a0, a1),
          KnownNat (BitSize (a0, a1))) =>
         BitPack (a, a0, a1) where
  type BitSize (a, a0,
                a1) = (+) (BitSize a) (BitSize (a0, a1))
  pack tup
    = pack (retup tup)
    where
        retup (a, a0, a1)
          = (a, (a0, a1))
  unpack x
    = let
        (a, y) = unpack x
        (a0, a1) = unpack y
      in (a, a0, a1)
  maybeUnpack x
    = let
        bvL :: BitVector (BitSize a)
        bvR :: BitVector (BitSize (a0, a1))
        (bvL, bvR) = split# x
      in
        case (maybeUnpack bvL, maybeUnpack bvR) of
          (Just a, Just (a0, a1))
            -> Just (a, a0, a1)
          _ -> Nothing
```

I've cleaned up the names for readability.

clash-prelude/src/Clash/Sized/Internal/Index.hs
```diff
@@ -214,6 +215,16 @@ unpack# :: KnownNat n => BitVector (CLogWZ 2 n 0) -> Index n
 unpack# (BV 0 i) = fromInteger_INLINE (naturalToInteger i)
 unpack# bv = undefError "Index.unpack" [bv]
 
+{-# INLINE maybeUnpack# #-}
+maybeUnpack# :: forall n. KnownNat n => BitVector (CLogWZ 2 n 0) -> Maybe (Index n)
+maybeUnpack# bv
+  | bound == 0 = Nothing
+  | bv <= maxBoundBV = Just (unpack# bv)
```
DigitalBrains1, line 226: I think this is good enough, but it is a bit of a pity that the error message isn't as straight-forward as the `unpack` case:

```haskell
>>> pack (maxBound :: Index 3)
0b10
>>> let bv = $(bLit "0.")
>>> unpack bv :: Index 3
*** Exception: X: Index.unpack called with (partially) undefined arguments: 0b0.
>>> maybeUnpack bv :: Maybe (Index 3)
*** Exception: X: Clash.Sized.BitVector.<= called with (partially) undefined arguments: 0b0. <= 0b10
```

Do we want to improve the error message with something funky like
```haskell
maybeUnpack# bv
  | bound == 0 = Nothing
  | !i <- unpack# bv
  , bv <= maxBoundBV = Just i
  | otherwise = Nothing
 where
  bound = natToInteger @n
  maxBoundBV = pack# (maxBound# @n)
```
?

That way `maybeUnpack#` still doesn't require a black box. The error message is still referencing `Index.unpack` though. We could also use [`clashSimulation`](https://hackage-content.haskell.org/package/clash-prelude-1.10.0/docs/Clash-Magic.html#v:clashSimulation) to only dive into the `BV` constructor in simulation and use the code as it is now for HDL generation! That might be even better.

tests/shouldwork/BitVector/MaybeUnpack.hs
```diff
@@ -0,0 +1,115 @@
...
+        P.<> ", but got "
+        P.<> P.show actual
+        )
+
+assertNothing :: P.String -> Maybe a -> IO ()
+assertNothing _ Nothing = pure ()
+assertNothing label (Just _) =
+  P.error (label P.<> ": expected Nothing, but got Just _")
+
+mainCommon :: IO ()
```
DigitalBrains1: Doesn't this `outputTest` just do some stuff that can be done in `clash-prelude/tests/Clash/Tests/BitPack.hs`, and in fact it already overlaps with that? I think any non-overlap is better put in the Tasty suite over in `clash-prelude`.

DigitalBrains1 (commented):
Looking good! Aside from the [true issue with `maybeUnpack` for `Char`](https://github.com/clash-lang/clash-compiler/pull/3177#discussion_r3305482042) I already commented on, I just wonder why a test is what it is, have a small suggestion regarding an error message and I wanted to show the reader what a piece of TH does, just as information.

clash-prelude/src/Clash/Sized/Internal/Index.hs
DigitalBrains1: Wait, I'm being silly, by forcing `unpack#` to WHNF before checking for out-of-bounds, it will bottom for out-of-bounds values through the check in `fromInteger_INLINE`.

But the following works in simulation and produces nice HDL (tested in GHC 9.12.4):

```haskell
maybeUnpack# bv
  | bound == 0 = Nothing
  | clashSimulation
  , BV mask _ <- bv
  , mask > 0 = undefError "Index.maybeUnpack" [bv]
  | bv <= maxBoundBV = Just (unpack# bv)
  | otherwise = Nothing
 where
  bound = natToInteger @n
  maxBoundBV = pack# (maxBound# @n)
```

clash-prelude/src/Clash/Class/BitPack/Internal/TH.hs
DigitalBrains1: I just started wondering why you give `bvL` and `bvR` type signatures. Surely from the fact that they unpack to `a` and `(a0, a1)`, GHC should be able to imply where to `split#` the bit vector? I copied the code in the previous message in the place of `deriveBitPackTuples`, removed the type signatures of `bvL` and `bvR` and sure enough it compiles and works:

```haskell
  maybeUnpack x
    = let
        (bvL, bvR) = split# x
      in
        case (maybeUnpack bvL, maybeUnpack bvR) of
          (Just a, Just (a0, a1))
            -> Just (a, a0, a1)
          _ -> Nothing
```

so it's possible to simplify this TH AST a bit.

tests/shouldwork/BitVector/MaybeUnpack.hs
rowanG077: This is done.

rowanG077:
Oke I think now this is in a good shape to be merged. I have retagged you @DigitalBrains1. 

I also included some hardware specific optimization. Which improved synthesis logic usage and depth in some cases when using yosys.

changelog/2026-07-23T11_50_16+02_00_add_maybeUnpack
```diff
@@ -0,0 +1,5 @@
+ADDED: `maybeUnpack` to `BitPack`, providing a synthesizable way to reject bit
```
DigitalBrains1: Since the `ADDED` will no longer be in front of the rest in the changelog (it just directs where in the changelog the text will end up), this sentence gets weird: _`maybeUnpack` to `BitPack`_. Please just repeat the word _added_.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -161,6 +165,54 @@ class KnownNat (BitSize a) => BitPack a where
    where
     (checkUnpackUndef unpack . resize -> sc, bFields) = split# b
 
+  -- | Attempt to convert a 'BitVector' to an element of type @a@. If the unpacking
```
DigitalBrains1, line 172: We can fix this in a followup, but the current documentation feels like it leaves some guesswork to the reader. `unpack` is documented as _Convert a `BitVector` to an element of type `a`_ and `maybeUnpack` is documented as _Attempt to convert a `BitVector` to an element of type `a`_, without really explaining the difference.

Maybe we should explain something like:

If there is no `x` such that `bv = pack x`, then `maybeUnpack bv` will return `Nothing`. Conversely, `unpack bv` can return invalid values or map to some other value `y` where `pack y` is not equal to `bv`.

And perhaps that can be phrased better. Documentation is pretty hard :-).

clash-prelude/src/Clash/Class/BitPack/Internal/TH.hs
```diff
@@ -86,25 +109,33 @@ deriveBitPackTuples bitPackName bitSizeName packName unpackName = do
...
+                    [ Match
+                        nestedPattern
+                        (NormalB flatExpression)
+                        []
+                    ]))
                 []
             ]
 
-    in InstanceD Nothing context instTy [bitSizeType, pack, unpack]
+        maybeUnpack =
```
DigitalBrains1, line 126: Ooh, neatly solved, nicely brief expression.

tests/shouldwork/BitVector/MaybeUnpack.hs
```diff
@@ -0,0 +1,75 @@
...
+         , Just (Wrapped 2)
+         , Just (2, 2)
+         , Just (2 :> 2 :> Nil)
+         , Just (2, 2, 2, 2)
+         )
+      :> (Nothing, Nothing, Nothing, Nothing, Nothing)
+      :> Nil
+      )
+  done = expectedOutput (topEntity <$> testInput)
+  clk = tbSystemClockGen (Clash.Prelude.not <$> done)
```
DigitalBrains1: Nit pick: there's surely no other `not` in scope? No need to qualify, right?

tests/shouldwork/CustomReprs/Deriving/MaybeUnpackDerivation.hs
```diff
@@ -0,0 +1,95 @@
...
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module MaybeUnpackDerivation where
+
+import           Data.Maybe                             (isNothing)
```
DigitalBrains1: Another nit pick: these days, we don't align import statements, we just use single spaces. To keep picking: there's also no `qualified` import asking for the horizontal space after `import` :-D.

DigitalBrains1 (commented):
Ah, some nice improvements! I just have some minor natural language suggestions.

I'm _requesting changes_ because I do think the changelog should be changed. I'd like the Haddock to be extended but we could also do that later. Do what you want with the nit picks.

Does the balancing of range checking now enable Yosys to fully eliminate the checks from a `fromJustX . maybeUnpack`, or does it just reduce the pain of it?

clash-prelude/src/Clash/Class/BitPack/Internal/TH.hs
DigitalBrains1: The new version is much sweeter.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
DigitalBrains1: So a rather surprising thing about `unpack` is that it can introduce logic. Say we want a registered input:

```haskell
data Color = Red | Green | Blue deriving (Generic, BitPack)

topEntity ::
  ... ->
  Signal System (BitVector 2) ->
  ...
topEntity clk rst en x = ...
 where
  x0 :: Signal System Color
  x0 = register clk rst en Red $ unpack x
```

This actually has logic between register and I/O-pins (thus also bypassing the usual dedicated I/O-register, routing the I/O directly into a LUT and using a regular register), though you might actually think that `unpack` preserves every bit and thus should be free in hardware.

It would be really nice if we could document that you should instead use

```haskell
  x0 = register clk rst en Red $ fromJustX $ maybeUnpack x
```

but can we? Given the decreased QoR in Yosys [mentioned above](https://github.com/clash-lang/clash-compiler/pull/3177#issuecomment-4543843650) it would seem this cannot actually be relied upon? Or will such a basic use correctly "optimise" the X away and is the issue only with more complicated patterns?

rowanG077:
I investigated a bit deeper regarding the resource usage on yosys of `fromJustX . maybeUnpack` and we definitely can't rely on it being free. In some cases HDL like this is generated for `fromJustX . maybeUnpack`

```Verilog
module top (
    input  wire [1:0] x,
    output wire [1:0] y
  );
    wire [1:0] chain = (x == 2'd0) ? 2'd0
                     : (x == 2'd1) ? 2'd1
                     : 2'd2;
    assign y = (x == 2'd3) ? 2'bxx : chain;
  endmodule
```

If the synthesizer were smart enough it could see that x = y is valid. But unfortunately yosys does a greedy simplification:

```Verilog
assign y = (x == 2'd3) ? 2'bxx : chain; => assign y = chain;
```

So you essentially get:

```Verilog
assign y = (x == 2'd0) ? 2'd0
                     : (x == 2'd1) ? 2'd1
                     : 2'd2;
```

Which maps `x == 2'd3` to `2'd2`  and thus it's no longer id and you actually get a LUT.

martijnbastiaan:
Is this due to Clash translating:

```haskell
data Abc = A | B | C

topEntity :: Abc -> Unsigned 2
topEntity = \case
  A -> 0
  B -> 1
  C -> 2
```

to:

```verilog
  always @(*) begin
    case(arg)
      2'b00 : result = 8'd10;
      2'b01 : result = 8'd20;
      default : result = 8'd30;
    endcase
  end
```

? It should probably translate it to:

```verilog
  always @(*) begin
    case(arg)
      2'b00 : result = 8'd10;
      2'b01 : result = 8'd20;
      2'b10 : result = 8'd30;
      default : result = 8'hxx;
    endcase
  end
```

rowanG077:
@martijnbastiaan I don't think so. If I test (the equivalent of your VHDL in Verilog):

```Verilog
case (x)
    2'b00 : y = 2'd0;
    2'b01 : y = 2'd1;
    2'b10 : y = 2'd2;
    default : y = 2'bxx;
endcase
```
 
 Even though it's semantically equivalent to the nested ternary yosys introduces 2 LUTs. Because it rewrites it to:
 
  ```Verilog
case (x)
    2'b00 : y = 2'd0;
    2'b01 : y = 2'd1;
    2'b10 : y = 2'd2;
    default : y = 2'd0;
endcase
```
 
Whereas the nested ternary case essentially collapses the final two case arms together.

To be clear. This means that that problem is independent of the "valid" bit of the `Maybe` and even a `unpack` variant that returns `x` on invalid bit patterns will not be free.

Edit: Neither me nor fable were able to construct a case where yosys would keep around the actual validation logic that feeds the valid bit with `fromJustX . maybeUnpack`.

**Comment by martijnbastiaan** [comment]
> @martijnbastiaan Yes
> 
> The `maybeUnpack.zip` contains the clash generated files for `unpack = fromJustX . maybeUnpack`. `unpack.zip` is the baseline.
> 
> [maybeUnpack.zip](https://github.com/user-attachments/files/28349455/maybeUnpack.zip) [unpack.zip](https://github.com/user-attachments/files/28349457/unpack.zip)

I finally did this and ends up with the exact same utilization. I also tested:

```
module top (
    input  wire [1:0] x,
    output wire [1:0] y
  );
    wire [1:0] chain = (x == 2'd0) ? 2'd0
                     : (x == 2'd1) ? 2'd1
                     : 2'd2;
    assign y = (x == 2'd3) ? 2'bxx : chain;
  endmodule
```

which Vivado, like Yosys, doesn't optimize. But.. it doesn't seem like `maybeUnpack.zip` contained any of those.

**Comment by martijnbastiaan** [comment]
For [test_cases.zip](https://github.com/user-attachments/files/30577516/test_cases.zip), Vivado seems a bit smarter than Yosys.

| File | Shape | Vivado 2025.2 | yosys 0.67 (`synth_ecp5`) |
|---|---|---|---|
| `baseline_nox.v` | no X — invalid pattern aliases to last value | 1 LUT | 1 LUT4 |
| `ternary_x.v` | `cond ? 'x : chain` (X in then-arm) | 1 LUT | 1 LUT4 |
| `case_x.v` | `case` with `default: 'x` | **0 LUTs** (`y = x`) | **2 LUT4** |
| `ternary_x_end.v` | ternary chain ending in `: 'x` | **0 LUTs** (`y = x`) | **1 LUT4** |
| `fromjustx_id.v` | `valid ? x : 'x` (passthrough payload) | 0 LUTs | 0 LUTs (`y = x`) |

**Comment by rowanG077** [comment]
TBH it means we cannot rely on such optimizations. Just an extra let confuses both yosys and the industrial strength synthesizer in vivado.

I think at least for the generic path we can omit any logic in generated HDL by using [`dontApplyInHDL`](https://hackage-content.haskell.org/package/clash-prelude-1.10.0/docs/Clash-Annotations-BitRepresentation-Deriving.html#v:dontApplyInHDL). That's the exact path that `BitPack` TH deriving uses.

**Comment by martijnbastiaan** [comment]
Hmmyeah. I do worry a bit. Clash is quite eager to generate arms with `x`s, and Yosys inability to deal with it hurts us. As in, yeah, we can work around it here (which I'm not against), but that still leaves all the other generic cases.

How did you end up generating the `chain` code from Clash?

**Comment by rowanG077** [comment]
> How did you end up generating the chain code from Clash?

It doesn't literally generate the ternaries. It generates separate case statement. I flattened that to ternaries so the code would be easier to understand. This clash code:

```Haskell
data Color = Red | Green | Blue
  deriving (Generic, NFDataX, BitPack)

topEntity :: BitVector 2 -> Color
topEntity = unpack
```

Generates a very noisy:

```Verilog
** GENERATED BY CLASH 1.11.0. DO NOT MODIFY.
*/
`default_nettype none
`timescale 100fs/100fs
module topEntity
    ( // Inputs
      input wire [1:0] b


      // Outputs
    , output wire [1:0] c$case_alt
    );
  wire [1:0] c$case_alt_1;
  wire [1:0] x;
  reg [2:0] c$case_alt_2;
  reg [2:0] c$$j_case_alt;
  reg [1:0] ds;
  wire signed [63:0] x1;
  wire signed [63:0] c$case_alt_selection_res;
  wire signed [63:0] c$$j_case_alt_selection_res;
  wire signed [63:0] ds_selection_res;
  wire [63:0] c$bv;
  wire signed [63:0] x1_projection;

  assign c$case_alt_1 = x[0:0] ? 2'd2 : 2'd1;

  assign c$case_alt = x[1:1] ? c$case_alt_1 : 2'd0;

  assign x = c$case_alt_2[1:0];

  assign c$case_alt_selection_res = (x1 < 64'sd3) ? 64'sd1 : 64'sd0;

  always @(*) begin
    case(c$case_alt_selection_res)
      64'sd1 : c$case_alt_2 = c$$j_case_alt;
      default : c$case_alt_2 = {1'b0,   {2 {1'bx}}};
    endcase
  end

  assign c$$j_case_alt_selection_res = (x1 < 64'sd1) ? 64'sd1 : 64'sd0;

  always @(*) begin
    case(c$$j_case_alt_selection_res)
      64'sd1 : c$$j_case_alt = {1'b1,
                                {1'b0,1'bx}};
      default : c$$j_case_alt = {ds[1:1],
                                 {1'b1,ds[0:0]}};
    endcase
  end

  assign ds_selection_res = (x1 < 64'sd2) ? 64'sd1 : 64'sd0;

  always @(*) begin
    case(ds_selection_res)
      64'sd1 : ds = {1'b1,   1'd0};
      default : ds = {1'b1,   1'd1};
    endcase
  end

  assign c$bv = ({62'b00000000000000000000000000000000000000000000000000000000000000,b});

  assign x1_projection = $unsigned(c$bv[0+:64]);

  assign x1 = x1_projection;


endmodule
`default_nettype wire
```

rowanG077:
I also implemented as proposal commits the path that generic generated `BitPack` instances do not do any logic in HDL. That is problematic in the case that a user writes a `BitPack` where the representation does not match clash internal representation, like:

```Haskell
data T1 = A | B deriving (Generic, NFDataX, Eq, Show, ShowX)

instance BitPack T1 where
  type BitSize T1 = 1
  pack A = 1
  pack B = 0
  unpack 1 = A
  unpack _ = B
  maybeUnpack = Just . unpack

topEntity :: BitVector 2 -> Unsigned 2
topEntity bv = case unpack bv of
 Just A -> 0
 Just B -> 1
 Nothing -> 2
```

Then Haskell simulation for `[0, 1, 2, 3]` will show `[2, 2, 1, 0]` and HDL sim will show `[2, 2, 0, 1]`.

rowanG077:
@martijnbastiaan @DigitalBrains1 I have split off the change that pack/unpack becomes logicless into https://github.com/clash-lang/clash-compiler/pull/3345

I think at this point we can merge this PR? Any remaining blockers?

martijnbastiaan (approved):
Not from my side :)

DigitalBrains1 (approved):

PR #3182, opened 2026-04-10
Add support for GHC 9.14

rowanG077:
Branched off from https://github.com/clash-lang/clash-compiler/pull/2941

A ton of bounds relaxation are necessary.

- [x] Run Test suite and fix issues
- [x] Add to test suite
- [ ] Debug HLS problems
- [X] ~~Smoke test clash-protocols~~
- [X] Smoke test `bittide-hardware`


martijnbastiaan:
@rowanG077 It's alive! 

I've done the very opinionated thing of removing `clash-term`, which has been sitting unused in the repo for quite a while now (though I do feel it's a cool project). But AFAIK it didn't do anything except for introducing a bunch of `allow-newer`s.. Should we just move it to a separate repo?

leonschoorl:
I would like to keep `clash-term` around.
I use it often to inspect or debug the clash transformation process.
And find it invaluable for that.
At least for me, it so much better and quicker then staring at the the output of `-fclash-debug-info AppliedTerm`

I'll see if I can't update it a bit, so we don't need all those `allow-newer`s anymore.

martijnbastiaan:
@mergify rebase

martijnbastiaan (approved):

DigitalBrains1 (changes requested):
So now we're back to testing 5 different GHC versions. Did you deliberately not add 9.14 to the on-push CI Cabal tests? I think it should be added.

It might be time to delegate one GHC version to only testing it in nightlies. But let's not make 9.14 that version. Maybe we can run `clash-testsuite` and Cabal tests for GHC 9.10 only on nightlies? That way, we test the oldest and the newest supported GHC versions fully, but leave out the middle one. This is why I think you should add the on-push CI Cabal tests for 9.14: we should test the outer limits, both the oldest and the newest supported GHC.

[edit]
On the other hand, Stackage LTS is at 9.10, and we suggest people use Stackage. Maybe we should move 9.8 to nightlies instead.
[/edit]

[edit 2]
I cannot get GitHub's _suggestion_ to work correctly, but if we want to move 9.8 to nightlies only, you should remove them in `on-push.yml` from the `cabal`, `packages_testsuite` and `running_testsuite` jobs. That way, all the tests like doctests and unit tests still run for 9.8, but I guess that's just something we'll have to accept. Forcing on-push CI to build on Nix with tests disabled doesn't seem a good way to do it.
[/edit 2]

martijnbastiaan:
Well I kinda lost track what our CI story is. But I'll take your comments and codify them in a (CI) script.

martijnbastiaan:
From @DigitalBrains1 I understand that the CI migration is still in flux, I'll hold off on any scripts until the migration is fully complete.

@DigitalBrains1: I've applied your suggestions and reorganized the commit history to make them a bit more self-contained. Maybe there's more to do, but I don't think it's worth the squeeze :).

martijnbastiaan:
@rowanG077 It's unclear to me what you mean with "Debug HLS problems". If there are any issues, could we please move it to an issue? I think this PR has been here for long enough..

DigitalBrains1 (approved):
Thanks!

rowanG077:
@martijnbastiaan HLS was not working. I will check again if that is still the case. If so I will move it to an issue. Regardless this can be merged.

PR #3183, opened 2026-04-10
Add `SomeBoundedSNat` and Hedgehog generator

JvWesterveld:

Useful to generate type level natural numbers in Hedgehog property-based tests, e.g. for DUT type variables. For instance, the following brings into scope the constraints `(KnownNat n, 1 <= n, n <= 8)`.
``` haskell
property $ do
  SomeBoundedSNat @n SNat <- forAll $ genSomeBoundedSNat @1 @8
  ...
```

A user might want to use e.g. `Range.constant` instead of the currently hardcoded `Range.linear`, but there's currently no `Range` argument as I'm not convinced it'd be an improvement.

## Still TODO:

### DIFF
--- a/clash-prelude-hedgehog/src/Clash/Hedgehog/Promoted/Nat.hs
+++ b/clash-prelude-hedgehog/src/Clash/Hedgehog/Promoted/Nat.hs
@@ -1,0 +1,79 @@
+{-|
+Copyright   : (C) 2026, QBayLogic B.V.
+License     : BSD2 (see the file LICENSE)
+Maintainer  : QBayLogic B.V. <devops@qbaylogic.com>
+
+Random generation of SNat.
+-}
+
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MagicHash #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskellQuotes #-}
+
+module Clash.Hedgehog.Promoted.Nat where
+
+import Clash.Prelude
+import Hedgehog (Gen)
+import qualified Data.String.Interpolate as I
+import qualified GHC.TypeNats as TypeNats
+import qualified Hedgehog.Gen as Gen
+import qualified Hedgehog.Range as Range
+
+-- | Like @SomeSNat@, but with bounds.
+data SomeBoundedSNat lower upperInclusive where
+  SomeBoundedSNat
+    :: forall n lower upperInclusive
+     . (lower <= n, n <= upperInclusive)
+    => SNat n
+    -> SomeBoundedSNat lower upperInclusive
+
+instance Show (SomeBoundedSNat lower upperInclusive) where
+  show (SomeBoundedSNat sn@SNat) = show ''SomeBoundedSNat <> " " <> show sn
+
+{- | Generate a v'SomeBoundedSNat' between the given bounds (inclusive).
+Uses 'Hedgehog.Range.linear' and shrinks to @lower@.
+
+For example, to bring into scope the constraints @(KnownNat n, 1 <= n, n <= 8)@:
+
+> SomeBoundedSNat (SNat :: SNat n) <- forAll (genSomeBoundedSNat @1 @8)
+-}
+genSomeBoundedSNat
+  :: (KnownNat lower, KnownNat upperInclusive, lower <= upperInclusive)
+  => Gen (SomeBoundedSNat lower upperInclusive)
+genSomeBoundedSNat = genSomeBoundedSNat# (\l -> Gen.integral . Range.linear l)
+
+-- | Like 'genSomeBoundedSNat' but does /not/ shrink.
+genSomeBoundedSNat_
+  :: (KnownNat lower, KnownNat upperInclusive, lower <= upperInclusive)
+  => Gen (SomeBoundedSNat lower upperInclusive)
+genSomeBoundedSNat_ = genSomeBoundedSNat# (\l -> Gen.integral_ . Range.linear l)
+
+{- | Generate a v'SomeBoundedSNat' between the given bounds (inclusive).
+To do so, uses a given term level generator.
+
+__NB__: Make sure the given generator respects the bounds that it is passed at
+the term level.
+-}
+genSomeBoundedSNat#
+  :: forall lower upperInclusive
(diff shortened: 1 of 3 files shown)


rowanG077 (approved):
Interesting does that work?

```
SomeBoundedSNat @n SNat <- forAll $ genSomeBoundedSNat @1 @8
```

I would expect you'd need a pattern match since now you introduce type variable `n` inside a body with a term level expression.

JvWesterveld:
Yes, with the `TypeAbstractions` extension. Alternatively use
```
SomeBoundedSNat (SNat :: SNat n) <- forAll $ genSomeBoundedSNat @1 @8
```

martijnbastiaan (approved):
🤩

DigitalBrains1:
Excellent idea of shoving those constraints into a GADT so you can type-check, thanks!

> A user might want to use e.g. `Range.constant` instead of the currently hardcoded `Range.linear`, but there's currently no `Range` argument as I'm not convinced it'd be an improvement.

So Tasty defines the concept of "test size" and starts with small tests working up to larger tests. Often when you are generating a number that influences the "size" of the test, you'd pick `Range.linear`, yet if the number had no effect on the size of the test, you'd regularly pick `Range.constant`. This "size" is a different dimension than the "shrinking" behaviour: test data generated with `Range.constant` can still shrink, and test data generated with `Range.linear` might not shrink. The shrinking is a property of the generator, whereas the range is an argument to the generator. So `Gen.integral (Range.constant 0 42)` will shrink, but `Gen.integral_ (Range.linear 0 42)` will not shrink.

Tasty will increase the size of the test until it finds a counterexample. Taking that counterexample, it will subsequently try to use shrinking to find a smaller counterexample. The two steps are sequential.

The current implementation scales the generated number with the size of the test and performs shrinking. Tests scale from `@lower` and shrink towards `@lower`. This is a common occurrence when the `SNat` influences the size of the test, but there are a lot more options: not scaling with size, not shrinking, different origin...

I'm in favour of adding the current version (although I'd add the sizing behaviour to the doc, currently it only documents shrinking), but I _do_ think having more control over the generated numbers is an improvement that we might add in a later PR. Perhaps make a variant where the generation of the number is left to the user, and making the `SNat` just checks that the provided number is within bounds and produces `SomeBoundedSNat`.

Alternatively, it could be modeled after `integral` and we could define a function that does shrink and a function that does not shrink with a `_` (underscore) appended to the name. However, after the number has been generated, it's set in stone. You can't use, for instance, generating `Range.constant` and then apply a transformation on the number to get the non-flat probability distribution that uses the search space better to find counterexamples. So that's why first generating the number, maybe tweak it some, and only then generating an `SNat` might be a better method.

DigitalBrains1:
@kloonbot run_ci 06a8118

JvWesterveld:
Right, you convinced me @DigitalBrains1. I added to the comment that it uses `Hedgehog.Range.linear` which implies the used sizing behaviour, added a non-shrinking variant `genSomeBoundedSNat_` and `genSomeBoundedSNat#` which needs to be passed a generator and allows customizing generation including sizing and shrinking behaviour.

Or do you think it's better to have the user run the generator themselves and pass the resulting value to `genSomeBoundedSNat#` instead of the generator? I.e. change the `(Natural -> Natural -> Gen Natural)` in the type to `Natural` or `Integral n => n`?

clash-prelude-hedgehog/src/Clash/Hedgehog/Promoted/Nat.hs
```diff
@@ -0,0 +1,81 @@
...
+
+instance Show (SomeBoundedSNat lower upperInclusive) where
+  show (SomeBoundedSNat sn@SNat) = show ''SomeBoundedSNat <> " " <> show sn
+
+{- | Generate a v'SomeBoundedSNat' between the given bounds (inclusive).
+Uses 'Hedgehog.Range.linear' and shrinks to @lower@.
+
+For example, to bring into scope the constraints @(KnownNat n, 1 <= n, n <= 8)@:
+
+>>> SomeBoundedSNat @n SNat <- Hedgehog.Gen.sample (genSomeBoundedSNat @1 @8)
```
JvWesterveld: I'm a little unhappy with the use of `Hedgehog.Gen.sample` here as it really shouldn't be used in Hedgehog tests. But just substituting `forAll` would probably fail `doctest-parallel`.

JvWesterveld:
I also added a `lower <= upperInclusive` constraint to all three functions.

clash-prelude-hedgehog/src/Clash/Hedgehog/Promoted/Nat.hs
martijnbastiaan: Yeah.. just replace it with a single `>` IMO. It's probably better to have a non-checked example than a misleading one.

martijnbastiaan:
@kloonbot run_ci a24a50734d63081caa76800152b85dec54682462

martijnbastiaan:
@JvWesterveld I've sent you an invitation to join the `Dev` team. That way you can push to branches directly and we don't need to invoke @kloonbot :).

JvWesterveld:
Thanks! I should probably still add a changelog entry, right?

martijnbastiaan:
Oh, woops, yes. Also we should probably split out the `*-hedgehog` packages.

JvWesterveld:
What would you like split out? Since `SNat` is currently defined in `Clash.Promoted.Nat` from clash-prelude, doesn't it makes sense for `SomeBoundedSNat` to be part of clash-prelude-hedgehog? Or is this looking forward to migrating to `GHC.TypeLits.SNat`?

martijnbastiaan:
Oh no sorry, that was just a loose comment on wanting to split out the packes from this repository, such that it can have its own changelog. Not an action point for this PR. Sorry, for the confusion.

martijnbastiaan:
@kloonbot run_ci 51184eaee1725c1b2c3b256e62d4aa9cced1a3b0

martijnbastiaan:
@JvWesterveld Can it be merged? :)

JvWesterveld:
@kloonbot run_ci 6df0a19ae80733bdce789f834c7dbad51f11602d

JvWesterveld:
@martijnbastiaan Yes!

PR #3186, opened 2026-04-12
CI: No Hackage candidate for published version

DigitalBrains1:
Since https://github.com/haskell/hackage-server/pull/1469 (Check candidate uploads against main index), Hackage no longer allows candidates for published versions.

But we use the `.ci/publish_sdist.sh` script from our stable branches to publish a candidate before we do the final release, so we do want it to run for nightlies on our stable branch. But we need to differentiate between runs that do not contain a new version and runs that do contain a new version.

So we do this check in the CI job, and exit with a special exit code to flag when we did not upload a candidate. This means normal nightlies will have an exclamation mark in an orange circle next to the `hackage-release-candidate`, indicating a failure that does not cause the pipeline as a whole to fail.

When the "special" nightly before a release is run, the developer inspecting the result should verify a candidate was published (check mark in green circle). In the usual case, where only one candidate will be uploaded before release, they will also necessarily notice that the candidate is simply not on Hackage if they forget to check the CI job's result.

Because we don't have control over the exit codes of any binaries we invoke, a trap handler is added in `.ci/functions.sh` that makes sure no other exits will have the exit code we reserve for allowed-to-fail. Currently, GitLab runners will not actually propagate the exit code of a failing step in the script, but this might very well change in an update. So we also proof our `.ci/setup.sh` script against such exits so they don't trigger unwantedly in the `hackage-release-candidate` job.

## Still TODO:

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,6 +93,12 @@ hackage-release-candidate:
     - if: $CI_PIPELINE_SOURCE == "schedule"
     - if: $CI_PIPELINE_SOURCE == "trigger"
 
+  allow_failure:
+    # .ci/publish_sdist.sh uses this magic exit code to signal allowed-to-fail.
+    # .ci/functions.sh defines a wrapper to prevent accidentally emitting this
+    # code.
+    exit_codes: 64
+
 # Release new version of Clash to Hackage
 hackage-release:
   extends: .hackage
--- a/.ci/functions.sh
+++ b/.ci/functions.sh
@@ -1,0 +1,10 @@
+exit_trap() {
+  local REAL_EXIT=$?
+
+  if [ "$REAL_EXIT" -eq 64 ]; then
+    echo "Process exited with status code 64, which we reserve to indicate" >&2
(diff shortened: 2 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -93,6 +93,9 @@ hackage-release-candidate:
     - if: $CI_PIPELINE_SOURCE == "schedule"
     - if: $CI_PIPELINE_SOURCE == "trigger"
 
+  allow_failure:
+    exit_codes: 64
```
martijnbastiaan, line 100: Could you comment what this magic exit code is? (Not in the PR, I mean as a comment in the YAML.)

martijnbastiaan (approved):

rowanG077 (approved):

.gitlab-ci.yml
rowanG077: Please reference the file where you conjure that one up. So it's not only explained but also easy to follow back to the source.

DigitalBrains1: That does make it more clear.

PR #3193, opened 2026-04-14
Add SaturatingNum instances for Clash.Num types

kleinreact:
The PR adds the missing `SaturatingNum` instances for `Erroring`, `Overflowing`, `Saturating`, `Wrapping`, and `Zeroing`.

## Still TODO:

### DIFF
--- a/clash-prelude/src/Clash/Num/Overflowing.hs
+++ b/clash-prelude/src/Clash/Num/Overflowing.hs
@@ -157,7 +157,9 @@ instance (Ord a, SaturatingNum a) => SaturatingNum (Overflowing a) where
     | otherwise
     = withOverflow (a || b)
    where
-    withOverflow = Overflowing $ satAdd mode x y
+    withOverflow =
+      let r = satAdd mode x y
+       in seq r $ Overflowing r
 
   {-# INLINE satSub #-}
   satSub mode (Overflowing x a) (Overflowing y b)
@@ -172,7 +174,9 @@ instance (Ord a, SaturatingNum a) => SaturatingNum (Overflowing a) where
     | otherwise
     = withOverflow (a || b)
    where
-    withOverflow = Overflowing (satSub mode x y)
+    withOverflow =
+      let r = satSub mode x y
+       in seq r $ Overflowing r
 
   {-# INLINE satMul #-}
   satMul mode (Overflowing x a) (Overflowing y b)
@@ -184,15 +188,21 @@ instance (Ord a, SaturatingNum a) => SaturatingNum (Overflowing a) where
     | otherwise
     = withOverflow (a || b)
    where
-    withOverflow = Overflowing (satMul mode x y)
+    withOverflow =
+      let r = satMul mode x y
+       in seq r $ Overflowing r
 
   {-# INLINE satSucc #-}
   satSucc mode (Overflowing x a) =
-    Overflowing (satSucc mode x) (a || x == maxBound)
+    let xSucc = satSucc mode x
+     in seq xSucc
+      $ Overflowing xSucc (a || x == maxBound)
 
   {-# INLINE satPred #-}
   satPred mode (Overflowing x a) =
-    Overflowing (satPred mode x) (a || x == minBound)
+    let xPred = satPred mode x
+     in seq xPred
+      $ Overflowing (satPred mode x) (a || x == minBound)
 
 instance (Enum a, Eq a, SaturatingNum a) => Enum (Overflowing a) where
   succ (Overflowing x a)
--- a/clash-prelude/src/Clash/Num/Overflowing.hs
+++ b/clash-prelude/src/Clash/Num/Overflowing.hs
@@ -143,6 +143,10 @@ instance (Bounded a) => Bounded (Overflowing a) where
 -- for the overflow check (depending on the utilized 'SaturationMode'
 -- or inner type), which is not guaranteed to be optimized away at a
 -- later synthesis stage.
+--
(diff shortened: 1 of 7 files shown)


clash-prelude/src/Clash/Num/Overflowing.hs
```diff
@@ -139,6 +139,57 @@ instance (Bounded a) => Bounded (Overflowing a) where
...
+    = withOverflow True
+
+    | y < 0
+    , x < satSub SatWrap minBound y
+    = withOverflow True
+
+    | otherwise
+    = withOverflow (a || b)
+   where
+    withOverflow = Overflowing $ satAdd mode x y
```
martijnbastiaan, line 160: I'm sure this works, but doesn't this mean that the underlying operation (`satAdd mode x y`) does an overflow check (depending on the mode of course) _and_ it happens here?

kleinreact: This is almost the same code as for the `Num` instance, only that instead of `SatWrap` the mode supplied by the user is used. My understanding is that the overflow shall be tracked independently of the mode selected by the user.

martijnbastiaan: Right, yes. I guess to put it another way, I'm trying to figure out whether this would (essentially) duplicate hardware. As in `satAdd mode x y` would (potentially) do an overflow check and then on top of that this logic would do it too. Is that true?

I'm also not saying I'd reject the PR based on it, but if it's true it's worth a comment in the instance.

kleinreact: Right, I get you point, but isn't this also the case for the `Num` instance then, e.g., the `SaturatingNum` instance of `Index` does an overflow check in case of `SatWrap` and the `Num` instance of `Overflowing` then does the same check again to track the bit. So if you say that's a problem to be resolved, then it requires some updates in more than just this case, I guess.

On the other hand, it is literally the same check, so I am pretty sure that the downstream tools will be able to optimize that away, because it resolves to just a simple Boolean gate optimization in the end.

martijnbastiaan: Oh yeah, I'm not saying resolve it, but I think it makes sense to add a Haddock comment on the instance and possibly even open an `enhancement` ticket. I'm sure there is a way we could do an API refactor to make the check shared, but that's way out-of-scope for this PR of course. (I'm way more pessimistic about downstream tools optimizations haha.)

kleinreact: I added said Haddock comment. The `enhancement` ticket can be added as soon as this PR has been merged.

martijnbastiaan (commented):
Could you please add unit tests for `SaturatingNum (Overflowing a)`? It's non-obvious to me that this is correct. No need to do that for the instances that just forward.

clash-prelude/src/Clash/Num/Overflowing.hs
martijnbastiaan: Great, thanks!

kleinreact:
> Could you please add unit tests for SaturatingNum (Overflowing a)?

Done. There indeed was an evaluation w.r.t. error propagation missing, which should be fixed now as well.

martijnbastiaan (approved):
Nice!

martijnbastiaan:
@rowanG077 What's the policy on backporting this? Strictly it's an API change, but in the past we've allowed backports like this.

DigitalBrains1:
Could you please add some additional documentation to `SaturatingNum (Overflowing ...`?

Using `satAdd SatBound` still reports overflow when `SatBound` triggers. This definitely sounds correct, but if I were looking at the instance, I'd think "Well `SatBound` bounds the result, so does that count as an overflow or not?". I'd very much appreciate some exposition to take away doubts.

```haskell
>>> satAdd SatBound 1 1 :: Overflowing (Unsigned 1)
Overflowing {fromOverflowing = 1, hasOverflowed = True}
```

DigitalBrains1:
> @rowanG077 What's the policy on backporting this? Strictly it's an API change, but in the past we've allowed backports like this.

Not Rowan, but... This is an API addition per the PVP, which is allowed. I would backport it in this case.

PVP:
> _Non-breaking change._ Otherwise, if only new bindings, types, classes, non-orphan instances or modules (but see below) were added to the interface, then A.B **MAY** remain the same but the new C **MUST** be greater than the old C.

DigitalBrains1:
I was thinking along these lines:

```diff
--- a/clash-prelude/src/Clash/Num/Overflowing.hs
+++ b/clash-prelude/src/Clash/Num/Overflowing.hs
@@ -143,6 +143,10 @@ instance (Bounded a) => Bounded (Overflowing a) where
 -- for the overflow check (depending on the utilized 'SaturationMode'
 -- or inner type), which is not guaranteed to be optimized away at a
 -- later synthesis stage.
+--
+-- The 'hasOverflowed' flag will still be set even when the 'SaturatingNum'
+-- operation has prevented the overflow:
+-- @hasOverflowed (satSucc SatBound maxBound) == True@
 instance (Ord a, SaturatingNum a) => SaturatingNum (Overflowing a) where
   {-# INLINE satAdd #-}
   satAdd mode (Overflowing x a) (Overflowing y b)
```

@kleinreact, if you want, I can push that change into this branch so you can merge it without effort.

kleinreact:
> Could you please add some additional documentation to SaturatingNum (Overflowing ...?

Done. @DigitalBrains1 Thanks for the suggestion.

PR #3194, opened 2026-04-14
Add `tested-with` for packages on Hackage

kleinreact:
The PR adds `tested-with` entries for packages on Hackage ~~and moves `CHANGELOG.md` to `extra-doc-files` as recommended by `cabal check`~~.

## Still TODO:

  - ~~[ ] Write a changelog entry (see changelog/README.md)~~
  - ~~[ ] Check copyright notices are up to date in edited files~~

### DIFF
--- a/clash-ghc/clash-ghc.cabal
+++ b/clash-ghc/clash-ghc.cabal
@@ -42,6 +42,10 @@ Copyright:            Copyright © 2012-2016, University of Twente,
                                   2017-2019, QBayLogic B.V., Google Inc.
 Category:             Hardware
 Build-type:           Simple
+tested-with:          GHC == 9.6.7,
+                      GHC == 9.8.4,
+                      GHC == 9.10.3,
+                      GHC == 9.12.4
 
 Extra-source-files:   README.md,
                       CHANGELOG.md,
--- a/clash-lib-hedgehog/clash-lib-hedgehog.cabal
+++ b/clash-lib-hedgehog/clash-lib-hedgehog.cabal
@@ -12,6 +12,10 @@ maintainer:         devops@qbaylogic.com
 copyright:          Copyright © 2021-2025, QBayLogic B.V.
 category:           Hardware
 build-type:         Simple
+tested-with:        GHC == 9.6.7,
+                    GHC == 9.8.4,
+                    GHC == 9.10.3,
+                    GHC == 9.12.4
 
 common basic-config
   default-language: Haskell2010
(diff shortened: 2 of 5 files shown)


martijnbastiaan (approved):
Hmm, what makes a `README.md` a source file but a `CHANGELOG.md` a doc file? :-) LGTM anyhow.

clash-ghc/clash-ghc.cabal
```diff
@@ -42,11 +42,15 @@ Copyright:            Copyright © 2012-2016, University of Twente,
...
 Build-type:           Simple
+tested-with:          GHC == 9.6.7
+                    , GHC == 9.8.4
+                    , GHC == 9.10.3
 
-Extra-source-files:   README.md,
-                      CHANGELOG.md,
-                      LICENSE_GHC,
-                      cbits/PosixSource.h
+Extra-source-files:   README.md
```
rowanG077: If we are changing formatting for these files we should be moving towards the [cabal-gild](https://hackage.haskell.org/package/cabal-gild) style.

rowanG077 (changes requested):

kleinreact:
> Hmm, what makes a README.md a source file but a CHANGELOG.md a doc file?

According to the documentation for `extra-doc-files`: _"and also copied to the html directory when Haddock documentation is generated"_. 

Also it is mentioned that `extra-source-files` are passed to `runhaskell Setup.hs sdist`, but `extra-doc-files` are not? So maybe that can cause problems in some specific cases.

> If we are changing formatting for these files we should be moving towards the [cabal-gild](https://hackage.haskell.org/package/cabal-gild) style.

As far as I can see there are more inconsistencies among our cabal files atm. Wouldn't it make more sense to update all of them in w.r.t. that in a separate PR then?

clash-ghc/clash-ghc.cabal
martijnbastiaan: Would it make sense to just do this in one go in a PR enforcing `cabal-gild`? Just so Felix doesn't have to actually read how the cabal-gild style works.

kleinreact:
I wonder whether [this](https://github.com/clash-lang/clash-compiler/actions/runs/24385541659/job/71218415785?pr=3194#step:4:5017) is a Nix bug. It looks to me like Nix is not able to resolve symbolic links in `extra-doc-files` but only in `extra-source-files` :upside_down_face:.

Shall I revert this change then and move it back to `extra-source-files` for the moment?

rowanG077:
> As far as I can see there are more inconsistencies among our cabal files atm. Wouldn't it make more sense to update all of them in w.r.t. that in a separate PR then?

Yes I agree!

kleinreact:
I reverted the changes w.r.t. `extra-doc-files`. @rowanG077 feel free to merge this if you consider it to be in the right shape.

rowanG077 (approved):

PR #3195, opened 2026-04-14
ci: use macos-latest

rowanG077:
Fixes GHC error https://github.com/clash-lang/clash-compiler/actions/runs/24382799695/job/71209826233?pr=3193

PR #3197, opened 2026-04-14
ci: use macos-26

rowanG077:
Should actually have been done by: https://github.com/clash-lang/clash-compiler/pull/3195

But `macos-latest` is not actually "latest", as that still refers to macos-15...


martijnbastiaan (approved):
SMH..

PR #3199, opened 2026-04-14
inline `localparam` in higher Vec higher order blackboxes

lmbollen:
What:
I inlined the `localparam` declarations in the higher order function blackboxes for Vectors.

Why:
The `localparam` usage causes problems where Vivado runs out of memory, takes ages to synthesize and straight up segfaults. We've found that with this patch vivado passes on the development branch of clash-compiler.

## Still TODO:


martijnbastiaan:
@lmbollen @rowanG077 Unfortunately (?) I cannot reproduce the Vivado failures we saw earlier. That is, if I compile with the latest `master`, everything works just fine over at `bittide-hardware`. Sorry for the noise everyone.

PR #3202, opened 2026-04-14
Fix clash-testsuite nix packaging

rowanG077:
This will allows running the clash-testsuite via the flake. Which is currently FUBAR.

TODO:

- [x] Look into why verilator output is different.

### DIFF
--- a/.ci/docker/Dockerfile
+++ b/.ci/docker/Dockerfile
@@ -49,14 +49,14 @@ RUN curl -L "https://github.com/steveicarus/iverilog/archive/${iverilog_version}
 
 FROM builder AS build-symbiyosys
 
-ARG DEPS_YOSYS="libffi-dev libreadline-dev pkg-config tcl-dev"
+ARG DEPS_YOSYS="libffi-dev libfl-dev libreadline-dev pkg-config tcl-dev"
 ARG DEPS_Z3="python3 python3-distutils"
 RUN apt-get install -y --no-install-recommends $DEPS_YOSYS $DEPS_Z3
 
-ARG yosys_version="8cfed1a97957e4c096d1e0a0304d978bcb27e116"
-RUN git clone https://github.com/YosysHQ/yosys.git yosys \
+ARG yosys_version="v0.64"
+RUN git clone --branch $yosys_version --depth 1 https://github.com/YosysHQ/yosys.git yosys \
  && cd yosys \
- && git checkout $yosys_version \
+ && git submodule update --init --recursive \
  && make PREFIX=$PREFIX -j$(nproc) \
  && make PREFIX=$PREFIX install \
  && cd .. \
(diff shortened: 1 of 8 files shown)


rowanG077:
Verilator bug: https://github.com/verilator/verilator/issues/7427

upstream PR: https://github.com/verilator/verilator/pull/7429

Edit: PR has been merged

rowanG077:
I vendored the patch. It might take a while for the full verilator fix merge -> nixpkgs upstream merge cycle. Till then I think it's best to vendor the patch.

DigitalBrains1:
You can find the deployment of new Docker images in this branch: https://github.com/clash-lang/clash-compiler/tree/peter/clash-testsuite-nix

Also, something different. Could you please split the `nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite` over a whole bunch of jobs like we do on GitLab CI? That way it'll run so much faster. And please run the Vivado tests only in nightlies, Vivado is sllloooowwwww. Also, these days dependencies of tasty tests actually works in `clash-testsuite`, so we can cut down on those nightly Vivado jobs, which currently duplicate a whole lot of tests from the non-Vivado jobs even though that's totally unnecessary.

I think the following would do nicely:
```shell
nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite -p .VHDL. --hide-successes --no-vivado
nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite -p .Verilog. --hide-successes --no-vivado
nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite -p .SystemVerilog. --hide-successes --no-vivado
```

and then in nightlies (which we should set up once we run `clash-testsuite` only in Nix):
```shell
nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite -p .VHDL. -p .Vivado. --hide-successes
nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite -p .Verilog. -p .Vivado. --hide-successes
nix run -L .#clashPackages-${{matrix.ghc}}.clash-testsuite -p .SystemVerilog. -p .Vivado. --hide-successes
```

[edit]
If I'm wrangling the names correctly, to run Vivado we need to run 997 test steps (just `clash (gen)` and `Vivado.sim [...]`), but currently on GitLab CI nightlies we actually run exactly 696 more: either `clashLibTest`, `outputTest` or `clash (gen)` for stuff that will not actually be run in Vivado. This is because we don't use `-p .Vivado.` to limit the tests because that didn't work before we had `sequentialTestGroup` (now `dependentTestGroup`).

It's probably not actually taking much extra time since Vivado severely dominates in run time. But it also doesn't serve any purpose whatsoever.
[/edit]

DigitalBrains1:
I also don't think `nix/overlay.nix` should force the `--no-modelsim` and `--no-vivado` flags for `clash-testsuite`. I do understand it won't work everywhere out of the box, but this outright prevents people from running these tests, which I think is undesirable.

jaschutte (approved):
Lgtm! Thanks you very much, the testsuite was a big headache for me haha. 

Same notes as Peter actually regarding removing the hardcoded `--no-model-sim` and `--no-vivado`.

Also the CI job could be a separate PR for when the nightlies get converted from Gitlab to Github.

PR #3211, opened 2026-04-21
Release v1.10.0

DigitalBrains1:
## Still TODO:


.mergify.yml
```diff
@@ -13,3 +13,15 @@ pull_request_rules:
         merge_conflict_style: diff3
         branches:
           - "1.8"
+
+  - name: backport patches to 1.10
```
rowanG077, line 17: I presume the 1.8 one will be removed on a follow up PR?

rowanG077 (approved):
Awesome!

.mergify.yml
DigitalBrains1: I suppose so. It's not important at all; up to a year ago, it still mentioned 1.2 as a backport target. We should keep 1.8 around for a bit just in case; if you want to formalise it, I'd say remove it when we release 1.12 but I think the overhead for this stuff is exceptionally low.

PR #3212, opened 2026-04-22
Remove GHC mmap crash workaround from Nix

DigitalBrains1:
GHC 9.6 and newer should no longer be affected by this. In CI, only the Nix jobs still carry this option. Let's remove it and once some time has passed without the bug happening, we can remove all the plumbing for it.

I did see this error pop up not too long ago, but most likely that was in Clash 1.8.

## Still TODO:

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -132,10 +132,9 @@ let
       clash-ghc =
         let
           unmodified =
-            hprev.callCabal2nixWithOptions
+            hprev.callCabal2nix
               "clash-ghc"
-              ../clash-ghc
-              "--flag workaround-ghc-mmap-crash" {
+              ../clash-ghc {
               inherit (hfinal) clash-lib clash-prelude;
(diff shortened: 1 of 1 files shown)


martijnbastiaan:
If I search for `workaround-ghc-mmap-crash` in this branch, I get a bunch of hits. Is that intended? I guess that's what you mean with plumbing? :)

DigitalBrains1:
Yes, I meant that the only place in CI that we still _use_ the flag is in our Nix CI, so that's where I remove its _use_. Once we determine that the flag doesn't serve a purpose, we can remove all the places where its functionality is implemented.

martijnbastiaan (approved):

DigitalBrains1:
It occurred to me that due to `tasty-flaky`, I don't think we'll ever see crashes inside `clash-testsuite` test steps anymore, even if the bug still occurs...

PR #3219, opened 2026-04-26
Add support for `ghdl-{gcc,mcode}` in `clash-testsuite`

martijnbastiaan:
I've tested this with all GHDLs supplied in the Ubuntu 22.04 repository, they all pass :). I've also verified that it fails when deliberately messing up a few tests.

## Still TODO:

### DIFF
--- a/tests/src/Test/Tasty/Ghdl.hs
+++ b/tests/src/Test/Tasty/Ghdl.hs
@@ -1,20 +1,26 @@
 {-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE QuasiQuotes #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeApplications #-}
 
 module Test.Tasty.Ghdl where
 
 import           Clash.Driver.Manifest     (Manifest(..), manifestFilename)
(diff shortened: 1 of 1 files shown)


rowanG077:
Why do we want this extra test surface?

martijnbastiaan:
We don't necessarily, but it has been an irritation point of mine that:

1. You get an error that `ghdl` isn't installed on your system.
2. `sudo apt install ghdl`
3. Still broken (with a terrible error message)

Note that it doesn't add anything to CI, it just gives people the option to run with the GCC/MCODE backend.

martijnbastiaan:
Of course any backend should yield the exact same result. Anything else would be a GHDL bug.

rowanG077:
But then wouldn't it be much simpler to just try `ghdl` and then fallback to trying `ghdl-llvm`, `ghdl-gcc` and `ghdl-mcode`? If someone wants to force a specific executable they can use PATH to set `ghdl`

Edit: Actually I see there are difference how `ghdl`is called depending on backend. So it can be that simple... too bad

rowanG077 (approved):

martijnbastiaan:
Yeah it's annoying :-(. Thanks for the review!

PR #3222, opened 2026-04-26
Prevent duplicate `attribute`s in VHDL backend

### LINKED ISSUE #3218: Clash can generate duplicate `attribute X : string` in VHDL
For example:

```haskell
module T3218 where

import Clash.Prelude
import Clash.Annotations.SynthesisAttributes

topEntity ::
  Clock System ->
  Reset System ->
  Enable System ->
  Signal System (BitVector 8) ->
  Signal System (BitVector 8) `Annotate` 'StringAttr "foo" "B1"
topEntity clk rst ena i = leds
 where
  leds :: Signal System (BitVector 8) `Annotate` 'StringAttr "foo" "B2"
  leds = withClockResetEnable clk rst ena $ register 0 (liftA2 xor i next)

  next = withClockResetEnable clk rst ena $ register 0 leds
{-# OPAQUE topEntity #-}
```

Then:

```
martijn@deskei:~/code/clash-compiler$ rg "attribute foo : string;" vhdl/3218.topEntity/topEntity.vhdl 
20:  attribute foo : string;
29:  attribute foo : string;
```

Which crashes GHDL:

```
/tmp/clash-test_T3218-f0ecee5e79ee9164/ghdl-topEntity/T3218.topEntity/topEntity.vhdl:29:13:error: identifier "foo already used for a declaration
  attribute foo : string;
                                ^
/tmp/clash-test_T3218-f0ecee5e79ee9164/ghdl-topEntity/T3218.topEntity/topEntity.vhdl:20:13:error: previous declaration: attribute "foo"
```

Note that this is not _quite_ the same as https://github.com/clash-lang/clash-compiler/issues/3217, although it has the same reproducer. For this issue, it's really about Clash generating invalid HDL.

martijnbastiaan:
Fixes #3218

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Backend/VHDL.hs
+++ b/clash-lib/src/Clash/Backend/VHDL.hs
@@ -970,14 +970,18 @@ entity c = do
 architecture :: Component -> VHDLM Doc
 architecture c = do {
   ; syn <- Ap hdlSyn
-  ; let attrs = case syn of
+  -- The entity declarative region is implicitly visible inside the
+  -- architecture, so re-declaring `attribute X : T;` here for an `X` already
+  -- declared by the entity produces a duplicate declaration that GHDL rejects.
+  -- Skip declarations for any (name, type) pair already emitted by the entity.
+  ; let (entityDeclared, attrs) = case syn of
                   -- See: [Note] Hack entity attributes in architecture
-                  Other -> declAttrs
-                  _     -> inputAttrs ++ outputAttrs ++ declAttrs
+                  Other -> (entityDeclaredAttrTypes c, declAttrs)
+                  _     -> (mempty, inputAttrs ++ outputAttrs ++ declAttrs)
   ; nest 2
       (("architecture structural of" <+> pretty (componentName c) <+> "is" <> line <>
        decls (declarations c)) <> line <>
-       if null attrs then emptyDoc else line <> line <> renderAttrs (TextS.pack "signal") attrs) <> line <>
+       if null attrs then emptyDoc else line <> line <> renderAttrsSkippingDecls (TextS.pack "signal") entityDeclared attrs) <> line <>
     nest 2
       ("begin" <> line <>
        insts (declarations c)) <> line <>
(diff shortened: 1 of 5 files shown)


clash-lib/src/Clash/Backend/VHDL.hs
```diff
@@ -1054,15 +1066,28 @@ renderAttrs
...
+renderAttrsSkippingDecls what skipDecls (attrMap -> attrs) =
   vcat $ sequence $ intersperse " " $ map renderAttrGroup (HashMap.toList attrs)
  where
   renderAttrGroup
     :: (TextS.Text, (TextS.Text, [(TextS.Text, TextS.Text)]))
     -> VHDLM Doc
   renderAttrGroup (attrname, (typ, elems)) =
-    ("attribute" <+> stringS attrname <+> colon <+> stringS typ <> semi)
-    <> line <>
+    (if HashMap.lookup attrname skipDecls == Just typ
```
rowanG077, line 1087: I'm not sure how GHDL works. But I'd expect it attrs with the same name are the problem and not just the same name + type? So for example declaring an `IntegerAttr` + `StringAttr` is still problematic. 

Edit: Yes confirmed that same attr name + different type errors still.

rowanG077 (changes requested):

clash-lib/src/Clash/Backend/VHDL.hs
martijnbastiaan: Yes, but, there is no coarse of action for Clash here. The user really requested to generate two (conflicting) synthesis attributes.

martijnbastiaan: I guess Clash could throw an error, but I'm almost thinking the error from GHDL would be more informative / provide more context.

rowanG077: Then shouldn't we error out if the duplicate attribute values are different as well?  Why does a different type mean it's conflicting, but two different values do not? I never used these attributes so it might be totally sensible.

rowanG077: I mean this is genuinely ambiguous. I'd rather clash at least warns about this, GHDL errors on it but other tools may not be so nice to users, worst case picking one attribute over the others.

martijnbastiaan: > Then shouldn't we error out if the duplicate attribute values are different as well? Why does a different type mean it's conflicting, but two different values do not? I never used these attributes so it might be totally sensible.

No, duplicate/different attribute values are fine. It's a bit like saying "there is a property KEEP of type bool" and then saying "foo's KEEP is true", "bar's KEEP is false", etc. The problem this PR tackles is just that there are two places you can say "there is a property KEEP of type bool" with one scope being available in the other scope, but not the other way around.

> I'd rather clash at least warns about this, GHDL errors on it but other tools may not be so nice to users, worst case picking one attribute over the others.

Fair enough, let's make it an error then! Conflicting types is always wrong AFAIK.

rowanG077: Ah yeah, you are right. I did not realize this attribute value was on a different thing at the netlist stage. I assumed normalization collapsed both attributes to be on the same thing, which is not the case.

rowanG077 (approved):

PR #3227, opened 2026-04-27
Add `--auto-detect-tools` to `clash-testsuite`

martijnbastiaan:
QoL change to the testsuite: I virtually never have ModelSim and Vivado in scope when running the testsuite, nor is it super interesting for me to test with them. I therefore always append `--no-vivado` and `--no-modelsim` to my invocation, with more flags depending on the system. After this patch, passing `--auto-detect-tools` to the test suite will make it probe the system and generate the right `--no-tool` flags automatically.

## Still TODO:

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -5,6 +5,7 @@ module Main (main) where
 
 import qualified Clash.Util.Interpolate    as I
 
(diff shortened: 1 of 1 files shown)


rowanG077 (approved):
LGTM, very handy!

PR #3235, opened 2026-04-29
Account for `ErrorCall` for `shift`/`rotate`

### LINKED ISSUE #3234: Evaluator doesn't account for errors in `shiftL#`
See:

https://github.com/clash-lang/clash-compiler/blob/ea6ca83346d495ca66c803c4b5ca95ef96d948e1/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs#L2923-L2929

Reproducer:

```haskell
import Clash.Prelude

topEntity :: BitVector 15
topEntity = (1 :: BitVector 15) `shiftL` negate 10
{-# OPAQUE topEntity #-}
```

There is precedent on how to handle this:

https://github.com/clash-lang/clash-compiler/blob/ea6ca83346d495ca66c803c4b5ca95ef96d948e1/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs#L4717-L4731

martijnbastiaan:
Fixes #3234

I will say this does call into question whether we should even do `error` (not `errorX`) in these functions. I guess that's for another issue?

## Still TODO:


christiaanb (commented):
Why not change the definition of `reduce` to go from just catching `XException` to catching all exceptions?

martijnbastiaan:
Mostly two things:

* I'm not sure what the behavior is with asynchronous exceptions, e.g. `SIGTERM`. Would we catch that and produce `x`s?
* It seems to me that the cost of getting it wrong is way too high. For better or for worse, we use `error` quite a bit around `clash-lib`. The more precise we are about where we expect an `ErrorCall` to be, the better. (IMO, I think the `catchErrorCall` should be even closer, e.g., `op !u !i _ = toInteger (catchErrorCall (Unsigned.rotateR# u i))`).

rowanG077:
In this case I think it's cleaner to use @christiaanb original patch of just changing the error calls in prelude to `errorX`. And I do also agree with @martijnbastiaan that unilaterally catching on `reduce` will hide logic errors in our own code that have nothing to do with evaluations at all. So I'm much more on-board with the proposal that we catch the exception much closer to the source.

martijnbastiaan:
Okay, is the change from `ErrorCall` to `XException` something we consider an API change? If yes, I think we should merge this, backport it, then remove `catchErrorCall` again and change `error` to `errorX` + changelog entry. (And _not_ backport that.)

rowanG077:
@martijnbastiaan That sounds good.

BTW this is a problem in more evaluator code, a non-exhaustive list:

  - `GHC.Prim.chr#`: uses `chr (fromInteger i)`.
  - A ton `quot`/`rem` variants don't catch divide by zero.
  - `(^)` reductions: negative exponents can throw.

martijnbastiaan:
Yeah fair. I'll set an AI on it and amend it to this PR.

martijnbastiaan:
I'm not entirely convinced that Clash actually sees all the cases now, I'll revert to draft and find out.

martijnbastiaan:
@rowanG077 It's good to go now :saluting_face:

martijnbastiaan:
Note that the test isn't exactly perfect, it critically relies on GHC not constant folding. This seems to not trigger with `-O0`. Clash itself reports:

```
Warning: caught exception: "'shiftL' undefined for negative number: -1" while trying to evaluate: 
shiftL# (removedArg @Natural) (fromInteger# @15 15 0 1) (I# (-1#))

Warning: caught exception: "'shiftR' undefined for negative number: -1" while trying to evaluate: 
shiftR# (removedArg @Natural) (fromInteger# @15 15 0 1) (I# (-1#))

Warning: caught exception: "'rotateL' undefined for negative number: -1" while trying to evaluate: 
rotateL# (removedArg @Natural) (fromInteger# @15 15 0 1) (I# (-1#))

Warning: caught exception: "'rotateR' undefined for negative number: -1" while trying to evaluate: 
rotateR# (removedArg @Natural) (fromInteger# @15 15 0 1) (I# (-1#))

Warning: caught exception: "'shiftL' undefined for negative number: -1" while trying to evaluate: 
shiftL# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'shiftR' undefined for negative number: -1" while trying to evaluate: 
shiftR# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'rotateL undefined for negative number: -1" while trying to evaluate: 
rotateL# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'rotateR' undefined for negative number: -1" while trying to evaluate: 
rotateR# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'shiftL' undefined for negative number: -1" while trying to evaluate: 
shiftL# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'shiftR' undefined for negative number: -1" while trying to evaluate: 
shiftR# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'rotateL' undefined for negative number: -1" while trying to evaluate: 
rotateL# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "'rotateR' undefined for negative number: -1" while trying to evaluate: 
rotateR# (removedArg @Natural) (fromInteger# @15 15 1) (I# (-1#))

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt8# 1#8 0#8

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt8# 1#8 0#8

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt8# 1#8 0#8

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt8# 1#8 0#8

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt16# 1#16 0#16

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt16# 1#16 0#16

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt16# 1#16 0#16

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt16# 1#16 0#16

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt32# 1#32 0#32

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt32# 1#32 0#32

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt32# 1#32 0#32

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt32# 1#32 0#32

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotInt64# 1#64 0#64

Warning: caught exception: "divide by zero" while trying to evaluate: 
remInt64# 1#64 0#64

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord8# 1##8 0##8

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord8# 1##8 0##8

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord8# 1##8 0##8

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord8# 1##8 0##8

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord16# 1##16 0##16

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord16# 1##16 0##16

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord16# 1##16 0##16

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord16# 1##16 0##16

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord32# 1##32 0##32

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord32# 1##32 0##32

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord32# 1##32 0##32

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord32# 1##32 0##32

Warning: caught exception: "divide by zero" while trying to evaluate: 
quotWord64# 1##64 0##64

Warning: caught exception: "divide by zero" while trying to evaluate: 
remWord64# 1##64 0##64
```

So it's okay for now, but if GHC ever _does_ constant fold, I don't think we'd see it.

rowanG077 (approved):
Awesome!

Edit: Oh interesting. What does GHC constant fold it to? I would expect a div by zero is a side effect and shouldn't be constant folded away. Or does it remove the `remWord64#` entirely from the core and replace it with the exception?

martijnbastiaan:
Oh huh, I would have expected it to replace it with an error call, but it doesn't!

```haskell
{-# LANGUAGE MagicHash #-}

import GHC.Exts (quotWord64#, wordToWord64#)
import GHC.Word (Word64(W64#), Word(W#))

w64 :: Word -> Word64
w64 (W# w) = W64# (wordToWord64# w)

x :: Word64
x = W64# (quotWord64# (wordToWord64# 1##) (wordToWord64# 0##))
{-# OPAQUE x #-}

main :: IO ()
main = print x
```

```
$ reset && ghc -O2 test.hs -ddump-simpl -fforce-recomp
[..]
-- RHS size: {terms: 7, types: 1, coercions: 0, joins: 0/0}
x [InlPrag=OPAQUE] :: Word64
[GblId]
x = case quotWord64# 1#Word64 0#Word64 of ds_dZ8 [OS=OneShot]
    { __DEFAULT ->
    GHC.Internal.Word.W64# ds_dZ8
    }
[..]
```

DigitalBrains1:
> I will say this does call into question whether we should even do `error` (not `errorX`) in these functions.

> In this case I think it's cleaner to use @christiaanb original patch of just changing the error calls in prelude to `errorX`.

I suspect there's been some discussion about this outside GitHub because I'm not really hearing any arguments in favour of against. Could you provide some context please?

The major difference between `error` and `errorX` is whether Clash should error out during Clash simulation or continue running. Did someone come with a compelling case where Clash simulation should rightfully continue even when a shift with a  negative amount is being evaluated? Maybe we could add the dual to `xToError`, a function which converts `ErrorCall` to `XException`. That still leaves all other exceptions, though; it might not make much sense. We could extend it to all exceptions that we currently render as an X in hardware. Other exceptions would probably lead to failing HDL generation anyway?

DigitalBrains1:
I think the messages in this PR deal with two related but distinct issues:

- Which exceptions we should throw in primitives
- Which exceptions we should handle in the primitive evaluator

My previous message was about the former, but we can make a related argument for the latter.

Would it be a good idea if the primitive evaluator in `clash-lib` catches all the specific exceptions that we currently turn into an X during HDL generation? That way, these X's would probably propagate the same whether we did constant evaluation or not: in both cases it's an X in HDL, either passed to HDL functions or reduced to a constant.

martijnbastiaan:
As for what error should be thrown in our primitives: I don't think the _why_ has been discussed (here or elsewhere). For me, we should use as much `XException`s as possible, as they can be `seqX`'d and therefore be used in `registers` and the like. The fact that an `XException` makes it way into a register does not invalidate a design, IMO. This is going to be tackled in a future PR:

>> Okay, is the change from `ErrorCall` to `XException` something we consider an API change? If yes, I think we should merge this, backport it, then remove `catchErrorCall` again and change `error` to `errorX` + changelog entry. (And _not_ backport that.)
>
> @martijnbastiaan That sounds good.

As for:

> Would it be a good idea if the primitive evaluator in clash-lib catches all the specific exceptions that we currently turn into an X during HDL generation?

I don't believe so, because `reduce` can't distinguish between an `ErrorCall` that was thrown by an evaluated primitive or by - say - a panic on a bit of unexpected Core. This is what I meant with my second point [here](https://github.com/clash-lang/clash-compiler/pull/3235#issuecomment-4350260346).

rowanG077:
The comment I was referring to by @christiaanb was on slack. But it basically just showed the code for changing the `error` to `errorX`. There was no further discussion if that is the real fix. As the point brought up by @martijnbastiaan that this would be an API change(not backportable) convinced me that in this PR we need to catch the `ErrorCall` anyway. Any discussion beyond that can go into a new issue/PR that might want to change these `error` calls to `errorX` and thus I didn't feel like it's productive to discuss that right now.

DigitalBrains1:
> Any discussion beyond that can go into a new issue/PR that might want to change these `error` calls to `errorX` and thus I didn't feel like it's productive to discuss that right now.

This PR didn't discuss those future changes in the format "we should discuss next steps". It just said "in the next PR we should change `error` to `errorX`." which was replied to with "That sounds good."

I would have appreciated a note in the form of "we should invite some broader discussion", then I just wouldn't have commented. Now I just read "Let's change `error`" "That sounds good". This to me very much sounded like I would one day open my GitHub notifications and find that PR opened, approved and merged without any discussion. So I spoke up.

martijnbastiaan:
FWIW, I thought it was pretty uncontroversial because @christiaanb suggested it and @rowanG077 agreed without discussion -- and I also thought it was "the obvious thing to do". It wasn't done with the intention of bypassing discussion. I'll report it as an issue.

PR #3244, opened 2026-05-01
Use better RTS options for clash executable

rowanG077:
Remove the RTS hooks that are not being used now anyway due to GHC churn. Instead set the flags in the executables themselves. GHC has since moved to the same system. These flags give a performance increase of ~25%.

### Original PR message so the discussion below makes sense:

The hook that existed was no longer called as the symbol changed. I fixed that, and added an explicit check on clash startup that this hook was called so we are no longer caught with our pants down.

I also added the equivalent of the `-A128M` option as that speeds up normalization considerably. Consequently I removed this flag from the rts options of the clash executables. This makes sure our users can still override it but we provide a sane default.

## Still TODO:


changelog/2026-05-01T14_08_40+02_00_clash_default_allocation_area
```diff
@@ -0,0 +1 @@
+CHANGED: Clash now sets its default RTS allocation area at startup instead of relying on Cabal `-with-rtsopts=-A...` options.
```
martijnbastiaan: Could you document you set it to `A128` and its consequences? (A 10-30% speedup.)

martijnbastiaan (approved):
Clever FFI stuff, I wouldn't have thought of that!

changelog/2026-05-01T14_08_40+02_00_clash_default_allocation_area
martijnbastiaan: Also, you do pay for it in (slightly?) higher memory usage. Not sure if that's significant enough to mention, but it might be worth to do so anyway if people start running out of memory all of a sudden.

rowanG077:
It seems the original `hschooks.c` was copied ages ago from GHC. Since then GHC themselves no longer use a hook and instead just build GHC with the right flags. They do this because the linking was a pain in the ass and breaking in a lot of cases. I don't want repeat the same mistakes the GHC devs have already found out. 

So I propose we simply set the flags in our `clash` and `clashi` build exes, and add the flags to the starter projects executables too.  

It's also interesting that with the hook enabled adding the equivalent `-A128M` results in a performance regression.

These are my (unscientific results) by synthesizing bittides `softUgnDemoTest` with `shake softUgnDemoTest:hdl`:

```
hook enabled, only maxStkSize override:               Total compilation took 13m43s
hook enabled, only idleGCDelayTime override:          Total compilation took 13m42s
hook disabled:                                        Total compilation took 13m37s
hook enabled, only minAllocAreaSize override:         Total compilation took 12m15s
hook enabled, only heapSizeSuggestionAuto override:   Total compilation took 11m37s
hook enabled, full + with minAllocAreaSize override:  Total compilation took 11m31s
hook enabled, full + no minAllocAreaSize override:    Total compilation took 10m18s
```

So basically the best is to use the flags GHC uses when running itself.

martijnbastiaan:
In that case, I guess I rescind my approval :-).

rowanG077:
Removed the hook completely and added a bolded section to the changelog entry.

TODO after this is merged: Add these flags to the starter projects.

rowanG077:
Fixes https://github.com/clash-lang/clash-compiler/issues/3252

clash-ghc/clash-ghc.cabal
```diff
@@ -79,24 +78,22 @@ flag workaround-ghc-mmap-crash
 executable clash
   Main-Is:            src-ghc/Batch.hs
   Build-Depends:      base, clash-ghc
-  GHC-Options:        -Wall -Wcompat -threaded -rtsopts
+  GHC-Options:        -Wall -Wcompat -threaded -rtsopts "-with-rtsopts=-K512M -H -I5 -T"
```
martijnbastiaan: I don't think these comments should go lost in the sands of time.

```suggestion
  GHC-Options:
    -- -T: Workaround for GHC #8754: if the GC stats aren't enabled because
    -- the -- compiler couldn't use -Bsymbolic to link the default hooks, then
    -- initialize them sensibly. 
    --
    -- -I5: See GHC #3408: the default idle GC time of 0.3s is too short on
    -- Windows where we receive console events once per second or so.
    --
    -- -H and -K512M: Trade slight memory increase for better performance in
    -- large designs.
    "-with-rtsopts=-K512M -H -I5 -T"
    -Wall
    -Wcompat
    -threaded
    -rtsopts
```

We don't have to add them everywhere, but I think it's a good idea to put them at least in one place.

changelog/2026-05-06T15_20_00+02_00_ghc_style_rts_defaults
```diff
@@ -0,0 +1 @@
+CHANGED: `clash` and `clashi` now use GHC-style RTS defaults through executable `-with-rtsopts` flags instead of linking RTS hooks. The existing unused hooks have been removed. **Existing users should set `-with-rtsopts` to `-K512M -H -I5 -T` if they use their own Clash executable. These flags has been shown to give a ~25% compilation speed up in larger designs.** [#3252](https://github.com/clash-lang/clash-compiler/issues/3252)
```
martijnbastiaan:
````suggestion
CHANGED: `clash` and `clashi` now use GHC-style RTS defaults through executable `-with-rtsopts` flags instead of linking RTS hooks. The existing unused hooks have been removed. **Existing users should set `-with-rtsopts` to `-K512M -H -I5 -T` if they use their own Clash executable. These flags has been shown to give a ~25% compilation speed up in larger designs.** If you used a Clash starter project to initialize your design, you should change:

```haskell
executable clash
  [..]
```

To:


```haskell
executable clash
  [..]
  ghc-options: "-with-rtsopts=-K512M -H -I5 -T"
```

See [#3252](https://github.com/clash-lang/clash-compiler/issues/3252).
````

martijnbastiaan (approved):
Sounds like a plan!

PR #3245, opened 2026-05-01
Memoize results in `flattenCallTree`

### LINKED ISSUE #3246: `flattenCallTree` does a lot of double-work
`flattenExpr` gets called for every to-be-inlined binder over here:

https://github.com/clash-lang/clash-compiler/blob/1710f799da1868af392895c14e0efcf5e03bed9d/clash-lib/src/Clash/Normalize.hs#L374

As it turns out, these binders are often the same. The `flattenExpr` therefore runs on the same term many times. Experiments show the whole normalization path becomes ~10% faster if we cache this.

martijnbastiaan:
Before this patch, `flatten` would get called many times on the same binder. Having a cache reduces duplicate work considerably. I've tested this on `bittide`, the generated HDL is byte-for-byte equal when compared to pre-patch and yields about a 10% speedup (18m16s -> 16m35s).

Fixes #3246

## Still TODO:
  - [x] Write up issue


rowanG077 (approved):
Yes, I like this!

PR #3250, opened 2026-05-06
Implement specialized `topdownFixR` that avoids redundant root restarts

### LINKED ISSUE #3249: Repeated top-down normalization restarts cause redundant traversal work
`repeatR (topdownR r)` restarts the full top-down traversal from the root after every successful rewrite:

https://github.com/clash-lang/clash-compiler/blob/1710f799da1868af392895c14e0efcf5e03bed9d/clash-lib/src/Clash/Rewrite/Combinators.hs#L116-L119

  Several normalization strategies use this pattern to reach a fixpoint, for example in `inlineAndPropagate`, `caseFlattening`, `dec`, and flattening-related cleanup.

This can redo a lot of traversal work. When a rewrite succeeds in a child term, the next possible parent-level rewrite can only be exposed along the path back to the root. Restarting the entire traversal from the root also revisits unrelated subtrees that could not have been affected by that child rewrite.

Experiments show normalization on larger designs improves roughly 20% by using a top-down fixpoint traversal that bubbles changes upward instead of restarting from the root after every rewrite.

  The proposed change is to add a rewrite combinator, currently called `topdownFixR`, for sites that semantically want `repeatR (topdownR r)`. It keeps the same fixpoint behavior for suitable single-node rewrites, but avoids full root restarts by rechecking only the affected path after child rewrites.

rowanG077:
With Bittide `softUgnDemoTest` it gives a 17% improvement, less than the reported issue because I still had a bug when reporting that number.

Before: Clash: Compiling Bittide.Instances.Hitl.SoftUgnDemo.TopEntity.softUgnDemoTest took 9m7s
After: Clash: Compiling Bittide.Instances.Hitl.SoftUgnDemo.TopEntity.softUgnDemoTest took 7m34s

Fixes https://github.com/clash-lang/clash-compiler/issues/3249

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize.hs
+++ b/clash-lib/src/Clash/Normalize.hs
@@ -75,7 +75,7 @@ import           Clash.Normalize.Transformations
 import           Clash.Normalize.Types
 import           Clash.Normalize.Util
 import           Clash.Rewrite.Combinators
-  ((>->), (!->), bottomupR, repeatR, topdownR)
+  ((>->), (!->), bottomupR, repeatR, topdownFixR)
 import           Clash.Rewrite.Types
   (RewriteEnv (..), RewriteState (..), bindings, debugOpts, extra,
    tcCache, topEntities, newInlineStrategy)
@@ -406,7 +406,10 @@ flattenCallTree cache (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
       else return (CBranch (nm,(Binding nm' sp inl pr newExpr r)) allUsed)
 
   flatten =
-    repeatR (topdownR (apply "appProp" appProp >->
+    -- topdownFixR reaches a fixpoint for the top-down propagation bundle.
+    -- Keep flattenLet in the outer fixed-point loop: flattening can expose
+    -- fresh propagation redexes for the next top-down pass.
+    repeatR (topdownFixR (apply "appProp" appProp >->
                apply "bindConstantVar" bindConstantVar >->
                apply "caseCon" caseCon >->
                (apply "reduceConst" reduceConst !-> apply "deadcode" deadCode) >->
--- a/clash-lib/src/Clash/Rewrite/Combinators.hs
+++ b/clash-lib/src/Clash/Rewrite/Combinators.hs
@@ -1,6 +1,6 @@
 {-|
   Copyright  :  (C) 2012-2016, University of Twente
-                         2021, QBayLogic B.V.
+                    2021-2026, QBayLogic B.V.
   License    :  BSD2 (see the file LICENSE)
   Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -16,6 +16,7 @@ module Clash.Rewrite.Combinators
   , bottomupR
   , repeatR
   , topdownR
+  , topdownFixR
   ) where
 
 import           Control.DeepSeq             (deepseq)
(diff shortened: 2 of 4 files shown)


martijnbastiaan (approved):
I'm not sure whether it is "CHANGED". Usually this prevent backporting, but I don't think that applies in this case.

Other than that, great fix!

rowanG077:
I first had it as `ADDED`, I will change it back.

DigitalBrains1:
That's a nice and very welcome improvement!

I do think _changed_ is a better match to what it does. Haskell PVP talks about API, this just isn't API, so there's no relevance of the Haskell PVP. For what it's worth, we've had _changed_ entries for Clash 1.8 in all minor bumps except 1.8.1 and 1.8.5.

clash-lib/src/Clash/Rewrite/Combinators.hs
```diff
@@ -118,6 +119,66 @@ topdownR :: Rewrite m -> Rewrite m
...
+topdownFixR r = go
+ where
+  go ctx term = do
+    term1 <- repeatR r ctx term
+    (term2, Monoid.getAny -> childChanged) <- Writer.listen (allR go ctx term1)
+    if childChanged
+      then do
+        (term3, Monoid.getAny -> parentChanged) <- Writer.listen (repeatR r ctx term2)
+        if parentChanged
+          then go ctx term3
```
christiaanb: In the recursive call to `go`, we will now retry all transformations we just applied to `term2`, i.e. the line:
```
term1 <- repeatR r ctx term
```
Even though we would know by now that those transformations won't fire.

Perhaps we can change `go` to:
```
  go tryParent ctx term = do
    term1 <- when tryParent (repeatR r ctx term)
    (term2, Monoid.getAny -> childChanged) <- Writer.listen (allR (go True) ctx term1)
    if childChanged
      then do
        (term3, Monoid.getAny -> parentChanged) <- Writer.listen (repeatR r ctx term2)
        if parentChanged
          then go False ctx term3
          else return term3
      else return term2
```
and change `topDownFixR` to:
```
topdownFixR r = go True
```

clash-lib/src/Clash/Normalize.hs
```diff
@@ -406,7 +406,10 @@ flattenCallTree cache (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
       else return (CBranch (nm,(Binding nm' sp inl pr newExpr r)) allUsed)
 
   flatten =
-    repeatR (topdownR (apply "appProp" appProp >->
+    -- The propagation bundle contains context-sensitive rewrites: for example,
+    -- caseCon/reduceConst can observe let-bound context through whnfRW. The
+    -- outer repeatR refreshes that context after changes.
+    repeatR (topdownFixR (apply "appProp" appProp >->
```
christiaanb, line 412: Does it really matter for `whnrRW` what shape that context has (transformed vs untransformed)? I'm somewhat surprised `caseCon` and `reduceConst` are affected. Do certain tests fail if you remove the `repeatR`?

rowanG077: No, no tests fail. Originally it was exactly what you describe. But I was worried it might matter. This was basically a conservative choice.

clash-lib/src/Clash/Rewrite/Combinators.hs
rowanG077: I will keep it, but it makes no difference in compilation time for bittide.

clash-lib/src/Clash/Normalize.hs
rowanG077: Actually yes, tests fail if you remove the outer `repeatR`. But it's because `flattenLet` exposes new opportunities which are then never re-written. Test `T3041` fails.

PR #3251, opened 2026-05-06
Replace `error` with `errorX` in Clash primitives 

### LINKED ISSUE #3247: Replace calls to `error` with `errorX` in our (unfoldable) primitives (?)
Primitives such as `shiftL#` use `error` on unexpected/illegal inputs:

https://github.com/clash-lang/clash-compiler/blob/1710f799da1868af392895c14e0efcf5e03bed9d/clash-prelude/src/Clash/Sized/Internal/Unsigned.hs#L493

A recent [bug report](https://github.com/clash-lang/clash-compiler/issues/3234) pointed out that the evaluator would crash when trying to constant fold such a primitive with an illegal input. This was fixed in PR https://github.com/clash-lang/clash-compiler/pull/3235, which made the evaluator test for expected exceptions.

However, it was pointed out that we could also just have replaced `error` with `errorX` in the primitives, because the evaluator already catches those for all primitives. This was [decided against](https://github.com/clash-lang/clash-compiler/pull/3235#issuecomment-4350606880) _for now_, because it would entail a non-backportable PR.

The cover letter of that PR says:

> I will say this does call into question whether we should even do `error` (not `errorX`) in these functions. I guess that's for another issue?

So, here is that other issue.

### LINKED ISSUE #3254: Primitive evaluator doesn't fold `XException` results into undefineds
See https://github.com/clash-lang/clash-compiler/pull/3251#issuecomment-4394657939.

martijnbastiaan:
Fixes #3247

Fixes #3254 

I've also put together a `HACKING.md`. I find that I keep explaining this to AI, which means our instructions for humans _also_ suck. I'll later amend this with instructions how to debug the transformation system if I get to fixing an issue in that realm again.

## Still TODO:

  - [x] Find out what the error is. I believe the problem is that `reduce` notices the `XException`, but doesn't replace the result with an undefined prim. It just warns and puts back the original term. @christiaanb Do you remember if that behavior is intentional?


christiaanb:
The behavior you describe was implemented as part of https://github.com/clash-lang/clash-compiler/pull/339 which predates the `makeUndefinedIf` function by a year: https://github.com/clash-lang/clash-compiler/pull/597. So the described behavior is definitely as intended when we implemented it 8 years, but then 7 years ago we forgot to give `XException` the same behavior as `DivByZero` exceptions: replacing them by an `undefined` term.

So I think we should drop behavior from 8 years ago, and use the behavior we introduced 7 years ago to convert caught exceptions to `undefined` terms.

christiaanb:
With regards to your `HACKING.md` file, do you think its contents could/should be merged with the existing documentation at https://github.com/clash-lang/clash-compiler/blob/master/ARCHITECTURE.md ?

martijnbastiaan:
> So I think we should drop behavior from 8 years ago, and use the behavior we introduced 7 years ago to convert caught exceptions to `undefined` terms.

Alright, will do! Thanks for the archeology.

> With regards to your `HACKING.md` file, do you think its contents could/should be merged with the existing documentation at https://github.com/clash-lang/clash-compiler/blob/master/ARCHITECTURE.md ?

I don't think so, but it's a very slight preference. If anyone has a strong opinion on it I'm happy to merge them.

martijnbastiaan:
@christiaanb Done.

rowanG077 (approved):
Approved!

DigitalBrains1 (approved):
Bis!

A note: this script `scripts/build-with-all-supported-ghcs` has never been on my radar. So everytime I changed wich GHC versions we use in CI, I don't update that file. It currently says GHC 9.6.7, 9.8.4, 9.10.1 and 9.10.2. Those last two should probably be 9.10.3 and 9.12.4.

martijnbastiaan:
> A note: this script `scripts/build-with-all-supported-ghcs` has never been on my radar. So everytime I changed wich GHC versions we use in CI, I don't update that file. It currently says GHC 9.6.7, 9.8.4, 9.10.1 and 9.10.2. Those last two should probably be 9.10.3 and 9.12.4.

Okay, I'll add an issue stating we should add some machinery/checking to keep them in check.

PR #3262, opened 2026-05-09
Bump dependency bounds hedgehog, QuickCheck

DigitalBrains1:
Our current bounds also prevent us from re-entering Stackage nightly; note, though, that Stackage merely requires QuickCheck 2.16 but Hackage already has 2.18, so I'm bumping to allow 2.18. I did test locally with stack and 2.16.

## Still TODO:


DigitalBrains1:
Since this is a requirement for inclusion in Stackage nightly, I intend to do a package revision on Hackage for 1.10.0 after this PR has been backported and I've run CI on the package revision (we don't do the latter automatically, but I can rig CI to effectively test it in a one-off run).

rowanG077 (approved):

PR #3268, opened 2026-05-10
GitLab CI: Move nearly everything to local runners

DigitalBrains1:
Only the "report status" jobs are kept on shared runners as they are truly tiny.

Our use of the shared runners has increased a lot lately. It's probably just a high number of PR's causing this. Moving these jobs to local runners will slow things down a bit because more jobs contend for the same limited set of runners, but the current excessive use of shared runners justifies this.

## Still TODO:


martijnbastiaan (approved):
LGTM. Let's not saddle GitLab with the costs either :).

PR #3270, opened 2026-05-11
Drop `data-binary-ieee754` in favor of `GHC.Float`

### LINKED ISSUE #3174: `data-binary-ieee754` has been deprecated
The `data-binary-ieee754` has been deprecated (cf. https://hackage.haskell.org/package/data-binary-ieee754).

I propose updating the dependencies using one of the recommended alternatives instead.

martijnbastiaan:
Fixes #3174

## Still TODO:

  - [x] Check on all supported GHCs
  - [x] ~~Write (regression) tests for all `GHC.Prim`s in the evaluator~~ I've got a better plan, but for another PR..

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
@@ -38,7 +38,6 @@ import qualified Control.Lens               as Lens
 import           Control.Monad.State.Strict (State, MonadState)
 import qualified Control.Monad.State.Strict as State
 import           Control.Monad.Trans.Except (runExcept)
-import           Data.Binary.IEEE754        (doubleToWord, floatToWord, wordToDouble, wordToFloat)
 import           Data.Bits
 import qualified Data.ByteString.Internal as BS
 import           Data.Char           (chr,ord)
@@ -358,17 +357,17 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
 
   $(namePat 'GHC.Prim.int2Float#)
     | [Lit (IntLiteral i)] <- args
-    -> reduce . Literal . FloatLiteral  . floatToWord $ fromInteger i
+    -> reduce . Literal . FloatLiteral  . castFloatToWord32 $ fromInteger i
   $(namePat 'GHC.Prim.int2Double#)
     | [Lit (IntLiteral i)] <- args
-    -> reduce . Literal . DoubleLiteral . doubleToWord $ fromInteger i
+    -> reduce . Literal . DoubleLiteral . castDoubleToWord64 $ fromInteger i
 
   $(namePat 'GHC.Prim.word2Float#)
     | [Lit (WordLiteral i)] <- args
-    -> reduce . Literal . FloatLiteral  . floatToWord $ fromInteger i
+    -> reduce . Literal . FloatLiteral  . castFloatToWord32 $ fromInteger i
   $(namePat 'GHC.Prim.word2Double#)
     | [Lit (WordLiteral i)] <- args
-    -> reduce . Literal . DoubleLiteral . doubleToWord $ fromInteger i
+    -> reduce . Literal . DoubleLiteral . castDoubleToWord64 $ fromInteger i
 
   $(namePat 'GHC.Prim.uncheckedIShiftL#)
     | [ Lit (IntLiteral i)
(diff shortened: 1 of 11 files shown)


clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -358,17 +357,17 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
 
   $(namePat 'GHC.Prim.int2Float#)
     | [Lit (IntLiteral i)] <- args
-    -> reduce . Literal . FloatLiteral  . floatToWord $ fromInteger i
+    -> reduce . Literal . FloatLiteral  . castFloatToWord32 $ fromInteger i
```
rowanG077, line 360: Is that really correct? 

GHCI tells me:

```Haskell
ghci> F# $ int2Float# (10# :: Int#)
10.0
ghci> castFloatToWord32 10
1092616192
```

rowanG077 (changes requested):
Not really changes, just want to confirm what is happening here is correct.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
martijnbastiaan: Oh boy, both before and after are wrong.

martijnbastiaan: No! Both before and after are right, haha. It's definitely confusing though. But:

* `int2Float#` is supposed to take the incoming integer and preserve its integeral value in `Float`. E.g., `1` -> `1.0`.
* `fromInteger i` does exactly this. E.g., `(fromInteger 1 :: Float) == 1.0`.
* `FloatLiteral` stores a `Float` in `BitVector`-like manner i.e. its bits. Both `floatToWord` and `castFloatToWord32` do what the name of the latter function suggests: essentially `pack`.

I think the confusion is that `floatToWord` really sounds like it should do that `1` -> `1.0` conversion, but it actually casts. This isn't actually documented though.. but it looks like:

```
fromFloat :: (F.Storable word, F.Storable float) => float -> word
fromFloat float = unsafePerformIO $ F.alloca $ \buf -> do
	F.poke (F.castPtr buf) float
	F.peek buf
```

rowanG077: Oke than I would propose to document this on the `FloatLiteral` and `DoubleLiteral` constructors as that is where I looked when checking this. Or going the other way and just storing `Float`/`Double` directly in these literals? But there probably is a good reason for not doing that.

DigitalBrains1: It's not documented at the constructors, but there's this immediately before the data type:
https://github.com/clash-lang/clash-compiler/blob/89c92c3e211a24d64aadb47362aefe928033f7ba/clash-lib/src/Clash/Core/Literal.hs#L30-L44

Perhaps we should make it stand out more:

```diff
diff --git a/clash-lib/src/Clash/Core/Literal.hs b/clash-lib/src/Clash/Core/Literal.hs
index 49bb253db..2cf843c76 100644
--- a/clash-lib/src/Clash/Core/Literal.hs
+++ b/clash-lib/src/Clash/Core/Literal.hs
@@ -58,8 +58,8 @@ data Literal
   | Word16Literal   !Integer
   | Word32Literal   !Integer
   | StringLiteral   !String
-  | FloatLiteral    !Word32
-  | DoubleLiteral   !Word64
+  | FloatLiteral    !Word32  -- See Note [Storage of floating point in Literal]
+  | DoubleLiteral   !Word64  -- See Note [Storage of floating point in Literal]
   | CharLiteral     !Char
   | NaturalLiteral  !Integer
   | ByteArrayLiteral !ByteArray
```

martijnbastiaan: I can make an extra comment to the data constructor, yeah. Is that enough Rowan?

rowanG077: yes that's fine.

martijnbastiaan: Done!

rowanG077 (approved):

PR #3275, opened 2026-05-21
Migrate on-push PR jobs from Gitlab over to Github

jaschutte:
We have self-hosted Github runners, however we still mirror the repository to Gitlab to run jobs on PRs. This migrates over the jobs that the Gitlab runners do over to Github. In specific; only the jobs performed on push are copied over. This includes the following set of jobs:

 - Testing if all clash-* projects successfully compile against all supported GHC versions in Cabal and Stack
 - Package & cache all clash-* projects using Nix (this also runs tests for those packages) for all supported GHC versions
 - Run the clash testsuite for VHDL, Verilog and SystemVerilog in all supported GHC versions

Nix automatically caches whatever it builds to our local cache, pushes to the master branch also get cached in Cachix.
Cabal caches only the clash dependencies (the key is the hash of the cabal.project file, I think this should be correct as its the only thing that changes the dependencies(?). If not, let me know and I will change it to something else. Maybe the `.cabal` files should be part of the key too?) It uses the local minio cache. 
Stack does not utilize caching.

## TODO:
 - [ ] Run clash-ffi tests(?)
 - [ ] Add `.cabal` to the key to the Cabal cache

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -27,11 +27,9 @@ jobs:
           9.6.7
         ]
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
-      - name: Setup cache
-        run: mkdir .cache
-      - uses: tespkg/actions-cache@v1
+      - uses: tespkg/actions-cache/restore@v1
         with:
           endpoint: "192.168.102.136"
           port: 9000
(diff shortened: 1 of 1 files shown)


jaschutte:
Another thing to note; this does not remove the old Gitlab pipeline yet. I think its best to do that all at once once all Gitlab jobs have been migrated.

martijnbastiaan:
Do you ensure memory limits on the host side? If not, you could do the poor man's thing we do over at `bittide-hardware`:

https://github.com/bittide/bittide-hardware/blob/a12b033591cedfe2403109016256e30e0ef61e13/.github/workflows/ci.yml#L77-L79

jaschutte:
We have configured the backend hosting the runners to limit the CPUs per job, memory I am not quite sure but it shouldn't be a difficult thing to add (iirc it's literally just an option inside of docker-compose).

jaschutte:
Closing as this is a PR from a fork and not from a main branch, therefore the new tests ironically enough don't run.

See #3289 for the new PR.

PR #3276, opened 2026-05-21
Add flag to en/disable the translation of  Integers and Naturals

leonschoorl:
## Still TODO:

  - [x] Fix the detection of Integers/Naturals inside function arguments to HO prims like in the  Scatter test
  - [ ] Add some failing test with:
    - [ ] HO prims with Integers hidden in a HO argument
    - [ ] nested HO functions like: `map (map illegal)`
  - [ ] Think if it's possible and makes sense to disable the `elimCaseBigNumInternals` transformation (added in PR #3172) when `-fclash-no-translate-integer-natural` is set
  - [ ] `blockRamU` and `blockRam1` (and possibly others?) use Integers internally, convert them to some better type

### DIFF
--- a/clash-lib/src/Clash/Netlist.hs
+++ b/clash-lib/src/Clash/Netlist.hs
@@ -33,12 +33,13 @@ import           Data.Bifunctor                   (first, second)
 import           Data.Char                        (ord)
 import           Data.Either                      (partitionEithers, rights)
 import           Data.Foldable                    (foldlM)
+import qualified Data.IntMap.Strict               as IntMap
 import           Data.List                        (elemIndex, partition)
 import           Data.List.Extra                  (zipEqual)
 import           Data.List.NonEmpty               (NonEmpty (..))
 import qualified Data.List.NonEmpty.Extra         as NE
 import           Data.Maybe
-  (listToMaybe, fromMaybe)
+  (catMaybes, listToMaybe, fromMaybe, mapMaybe, maybeToList)
 import qualified Data.Map.Ordered                 as OMap
 import qualified Data.Set                         as Set
 import qualified Data.Text                        as StrictText
@@ -74,6 +75,7 @@ import           Clash.Core.VarEnv
    lookupVarEnv')
 import           Clash.Driver.Types               (BindingMap, Binding(..), ClashEnv(..), ClashOpts (..))
 import           Clash.Netlist.BlackBox
+import           Clash.Netlist.BlackBox.Types     (BlackBoxTemplate)
 import qualified Clash.Netlist.Id                 as Id
 import           Clash.Netlist.Types              as HW
 import           Clash.Netlist.Util
@@ -837,7 +839,8 @@ mkExpr :: HasCallStack
        -> NetlistMonad (Expr,[Declaration]) -- ^ Returned expression and a list of generate BlackBox declarations
 mkExpr _ _ _ (stripTicks -> Core.Literal l) = do
   iw <- Lens.view intWidth
-  return (mkLiteral iw l, [])
+  translBigNums <- Lens.view translateBigNums
+  return (mkLiteral iw translBigNums l, [])
 
 mkExpr bbEasD declType bndr app =
  let (appF,args,ticks) = collectArgsTicks app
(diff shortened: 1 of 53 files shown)


clash-lib/src/Clash/Netlist.hs
```diff
@@ -1148,3 +1149,93 @@ mkDcApplication declType dstHTypes (MultiId argNms) _ args = do
...
+    CtxSignal nm -> " in signal " <> showIdentifier nm
+    CtxBlackBox nm -> " blackbox " <> StrictText.unpack nm
+    CtxBbInput n -> " in input " <> show n <> " to"
+    CtxBbResult n -> " in result " <> show n <> " of"
+
+showIdentifier :: Identifier -> String
+showIdentifier = StrictText.unpack . Id.toText
+
+-- Checks the netlist for Integer/Natural usage
+checkComponent :: Component -> [ErrorMsg]
```
rowanG077, line 1371: I think the problem why it doesn't detect `Vec 5 Integer` is because it doesn' recurse into `HWType` here.

rowanG077 (commented):
I think the approach is good. Just some missing things

clash-lib/src/Clash/Netlist.hs
leonschoorl: Thanks, I totally forgot to do that.

I've made checkHWType recurse into nested HWTypes.
That does allow it to detect an intermediate signal like `Vec 5 Integer` in something like:
```haskell
topEntity =
  map @_ @_ @5 (fromInteger @Int)
  .
  map (toInteger @Int)
```

But the problems which I was working on was:
```haskell
topEntity =
  \(xs :: Vec 2 Int) -> map illegal xs

illegal :: Int -> Int
illegal = fromInteger . (+1) . toInteger
```

Which isn't fixed by this better checkHWType check, because there is no intermediate vector with integers for it to find.

clash-lib/src/Clash/Netlist.hs
```diff
@@ -1148,3 +1151,136 @@ mkDcApplication declType dstHTypes (MultiId argNms) _ args = do
...
+     Nothing -> checkExpr ctx e
+   checkPort :: [NetlistContext] -> PortDirection -> (Identifier,HWType) -> Maybe ErrorMsg
+   checkPort ctx portDirection (portNm,ty) = checkHWType (CtxPort portDirection portNm:ctx) ty
+   checkDecl :: [NetlistContext] -> Declaration -> [ErrorMsg]
+   checkDecl ctx x = case x of
+     Assignment nm _ expr -> checkExpr (CtxSignal nm : ctx) expr
+     CondAssignment nm ty scrut scrutTy alts ->
+       let ctx' = CtxSignal nm : ctx
+       in catMaybes [checkHWType ctx' ty, checkHWType ctx' scrutTy] <> checkExpr ctx' scrut <> concatMap (checkExpr ctx' . snd) alts
+     InstDecl {} -> [] -- will be checked when we check that components netlist
```
rowanG077, line 1432: Should the generic parameters be checked here as well? Those do not get checked when checking the netlist.

clash-lib/src/Clash/Netlist/BlackBox.hs
```diff
@@ -296,7 +296,7 @@ mkArgument bbName bndr declType nArg e = do
         (C.Var v,[],_) -> do
           return ((Identifier (Id.unsafeFromCoreId v) Nothing,hwTy,False),[])
         (C.Literal l,[],_) ->
-          return ((mkLiteral iw l,hwTy,True),[])
+          return ((mkLiteral iw True l,hwTy,True),[])
```
rowanG077: This is not what we want right? Now it is always translated regardless of the flag.

rowanG077 (changes requested):

rowanG077:
Beside the comments I was not able to find glaring problems with this.

clash-lib/src/Clash/Netlist.hs
leonschoorl: You are correct that it's not check right now.

But I'm thinking generics one of the few places where the `integer` type would be ok.

PR #3280, opened 2026-05-28
Add `__debug_hash` to Clash manifest files

martijnbastiaan:
I'm trying to find out why we keep seeing cache invalidations over at `bittide-hardware` even though all the files Clash outputs (except for the manifest file) are byte identical. After writing the patch I figured it is probably useful to have it in there for future debug runs.

## Still TODO:


rowanG077 (commented):
Perhaps it would be good to only have them in the JSON when debugging `-g` is enabled?

martijnbastiaan:
Oh right. I'm not sure it fits in `-g`, but perhaps `-fclash-debug-manifest-hash`?

rowanG077:
Yes, that's fine.

martijnbastiaan:
Done.

rowanG077 (approved):

PR #3286, opened 2026-06-01
Update Github action versions

leonschoorl:
We were still using old versions of actions that depend on Node.js 20
Which is deprecated and support will be dropped soon https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

I've opted to use commit hashes instead of tags as precaution against supply chain attacks.
## Still TODO:

  - ~Write a changelog entry (see changelog/README.md)~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,8 +44,8 @@ jobs:
             ghc: "9.12"
 
     steps:
-      - uses: actions/checkout@v5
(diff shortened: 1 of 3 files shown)


rowanG077 (approved):

martijnbastiaan:
> I've opted to use commit hashes instead of tags as precaution against supply chain attacks.

We can force this in the org settings too. It seems sensible.

PR #3289, opened 2026-06-02
Migrate on-push PR jobs from Gitlab over to Github

jaschutte:
We have self-hosted Github runners, however we still mirror the repository to Gitlab to run jobs on PRs. This migrates over the jobs that the Gitlab runners do over to Github. In specific; only the jobs performed on push are copied over. This includes the following set of jobs:

 - Testing if all clash-* projects successfully compile against all supported GHC versions in Cabal
 - Testing if all clash-* projects succesfully compile with the `stack.yaml` in the root
 - Migrates the Haddock tests from Gitlab over to Github  
 - Package & cache all clash-* projects using Nix (this also runs tests for those packages) for all supported GHC versions
 - Run the clash testsuite for VHDL, Verilog and SystemVerilog in all supported GHC versions

Nix automatically caches whatever it builds to our local cache, pushes to the master and release branch also get cached in Cachix.

Original PR: #3275

### DIFF
--- a/.ci/gh-setup.sh
+++ b/.ci/gh-setup.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -xou pipefail
+
+# Make sure everything is up-to-date & installing missing dependencies
+sudo apt-get update
+sudo apt-get install libtinfo-dev -y
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+CI_DIR=$GIT_ROOT/.ci
+cd CI_DIR
+
+# Get the proper amount of cores we can use and configure ghc to use them
+THREADS=$($CI_DIR/effective_cpus.sh)
+sed "7s/-j4/-j$THREADS/g" cabal.project.local
+cp $CI_DIR/cabal.project.local $GIT_ROOT
+
+echo "--- cabal.project.local ---"
+cat $GIT_ROOT/cabal.project.local
+
+echo "--- thread count ---"
+echo "Using $THREADS threads"
--- a/.ci/gh-setup.sh
+++ b/.ci/gh-setup.sh
@@ -2,19 +2,21 @@
 set -xou pipefail
 
 # Make sure everything is up-to-date & installing missing dependencies
-sudo apt-get update
-sudo apt-get install libtinfo-dev -y
+sudo apt-get update || exit $?
+sudo apt-get install libtinfo-dev -y || exit $?
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
+[[ $? -ne 0 ]] && exit 1
 CI_DIR=$GIT_ROOT/.ci
 
 # Get the proper amount of cores we can use and configure ghc to use them
 THREADS=$($CI_DIR/effective_cpus.sh)
+[[ $? -ne 0 ]] && exit 1
 sed <$CI_DIR/cabal.project.local.in >$GIT_ROOT/cabal.project.local "
-    s/__THREADS__/$THREADS/"
+    s/__THREADS__/$THREADS/" || exit $?
 
 echo "--- cabal.project.local ---"
-cat $GIT_ROOT/cabal.project.local
+cat $GIT_ROOT/cabal.project.local || exit $?
 
 echo "--- thread count ---"
 echo "Using $THREADS threads"
--- a/.ci/gh-setup.sh
+++ b/.ci/gh-setup.sh
@@ -7,12 +7,11 @@ sudo apt-get install libtinfo-dev -y
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
 CI_DIR=$GIT_ROOT/.ci
-cd CI_DIR
 
 # Get the proper amount of cores we can use and configure ghc to use them
 THREADS=$($CI_DIR/effective_cpus.sh)
-sed "7s/-j4/-j$THREADS/g" cabal.project.local
-cp $CI_DIR/cabal.project.local $GIT_ROOT
+sed <$CI_DIR/cabal.project.local.in >$GIT_ROOT/cabal.project.local "
+    s/__THREADS__/$THREADS/"
 
 echo "--- cabal.project.local ---"
 cat $GIT_ROOT/cabal.project.local
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -1,4 +1,4 @@
-name: Nix Packaging
+name: On push CI
 
 on: push
 
@@ -7,6 +7,99 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Nix by default runs tests, meaning that there's not much point in re-running tests in Cabal/Stack
+  # So for these we do a simple compilation test instead, checking that they still compile with
+  # whatever changes got made
+
+  cabal:
+    name: Cabal compilation test
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: [
+          9.12.4,
+          9.10.3,
+          9.8.4,
+          9.6.7
+        ]
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup cache
+        run: mkdir .cache
+      - uses: tespkg/actions-cache@v1
+        with:
+          endpoint: "192.168.102.136"
+          port: 9000
+          insecure: true
+          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
+          secretKey: "${{ secrets.S3_SECRET_KEY }}"
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-checkout-cabal
+          path: .cache
+
+      # Perhaps we should preinstall these on the container?
+      # This step always takes a minute...
+      # We'd lose the ability to select the version of cabal inside of the runner though
+      - name: Install GHC
+        uses: haskell-actions/setup@v2
+        with:
+          ghc-version: ${{ matrix.ghc }}
+          cabal-version: latest
+
+      - name: Version information
+        run:
+          cabal --version
+
+  stack:
+    name: Stack compilation test
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: [
+          9.12.4,
+          9.10.3,
+          9.8.4,
+          9.6.7
+        ]
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Setup cache
+        run: mkdir .cache
+      - uses: tespkg/actions-cache@v1
+        with:
+          endpoint: "192.168.102.136"
+          port: 9000
+          insecure: true
+          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
+          secretKey: "${{ secrets.S3_SECRET_KEY }}"
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-checkout-stack
+          path: ~/.stack
+
+      # Perhaps we should preinstall these on the container?
+      # This step always takes a minute...
+      # We'd lose the ability to select the version of cabal inside of the runner though
+      - name: Install GHC
+        uses: haskell-actions/setup@v2
+        with:
+          ghc-version: ${{ matrix.ghc }}
+          enable-stack: true
+          stack-version: latest
+
+      - name: Version information
+        run:
+          stack --version
+
   packages_common:
     name: Build Common (${{matrix.ghc}})
     runs-on: self-hosted
@@ -39,7 +132,6 @@ jobs:
       - run: |
           nix build -L .#clashPackages-${{matrix.ghc}}.clash-prelude
           attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-prelude)
-
       - run: |
           nix build -L .#clashPackages-${{matrix.ghc}}.clash-lib
           attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-lib)
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -24,11 +24,9 @@ jobs:
           9.6.7
         ]
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
-      - name: Setup cache
-        run: mkdir .cache
-      - uses: tespkg/actions-cache@v1
+      - uses: tespkg/actions-cache/restore@v1
         with:
           endpoint: "192.168.102.136"
           port: 9000
@@ -39,8 +37,12 @@ jobs:
           region: local
           use-fallback: false
           retry: false
-          key: ${{ runner.os }}-checkout-cabal
-          path: .cache
+          # The cabal.project file contains the index state & hashes
+          # Since we only cache the dependencies, and this is what determines what dependencies are grabbed
+          # hashing the cabal.project file should suffice in determining if the dependencies changed
+          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles("cabal.project") }
+          restore-keys: ${{ runner.os }}-${ matrix.ghc }-cabal-
+          path: ~/.cabal
 
       # Perhaps we should preinstall these on the container?
       # This step always takes a minute...
@@ -55,6 +57,36 @@ jobs:
         run:
           cabal --version
 
+      - name: Install system dependencies
+        run: |
+          sudo apt-get install libtinfo-dev -y
+          sudo apt-get install tree -y
+          sudo apt-get install iverilog -y
+
+      - name: Build dependencies
+        run: |
+          cabal v2-build all --dependencies-only
+
+      - uses: tespkg/actions-cache/save@v1
+        with:
+          endpoint: "192.168.102.136"
+          port: 9000
+          insecure: true
+          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
+          secretKey: "${{ secrets.S3_SECRET_KEY }}"
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles("cabal.project") }
+          path: ~/.cabal
+
+      - name: Build
+        run: |
+          # clash-cosim really wants to be built first
+          cabal v2-build clash-cosim
+          cabal v2-build all
+
   stack:
     name: Stack compilation test
     runs-on: self-hosted
@@ -68,23 +100,7 @@ jobs:
           9.6.7
         ]
     steps:
-      - uses: actions/checkout@v5
-
-      - name: Setup cache
-        run: mkdir .cache
-      - uses: tespkg/actions-cache@v1
-        with:
-          endpoint: "192.168.102.136"
-          port: 9000
-          insecure: true
-          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
-          secretKey: "${{ secrets.S3_SECRET_KEY }}"
-          bucket: runner
-          region: local
-          use-fallback: false
-          retry: false
-          key: ${{ runner.os }}-checkout-stack
-          path: ~/.stack
+      - uses: actions/checkout@v6
 
       # Perhaps we should preinstall these on the container?
       # This step always takes a minute...
@@ -96,9 +112,20 @@ jobs:
           enable-stack: true
           stack-version: latest
 
+
       - name: Version information
         run:
           stack --version
+
+      - name: Install system dependencies
+        run: sudo apt-get install libtinfo-dev -y
+
+      # From what I read online, stack is difficult to cache and also generates huge build artifacts
+      # So maybe it's better not to cache the stack builds? The current Gitlab runners don't either
+      - name: Build
+        run:
+          stack build -j$(nproc) --pedantic
+
 
   packages_common:
     name: Build Common (${{matrix.ghc}})
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -126,8 +126,6 @@ jobs:
         run: .ci/gh-setup.sh
 
       - name: Build & check documentation
-        env:
-          PKG: ${{ matrix.pkg }}
         run: .ci/build_docs.sh
 
   stack:
@@ -247,7 +245,7 @@ jobs:
           attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-ghc)
 
       - name: Cachix
-        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/v1.10.0'
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/1.10'
         env:
           CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_SECRET }}
         run: |
@@ -291,7 +289,7 @@ jobs:
         run: attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.${{matrix.package}})
 
       - name: Cachix
-        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/v1.10.0'
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/1.10'
         env:
           CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_SECRET }}
         run: nix path-info .#clashPackages-${{matrix.ghc}}.${{matrix.package}} | cachix push clash-lang
@@ -324,7 +322,7 @@ jobs:
         run: attic push public $(nix path-info .#devShells.x86_64-linux.${{matrix.ghc}})
 
       - name: Cachix
-        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/v1.10.0'
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/1.10'
         env:
           CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_SECRET }}
         run: nix path-info .#devShells.x86_64-linux.${{matrix.ghc}} | cachix push clash-lang
@@ -372,8 +370,38 @@ jobs:
             ${{ runner.os }}-master-cabal-${{ matrix.ghc }}
           path: ${{ steps.ghc-install.outputs.cabal-store }}
 
-      - name: Setup
-        run: .ci/gh-setup.sh
+      - name: Cachix
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/1.10'
+        env:
+          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_SECRET }}
+        run: |
+          nix path-info .#clashPackages-${{matrix.ghc}}.clash-testsuite-debug | cachix push clash-lang
+
+  running_testsuite:
+    name: Running clash-testsuite (${{matrix.ghc}}, ${{matrix.hdl}})
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        hdl: [
+          'VHDL',
+          'Verilog',
+          'SystemVerilog'
+        ]
+        ghc: [
+          ghc9124,
+          ghc9103,
+          ghc984,
+          ghc967
+        ]
+    needs: packages_testsuite
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Authenticate cache
+        env:
+          ATTIC_AUTH_TOKEN: ${{ secrets.ATTIC_SECRET }}
+        run: .ci/attic-auth.sh
 
       - name: Run clash-testsuite
         run: cabal v2-run clash-testsuite -- -j$(.ci/effective_cpus.sh) --hide-successes -p .${{matrix.hdl}}. --no-vivado --no-modelsim
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -40,7 +40,7 @@ jobs:
           # The cabal.project file contains the index state & hashes
           # Since we only cache the dependencies, and this is what determines what dependencies are grabbed
           # hashing the cabal.project file should suffice in determining if the dependencies changed
-          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles("cabal.project") }
+          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles("cabal.project", "*/*.cabal") }
           restore-keys: ${{ runner.os }}-${ matrix.ghc }-cabal-
           path: ~/.cabal
 
@@ -287,7 +287,6 @@ jobs:
     name: Running clash-testsuite (${{matrix.ghc}}, ${{matrix.hdl}})
     runs-on: self-hosted
     strategy:
-      # If caching fails for one package, then we are still interested in caching other packages
       fail-fast: false
       matrix:
         hdl: [
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -342,4 +342,23 @@ jobs:
 
       - name: Run clash-testsuite
         run: |
-          nix run .#clashPackages-${{matrix.ghc}}.clash-testsuite -- -j$(nproc) --hide-successes -p .${{matrix.hdl}} --no-vivado --no-modelsim
+          nix run .#clashPackages-${{matrix.ghc}}.clash-testsuite -- -j$(nproc) --hide-successes -p .${{matrix.hdl}}. --no-vivado --no-modelsim
+
+  all:
+    name: All on-push jobs finished
+    if: ${{ !cancelled() }}
+    needs: [
+      cabal,
+      stack,
+      packages_common,
+      packages_other,
+      developer_shells,
+      packages_testsuite,
+      running_testsuite
+    ]
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/checkout@v6
+      - uses: clash-lang/all-jobs-ok@v2
+        with:
+          needs: ${{ toJSON(needs) }}
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -193,8 +193,7 @@ jobs:
             "clash-prelude-hedgehog",
             "clash-profiling",
             "clash-profiling-prepare",
-            "clash-term",
-            "clash-testsuite"
+            "clash-term"
         ]
     needs: packages_common
     steps:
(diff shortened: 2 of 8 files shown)


jaschutte:
I noticed that it failed to retrieve the cache due to lacking permissions, I've added the keys and this should be resolved on the next run.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,320 @@
...
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          # The cabal.project file contains the index state & hashes
+          # Since we only cache the dependencies, and this is what determines what dependencies are grabbed
+          # hashing the cabal.project file should suffice in determining if the dependencies changed
+          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles("cabal.project", "*/*.cabal") }
+          restore-keys: ${{ runner.os }}-${ matrix.ghc }-cabal-
+          path: ~/.cabal
```
martijnbastiaan: New Cabals conform to freedesktop standards, so:

```
~/.config/cabal
~/.local/state/cabal
~/.cache/cabal
```

I see you're using `haskell-actions` though, so you can use the `cabal-store` [output](https://github.com/haskell-actions/setup#outputs). (Though this wouldn't include `.cache/cabal`, which I still think is a very good idea to include.)

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,320 @@
...
+
+      - name: Version information
+        run:
+          cabal --version
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get install libtinfo-dev -y
+          sudo apt-get install tree -y
+          sudo apt-get install iverilog -y
```
martijnbastiaan:
```suggestion
          sudo apt-get install tree iverilog libtinfo-dev -y
```

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,320 @@
...
+
+      - name: Build dependencies
+        run: |
+          cabal v2-build all --dependencies-only
+
+      - uses: tespkg/actions-cache/save@v1
+        with:
```
martijnbastiaan, line 67: Given that this is a fairly obscure package and it gets secrets to local machines, could you please hash-lock it?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,320 @@
...
+          port: 9000
+          insecure: true
+          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
+          secretKey: "${{ secrets.S3_SECRET_KEY }}"
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles("cabal.project") }
+          path: ~/.cabal
```
martijnbastiaan: Same as earlier comment. Perhaps there is a way to deduplicate these sections?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,320 @@
...
+          stack --version
+
+      - name: Install system dependencies
+        run: sudo apt-get install libtinfo-dev -y
+
+      # From what I read online, stack is difficult to cache and also generates huge build artifacts
+      # So maybe it's better not to cache the stack builds? The current Gitlab runners don't either
+      - name: Build
+        run:
+          stack build -j$(nproc) --pedantic
```
martijnbastiaan: We've got enough disk space :-). Use the `stack-root` [output](https://github.com/haskell-actions/setup#outputs).

martijnbastiaan (changes requested):

.github/workflows/on-push.yml
DigitalBrains1: > The current Gitlab runners don't either

Drive-by comment... they most certainly do, unless the path we use is wrong. We cache `/root/.stack` on GitLab CI. That's where Stack stores stuff, right?

martijnbastiaan: Yup

DigitalBrains1:
Could you please indicate when you want reviews? I notice not all that we discussed has been implemented yet, so it's not really obvious to me when you'd like to have the feedback.

.github/workflows/on-push.yml
jaschutte: Oh I missed that then! :sweat_smile: 
I've added the stack caching to the pipeline

.github/workflows/on-push.yml
jaschutte: I've changed the key to use cabal-store output rather than the files that determine up the output, which makes a lot more sense.

jaschutte:
If people have the time could another round of reviews come about? I (finally..) got around to applying the comments.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          9.10.3,
+          9.8.4,
+          9.6.7
+        ]
+    steps:
+      - uses: actions/checkout@v6
+
+      # Perhaps we should preinstall these on the container?
+      # This step always takes a minute...
+      # We'd lose the ability to select the version of cabal inside of the runner though
```
martijnbastiaan, line 31: I think you can do this: https://github.com/marketplace/actions/build-and-push-docker-images. But I don't think you should do that in this PR, just make an issue.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+    name: Cabal compilation test
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: [
+          9.12.4,
+          9.10.3,
+          9.8.4,
+          9.6.7
```
martijnbastiaan, line 24: Again not for this PR, but perhaps do https://github.com/clash-lang/checked-literals/blob/main/ghc-versions.json to dedup.

martijnbastiaan (approved):
Nice!

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          endpoint: "192.168.102.136"
+          port: 9000
+          insecure: true
+          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
+          secretKey: "${{ secrets.S3_SECRET_KEY }}"
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-${ matrix.ghc }-cabal-${ hashFiles( ${{ steps.ghc-install.outputs.cabal-store }}, "~/.cache/cabal" ) }
```
DigitalBrains1: What exactly is the effect of making a `hashFiles` of the directories `~/.cabal/store` and `~/.cache/cabal`? These don't exist yet when the cache is going to be restored, right? Would the keys even match between what is restored and what is saved? I don't think they would, you specify the key for both steps, so I expect they compute them separately for restore and save. Why is `~/.cache/cabal` useful as a cache discriminator? Why is `~/.cache/cabal` not in the cache? FWIW, I think we cache the equivalent of `~/.cache/cabal` on GitLab, as we use the old-style directory layout there and that stores stuff in `~/.cabal` which we cache wholesale.

I know that's a whole bunch of questions but I really don't get it :-D

I note this here at this job, but it goes for the other jobs as well.

<details><summary>Directory locations for Cabal</summary>

If I make a dir `~/.cabal` I get:
```
$ cabal path
Config file path source is default config file.
Config file not found: /home/peter/.cabal/config
Writing default configuration to /home/peter/.cabal/config
compiler-flavour: ghc
compiler-id: ghc-9.12.4
compiler-abi-tag: ghc-9.12.4-876d
compiler-path: /home/peter/.ghcup/bin/ghc
compiler-store-path: /home/peter/.cabal/store/ghc-9.12.4-876d
cache-home: /home/peter/.cabal
remote-repo-cache: /home/peter/.cabal/packages
logs-dir: /home/peter/.cabal/logs
store-dir: /home/peter/.cabal/store
config-file: /home/peter/.cabal/config
installdir: /home/peter/.cabal/bin
```

If I don't, I get:
```
$ cabal path
Config file path source is default config file.
Config file not found: /home/peter/.config/cabal/config
Writing default configuration to /home/peter/.config/cabal/config
compiler-flavour: ghc
compiler-id: ghc-9.12.4
compiler-abi-tag: ghc-9.12.4-876d
compiler-path: /home/peter/.ghcup/bin/ghc
compiler-store-path: /home/peter/.local/state/cabal/store/ghc-9.12.4-876d
cache-home: /home/peter/.cache/cabal
remote-repo-cache: /home/peter/.cache/cabal/packages
logs-dir: /home/peter/.cache/cabal/logs
store-dir: /home/peter/.local/state/cabal/store
config-file: /home/peter/.config/cabal/config
installdir: /home/peter/.local/bin
```
</details>

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          path: ${{ steps.ghc-install.outputs.cabal-store }}
+
+      - name: Version information
+        run:
+          cabal --version
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install libtinfo-dev tree iverilog -y
```
DigitalBrains1: I don't think you're using IVerilog. Looking at your Stack job, I doubt you need `tree`.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          stack --version
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install libtinfo-dev -y
+
+      - name: Build Dependencies
+        run:
+          stack build -j$(nproc) --pedantic --dependencies-only
```
DigitalBrains1: I don't think dependencies are built with `--pedantic`. It would almost certainly error; even just building the dependencies you can see warnings emitted so they clearly haven't been made errors. So I'm almost certain `--pedantic` is a no-op with `--dependencies-only`.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-${ matrix.ghc }-stack-${ hashFiles( ${{ steps.ghc-install.outputs.stack-root }} ) }
+          path: ${{ steps.ghc-install.outputs.stack-root }}
+
+      - name: Build
+        run:
+          stack build -j$(nproc) --pedantic
```
DigitalBrains1:
```suggestion
      - name: Build
        run:
          # Note: the --pedantic switch adds -Wall -Werror to the options passed to GHC. Options
          # specified in stack.yaml (like -Wcompat) are still passed; it is cumulative. Future
          # versions of Stack might add behavior to --pedantic.
          stack build -j$(nproc) --pedantic
```

As the author of that comment I might be biased, but I think it is worth it to preserve the comment.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-prelude)
+      - run: |
+          nix build -L .#clashPackages-${{matrix.ghc}}.clash-lib
+          attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-lib)
+      - run: |
+          nix build -L .#clashPackages-${{matrix.ghc}}.clash-ghc
+          attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-ghc)
+
+      - name: Cachix
+        if: github.ref == 'refs/heads/master'
```
DigitalBrains1: Shouldn't we publish for `1.10` as well?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+      - uses: actions/checkout@v6
+
+      - name: Authenticate cache
+        env:
+          ATTIC_AUTH_TOKEN: ${{ secrets.ATTIC_SECRET }}
+        run: .ci/attic-auth.sh
+
+      - name: Run clash-testsuite
+        run: |
+          nix run .#clashPackages-${{matrix.ghc}}.clash-testsuite -- -j$(nproc) --hide-successes -p .${{matrix.hdl}} --no-vivado --no-modelsim
```
DigitalBrains1: I'd like to make a little autistic suggestion about something that's been odd to me since forever and we now have the option of fixing. Why would we match on `.${{matrix.hdl}}` instead of on `.${{matrix.hdl}}.`? Surely anchoring it on the right makes as much sense as anchoring it on the left. It has no difference in which tests are selected, but when we are rewriting stuff anyways let's fix this little thing.
```suggestion
          nix run .#clashPackages-${{matrix.ghc}}.clash-testsuite -- -j$(nproc) --hide-successes -p .${{matrix.hdl}}. --no-vivado --no-modelsim
```

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          cabal --version
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install libtinfo-dev tree iverilog -y
+
+      - name: Build dependencies
+        run: |
+          cabal v2-build all --dependencies-only
```
DigitalBrains1: I think you're forgetting to copy `.ci/cabal.project.local`.

DigitalBrains1 (changes requested):
Thanks for iterating on further refinements!

* I think you forgot the mandatory _All jobs finished_ construction
* We should check that external contributions do not trigger these CI jobs as they run on local runners
* Do you intend to keep the GitLab Haddock test and only port that over later? Maybe we can find a way to restore the functionality that suddenly broke quite a while ago. We used to be able to browse the built documentation in an artifact on the web. Unfortunately, GitHub doesn't let you browse artifacts the way GitLab does. [edit] Restoring a browsable version of the artifact is definitely something for a future PR, I just loved that feature so much I can't shut up about it.[/edit]
* Do you intend to port over the checks in `.ci/setup.sh` later? Did you check that the modifications to the Cabal configuration in that script are no longer useful for the new Cabal job you define here?
* Did you check that the `ghc-pkg recache` commands we run on GitLab are no longer necessary? Could very well be, since we now use the modern Cabal directory layout and so we cache different things. Or maybe they're only applicable to the build artifact, not the cache, and we don't have a build artifact anymore. I'd just like it noted explicitly as part of the PR. Please put it in a commit message that gets merged, for posterity.
* Shouldn't you make a job for `.ci/build_clash_dev.sh`?
* Did you deliberately go from just one Stack test to four different snapshots?

DigitalBrains1:
The PR cover letter says:

> Cabal caches only the clash dependencies (with `cabal.project` and `*/*.cabal` hashed as keys).

Don't forget to update that once we've hashed out (heh) what the key should contain.

>  Stack does not utilize caching.

It does now! :-)

DigitalBrains1:
Ah, and there's another check that hasn't been ported:

`.ci/build.sh`:
```sh
# test that we can create a build plan with the index-state in cabal.project
```

Although that test only actually does anything once you copy `.ci/cabal.project.local` :joy:

.github/workflows/on-push.yml
DigitalBrains1: There's something funny going on. Why does `${{ steps.ghc-install.outputs.cabal-store }}` equal `/home/gha/.cabal/store` in practice in the CI run for this PR? That sounds like an old-style layout, where did that come from, shouldn't it be the new style?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+          endpoint: "192.168.102.136"
+          port: 9000
+          insecure: true
+          accessKey: "${{ secrets.S3_ACCESS_KEY }}"
+          secretKey: "${{ secrets.S3_SECRET_KEY }}"
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-${ matrix.ghc }-stack-${ hashFiles( ${{ steps.ghc-install.outputs.stack-root }} ) }
```
DigitalBrains1: `${{ steps.ghc-install.outputs.stack-root }}` seems to be the empty string in practice.

(and now that I have a comment on this line anyway: this is the other instance of where I am wondering why you are trying to hash a directory into the cache key)

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,341 @@
...
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: [
+          9.12.4,
+          9.10.3,
+          9.8.4,
+          9.6.7
+        ]
```
DigitalBrains1: Wait wait wait wait... you're just running the same Stack snapshot four times! I thought you replaced `stack.yaml` by `.ci/stack-X.Y.yaml` but you don't, so you just run the same snapshot every time!

FWIW, I think we can just keep it as on GitLab and only run one snapshot, and keep running the four snapshots on macOS and Windows runners (although currently we don't run any Windows because it is slow as molasses in January).

DigitalBrains1: If you use `stack-no-global: true` for `haskell-actions/setup` that'll speed things up by avoiding installing stuff we're not using anyway.

DigitalBrains1: On macOS and Windows, we have this:

https://github.com/clash-lang/clash-compiler/blob/df6369060086e126d8a4cc5e6f77622aa190f71c/.github/workflows/ci.yml#L94-L98

Maybe we should use that for Linux as well?

Although `shell: bash` would be redundant, please don't copy that line :-)

.github/workflows/on-push.yml
DigitalBrains1: I kinda get the impression the CI jobs are taking the name of the key rather literally...

```
key: Linux-${ matrix.ghc }-cabal-${ hashFiles( /home/gha/.cabal/store, "~/.cache/cabal" ) }
```

It substitutes `runner.os` and `...cabal-store` but not the others. Ah... aren't you missing a whole bunch of curly brackets?
```suggestion
          key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-${{ hashFiles( ${{ steps.ghc-install.outputs.cabal-store }}, "~/.cache/cabal" ) }}
```

DigitalBrains1:
@martijnbastiaan Something to think about as well...

We will run `clash-testsuite` in Nix now. And we build with the default Nix options and configuration. But on GitLab, with Cabal, we build `clash-lib` with the Cabal flag `debug` asserted, which means by the time we run `clash-testsuite`, it will use a `clash-lib` which has some extra debugging enabled, which we lose with the current Nix setup. I'm inclined to do a separate build for `clash-testsuite` specifically, which has this Cabal flag asserted, so we get the extra validation that the flag enables in `clash-lib`.

~~(Conversely, I think we can drop the `debug` flag from `.ci/cabal.project.local` as we will only do a build test now in Cabal.)~~

[edit]
I discussed it with @martijnbastiaan and he agreed it'd be good to not only build the packages with the regular flags, but also build a version where `clash-lib` has the `debug` asserted, and use that in `clash-testsuite`. He also pointed out the `debug` flag is already in `cabal.project`, so my suggestion of removing it from `cabal.project.local` is moot; scratch that.
[/edit]

.github/workflows/on-push.yml
jaschutte: Oh you're completely right, I don't know how I didn't notice that! I suppose I just saw 'cache restored succesfully' and moved on haha.

jaschutte:
> Do you intend to keep the GitLab Haddock test and only port that over later? Maybe we can find a way to restore the functionality that suddenly broke quite a while ago. We used to be able to browse the built documentation in an artifact on the web. Unfortunately, GitHub doesn't let you browse artifacts the way GitLab does. [edit] Restoring a browsable version of the artifact is definitely something for a future PR, I just loved that feature so much I can't shut up about it.[/edit]

Iirc Nix already runs these tests when the Nix packages get built, I didn't see the need to re-run haddock tests in Cabal

DigitalBrains1:
> Iirc Nix already runs these tests when the Nix packages get built

We manually grep for warnings emitted by Haddock to detect issues. I don't think Nix does that (it'd be a fun feature except for those pesky false positives!). I think you misinterpreted my words _the GitLab Haddock test_ as _doctests_, which are completely different concepts. I was referring to the CI test script we wrote. Please see https://github.com/clash-lang/clash-compiler/blob/f002b5612333963d06e3e649b80b9a5fc5db16cb/.gitlab-ci.yml#L71-L80 and [`.ci/build_docs.sh`](https://github.com/clash-lang/clash-compiler/blob/f002b5612333963d06e3e649b80b9a5fc5db16cb/.ci/build_docs.sh).

jaschutte:
Ah I see, yes I did misinterpret! I will add those.
You can follow some of my progress [in my fork](https://github.com/jaschutte/rich-clash-compiler/tree/refs/heads/gl-to-gh-migration-on-push-localfork), I will bring over the commits over to this branch once I got it not erroring constantly.

jaschutte:
> Did you check that the ghc-pkg recache commands we run on GitLab are no longer necessary? Could very well be, since we now use the modern Cabal directory layout and so we cache different things. Or maybe they're only applicable to the build artifact, not the cache, and we don't have a build artifact anymore. I'd just like it noted explicitly as part of the PR. Please put it in a commit message that gets merged, for posterity.

I checked just now- it seems to work fine without ghc-pkg recache. Although as you mention, there's no build artifact right now so maybe that changes things? For now though it seems to grab (and use) the cache without problems.

.github/workflows/on-push.yml
jaschutte: What does this do exactly? Also looking at the issue linked it seems to have been closed in 2024. I've not seen stack fail to initialize either, so I think this has been resolved?

EDIT: I will change the stack job to only run one GHC version with stack (with only one snapshot). And if I stand correctly we will keep the on gitlab for now?

DigitalBrains1: > What does this do exactly?

It retries up to 20 times to properly initialize Stack, to work around an issue downloading the package list from Hackage. I think `build base` is a no-op as it is a bundled library; it just keeps prodding Stack to do its initial actions. Why on a single CPU? No idea.

> Also looking at the issue linked it seems to have been closed in 2024.

Without any reason given. The reason might have also been "this is a bug in another program", which is what is said in that issue report. So it doesn't necessarily mean the issue is fixed, just that it's closed.

> I've not seen stack fail to initialize either, so I think this has been resolved?

I don't think it occurred all that often... and we've lately seen issues at certain times of day, probably related to overloading of the Hackage web server. But we can keep it out for now if you really want. I would keep it, since it is so trivial and the only extra load it gives is the extra "step" in the CI job, which I believe might do some Docker thingies. Quite negligible.

> And if I stand correctly we will keep the on gitlab for now?

I do not understand what you mean, could you phrase it differently?

Ah... re-reading my own initial message reveals an almost but not quite identically phrased... phrase. What I was suggesting, was to do in the new GitHub CI for Stack the same we have done on GitLab: run one Stack test with the `stack.yaml` in the root of the repo. And keep the Windows and macOS jobs we already have on GitHub to test the other Stack snapshots we keep in `.ci/stack-*.yaml`.

DigitalBrains1: Hm. It suddenly occurs to me we don't do the retrying on GitLab and that generally works out nonetheless. I was looking at both the GitLab and the GitHub way of running Stack and noticed that step, which seemed useful. Well... I still think it is a harmless thing that might have a use, so why not do it.

jaschutte:
In the latest set of commits I have:
- Rebased to be up-to-date with master
- Hardcoded the stack root path, as stack-root was indeed being set to nil
- Changed the caching keys from being output dependent (I don't know why I thought this would work) to being dependent on select input files
- Made cachix also cache on 1.10 branch (untested)
- Added the 'all' job (modified the all_checks.py script to read in an environment variable for what .yml script to check)
- Only having one Stack template being tested on linux
- Plenty of more minor changes (such as removing the --pendantic in combination with --dependencies-only flag) 

Writing a quick TODO list:

- [x] Fix the Haddock tests failing (I'm not quite sure why this is happening maybe the dependencies are being generated without documentation? When I run the original `build_docs.sh` script locally it fails too.)
- [x] Add the Stack waiting job
- [x] ~~Make only contributors be able to run this workflow (was this not an option inside of Github repos? I'll go and double check)~~
- [x] Rename the all job so we don't have 2 different 'all completed' jobs
- [x] Update the opening PR message to reflect the current state of the PR

If I missed anything, let me know

[edit]
I checked our repository settings and workflows already require explicit permission to run on commits and forks for outside contributors. I don't think this something you can do per-workflow basis either. Well, not *easily*. You can of course use the web API and conjure something up like that but I don't think that's necessary. 
[/edit]

DigitalBrains1:
> [edit]
> I checked our repository settings and workflows already require explicit permission to run on commits and forks for outside contributors.

Right, I think we changed that recently in anticipation of what we do in this PR. It did not use to be that way. I was asking to make sure we had actually did that, and I guess we did. We can do a normal PR in the form of an external PR one of these days and see that it doesn't trigger the new CI. But GitHub does have more distinctions between first-time and recurring contributors and stuff like that, so we can only test the "external" bit easily.

> I don't think this something you can do per-workflow basis either. Well, not _easily_. You can of course use the web API and conjure something up like that but I don't think that's necessary.
> [/edit]

It is a pity we lose the _Build and Test_ CI in that case, which we had explicitly set up to run on external contributions so you have quick feedback even before one of the people with privileges kicks off the GitLab CI.

jaschutte:
Okay! I think I've addressed every commit in the latest patch of commits. It *should* be ready to go now (assuming CI passes haha).

.ci/all_check.py
DigitalBrains1, line 1: This file can be removed now, right?

.ci/gh-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+sudo apt-get update
+sudo apt-get install libtinfo-dev -y
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
+CI_DIR=$GIT_ROOT/.ci
+cd CI_DIR
+
+# Get the proper amount of cores we can use and configure ghc to use them
+THREADS=$($CI_DIR/effective_cpus.sh)
+sed "7s/-j4/-j$THREADS/g" cabal.project.local
```
DigitalBrains1: If I look at the _intention_ rather than the actual code, I find this rather brittle. I'd rather do it like I wrote for `clash-protocols`:
https://github.com/clash-lang/clash-protocols/blob/main/.ci/cabal.project.local.in
https://github.com/clash-lang/clash-protocols/blob/main/.ci/apply_settings.sh

So we're not trying to match on some text that somebody might change, like the `-j4`. My suggestion is here:
https://github.com/clash-lang/clash-compiler/tree/peter/gl-to-gh-migration-on-push
It includes two commits dealing with separate issues, both are separate suggestions. Also, some flags in `.ci/cabal.project.local` are a bit outdated. I think all the lines `flags:` can go. They either deal with the mmap crash which we don't deal with on CI anymore, or they specify flags that are either already in cabal.project or simply the default value of the flag anyway.

Apart from that, there are several mistakes here. One is that you specify it should happen on line 7 only, but line 7 does not contain the mentioned text at all. Secondly, in fact, `.ci/cabal.project.local` does not contain a `-jN` option at all... Thirdly, you forgot a dollar sign in the `cd` command a few lines up, so this is still running in the root of the repo and not finding the file anyway.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+
+  cabal-haddock:
+    name: Haddock tests
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        # Documentation is built at 9.6.7, testing on other versions therefore does not make a lot
+        # of sense
+        ghc: [ 9.6.7 ]
```
DigitalBrains1: So why keep it a matrix? Just put it in
```
      - name: Install GHC
        id: ghc-install
        uses: haskell-actions/setup@v2
        with:
          # Documentation is built at 9.6.7, testing on other versions therefore does not make a lot
          # of sense
          ghc-version: 9.6.7
          cabal-version: latest
```

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+          PKG: ${{ matrix.pkg }}
+        run: .ci/build_docs.sh
+
+  stack:
+    name: Stack compilation test
+    runs-on: self-hosted
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: [ 9.12.4 ]
```
DigitalBrains1: I'd go even further than "so why keep it a matrix" here. It is downright misleading! Stack actually uses GHC 9.10.3,  and this is dictated by `stack.yaml`. This variable is meaningless, and correcting it to 9.10.3 doesn't improve it much because someone will have to keep this variable which has no effect on Stack up-to-date with the actual version used by Stack. Please remove it.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+      # Perhaps we should preinstall these on the container?
+      # This step always takes a minute...
+      # We'd lose the ability to select the version of cabal inside of the runner though
+      - name: Install GHC
+        uses: haskell-actions/setup@v2
+        with:
+          ghc-version: ${{ matrix.ghc }}
+          enable-stack: true
+          stack-no-global: true
+          stack-version: latest
```
DigitalBrains1: I think `ghc-version` is a no-op in combination with `stack-no-global: true`. Please remove it.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+          endpoint: '192.168.102.136'
+          port: 9000
+          insecure: true
+          accessKey: '${{ secrets.S3_ACCESS_KEY }}'
+          secretKey: '${{ secrets.S3_SECRET_KEY }}'
+          bucket: runner
+          region: local
+          use-fallback: false
+          retry: false
+          key: ${{ runner.os }}-${{ github.ref_name }}-stack-${{ matrix.ghc }}-${{ hashFiles('.ci/cabal.project.local', '.github/workflow/on-push.yml', 'cabal.project', '**/*.cabal' ) }}
```
DigitalBrains1: Stack isn't affected by `${{ matrix.ghc}}`, `.ci/cabal.project.local` or `cabal.project`. It _is_ however strongly influenced by `stack.yaml`. 

```suggestion
          key: ${{ runner.os }}-${{ github.ref_name }}-stack-${{ hashFiles('stack.yaml', '.github/workflow/on-push.yml', '**/*.cabal' ) }}
```

This suggestion has been applied in my suggestion branch because I had to change the line anyway.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+      - name: Local cache
+        run: attic push public $(nix path-info .#devShells.x86_64-linux.${{matrix.ghc}})
+
+      - name: Cachix
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/v1.10.0'
+        env:
+          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_SECRET }}
+        run: nix path-info .#devShells.x86_64-linux.${{matrix.ghc}} | cachix push clash-lang
+
+  packages_testsuite:
```
DigitalBrains1, line 348: Could you please have `clash-testsuite` use a `clash-lib` that has been built with `flags: +debug` as suggested [here](https://github.com/clash-lang/clash-compiler/pull/3289#issuecomment-5141875644)?

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+          attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-prelude)
+      - run: |
+          nix build -L .#clashPackages-${{matrix.ghc}}.clash-lib
+          attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-lib)
+      - run: |
+          nix build -L .#clashPackages-${{matrix.ghc}}.clash-ghc
+          attic push public $(nix path-info .#clashPackages-${{matrix.ghc}}.clash-ghc)
+
+      - name: Cachix
+        if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/v1.10.0'
```
DigitalBrains1: Ah, this won't work. There isn't a branch named `v1.10.0`, that's a tag. And we treat the tags as immutable, so we will never again "push" to that tag, so there's no trigger anyway.

I was asking whether we also wanted to push to Cachix for pushes to the `1.10` branch. Note I'm asking, not suggesting. I'm fine with not doing it, I just wanted you to consider it.

[edit]
Ah no, I see I did actually _suggest_ rather than _ask_. Trying to remember why, I think back then I thought I remembered discussing this and that we were inclined to add it. But perhaps I'm simply confused with the decision to push actual _releases_, not the `1.10` branch; I just don't know anymore.
[/edit]

DigitalBrains1 (changes requested):
So close, but still not done :-)

DigitalBrains1:
Ah, also, some flags in `.ci/cabal.project.local` are a bit outdated. I think all the lines `flags:` can go. They either deal with the mmap crash which we don't deal with on CI anymore, or they specify flags that are either already in `cabal.project` or simply the default value of the flag anyway.

.ci/gh-setup.sh
DigitalBrains1: Please consider the commit messages in that [peter/gl-to-gh-migration-on-push](https://github.com/clash-lang/clash-compiler/tree/peter/gl-to-gh-migration-on-push) part of the review :-).

.ci/all_check.py
jaschutte: No since regular (windows,mac) CI step still uses it

.ci/gh-setup.sh
jaschutte: I fully agree with this, it's much more robust than just relying on `-j4` never being modified. Thank you very much! (I'm shamelessly going to cherry picking your commits >:) )

.github/workflows/on-push.yml
jaschutte: I think it makes sense to a cache a release anyhow, so I would've agreed anyhow :)

.ci/all_check.py
DigitalBrains1: Okay but then it can be kept unchanged. There's no need for this PR to touch this file.

.github/workflows/on-push.yml
```diff
@@ -0,0 +1,405 @@
...
+      - name: Version information
+        run:
+          cabal --version
+
+      - name: Setup
+        run: .ci/gh-setup.sh
+
+      - name: Build & check documentation
+        env:
+          PKG: ${{ matrix.pkg }}
```
DigitalBrains1: This env var is a remnant of an earlier incarnation of this job; it can be removed.

DigitalBrains1 (approved):
:tada:

martijnbastiaan:
Should I remove GitLab as a required check? :)

DigitalBrains1:
Let's wait until this is merged. Then let's race who's the first to remove the requirement! :racehorse:

PR #3296, opened 2026-06-12
Limit Clash invocations to 2G of RAM in testsuite

martijnbastiaan:
I keep making this patch locally :-).

## Still TODO:


rowanG077 (approved):

martijnbastiaan:
Btw, this limit was chosen somewhat arbitrary. I just picked "1G" and all tests still passed. If it turns out that some debugging flags would push it over, it also makes sense to me to just bump it to "2G".

martijnbastiaan:
Turns out older GHCs are less memory efficient. Bumping it to 2G!

PR #3297, opened 2026-06-12
Prune closure + parallelize closure loading

martijnbastiaan:
Time measurements on `bittide`:

| Patch                     | Cumulative loading time |
|---------------------------|-------------------------|
| Baseline                  | 22.135s                 |
| Prune                     | 19.357                  |
| Prune + parallel GHC2Core | 11.897s                 |

This does not offer any performance benefits for `clash-testsuite`.

For bittide: pruning the core results in slightly different HDL (names). The parallelization patch does not change the HDL at all.

## Still TODO:

  - [x] Patch cleanup
  - [X] Cleanup commit messages
  - [x] Test on more GHCs

### DIFF
--- a/changelog/2026-06-12T12_52_51+02_00_prune_closure_before_loading
+++ b/changelog/2026-06-12T12_52_51+02_00_prune_closure_before_loading
@@ -1,0 +1,1 @@
+ADDED: Clash now prunes the binders it collects from top modules when loading from precompiled core. For larger projects, this can shave off 15% of package loading times. See [#3298](https://github.com/clash-lang/clash-compiler/issues/3298).
--- a/changelog/2026-06-12T13_01_29+02_00_parallelize_ghc2core
+++ b/changelog/2026-06-12T13_01_29+02_00_parallelize_ghc2core
@@ -1,0 +1,1 @@
+ADDED: Clash now converts GHC Core to its own Clash Core in parallel. For larger projects, this can shave off ~40% of load times. See [#3299](https://github.com/clash-lang/clash-compiler/issues/3299).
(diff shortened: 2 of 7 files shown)


rowanG077 (commented):
Hmmm I'm not too sure whether we should piggyback of `main-is` as is.

The reason for that is we don't really use the main terminology but rather `topEntity`.

The filter is also not good enough yet as the code below synthesized with `-main-is topEntity`

```Haskell
import Clash.Prelude

topEntity :: Bool -> Bool
topEntity = not
{-# OPAQUE topEntity #-}

-- This magic top-entity name is intentionally outside 'topEntity's closure.
-- With -main-is topEntity, Clash should not require this binding to be loaded.
testBench = pure True
{-# OPAQUE testBench #-}
``` 

Fails with:

```
No top entity called 'Name {nameSort = User, nameOcc = "MainIsPrunedMagic.testBench", nameUniq = 8214565720323787838, nameLoc = UnhelpfulSpan UnhelpfulNoLocationInfo}
' found. Make sure you are compiling with the '-fexpose-all-unfoldings' flag.
```

martijnbastiaan:
> Hmmm I'm not too sure whether we should piggyback of main-is as is.

Fair! Though that's already in `master`, this PR doesn't introduce that.

martijnbastiaan:
I've amended the filter, but I've got to think a little bit about whether the current approach makes sense as it feels a bit like a bandaid. Still, if you feel it's fine anyway, then we can merge (after `fixup`ing the last commit).

martijnbastiaan:
Current patches LGTM @rowanG077.

rowanG077 (approved):

PR #3315, opened 2026-07-06
Fix vhdl prims integerToNaturalThrow and integerToNaturalClamp

leonschoorl:
Somehow some Verilog literal syntax slipped into these VHDL prims
Along with a stray `]`.
And these primitives were never tested.

All this is now fixed and these primitives are now included in the existing Naturals test.

## Still TODO:

  - [ ] ~Check copyright notices are up to date in edited files~


rowanG077 (approved):

rowanG077:
Could you add a regression test for this?

martijnbastiaan:
Plus probably an issue report for the changelog?

leonschoorl:
@martijnbastiaan Could it be ok to like to the PR instead?
It feels a bit pointless to open an issue after the fact

rowanG077:
For me it's not a must to have an issue per PR. It's of course better and cleaner but to make an issue after the fact seems redundant. Especially since it's very clear here what the problem is.

martijnbastiaan:
I like it because it's a natural place to show symptons, which are then Googlable. Also, to me it's usually nice to have an obvious place to add a regression test. Anyway, just my thoughts.

PR #3320, opened 2026-07-10
Align `rom` interfaces

kleinreact:
Updates the interface of `Clash.Prelude.ROM.rom` to be in line with the one of `Clash.Explicit.ROM.rom`.

## Still TODO:

### DIFF
--- a/changelog/2026-07-10T16_32_36+02_00_fix_rom_interface
+++ b/changelog/2026-07-10T16_32_36+02_00_fix_rom_interface
@@ -1,0 +1,1 @@
+CHANGED: The interface of `Clash.Prelude.ROM.rom` has been aligned with the one of `Clash.Explicit.ROM.rom`. Arguments are no longer `Unsigned`, but `Enum` based.
--- a/clash-prelude/src/Clash/Prelude/ROM.hs
+++ b/clash-prelude/src/Clash/Prelude/ROM.hs
@@ -2,7 +2,7 @@
 Copyright  :  (C) 2015-2016, University of Twente,
                   2017     , Google Inc.
                   2019     , Myrtle Software Ltd,
-                  2021-2022, QBayLogic B.V.
+                  2021-2026, QBayLogic B.V.
 License    :  BSD2 (see the file LICENSE)
 Maintainer :  QBayLogic B.V. <devops@qbaylogic.com>
 
@@ -131,6 +131,9 @@ asyncRom# content = safeAt
 -- * __NB__: Initial output value is /undefined/, reading it will throw an
 -- 'Clash.XException.XException'
 --
+-- This primitive is particularly designed to enable mapping to a block RAM by
+-- downstream synthesis tools.
+--
 -- === See also:
 --
 -- * See "Clash.Sized.Fixed#creatingdatafiles" and "Clash.Prelude.BlockRam#usingrams"
@@ -139,17 +142,17 @@ asyncRom# content = safeAt
 -- is constructed. See 'Clash.Prelude.ROM.File.romFile' and
 -- 'Clash.Prelude.ROM.Blob.romBlob' for different approaches that scale well.
 rom
-  :: forall dom n m a
-   . ( NFDataX a
+  :: forall dom n addr a
+   . ( Enum addr
+     , NFDataX a
      , KnownNat n
-     , KnownNat m
      , HiddenClock dom
      , HiddenEnable dom  )
   => Vec n a
   -- ^ ROM content, also determines the size, @n@, of the ROM
   --
   -- __NB__: __MUST__ be a constant
-  -> Signal dom (Unsigned m)
+  -> Signal dom addr
   -- ^ Read address @r@
   -> Signal dom a
   -- ^ The value of the ROM at address @r@ from the previous clock cycle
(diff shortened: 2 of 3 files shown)


martijnbastiaan:
I've made suggestions like this before (make `!!` operate on `Index`), it wasn't met with enthusiasm. I can't find it anywhere though. Perhaps you can open a general discussion on what we want with functions like this?

-----

As far as my review goes for this PR:

* Have you checked whether synthesis tools still infer efficient RAMs after this change? They can be very picky.
* Could you amend the PR title to something like ``Use `Index n` in `rom` and friends``?

DigitalBrains1:
Literally the only two memories that don't use `Enum addr` are the way newer `trueDualPortBlockRam` and `Clash.Prelude.ROM.rom`. Even `Clash.Explicit.ROM.rom` uses `Enum addr`! It seems that commit https://github.com/clash-lang/clash-compiler/commit/31cfd6dd63777d56307c9ff67b8706ab30ca380b changed them all from `Unsigned m` to `Enum addr` but missed precisely this single one.

So it seems to me that if we change anything, we should either change `Clash.Prelude.ROM.rom` to use `Enum addr` as was probably intended to get it on par with the other functions, or we change all those `Enum addr`s to `Index`, which indeed has come up before and was not universally praised.

My personal opinion, btw, is that `Index` would be better for these memories. I can't recall if changing `Enum` to `Index` for memories specifically was something people objected to? For that, a general discussion might be a good idea.

The memories that are an exact power of two use a fully constrained `Unsigned` as address lines, btw. I'm just considering "arbitrarily-sized" memories.

kleinreact:
Thanks for the feedback. I primarily was concerned about the confusing `Unsigned m` type interface. I updated the PR such that `Clash.Prelude.ROM.rom` now is in line with `Clash.Explicit.ROM.rom`. 

I still would prefer an `Index`-based interface here, but I totally agree that such changes deserves a more elaborated analysis and discussion at first. In a nutshell, I think that a Clash user should never be enforced to work with types like `Int` or `Integer`, which simply are out-of-place when it comes to hardware development. Although it might look more familiar to beginners at first, I still think that using proper bus sizes is one of the crucial points that you have to learn when working in this area right from the start.

Anyway, more personal opinions to be found in a dedicated discussion. This PR is intended to be minimally invasive in that regard.

rowanG077:
I agree that index is much nicer. I have also gotten triggered in the past by `(!!)`, `gather`,  `rotate` and `shift` family for `Vec` among others. Just like @martijnbastiaan I vaguely remember changing them was not wanted. But to be clear I would prefer `Index` for these functions.

martijnbastiaan:
I've opened a discussion + poll here: https://github.com/clash-lang/clash-compiler/discussions/3321.

changelog/2026-07-10T16_32_36+02_00_fix_rom_interface
```diff
@@ -0,0 +1 @@
+CHANGED: The interface of `Clash.Prelude.ROM.rom` has been aligned with the one of `Clash.Explicit.ROM.rom`.
```
martijnbastiaan:
```suggestion
CHANGED: The interface of `Clash.Prelude.ROM.rom` has been aligned with the one of `Clash.Explicit.ROM.rom`. Arguments are no longer `Unsigned`, but `Enum` based.
```

clash-prelude/src/Clash/Prelude/ROM.hs
```diff
@@ -130,6 +130,8 @@ asyncRom# content = safeAt
 -- * __NB__: Read value is delayed by 1 cycle
 -- * __NB__: Initial output value is /undefined/, reading it will throw an
 -- 'Clash.XException.XException'
+-- * __NB__: The primitive is particularly designed for being mapped to a block ram by
+-- downstream synthesis tools.
```
DigitalBrains1: The documentation for
* `Clash.Explicit.ROM.rom`
* `Clash.Explicit.ROM.romPow2`
* `Clash.Prelude.ROM.rom`
* `Clash.Prelude.ROM.romPow2`

is, unsurprisingly, all virtually identical. Could you please add this comment to them all instead of just this one?

As I'm requesting changes anyway, I'd like to suggest this instead:

```suggestion
--
-- This primitive is particularly designed to enable mapping to a block RAM by
-- downstream synthesis tools.
```

I don't think this is a caveat like the other _NB_ s are. Also, a synthesis tool doesn't _necessarily_ map it to a block RAM; it just makes it possible if that is an efficient result.

DigitalBrains1 (changes requested):
The code looks good to me, thanks! But I'd like to suggest some changes to the new Haddock.

martijnbastiaan (approved):
Very small comment, otherwise LGTM.

~~Did you test whether this still gets inferred to efficient memories? Especially for VHDL where the indexing gets its own type I worry it might not -- because synthesis tooling sucks.~~ Ah nevermind, the other memories already do this, so I guess it's fine. (Also probably worth an issue stating we should have an automated test for this.)

clash-prelude/src/Clash/Prelude/ROM.hs
kleinreact: Great point. Added the note to all of them.

DigitalBrains1 (approved):

PR #3323, opened 2026-07-13
Bunch of performance fixes

martijnbastiaan:
I put Fable to work over the last 24 hours and it came up with these patches. I've confirmed that it passes the `clash-testsuite` on all supported GHCs and that it produces bit-for-bit equal HDL at every commit over at `bittide-hardware`. Benefits measured on a 4-core VM running on a AMD Ryzen 7 3700X (`wireDemoTest` @ `bittide-hardware`).

Per commit measurements:

| patch | runtime | net diff | net diff % |
|-------|--------:|---------:|-----------:|
| (base: master) | 8m23s | — | — |
| Evaluator: stop projecting the global binding map on every whnf' call | 6m07s | −136s | **−27.0%** |
| Evaluator: peel literal unconcatBitVector# without residual split# calls | 5m14s | −53s | **−14.4%** |
| ~~Dispatch rewrite bundles on the node constructor~~ | ~~4m18s~~ | ~~−56s~~ | ~~**−17.8%**~~ |
| Skip reduceConst on inner nodes of application spines | 4m08s | −10s | **−3.9%** |
| Cache Core-type to HWType translation across the rewrite session | 4m04s | −4s | **−1.6%** |
| Skip removeUnusedExpr on inner nodes of application spines | 3m56s | −8s | **−3.3%** |
| Speed up GHC Core to Clash Core conversion | 3m44s | −12s | **−5.1%** |

Every row produced bit-identical HDL to the previous branch tip. Total: 8m23s -> 3m44s (2.25x).

Edit: I've kicked one commit out due to maintainability issues, but I haven't updated the measurements. Overall picture still applies though.

## Still TODO:

  - [X] Go over it manually (@martijnbastiaan)
  - [x] Split out performance benefits per commit
  - [X] Fix up the annoying AI habit of leaving code comments that talk about the past ("this piece of code is the same as it was before", OH THANKS)
  - [X] Trim the references to `bittide-hardware` as a concept that every Clash developer is just supposed to know
  - [X] Decide whether to leave timing instrumentation in
  - [x] Remove "Dispatch rewrite bundles on the node constructor" (PR will follow)
  - [x] Touch up commit messages and documentation to be more concise / less LLMy

### DIFF
--- a/changelog/2026-07-12T22_45_00+02_00_evaluator_no_global_heap_copy
+++ b/changelog/2026-07-12T22_45_00+02_00_evaluator_no_global_heap_copy
@@ -1,0 +1,1 @@
+CHANGED: The compile-time evaluator no longer projects the entire global binding map into a fresh heap on (virtually) every invocation. We've observed a 25% Clash compile speedups for larger designs.
--- a/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
@@ -4580,38 +4580,51 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
          n' | Right m <- runExcept (tyNatSize tcm mTy) ->
           let Just vecTc  = UniqMap.lookup vecTcNm tcm
               [_,consCon] = tyConDataCons vecTc
-              tupTcNm     = ghcTyconToTyConName (tupleTyCon Boxed 2)
-              Just tupTc  = UniqMap.lookup tupTcNm tcm
-              [tupDc]     = tyConDataCons tupTc
-              splitCall   =
-                mkApps (bvSplitPrim bvTcNm)
-                       [ Right (mkTyConApp typeNatMul [LitTy (NumTy (n'-1)),mTy])
-                       , Right mTy
-                       , Left (Literal (NaturalLiteral ((n'-1)*m)))
-                       , Left (valToTerm bv)
-                       ]
               mBVTy       = mkTyConApp bvTcNm [mTy]
-              n1BVTy      = mkTyConApp bvTcNm
-                              [mkTyConApp typeNatMul
-                                [LitTy (NumTy (n'-1))
-                                ,mTy]]
-              -- Guaranteed no capture, so okay to use unsafe name generation
-              xNm         = mkUnsafeSystemName "x" 0
-              bvNm        = mkUnsafeSystemName "bv'" 1
-              xId         = mkLocalId mBVTy xNm
-              bvId        = mkLocalId n1BVTy bvNm
-              tupPat      = DataPat tupDc [] [xId,bvId]
-              xAlt        = (tupPat, (Var xId))
-              bvAlt       = (tupPat, (Var bvId))
+              n1MTy       = mkTyConApp typeNatMul [LitTy (NumTy (n'-1)),mTy]
+              n1BVTy      = mkTyConApp bvTcNm [n1MTy]
 
-          in  reduce $ mkVecCons consCon (mkTyConApp bvTcNm [mTy]) n'
-                (Case splitCall mBVTy [xAlt])
+              (hd, tl) = case bitVectorLiteral bv of
+                -- Fast path for literals
+                Just (mski, i) ->
+                  let sh     = fromInteger ((n'-1) * m)
+                      loMask = bit sh - 1
+                  in  ( mkBitVectorLit mBVTy mTy m
+                          (mski `shiftR` sh) (i `shiftR` sh)
+                      , mkBitVectorLit n1BVTy n1MTy ((n'-1)*m)
+                          (mski .&. loMask) (i .&. loMask)
+                      )
+                Nothing ->
+                  let tupTcNm     = ghcTyconToTyConName (tupleTyCon Boxed 2)
+                      Just tupTc  = UniqMap.lookup tupTcNm tcm
+                      [tupDc]     = tyConDataCons tupTc
+                      splitCall   =
+                        mkApps (bvSplitPrim bvTcNm)
+                               [ Right n1MTy
+                               , Right mTy
+                               , Left (Literal (NaturalLiteral ((n'-1)*m)))
+                               , Left (valToTerm bv)
+                               ]
+                      -- Guaranteed no capture, so okay to use unsafe name
+                      -- generation
+                      xNm         = mkUnsafeSystemName "x" 0
+                      bvNm        = mkUnsafeSystemName "bv'" 1
+                      xId         = mkLocalId mBVTy xNm
+                      bvId        = mkLocalId n1BVTy bvNm
+                      tupPat      = DataPat tupDc [] [xId,bvId]
+                      xAlt        = (tupPat, (Var xId))
+                      bvAlt       = (tupPat, (Var bvId))
+                  in  ( Case splitCall mBVTy [xAlt]
+                      , Case splitCall n1BVTy [bvAlt]
+                      )
+
+          in  reduce $ mkVecCons consCon mBVTy n' hd
                 (mkApps (Prim pInfo)
                         [ Right (LitTy (NumTy (n'-1)))
                         , Right mTy
                         , Left (Literal (NaturalLiteral (n'-1)))
                         , Left (valToTerm km)
-                        , Left (Case splitCall n1BVTy [bvAlt])
+                        , Left tl
                         ])
          _ -> Nothing
   "Data.Text.Show.$wunpackCStringAscii#"
--- a/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
@@ -1,8 +1,8 @@
 {-|
   Copyright   :  (C) 2013-2016, University of Twente,
                      2016-2017, Myrtle Software Ltd,
-                     2017-2022, Google Inc.
-                     2021-2025, QBayLogic B.V.,
+                     2017-2022, Google Inc.,
+                     2021-2026, QBayLogic B.V.
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
 -}
@@ -79,7 +79,7 @@ import GHC.Core.FamInstEnv
   ( FamInst (..), FamInstEnvs
   , familyInstances, normaliseType, emptyFamInstEnvs, topReduceTyFamApp_maybe
   )
-import GHC.Data.FastString (unpackFS, bytesFS)
+import GHC.Data.FastString (FastString, mkFastString, unpackFS, bytesFS)
 import GHC.Types.Id (isDataConId_maybe)
 import GHC.Types.Id.Info (IdDetails (..), unfoldingInfo)
 import GHC.Types.Literal (Literal (..), LitNumType (..), literalType)
@@ -87,7 +87,7 @@ import GHC.Unit.Module (moduleName, moduleNameString)
 import GHC.Types.Name
   (Name, nameModule_maybe, nameOccName, nameUnique, getSrcSpan)
 import GHC.Builtin.Names  (integerTyConKey, naturalTyConKey)
-import GHC.Types.Name.Occurrence (occNameString)
+import GHC.Types.Name.Occurrence (occNameFS, occNameString)
 import GHC.Data.Pair (Pair (..))
 import GHC.Types.SrcLoc (SrcSpan (..), isGoodSrcSpan)
 import GHC.Core.TyCon
@@ -131,6 +131,9 @@ data GHC2CoreState
   = GHC2CoreState
   { _tyConMap :: C.UniqMap TyCon
   , _nameMap  :: HashMap Name Text
+  , _varTypeMap :: HashMap Name C.Type
+  -- ^ Cache for converted types of global variables ('varType'). See
+  -- 'coreToVarType'.
   }
 
 makeLenses ''GHC2CoreState
@@ -144,7 +147,7 @@ data GHC2CoreEnv
 makeLenses ''GHC2CoreEnv
 
 emptyGHC2CoreState :: GHC2CoreState
-emptyGHC2CoreState = GHC2CoreState mempty HashMap.empty
+emptyGHC2CoreState = GHC2CoreState mempty HashMap.empty HashMap.empty
 
 newtype SrcSpanRB = SrcSpanRB {unSrcSpanRB :: SrcSpan}
 
@@ -505,7 +508,7 @@ coreToTerm primMap unlocs = term
     var x = do
         xPrim <- if isGlobalId x then coreToPrimVar x else coreToVar x
         let xNameS = C.nameOcc xPrim
-        xType  <- coreToType (varType x)
+        xType  <- coreToVarType x
         case isDataConId_maybe x of
           Just dc -> case lookupPrim xNameS of
             Just p  ->
@@ -867,14 +870,21 @@ coreToAttrs' [k, a, attrs] = do
 coreToAttrs' illegal =
   error $ "Unexpected type args to Annotate: " ++ show (map (showPprUnsafe) illegal)
 
+annotateOccFS :: FastString
+annotateOccFS = mkFastString "Annotate"
+{-# NOINLINE annotateOccFS #-}
+
 -- | If this type has an annotate type synonym, return list of attributes.
 coreToAttrs :: Type -> C2C [Attr Text]
-coreToAttrs (TyConApp tycon kindsOrTypes) = do
-  name' <- typeConstructorToString tycon
+coreToAttrs (TyConApp tycon kindsOrTypes)
+  -- Cheap pre-check on the interned occurrence name; only build the qualified
+  -- name when it can possibly be Clash.Annotations.SynthesisAttributes.Annotate.
+  | occNameFS (nameOccName (tyConName tycon)) == annotateOccFS = do
+      name' <- typeConstructorToString tycon
 
-  if name' == show ''Annotate
-  then coreToAttrs' kindsOrTypes
-  else return []
+      if name' == show ''Annotate
+      then coreToAttrs' kindsOrTypes
+      else return []
 
 coreToAttrs _ =
     return []
@@ -919,7 +929,7 @@ coreToType' (TyConApp tc args)
                         foldl C.AppTy <$> coreToType synTy' <*> mapM coreToType remArgs
                       _ -> do
                         tcName <- coreToName tyConName tyConUnique qualifiedNameString tc
-                        tyConMap %= (C.insert tcName tc)
+                        tyConMap %= C.insertIfAbsent tcName tc
                         C.mkTyConApp <$> (pure tcName) <*> mapM coreToType args
 coreToType' (ForAllTy (Bndr tv _) ty)   = C.ForAllTy <$> coreToTyVar tv <*> coreToType ty
 -- TODO: save the distinction between => and ->
@@ -944,13 +954,25 @@ coreToTyVar tv =
 coreToId :: Id
          -> C2C C.Id
 coreToId i = do
-  C.mkId <$> coreToType (varType i) <*> pure scope <*> coreToVar i
+  C.mkId <$> coreToVarType i <*> pure scope <*> coreToVar i
  where
   scope = if isGlobalId i then C.GlobalId else C.LocalId
 
 coreToVar :: Var
           -> C2C (C.Name a)
 coreToVar = coreToName varName varUnique qualifiedNameStringM
+
+-- | Convert the type of a variable, memoized on the variable's name.
+--
+-- Only global variables are memoized: their names come from GHC's name cache
+-- and are globally unique, so the same name always refers to the same
+-- variable, whose type is intrinsic. Names of local variables (e.g. binders
+-- instantiated when loading unfoldings from interface files) can collide
+-- between declarations and must not share cache entries.
+coreToVarType :: Var -> C2C C.Type
+coreToVarType v
+  | isGlobalId v = makeCached (varName v) varTypeMap (coreToType (varType v))
+  | otherwise = coreToType (varType v)
 
 coreToPrimVar :: Var
               -> C2C (C.Name C.Term)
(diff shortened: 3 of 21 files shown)


martijnbastiaan:
I'm pleased to announce that this turns out to be 2m11s on the actual CI runners, down from 18m16s according to [this PR](https://github.com/clash-lang/clash-compiler/pull/3245).

changelog/2026-07-12T22_45_00+02_00_evaluator_no_global_heap_copy
```diff
@@ -0,0 +1 @@
+CHANGED: The compile-time evaluator no longer copies the entire global binding map on every invocation. On designs that trigger the evaluator often (e.g. large designs with many vector primitives) this can cut normalization time by 25% or more.
```
martijnbastiaan: Though this patch looks good to me, I'm not sure about the description here (and in the commit message). It seems to me that this is faster because the maps it now inserts/deletes into/from are much much smaller.

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -133,6 +138,33 @@ findAccidentialShadows =
   findDups ids = filter ((1 <) . length) (group (sort ids))
 
 
+-- TEMPORARY instrumentation (perf investigation): cumulative wall-clock time
+-- per transformation, printed at the end of the rewrite session.
+{-# NOINLINE rewriteTimers #-}
+rewriteTimers :: IORef (HashMap.HashMap Text (Word64, Word64))
+rewriteTimers = unsafePerformIO (newIORef HashMap.empty)
```
martijnbastiaan: Let's remove the timer instrumentation. We could have also used `SCC` and GHC profiling?

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -4565,6 +4565,38 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
                                 ])
                  ]
          _ -> Nothing
+  -- Fast path: unconcatting a literal BitVector peels off the head and tail
+  -- as literals directly. The general case below leaves residual split#
+  -- calls in both the head and the tail, which later rewrite passes have to
+  -- evaluate separately (and repeatedly).
+  $(namePat 'Clash.Sized.Vector.unconcatBitVector#)
```
martijnbastiaan: Honestly, pretty cool patch, but it should have a test associated with it.

clash-lib/src/Clash/Debug.hs
```diff
@@ -3,10 +3,33 @@
...
+import Data.Text (Text)
+import Data.Word (Word64)
 import Debug.Trace
+import System.IO.Unsafe (unsafePerformIO)
+
+-- TEMPORARY instrumentation (perf investigation): global event counters,
+-- printed at the end of the rewrite session when CLASH_TIMERS is set.
+{-# NOINLINE debugCounters #-}
+debugCounters :: IORef (HashMap Text Word64)
+debugCounters = unsafePerformIO (newIORef HashMap.empty)
```
martijnbastiaan: Again, this commit should probably go

clash-lib/src/Clash/Normalize.hs
```diff
@@ -409,13 +409,16 @@ flattenCallTree cache (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
...
+    --
+    -- The bundle dispatches on the node constructor so only transformations
+    -- that can fire on that constructor are attempted; the skipped ones are
+    -- guaranteed no-ops there (checked against each transformation's entry
+    -- patterns). A node whose constructor changes is re-attempted by
+    -- topdownFixR's repeatR and then re-dispatched on its new shape.
+    repeatR (topdownFixR flattenStep >->
+             bottomupR (\ctx e -> case e of
+               Let {} -> apply "flattenLet" flattenLet ctx e
+               _ -> pure e)) !->
```
martijnbastiaan: I don't understand this, why would this be faster than just trying `flattenLet`?

clash-lib/src/Clash/Normalize.hs
```diff
@@ -409,13 +409,16 @@ flattenCallTree cache (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
...
-               apply "reduceNonRepPrim" reduceNonRepPrim >->
-               apply "removeUnusedExpr" removeUnusedExpr) >->
-             bottomupR (apply "flattenLet" flattenLet)) !->
+    --
+    -- The bundle dispatches on the node constructor so only transformations
+    -- that can fire on that constructor are attempted; the skipped ones are
+    -- guaranteed no-ops there (checked against each transformation's entry
+    -- patterns). A node whose constructor changes is re-attempted by
+    -- topdownFixR's repeatR and then re-dispatched on its new shape.
+    repeatR (topdownFixR flattenStep >->
```
martijnbastiaan: This is pretty cool, but it messes with the order of transformations. @christiaanb Do you have an opinion on this?

clash-lib/src/Clash/Normalize.hs
```diff
@@ -427,6 +430,35 @@ flattenCallTree cache (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
...
+
+  bundleTyApp =
+    apply "appProp" appProp >->
+    apply "removeUnusedExpr" removeUnusedExpr
+
+  bundleCase = apply "caseCon" caseCon
+
+  bundleLet = apply "bindConstantVar" bindConstantVar
+
+  bundlePrim = apply "removeUnusedExpr" removeUnusedExpr
```
martijnbastiaan: I don't like this: correct usage depends on the definition of the transformation. I.e., if someone where to extend `removeUnusedExpr` to match on more than it does now, it breaks. This should be refactored such that the transformations don't match on the constructor anymore, but receive fields directly. That way, it cannot be messed up anymore.

Perhaps this concept should also be taken further: we could have `applyLet`, `applyApp`, etc. The `apply*` functions could propagate this upwards, so combinators like `topDownR` would know what do dispatch -- therefore only matching on constructors once.

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -77,40 +78,84 @@ constantPropagation =
     -- The outer repeatR is still needed: inlineNR is a full traversal whose
     -- results can only be processed by re-running the top-down bundle from the
     -- new root.
-    inlineAndPropagate = repeatR (topdownFixR (applyMany transPropagateAndInline) >-> inlineNR)
+    inlineAndPropagate = repeatR (topdownFixR propInlineStep >-> inlineNR)
```
martijnbastiaan: Same commentary as previous "shape dispatch" commit

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -55,10 +55,14 @@ normalization =
...
                  topdownSucR (apply "inlineCleanup" inlineCleanup) !->
-                 innerMost (applyMany [("caseCon"        , caseCon)
-                                      ,("bindConstantVar", bindConstantVar)
-                                      ,("letFlat"        , flattenLet)])
+                 innerMost cleanupStep
                  >-> rmDeadcode >-> letTL
+    -- Entry shapes: caseCon: Case; bindConstantVar, letFlat: Let.
+    cleanupStep ctx e = case e of
+      Case {} -> apply "caseCon" caseCon ctx e
+      Let {} -> (apply "bindConstantVar" bindConstantVar >->
```
martijnbastiaan: And once more same commentary

clash-lib/src/Clash/Normalize/Strategy.hs
```diff
@@ -41,20 +41,34 @@ normalization =
   xOptim >-> rmDeadcode >->
   cleanup >-> bindSimIO >-> recLetRec >-> splitArgs
   where
-    multPrim   = topdownR (apply "setupMultiResultPrim" setupMultiResultPrim)
-    anf        = topdownR (apply "nonRepANF" nonRepANF) >-> apply "ANF" makeANF >-> topdownR (apply "caseCon" caseCon)
+    -- The shape gates (onApp, onCase, ...) skip transformations on nodes
+    -- their entry patterns cannot match; see each transformation's entry
+    -- clauses.
+    multPrim   = topdownR (onPrim (apply "setupMultiResultPrim" setupMultiResultPrim))
+    anf        = topdownR (onApp (apply "nonRepANF" nonRepANF)) >-> apply "ANF" makeANF >-> topdownR (onCase (apply "caseCon" caseCon))
```
martijnbastiaan: Same.

I'm also happy for all these "shape" commits to go onto a single heap.

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -10,6 +10,7 @@
 -}
 
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE NondecreasingIndentation #-}
```
martijnbastiaan: Please no

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -158,6 +162,7 @@ reduceConst _ e = return e
 reduceNonRepPrim :: HasCallStack => NormRewrite
 reduceNonRepPrim c@(TransformContext _ ctx) e@(App _ _) | (Prim p, args, ticks) <- collectArgsTicks e = do
   tcm <- Lens.view tcCache
+  if not (reduceNonRepPrimCouldFire tcm p) then return e else do -- NondecreasingIndentation
```
martijnbastiaan: Please just indent

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -542,6 +547,56 @@ reduceNonRepPrim c@(TransformContext _ ctx) e@(App _ _) | (Prim p, args, ticks)
          else return False
 
 reduceNonRepPrim _ e = return e
+
+-- | Cheap pre-check for 'reduceNonRepPrim': it can only fire when the
+-- primitive is one of the primitives it specifically handles, or when the
+-- (ultimate) result type of the application can be @Vec 0 a@. The latter is
```
martijnbastiaan: The fragility of this setup is not worth the meager speedup. Let's remove the commit.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -970,6 +976,55 @@ acmpTerm' inScope = go (mkRnEnv inScope)
     Case {}    -> 11
     Tick {}    -> 12
 
+-- | A cheap, alpha-invariant fingerprint of a term: alpha-equivalent terms
+-- (in the sense of 'aeqTerm') have equal fingerprints. The converse does not
+-- hold, so use the fingerprint as a pre-filter and verify candidates with
+-- 'aeqTerm'. Types and ticks do not contribute to the fingerprint.
+aeqTermFingerprint :: Term -> Int
+aeqTermFingerprint = go 0 emptyVarEnv 17
```
martijnbastiaan: This should use the "hash" terminology. Speaking of.. Maybe we should just implement `Hashable Term` with https://arxiv.org/pdf/2105.02856?

clash-lib/src/Clash/Normalize/Strategy.hs
martijnbastiaan: Done

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Dropped this commit, let's implement 2105.02856 in another PR.

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: I'll pick this up in another PR, as it's getting quite a bit much for this one.

clash-lib/src/Clash/Normalize.hs
martijnbastiaan: The idea here is that `apply` itself is pretty expensive. I don't think doing this is the right move though, as it bypasses `apply`'s debug records.

martijnbastiaan:
These content of the commits now look good to me. Code gets slightly more complicated, but given the performance wins I think it's worth it.

I've yet to touch up the commit messages / documentation, which are verbose in a way that I feel isn't helpful.

CI now also fails, due to clash-lib tests not working on macos. I think we haven't seen this before because there are no other clash-lib tests in the Vector testsuite, which I think we limit the tests to on CI (on mobile at the moment, so hard to check). @rowanG077 offered to look into it on his macos device, which I happily accepted :).

martijnbastiaan:
> @rowanG077 offered to look into it on his macos device, which I happily accepted :).

In the meantime I've CPPed it out, as I believe this is an issue outside of this PR.

christiaanb (commented):
The details of the commit message for "Evaluator: use `BindingMap` in evaluator as-is" are kinda bogus. The changelog message is better.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -4565,6 +4565,38 @@ ghcPrimStep tcm isSubj pInfo tys args mach = case primName pInfo of
...
+           tl      = mkBitVectorLit n1BVTy n1MTy ((n-1)*m)
+                       (mski .&. loMask) (i .&. loMask)
+       in reduce $ mkVecCons consCon mBVTy n hd
+            (mkApps (Prim pInfo)
+                    [ Right (LitTy (NumTy (n-1)))
+                    , Right mTy
+                    , Left (Literal (NaturalLiteral (n-1)))
+                    , Left (valToTerm km)
+                    , Left tl
+                    ])
```
christiaanb: Can't this logic be merged into the alternative of line 4580?

martijnbastiaan:
I don't think the commit message is bogus, could you explain? Why wouldn't a `map` over a large strict map cause a bunch of work?

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -919,7 +929,7 @@ coreToType' (TyConApp tc args)
                         foldl C.AppTy <$> coreToType synTy' <*> mapM coreToType remArgs
                       _ -> do
                         tcName <- coreToName tyConName tyConUnique qualifiedNameString tc
-                        tyConMap %= (C.insert tcName tc)
+                        tyConMap %= (\m -> if tcName `C.elem` m then m else C.insert tcName tc m)
```
christiaanb: It's weird that his line is supposedly getting the best speedup... due to sharing... `C.elem` is implemented using https://hackage-content.haskell.org/package/containers-0.8/docs/Data-IntMap-Strict.html#v:member which as O(min(n,W)) compute complexity. `C.insert` is implemented using https://hackage-content.haskell.org/package/containers-0.8/docs/Data-IntMap-Strict.html#v:insert which also has O(min(n,W)) compute complexity.

So from a compute point of view, when the `tcName` is not present, you are paying the O(min(n,W)) twice. Meaning that from a compute point of view, this change is objectively bad.

That means that the entire speedup is attributed to allocation and GC... would be worth it to check if this change really as the speedup contribution Claude claims it has; along-side a alloc+GC-time spend in both cases.

If this turns out to be true, then I really need to re-adjust my thinking about performance of Haskell code.

martijnbastiaan: Will do!

christiaanb:
> I don't think the commit message is bogus, could you explain? Why wouldn't a `map` over a large strict map cause a bunch of work?

The `Machine` data type is lazy, so the `map` over the large strict IntMap isn't performed until it is acually needed. The current phrasing makes it sound like we are operating in a strict language. I guess what does hold, is that once the IntMap is needed for some computation, whether it be a lookup or insertion, the entire IntMap is walked due to the strict constructors of `IntMap` https://hackage-content.haskell.org/package/containers-0.8/docs/src/Data.IntMap.Internal.html#IntMap (which is used by both the "Lazy" and "Strict" IntMap API). So I guess, it's more the "as if operating in a strict language" that I think is the bogus part rather than the actual effect achieved.

On top of that, the changelog says
"updates to the global heap during evaluation now go through a small overlay instead of a full-size map." and I wonder whether the speedup shouldn't be mainly contributed to this, rather than the one-time walk of the `IntMap`.

martijnbastiaan:
> So I guess, it's more the "as if operating in a strict language" that I think is the bogus part rather than the actual effect achieved.

That's fair!

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
martijnbastiaan: > So from a compute point of view, when the tcName is not present, you are paying the O(min(n,W)) twice.

So the idea is that for the overwhelming majority of cases it _is_ present, and insert makes you pay GC/mutator cost by rebuilding. This contributes about 1s / 0.5GB alloc.

This is definitely bullshit though:

> This is where most of the `makeAllTyCons` win comes from.

It's about ~10% of that patch.

martijnbastiaan: So IMO the commit message is just wrong. It should just say "avoids rebuilding". I'm also fine with just dropping this change, its effect is somewhat small (though real, I suppose).

martijnbastiaan:
> On top of that, the changelog says "updates to the global heap during evaluation now go through a small overlay instead of a full-size map." and I wonder whether the speedup shouldn't be mainly contributed to this, rather than the one-time walk of the IntMap.

Turns out it really is due to just the `mapVarEnv`. If you keep the `mapVarEnv` _and_ introduce the extra layer, things get slower (+5s) when compared to `HEAD`.  If you drop the `mapVarEnv` allocations drop from 1390GB to 687GB and runtime from 6m20s to 4m21s. FWIW, this comment:

> IntMap's spine-strict Bin means the first global-heap access in a whnf' call forces a complete spine rebuild, and the numbers show that happened on essentially every call: the extra 704 GB over ~5M evaluator invocations is ~140 KB/call, which matches one full spine copy of a few-thousand-entry map each time. And again the cost is 93% mutator, 7% GC.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
martijnbastiaan: Yeah I think so, I'll have a look.

martijnbastiaan: Done

martijnbastiaan:
@christiaanb I've amended the commits and changelogs. The commit now reads:

> Previously, the global `BindingMap` would be converted to a `PureHeap`
> on (virtually) every call to the evaluator:
> 
>     gh = mapVarEnv bindingTerm bm
> 
> The WHNF evaluator is invoked millions of times during normalization on
> larger designs, so made up a fair share of normalization time.

With the "virtually" in there, does it still read imperative to you? FWIW, I don't really mind it -- given that we're writing Haskell I don't thing it needs elaborating that this _actually_ only happens when needed.

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
rowanG077: I'd keep the change, but put a comment above that line. Because as it stands that line is surprising to read.

clash-lib/src/Clash/Normalize/Transformations/Case.hs
```diff
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017-2022, Google Inc.,
-                    2021-2024, QBayLogic B.V.
+                    2021-2024,2026, QBayLogic B.V.
```
rowanG077: We usually just update the upper bound. So it should be `2021-2026`.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017-2022, Google Inc.,
-                    2021-2024, QBayLogic B.V.
+                    2021-2024,2026, QBayLogic B.V.
```
rowanG077: Same here.

clash-lib/src/Clash/Normalize/Transformations/Letrec.hs
```diff
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017-2018, Google Inc.,
-                    2021-2022, QBayLogic B.V.
+                    2021-2022,2026, QBayLogic B.V.
```
rowanG077: Same here

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016-2017, Myrtle Software Ltd,
                     2017-2018, Google Inc.,
-                    2021-2022, QBayLogic B.V.
+                    2021-2022,2026, QBayLogic B.V.
```
rowanG077: And here

clash-lib/src/Clash/Rewrite/Types.hs
```diff
@@ -2,7 +2,7 @@
   Copyright  :  (C) 2012-2016, University of Twente,
                     2016     , Myrtle Software Ltd,
                     2017     , Google Inc.,
-                    2021-2022, QBayLogic B.V.
+                    2021-2022,2026, QBayLogic B.V.
```
rowanG077: And here. Well I won't call out anymore as they are easily findable.

rowanG077 (approved):
Actually looks like good code to me. Some really tiny things really I found.

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
martijnbastiaan: Yeah I'll pull it out into an `insertUnlessPresent`.

martijnbastiaan:
All done. I think we can merge if there are no objections to the commit message of "Evaluator: use BindingMap in evaluator as-is" anymore.

christiaanb (approved):
I’m good with merging it now.

Issue #3325, opened 2026-07-16
Allow ~SYMs to be named


leonschoorl:
When looking at #3273 I had [this idea](https://github.com/clash-lang/clash-compiler/issues/3273#issuecomment-4969019939) to extend `~SYM[]` to make it nicer to use.

Currently `~SYM` works like this:
```
  signal ~GENSYM[foo][3] : ..;
  signal ~GENSYM[bar][4] : ..;

  .. <= ~SYM[3] - ~SYM[4];
```
As a blackbox writer you use `~GENSYM[foo][3]` define symbol 3 and name it `foo`.
And them you can refer to it by using `~SYM[3]`.
In [complex blackboxes](https://github.com/clash-lang/clash-compiler/blob/62938cad423c88d1732e361b1c05582ebecdd3d4/clash-lib/prims/vhdl/GHC_Prim.primitives.yaml#L474-L513) you can have a lot of these floating around, which you have to keep track of when reading/writing these blackboxes.

The core of the idea is to allow `~SYM[]` to refer to names. Because they're just way easier to keep track of then number.
```
  signal ~SYM[foo] : ..;
  signal ~SYM[bar] : ..;

  .. <= ~SYM[foo] - ~SYM[bar];
```

But implement it in such a way that the old behavior keep working.

leonschoorl:
Implemented in #3326

PR #3326, opened 2026-07-20
Extend the blackbox template tag ~SYM to be able to accept a name

leonschoorl:
Instead of doing:
```
	signal ~GENSYM[foo][3] : ..;
	signal ~GENSYM[bar][4] : ..;

	.. <= ~SYM[3] - ~SYM[4];
```
You can now refer to `SYM`bols by name:
```
	signal ~SYM[foo] : ..;
	signal ~SYM[bar] : ..;

	.. <= ~SYM[foo] - ~SYM[bar];
```
The old behaviour with GENSYM and the numeric references is still fully supported.

Implements #3325

## Still TODO:

### DIFF
--- a/changelog/2026-07-20T16_21_02+02_00_named_SYMs
+++ b/changelog/2026-07-20T16_21_02+02_00_named_SYMs
@@ -1,0 +1,21 @@
+ADDED: In blackboxes `~SYM`s can now be referenced by name
+
+Instead of doing:
+```
+	signal ~GENSYM[foo][3] : ..;
+	signal ~GENSYM[bar][4] : ..;
+
+	.. <= ~SYM[3] - ~SYM[4];
+```
+
+You can now refer to SYMbols by name:
+```
+	signal ~SYM[foo] : ..;
+	signal ~SYM[bar] : ..;
+
+	.. <= ~SYM[foo] - ~SYM[bar];
+```
+
+The old behaviour with GENSYM and the numeric references is still fully supported.
+
+See [#3325](https://github.com/clash-lang/clash-compiler/issues/3325).
--- a/clash-lib/src/Clash/Netlist/BlackBox/Parser.hs
+++ b/clash-lib/src/Clash/Netlist/BlackBox/Parser.hs
@@ -1,7 +1,7 @@
 {-|
   Copyright  :  (C) 2012-2016, University of Twente,
                     2017     , Myrtle Software Ltd,
-                    2021-2022, QBayLogic B.V.
+                    2021-2026, QBayLogic B.V.
                     2022     , Google Inc.
   License    :  BSD2 (see the file LICENSE)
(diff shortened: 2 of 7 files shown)


leonschoorl:
One thing that might need some more thought/discussion is what should happen when someone (accidentally) mixes the old and the new style.

```
	signal ~GENSYM[foo][3] : ..;

	.. <= ~SYM[foo];
```

Right now clash considers these too symbols different, and will generate:
```
	signal foo : ..;

	.. <= foo_1;
```

Which is probably not what the user intended.

rowanG077:
I'm in favor of erroring. Another avenue is to not overload `SYM` but introduce a new concept for this.

leonschoorl:
Mixing GENSYM and SYM like this:
```
~GENSYM[levels][1]
...
~SYM[levels]
```
Now errors with:
```
<no location info>: error:
    Other error:
    In BlackBox Test.myBlackbox:  ~SYM[foo] uses the same name as previously used in ~GENSYM.
    Don't mix numbered and named symbols.
```

And the opposite order generates:
```
<no location info>: error:
    Other error:
    In BlackBox Test.myBlackbox: the name of ~GENSYM[foo][1] would overwrite an earlier ~SYM[foo].
    Don't mix numbered and named symbols.
```

rowanG077 (approved):
Implementation looks good to me!

I just want some tests, showing the error route and the correct route.

rowanG077 (changes requested):
Sorry I mean requested changes :)

jaschutte:
I am a big fan of this change as directly referring to symbols by name is a lot less error prone.
Once this lands, perhaps we should update the [clash tutorial](https://hackage.haskell.org/package/clash-prelude-0.10.7/docs/CLaSH-Tutorial.html) to use this new behavior too? I know the old syntax is still valid, but I feel like the tutorial should present the most 'obvious' way to do something.

leonschoorl:
@jaschutte The tutorial doesn't exist anymore in clash-prelude, but the relevant part of it now over at https://docs.clash-lang.org/compiler-user-guide/developing-hardware/primitives.html

leonschoorl:
I was looking at updating that documentation, but it does something that this PR (currently) can't do: `signal ~GENSYM[~RESULT_RAM][2]`
It uses template holes inside of inside of the name.
For simplicity I made `~SYM[]` only accept simple strings as names, only containing letters, numbers and underscores.

Maybe I should extend it to do the same as GENSYM?

rowanG077:
Ah yes I think it's unnatural if that isn't supported.

leonschoorl:
I've added test for the errors when you use GENSYM and SYM with the same name.
And I've updated SYM to accept the same things in the name as GENSYM already accepted.

rowanG077 (approved):

rowanG077 (approved):

PR #3327, opened 2026-07-22
Skip type traversal in identifier-only free-variable folds

martijnbastiaan:
`termFreeVars'` descends into the type of every variable occurrence and every binder it passes. For folds that only select identifiers -- `freeIds`, `freeLocalIds`, `globalIds` -- those excursions can never yield an occurrence.

Yields about ~2% faster normalization.

## Still TODO:


christiaanb (approved):
Perhaps add a changelog entry that we get a ~2% compile-time improvement?

PR #3331, opened 2026-07-23
Use `HashMap` for name lookups in `clash-ghc`'s evaluator

martijnbastiaan:
Replace big `case`-expressions on `String` with a `HashMap` lookups. Additionally, split out the gigantic file into multiple smaller ones for compilation speed benefits.

The steps for this PR were very mechanical: you can check out the refactor scripts in the relevant commits.

# Benchmark
|                   | **Before** | **After** | **Ratio** |
|-------------------|------------|-----------|-----------|
| **Loading**       | 17.322     | 17.544    | 1.01      |
| **Normalization** | 242        | 208       | 0.86      |
| **Netlist**       | 3.201      | 3.301     | 1.03      |
| **Total**         | 264        | 230       | 0.87      |

Splitting up the primitive module doesn't yield runtime benefits, but makes compiling `clash-ghc` ~20% faster.

# Reviewer
1. We might want to squash the commit "Trim imports and pragmas of split evaluator modules" into its parent. The reason I didn't do this is because the parent commit is the result of a script, while the trim-commit is "manual" work. If logical chunks of work are valued over traceability, I'd squash them. Otherwise let it be.

2. Also, I'm not a big fan of `RecordWildCard`, but it was a "nice" way to make this exercise mechanical. 

3. This started out as a `FastString` based approach, but that didn't turn out to be necessary. There are still a few minor cleanup commits in this PR, which we might want to throw out (or keep).

4. I didn't look to closely at the (AI generated) refactor scripts, as the most difficult part was parsing -- which was verified by roundtripping it.

## Still TODO:

  - [x] Fix warnings on "other" GHCs
  - [x] Go over it manually (@martijnbastiaan)
  - [x] Investigate whether the speedup is due to `FastString` or simply switching to a `UniqueMap` (`IntMap`) lookup
    - It was essentially all due to the linear evaluator lookup


clash-lib/src/Clash/Unique.hs
```diff
@@ -64,6 +65,14 @@ instance Uniquable Unique where
   getUnique = id
   setUnique = flip const
 
+-- | Uses the 'GHC.Types.Unique.Unique' GHC already assigned the 'FastString'
+-- through its global intern table; no hashing or string comparison involved.
+-- There is no way to construct a 'FastString' from a 'Unique' alone, so
+-- 'setUnique' is not supported.
+instance Uniquable FastString where
+  getUnique = fromGhcUnique . GHC.getUnique
```
rowanG077: I don't think this instance is used anywhere? I removed it locally without issues. Considering it's partial I'd say it's better to remove it.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitives.hs
```diff
@@ -0,0 +1,82 @@
...
+import qualified Clash.GHC.Evaluator.Primitives.GHC.Prim as GHC.Prim
+import qualified Clash.GHC.Evaluator.Primitives.GHC.PrimopWrappers as GHC.PrimopWrappers
+import qualified Clash.GHC.Evaluator.Primitives.GHC.Real as GHC.Real
+import qualified Clash.GHC.Evaluator.Primitives.GHC.TypeLits as GHC.TypeLits
+import qualified Clash.GHC.Evaluator.Primitives.GHC.TypeNats as GHC.TypeNats
+import qualified Clash.GHC.Evaluator.Primitives.GHC.Types as GHC.Types
+import qualified Clash.GHC.Evaluator.Primitives.GHC.Word as GHC.Word
+
+ghcPrimStepImpls :: HashMap.HashMap Text PrimStep
+ghcPrimStepImpls = HashMap.fromList $ concat
```
rowanG077: This will remove duplicate names. In fact there is a duplicate name now.

`GHC.Base.dataToTag#` and `GHC.Magic.dataToTag#` both become  `GHC.Magic.dataToTag#`.

Especially considering we have a ton of guards for different GHC versions I would like it if this function returns an error if there are dups.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitive.hs
```diff
@@ -239,5693 +136,8 @@ ghcPrimUnwind tcm p tys vs (collectValueTicks -> (v, ts)) (e:es) m
...
-    n       = TyCon.tyConName tc
-
-svoid :: (State# RealWorld -> State# RealWorld) -> IO ()
-svoid m0 = IO (\s -> case m0 s of s' -> (# s', () #))
-
-isTrueDC,isFalseDC :: DataCon -> Bool
-isTrueDC dc  = dcUniq dc == fromGhcUnique trueDataConKey
-isFalseDC dc = dcUniq dc == fromGhcUnique falseDataConKey
+ghcPrimStep tcm isSubj pInfo tys args mach =
+  case HashMap.lookup (primName pInfo) ghcPrimStepImpls of
```
rowanG077, line 141: wow so clean, so nice.

rowanG077 (changes requested):
I was not looking forward to reviewing this PR. but it was actually not that terrible.

Some small things I'd like to see changed. 

Regarding your points:

> We might want to squash the commit "Trim imports and pragmas of split evaluator modules" into its parent. The reason I didn't do this is because the parent commit is the result of a script, while the trim-commit is "manual" work. If logical chunks of work are valued over traceability, I'd squash them. Otherwise let it be.

I prefer traceability. Especially in terms of mechanically generated code vs manual fixups.

> Also, I'm not a big fan of RecordWildCard, but it was a "nice" way to make this exercise mechanical.

I'm not too against it. We could do another mechanical rewrite that removes all the wildcards, but it's not a requirement to get this in for me.  

I have added a new commit, cbbd3df9df0d11bca4cb8948de3f94fed2ec9cdd,  that removes about ~1.5K additional lines by making a small helper. But does nothing else. If you are against it feel free to drop that commit.

clash-ghc/src-ghc/Clash/GHC/Evaluator/Primitives.hs
martijnbastiaan: ~~Makes sense! I'll add this as a separate commit.~~ Done

clash-lib/src/Clash/Unique.hs
martijnbastiaan: Yeah I'll drop the commit

martijnbastiaan:
> I have added a new commit, https://github.com/clash-lang/clash-compiler/commit/cbbd3df9df0d11bca4cb8948de3f94fed2ec9cdd, that at removes about ~1.5K additional lines by making a small helper. But does nothing else. If you are against it feel free to drop that commit.

Looks good. I'll fix the two minor things and re-request a review. (Probably before the end of the day.)

martijnbastiaan:
Done @rowanG077

rowanG077 (approved):

PR #3332, opened 2026-07-24
nix: avoid building local packages in dev shells

rowanG077:
This fixes a longstanding annoyance of mine. Entering a `nix develop` shell required a build of most local packages because the way it is setup is that we use the `*.env` environments of all packages. This puts all dependencies in the develop shell, which includes the local packages themselves.

using `shellFor` + `packages` automatically filters out the packages in the `packages` list, which means it now puts all haskell dependencies in the shell except the local packages.

### DIFF
--- a/nix/devshell.nix
+++ b/nix/devshell.nix
@@ -3,20 +3,22 @@ compilerVersion:
 let
   clashPkgs = pkgs."clashPackages-${compilerVersion}";
 in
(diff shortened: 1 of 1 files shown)


martijnbastiaan (approved):
RIP HLS

rowanG077:
Nah, it was just duplicated in the list and you can't see it in the diff

PR #3330, opened 2026-07-23
Remove `clash-cosim` and `clash-ffi` packages

rowanG077:
They are moved to standalone archived repositories.

## Still TODO:

  - [x] Write a changelog entry (see changelog/README.md)
  - [x] Make seperate repos

PR #3328, opened 2026-07-23
Execute transformations based on node constructor

martijnbastiaan:
The basic idea of this PR is that the "transformation engine" now keeps track of what constructors transformations are interested in. It uses that to match on the constructor _once_ and then only dispatch the transformations that are interested in it.

See https://clash-lang.github.io/clash-benchmarks/?machine=oele&branch=clash-lang/clash-compiler@perf/strategy-dsl.

## Still TODO:

  - [x] Write a changelog entry (see changelog/README.md)
  - [x] Check copyright notices are up to date in edited files
  - [x] Go over code manually (@martijnbastiaan)
  - [X] Evaluate whether TH approach makes sense
    - Yes, this yields another 10-20% perf (depending on benchmark)

PR #3333, opened 2026-07-25
Even more performance patches

martijnbastiaan:
Second installment! I put Fable to work over the last few days and it came up with these patches. I've confirmed that it passes the `clash-testsuite` on all supported GHCs and that it produces bit-for-bit equal HDL at every commit over at `bittide-hardware`. Benefits measured on a 4-core VM running on a AMD Ryzen 7 3700X (`wireDemoTest` @ `bittide-hardware`).

| Commit title | Loading | Normalization | Total | Delta | Delta (%) |
| --- | ---: | ---: | ---: | ---: | ---: |
| Baseline: Merge pull request #3330 | 17.473s | 4m04.0s | 4m28.0s | — | — |
| [Compile the normalization strategies from StrategyDSL specs](https://github.com/clash-lang/clash-compiler/pull/3328) | 17.552s | 2m45.0s | 3m10.0s | −1m18.0s | −29.1% |
| [Use `HashMap` lookup in `ghcPrimStep`](https://github.com/clash-lang/clash-compiler/pull/3331) | 17.571s | 2m09.0s | 2m34.0s | −36.000s | −18.9% |
| [Compile primitives in a single shared hint session](https://github.com/clash-lang/clash-compiler/pull/3337) | 14.211s | 2m09.0s | 2m31.0s | −3.000s | −1.9% |
| Use a map index in CSE's `reduceBinders` | 14.122s | 2m11.0s | 2m32.0s | +1.000s | +0.7% |
| Stop measuring term size once it exceeds the inline limit | 14.111s | 2m09.0s | 2m31.0s | −1.000s | −0.7% |
| Track visited binders in a `VarSet` when flattening call trees | 14.186s | 2m11.0s | 2m32.0s | +1.000s | +0.7% |
| Key the HWType translation cache on structural equality | 14.197s | 2m02.0s | 2m24.0s | −8.000s | −5.3% |
| [Dispatch `reduceNonRepPrim` through a `HashMap`](https://github.com/clash-lang/clash-compiler/pull/3339) | 14.129s | 1m54.0s | 2m15.0s | −9.000s | −6.2% |
| Cache blackbox `usedArguments` per primitive name | 14.269s | 1m53.0s | 2m15.0s | +0.000s | +0.0% |
| Speed up free-variable checks when inlining and lifting | 14.180s | 1m49.0s | 2m10.0s | −5.000s | −3.7% |
| Skip the `flattenCheap` rewrite when nothing was inlined | 14.241s | 1m49.0s | 2m10.0s | +0.000s | +0.0% |
| Share the evaluator record in primitive evaluation rules | 14.172s | 1m48.0s | 2m09.0s | −1.000s | −0.8% |
| Preserve sharing when inlining globals in `flattenCallTree` | 14.214s | 1m49.0s | 2m10.0s | +1.000s | +0.8% |
| [Only consider spine roots in `reduceNonRepPrim`](https://github.com/clash-lang/clash-compiler/pull/3338) | 14.287s | 1m45.0s | 2m06.0s | −4.000s | −3.1% |
| [Run evaluator-backed transformations once per flatten iteration](https://github.com/clash-lang/clash-compiler/pull/3338) | 14.267s | 1m12.0s | 1m33.0s | −33.000s | −26.2% |
| Run cheap transformation guards before expensive ones | 14.254s | 1m08.0s | 1m29.0s | −4.000s | −4.3% |
| Prefilter `liftBinding`'s alpha-equivalence scan on a type fingerprint | 14.233s | 1m07.0s | 1m30.0s | +1.000s | +1.1% |
| Reuse the machine's in-scope set in evaluator substitutions | 14.236s | 1m04.0s | 1m25.0s | −5.000s | −5.6% |
| Prune the pure heap to reachable bindings in `bindPureHeap` | 14.263s | 1m04.0s | 1m25.0s | +0.000s | +0.0% |
| [Memoize GHC-to-Clash type conversion on the GHC type](https://github.com/clash-lang/clash-compiler/pull/3340) | 6.251s | 1m04.0s | 1m17.0s | −8.000s | −9.4% |
| Skip SCC sort for known-independent let bindings | 6.244s | 1m01.0s | 1m14.0s | −3.000s | −3.9% |
| Index liftBinding candidates by type fingerprint | 6.228s | 58.671s | 1m11.0s | −3.000s | −4.1% |
| [Use sharing-preserving substitution in appProp beta reduction](https://github.com/clash-lang/clash-compiler/pull/3341) | 6.231s | 55.440s | 1m08.0s | −3.000s | −4.2% |

I've ran it on our CI runners too, which came back as 4.145s loading, 34.747s normalization, 43.666s total.

# Martijn's review filter
```mermaid
flowchart TD
    B{Performance benefit?}
    B -->|Yes| C[Review as usual]
    B -->|No| D{BigO benefit?}
    E -->|Yes| F
    D -->|No| F[Reject]
    D -->|Yes| E{Sizes we control?}
    E -->|No| C
```

# Martijn's review list
|                             Commit title                             | Approval | Comments |
|--------------------------------------------------------------------|----------|----------|
| Compile the normalization strategies from StrategyDSL specs          |          | See #3328         |
| Use HashMap lookup in ghcPrimStep                                    |          | See #3331         |
| Compile primitives in a single shared hint session                   | 🦭         | Minor inline comments         |
| Use a map index in CSE's reduceBinders                               | 🦭/🔧         | Minor comments. Even though it doesn't give a speedup _right now_, we don't control the number of binders so we shouldn't do linear lookups. We probably also shouldn't do `Ord` lookups, so this should take a more principled approach and go into a separate PR.         |
| Stop measuring term size once it exceeds the inline limit            | 🦭         | Sensible change, but I'd like to take the opportunity to add some comments, so let's move it to a separate PR.         |
| Track visited binders in a VarSet when flattening call trees         | 🦭          | Doesn't give a speedup right now, but we don't control the number of binders to non-quadratic seems the way to go.         |
| Key the HWType translation cache on structural equality              | ❓         | Yes, but only if alpha-eq hashing doesn't work out.         |
| Dispatch reduceNonRepPrim through a HashMap                          | 🦭         | Minor inline comments         |
| Cache blackbox usedArguments per primitive name                      | 🦭         | No direct performance benefit, but it seems sensible. Newer code also looks nicer.         |
| Speed up free-variable checks when inlining and lifting              | 🔧         | Not like this         |
| Skip the flattenCheap rewrite when nothing was inlined               | 🔧         | Not like this         |
| Share the evaluator record in primitive evaluation rules             | ❌         | Stupid.         |
| Preserve sharing when inlining globals in flattenCallTree            | ❌         | Doesn't seem to offer a benefit, seems dangerous, and increases LOC.         |
| Only consider spine roots in reduceNonRepPrim                        | ❓         | I don't understand this.        |
| Run evaluator-backed transformations once per flatten iteration      | ❓         | I can't really reason about this. Given that it survives `bittide-hardware`, the test suite, and that it offer such a big performance benefit I'm inclined to approve.         |
| Run cheap transformation guards before expensive ones                | 🦭         | Minor comments         |
| Prefilter liftBinding's alpha-equivalence scan on a type fingerprint | ❌         | Increased LOC, no measurable performance benefit.         |
| Reuse the machine's in-scope set in evaluator substitutions          | ❓         | Yes? I don't fully understand yet.         |
| Prune the pure heap to reachable bindings in bindPureHeap            | ❌        | Cute, but I don't see how this would really improve anything given that `removeUnusedBinders` runs _right_ after it.         |
| Memoize GHC-to-Clash type conversion on the GHC type                 | TBD         |          |
| Skip SCC sort for known-independent let bindings                     | TBD         |          |
| Index liftBinding candidates by type fingerprint                     | TBD         |          |
| [Use sharing-preserving substitution in appProp beta reduction        | ❌         | Hot garbage. Not worth saving in its current incarnation.](https://github.com/clash-lang/clash-compiler/pull/3341)         |

## Still TODO:

  - [ ] Go over each patch manually (@martijnbastiaan)
  - [ ] Kick out non-contributing patches
  - [ ] Merge / reject https://github.com/clash-lang/clash-compiler/pull/3328
    - [ ] Rebase
  - [ ] Merge / reject https://github.com/clash-lang/clash-compiler/pull/3331
    - [ ] Rebase


clash-lib/src/Clash/Driver.hs
```diff
@@ -517,66 +517,177 @@ generateHDL env design hdlState typeTrans peEval eval mainTopEntity startTime =
...
+
+-- | Module the function of an 'InterpFunctionRequest' lives in, as a
+-- qualified module name
+ifrQualMod :: InterpFunctionRequest -> Hint.ModuleName
+ifrQualMod = intercalate "." . ifrModNames
+
+-- | The result of interpreting an 'InterpFunctionRequest'
+data InterpFunctionResult
+  = InterpBBF BlackBoxFunction
+  | InterpTF TemplateFunction
```
martijnbastiaan, line 550:
```suggestion
-- | The result of interpreting an 'InterpretFunctionRequest'
data InterpretFunctionResult
  = InterpretBbf BlackBoxFunction
  | InterpretTf TemplateFunction
```

Don't save two characters, just write it out. Use Rust-style capitalization. Both comments go for the rest of the module.

clash-lib/src/Clash/Driver.hs
```diff
@@ -517,66 +517,177 @@ generateHDL env design hdlState typeTrans peEval eval mainTopEntity startTime =
...
+  , ifrType :: InterpFunctionType
+    -- ^ Type to interpret the function at
+  , ifrSource :: Maybe Text
+    -- ^ Inline Haskell source of the module, if the primitive provided one
+  } deriving (Eq, Show, Generic, Hashable)
+
+-- | Module the function of an 'InterpFunctionRequest' lives in, as a
+-- qualified module name
+ifrQualMod :: InterpFunctionRequest -> Hint.ModuleName
+ifrQualMod = intercalate "." . ifrModNames
```
martijnbastiaan, line 545: `interpFunctionRequestToModuleName`? Yeah it's a lot, but at least then you don't have to scroll up to know what the heck `ifrQualMod` is supposed to mean.

clash-lib/src/Clash/Driver.hs
```diff
@@ -517,66 +517,177 @@ generateHDL env design hdlState typeTrans peEval eval mainTopEntity startTime =
...
+    Hint.setImports
+      ["Clash.Netlist.Types", "Clash.Netlist.BlackBox.Types", ifrQualMod req]
+    interpret req
+
+  interpretLocal defaultPath req = withSourceDir $ \extraDirs -> do
+    Hint.set [Hint.searchPath Hint.:= (extraDirs ++ idirs ++ defaultPath)]
+    Hint.loadModules [ifrQualMod req]
+    Hint.setImports
+      ["Clash.Netlist.BlackBox.Types", "Clash.Netlist.Types", ifrQualMod req]
+    interpret req
```
martijnbastiaan, line 677: Honestly the fact that this has these ad-hoc imports signals that this feature wasn't a good idea. Let's add a TODO to look into deprecating it.

clash-lib/src/Clash/Driver.hs
```diff
@@ -517,66 +517,177 @@ generateHDL env design hdlState typeTrans peEval eval mainTopEntity startTime =
...
+hintLanguageExtensions =
+  map Hint.asExtension $
+    map show wantedLanguageExtensions ++
+    map ("No" ++ ) (map show unwantedLanguageExtensions)
+
+-- | The functions a primitive needs interpreted before it can be compiled by
+-- 'compilePrimitiveWith'. Functions in 'knownBlackBoxFunctions' or
+-- 'knownTemplateFunctions' need no interpretation and yield no request. NB:
+-- keep this function in sync with the interpreter results consumed by
+-- 'compilePrimitiveWith'.
```
martijnbastiaan, line 567: What do you mean "keep in sync with"? Maybe just drop the comment, it seems confusing.

clash-lib/src/Clash/Driver.hs
```diff
@@ -517,66 +517,177 @@ generateHDL env design hdlState typeTrans peEval eval mainTopEntity startTime =
...
+        [InterpFunctionRequest modNames funcName InterpTemplateFunction Nothing]
+   where
+    fullName = intercalate "." modNames ++ "." ++ funcName
+neededInterpRequests (Primitive {}) = []
+
+-- | Interpret the given functions in a single shared Hint session. Starting a
+-- session is expensive (it initializes a full GHC session, including reading
+-- all package databases), so sharing one session over all requests is much
+-- faster than a session (or two) per function. For each request this action
+-- tries two things, mirroring the historical per-function behavior:
```
martijnbastiaan, line 594:
```suggestion
-- faster than a session per function. For each request this action
-- tries two things:
```

Don't comment on "historical things". Code comments should explain how it is NOW, unless there is truly some arcane reason for things being the way they are -- in that case it's fine to comment on its history.

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -56,28 +57,33 @@ import Clash.Rewrite.Util (changed, isUntranslatableType, setChanged, whnfRW)
...
-  , NoDeDup `notElem` ticks
-  , Just (i1,_) <- List.find ((== e) . snd) processed
-  = do
-    let subst1 = extendIdSubst subst i (Var i1)
-    setChanged
-    reduceBinders subst1 processed rest
-  | otherwise
-  = reduceBinders subst ((i,e):processed) rest
+reduceBinders subst0 processed0 =
+  go subst0 (Map.fromList [(e, i) | (i, e) <- reverse processed0]) processed0
```
martijnbastiaan, line 69:
```suggestion
reduceBinders subst0 bindings =
  go subst0 (Map.fromList [(e, i) | (i, e) <- reverse bindings]) bindings
```

With `go` being the recursive step this is much more natural.

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -56,28 +57,33 @@ import Clash.Rewrite.Util (changed, isUntranslatableType, setChanged, whnfRW)
...
+ where
+  -- The index maps a processed binding's RHS to its binder, keyed on
+  -- alpha-equivalence ('Ord' on 'Term'). It replaces a linear scan over the
+  -- processed bindings that did an alpha-equivalence comparison per entry.
+  -- Later insertions shadow earlier ones, like the head-first linear scan
+  -- they replace did.
```
martijnbastiaan, line 75: Again, don't comment on previous iteration of the code.

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -56,28 +57,33 @@ import Clash.Rewrite.Util (changed, isUntranslatableType, setChanged, whnfRW)
...
+  go !subst _index processed [] = return (subst,processed)
+  go !subst index processed ((i,substTm "reduceBinders" subst -> e):rest)
+    | (_,_,ticks) <- collectArgsTicks e
+    , NoDeDup `notElem` ticks
+    , Just i1 <- Map.lookup e index
```
martijnbastiaan, line 80: IMO we should just implement https://arxiv.org/abs/2105.02856 and use a `HashMap`.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -666,7 +666,7 @@ inlineSmallWorker _ e@(collectArgsTicks -> (Var f,args,ticks)) = do
         -- Don't inline recursive expressions
         Just b -> do
           isRecBndr <- isRecursiveBndr f
-          if not isRecBndr && not (isNoInline (bindingSpec b)) && termSize (bindingTerm b) < sizeLimit
+          if not isRecBndr && not (isNoInline (bindingSpec b)) && termSizeSmallerThan sizeLimit (bindingTerm b)
```
martijnbastiaan, line 669: For a separate commit: we should probably do the cheap check first.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -747,7 +747,7 @@ inlineWorkFreeWorker _ e@(Var f) = do
               sizeLimit <- Lens.view inlineWFCacheLimit
               -- caching only worth it from a certain size onwards, otherwise
               -- the caching mechanism itself brings more of an overhead.
```
martijnbastiaan, line 749: No good place to comment, so doing it here: I was (perhaps am) confused about the existence of `inlineSmall`. It seems like this is only relevant "mid normalization", but flattening will _always_ inline in the end. That's probably worth a comment.

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -129,7 +130,34 @@ newtype NetlistMonad a =
...
+-- | Cache for Core-type to 'HWType' translation, keyed structurally. See
+-- 'StructuralType'.
+type HWMap = HashMap StructuralType (Either String FilteredHWType)
+
+-- | A 'Type' keyed for cache lookups: 'Eq' and 'Hashable' compare types
+-- structurally instead of up to alpha-equivalence. Alpha-comparison computes
+-- the free variables of both types on every comparison, which dominated the
+-- cost of 'HWMap' lookups. Types that differ only in the names of bound
+-- variables get separate — but semantically identical — cache entries, which
+-- is harmless for a cache.
```
martijnbastiaan, line 142:
```suggestion
-- | A 'Type' keyed for cache lookups: 'Eq' and 'Hashable' compare types
-- structurally instead of up to alpha-equivalence. This can be useful for
-- caches, where it is more important to do quick lookups than for the map
-- to be alpha-equivalent duplicate free.
```

clash-lib/src/Clash/Netlist/Types.hs
```diff
@@ -129,7 +130,34 @@ newtype NetlistMonad a =
...
+newtype StructuralType = StructuralType Type
+
+instance Eq StructuralType where
+  StructuralType t1 == StructuralType t2 = eqType t1 t2
+
+instance Hashable StructuralType where
+  hashWithSalt salt0 (StructuralType ty0) = go salt0 ty0
```
martijnbastiaan, line 149: I like the idea of this but it is also slightly surprising that there would be "duplicate" types in the resulting map. I'd like to see if we can get similar speedups with https://arxiv.org/abs/2105.02856.

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -178,393 +164,613 @@ reduceNonRepPrim = transform "reduceNonRepPrim" (onApp 'reduceNonRepPrimWorker)
...
+    TyConApp tcNm _
+      | nameOcc tcNm == vecTcName -> True
+      | otherwise -> case UniqMap.lookup tcNm tcm of
+          -- Type families might reduce to a Vec
+          Just FunTyCon{} -> True
+          Just _ -> False
+          Nothing -> True
+    OtherType otherTy -> case otherTy of
+      LitTy _ -> False
+      _ -> True
```
martijnbastiaan, line 246: This looks complicated to me. Why not rely on `zeroLengthVecTerm`?

clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -178,393 +164,613 @@ reduceNonRepPrim = transform "reduceNonRepPrim" (onApp 'reduceNonRepPrimWorker)
...
-        -> case runExcept (tyNatSize tcm nTy) of
-          Right n -> do
-            shouldReduce1 <- List.orM [ pure ultra
-                                 , shouldReduce ctx
+reduceNonRepPrimWorker c e _appFunction _appArgument
+  | (Prim p, args, ticks) <- collectArgsTicks e
+  = do
+    tcm <- Lens.view tcCache
+    case HashMap.lookup (primName p) reduceNonRepPrimImpls of
+      Just handler -> do
```
martijnbastiaan, line 172: The double-check on `zeroLengthVecTerm` is confusing to me -- it happens when the primitive is found _and_ when it is not found. Why not wrap the whole thing in a check on `zeroLengthVecTerm` then?

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -631,11 +632,18 @@ liftBinding (var@Id {varName = idName} ,e) = do
       -- Create a new body that abstracts over the free variables
       newBody = mkTyLams (mkLams e' boundFVs) boundFTVs
 
-  -- Check if an alpha-equivalent global binder already exists
-  aeqExisting <- (UniqMap.elems . UniqMap.filter ((`aeqTerm` newBody) . bindingTerm)) <$> Lens.use bindings
+  -- Check if an alpha-equivalent global binder already exists. Terms can
+  -- only be alpha-equivalent if their types are, so candidates are
+  -- prefiltered on alpha-equality of their (much smaller) types. The scan
```
martijnbastiaan, line 637:
```suggestion
  -- prefiltered on alpha-equality of their (typically much smaller) types. The scan
```

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -883,6 +885,27 @@ aeqType t1 t2 = acmpType' rnEnv t1 t2 == EQ
  where
   rnEnv = mkRnEnv (mkInScopeSet (freeVarsOf [t1,t2]))
 
+-- | A fingerprint of a 'Type' modulo alpha-equivalence: alpha-equivalent
+-- types have equal fingerprints. The converse does not hold, so the
+-- fingerprint is only usable to prefilter candidates for the more expensive
+-- 'aeqType'. All type variables hash alike, and annotation attributes are
+-- ignored; both make the fingerprint coarser, never unsound.
+typeAlphaFingerprint :: Type -> Int
```
christiaanb, line 893: The name of this function is just wrong as it has nothing to do with alpha-equivalence. The fingerprint is just oblivious to variable references. A better name would me something like `typeFingerprintVarOblivious`. Yes, it's long name, and that's the whole point: it should instill the urge to any (human) reader of code using this function to want to figure out what this function is actually about.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -883,6 +885,27 @@ aeqType t1 t2 = acmpType' rnEnv t1 t2 == EQ
...
+typeAlphaFingerprint = go (0x2f9be6cc :: Int)
```
christiaanb, line 894: What's the origin of this specific seed?

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -655,6 +656,88 @@ substGlobalsInTerm globals = \term -> fromMaybe term (go term)
...
+-- replacement term: no binder in the traversed term can then capture a
+-- variable of the replacement term, so binders never have to be renamed —
+-- where they would, 'substTm' must be used instead.
+--
+-- Sharing-preserving: a subterm in which no substitution takes place is
+-- returned as the original subterm instead of being reallocated, and types
+-- are never touched, unlike 'substTm', which rebuilds every node it visits.
+-- Binders shadowing the substituted variable end the substitution in their
+-- scope.
+substLocalIdInTerm
```
christiaanb, line 671: I would just call this function `unsafeSubstAbortsOnShadow`, as that what it basically is: substitution that is not capture-avoiding and hence unsafe in the general case, and on top of that, the substitution is completely and fully aborted if the expression has a binder that shadows the variable for which we are substituting.

clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -631,11 +632,18 @@ liftBinding (var@Id {varName = idName} ,e) = do
...
+  -- also stops at the first match, where it used to filter the whole binding
+  -- map with a term alpha-equality check per entry.
+  let aeqExisting = List.find
+        (\b -> varType (bindingId b) `aeqType` newBodyTy
+                 && bindingTerm b `aeqTerm` newBody)
+        (UniqMap.elems binders)
```
martijnbastiaan, line 643: This commit seems _meh_ to me. We don't control binders (recurring theme in this PR), so this will break complexity wise whenever there happen to be many binders with the same type. I'd like to see a refactor that turns `bindings` into its own type that keeps an index on `Uniq` (`IntMap`-based) and its `Term` (`HashMap` based).

clash-lib/src/Clash/Normalize.hs
```diff
@@ -403,11 +403,15 @@ flattenCallTree cache (CBranch (nm,(Binding nm' sp inl pr tm r)) used) = do
...
-         newExpr' <- rewriteExpr ("flattenCheap",flatten) (showPpr nm, tm1) (nm', sp)
-         return (CBranch (nm,(Binding nm' sp inl pr newExpr' r)) (concat allUsed'))
+         case Maybe.catMaybes toInline' of
+           -- Nothing to inline: substituting and re-running the flatten
+           -- strategy would be working on an unchanged expression.
+           [] -> return (CBranch (nm,(Binding nm' sp inl pr newExpr r)) allUsed)
+           toInline1 -> do
+             let subst' = extendGblSubstList (mkSubst emptyInScopeSet) toInline1
+             let tm1 = substTm "flattenCallTree.flattenCheap" subst' newExpr
+             newExpr' <- rewriteExpr ("flattenCheap",flatten) (showPpr nm, tm1) (nm', sp)
```
martijnbastiaan, line 413: This commit doesn't offer a measurable speedup. I also think this fix is in the wrong place: `substTm` should just do nothing when it gets and empty substitution. I'd be fine accepting this if it were just that -- even if it doesn't immediately offer a speedup.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -655,6 +656,88 @@ substGlobalsInTerm globals = \term -> fromMaybe term (go term)
...
+-- respect to an in-scope set that contains the free variables of the
+-- replacement term: no binder in the traversed term can then capture a
+-- variable of the replacement term, so binders never have to be renamed —
+-- where they would, 'substTm' must be used instead.
+--
+-- Sharing-preserving: a subterm in which no substitution takes place is
+-- returned as the original subterm instead of being reallocated, and types
+-- are never touched, unlike 'substTm', which rebuilds every node it visits.
+-- Binders shadowing the substituted variable end the substitution in their
+-- scope.
```
christiaanb, line 670: That's not how the function is behaving, I see this in the implementation:
```
    Lam v e
      | v `shadows` target -> Nothing
      | otherwise -> Lam v <$> go e
```
i.e. when there is any shadowing happening in the term in which the substitution is taking place, the original term is returned. The description makes it sound like substitution did take place, just not beyond the shadowing.

So from the documentation it sounds like:
```
substLocalIdInTerm x u (let y = x in \x -> let x = y in x)
```
we would get back the term:
```
let y = u in \x -> let x = y in x
```
but what the the implementation actually returns is:
```
let y = x in \x -> let x = y in x
```
i.e. the original term.

clash-lib/src/Clash/Normalize/Transformations/Inline.hs
```diff
@@ -652,27 +654,37 @@ inlineSmall =
...
+              , termSizeSmallerThan sizeLimit (bindingTerm b)
+              -> do
+                -- Don't inline recursive expressions
+                isRecBndr <- isRecursiveBndr f
+                if isRecBndr
+                   then return e
+                   else do
+                     untranslatable <- isUntranslatable True e
+                     if untranslatable
+                        then return e
```
martijnbastiaan, line 682: Use [`orM`](https://hackage-content.haskell.org/package/extra-1.8.1/docs/Data-Foldable-Extra.html#v:orM) to prevent nesting.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -655,6 +656,88 @@ substGlobalsInTerm globals = \term -> fromMaybe term (go term)
...
+substLocalIdInTerm
+  :: Id
+  -- ^ The local variable to substitute
+  -> Term
+  -- ^ The term to replace it with
+  -> Term
+  -- ^ The term to substitute in
+  -> Term
+substLocalIdInTerm target payload = \term -> fromMaybe term (go term)
```
christiaanb, line 679: The `Maybe` is just doing too much! It captures both the fact whether substitution took place and whether there was any shadowing with regards to the variable for which we are substituting.

I understand that if you don't go beyond a certain place because of shadowing, than no substitution took place there. But I think we should disambiguate the two such that for:
```
substLocalIdInTerm x u (let y = x in \x -> let x = y in x)
```
we do actually return:
```
let y = u in \x -> let x = y in x
```
I am very surprised that we are not seeing any bugs in how this function is used in `appProp`.

christiaanb (commented):
I understand the intention of ec4b14c5a159e2c03e6f5d181ab1160d48bc24c9 "Use sharing-preserving substitution in appProp beta reduction" But the implementation is just wrong in so many self-contradicting ways.

martijnbastiaan:
Just to be clear, this is still WIP @christiaanb, I haven't gotten to all commits yet. But thanks for your comments :).

martijnbastiaan (commented):
First few commits.. few more to go.

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Yeah, or `hashTypeFast` or something.

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: > the substitution is completely and fully aborted if the expression has a binder that shadows the variable for which we are substituting.

Which seems like an intensely dumb thing to do anyway. The precondition is that this shouldn't happen, so it should error if it does!

martijnbastiaan:
> I understand the intention of https://github.com/clash-lang/clash-compiler/commit/ec4b14c5a159e2c03e6f5d181ab1160d48bc24c9 "Use sharing-preserving substitution in appProp beta reduction" But the implementation is just wrong in so many self-contradicting ways.

It seems very stupid, I'll drop it. What's worth trying is that substitution checks for shadowing and falls back to the expensive path when that happens. Depending on how often shadowing actually happens, this might save quite a bit.

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: I mean, the codebase used to be littered with deshadowing issues and they only came up ever so often :-).

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: But also, the core of the problem here is that each `==` involves free var lookup. If we'd implement https://arxiv.org/abs/2105.02856 and use proper `HashMap`s, this would just become "normal" intuitive code.

clash-lib/src/Clash/Core/Subst.hs
christiaanb: Well, the precondition just states it’s been deshadowed w.r.t. the inscopeset of the context, which is sufficient to prevent capture (of free variable references in the substituted-for expression). That doesn’t mean that the variable “we pick” to substitute for is not being shadowed by the binders in the expression we are substituting in.

clash-lib/src/Clash/Normalize.hs
christiaanb: An empty substitution, but potentially non-empty inscopeset, is how we implement deshadowing currently. So you cannot change `substTm` like that without consequences.

christiaanb:
> > I understand the intention of [ec4b14c](https://github.com/clash-lang/clash-compiler/commit/ec4b14c5a159e2c03e6f5d181ab1160d48bc24c9) "Use sharing-preserving substitution in appProp beta reduction" But the implementation is just wrong in so many self-contradicting ways.
> 
> It seems very stupid, I'll drop it. What's worth trying is that substitution checks for shadowing and falls back to the expensive path when that happens. Depending on how often shadowing actually happens, this might save quite a bit.

No, the entire shadowing check seems non-sensible anyhow. In line
https://github.com/clash-lang/clash-compiler/commit/ec4b14c5a159e2c03e6f5d181ab1160d48bc24c9#diff-38424dec2448cd00ad74cbd954d63cd04bc9d70ae89eaf7eecec8087669d2a33L205
we start with deshadowing the expression in the function position, `fun` with regards to the InScopeSet. This is sufficient to prevent capture of any free variables of the expression in the argument position. However, it also means that none of the binders in the function expression shadow one another any longer!

Then in line
https://github.com/clash-lang/clash-compiler/commit/ec4b14c5a159e2c03e6f5d181ab1160d48bc24c9#diff-38424dec2448cd00ad74cbd954d63cd04bc9d70ae89eaf7eecec8087669d2a33R223
We start the subsitution `substLocalIdInTerm v arg e`. Those arguments come from line:
https://github.com/clash-lang/clash-compiler/commit/ec4b14c5a159e2c03e6f5d181ab1160d48bc24c9#diff-38424dec2448cd00ad74cbd954d63cd04bc9d70ae89eaf7eecec8087669d2a33R214
Basically, that `Id`, the `v` is the binder of the lambda-expression `Lam v e` that is the earlier-mentioned function expression `fun`; the one that has been deshadowed! That should mean that there should be now bindings in `e` that shadow `v`!

So I'm fine with an `unsafeSubsTm` that assumes an expression that is fully deshadowed with regards to the free variables in the replacement term. Meaning that:
1. It is fully deshadowed
2. There should be no possibility for free variable capture in the replacement term.

Then `unsafeSubstTm` can just do the "naive" subsitution. And on top of that, let's say that the local variable `v` to substitute does not occur free in `e`, the term to substitute in, that we just return the unchanged `e`. Basically, `unsafeSubstTm` just takes the structure from `substGlobalsInTerm` but replaces the line `| isGlobalId v -> lookupVarEnv v globals` with `| v == target -> Just payload`. Actually what we should do, is create:

```haskell
unsafeSubstTm
  :: VarEnv Term
  -- ^ Substitution: global variable to replacement term
  -> VarEn Term
  -- ^ Substitution: local variable to replacement term
  -> Term
  -- ^ Term to subsitute in
  -> Term
unsafeSubtTm globals locals = \term -> fromMaybe term (go term)
 where
  go :: Term -> Maybe Term
  go = \case
    Var v
      | isGlobalId v -> lookupVarEnv v globals
      | otherwise -> lookupVarEnv v locals
    Lam v e -> Lam v <$> go e
    TyLam tv e -> TyLam tv <$> go e
    App l r -> case (go l, go r) of
      (Nothing, Nothing) -> Nothing
      (l1, r1) -> Just (App (fromMaybe l l1) (fromMaybe r r1))
    TyApp e ty -> (`TyApp` ty) <$> go e
    Let bs body -> case (goBind bs, go body) of
      (Nothing, Nothing) -> Nothing
      (bs1, body1) -> Just (Let (fromMaybe bs bs1) (fromMaybe body body1))
    Case subject ty alternatives ->
      case (go subject, goList goAlternative alternatives) of
        (Nothing, Nothing) -> Nothing
        (subject1, alternatives1) ->
          Just (Case (fromMaybe subject subject1) ty
                     (fromMaybe alternatives alternatives1))
    Cast e t1 t2 -> (\e1 -> Cast e1 t1 t2) <$> go e
    Tick tickInfo e -> case (goTickInfo tickInfo, go e) of
      (Nothing, Nothing) -> Nothing
      (tick1, e1) -> Just (Tick (fromMaybe tickInfo tick1) (fromMaybe e e1))
    -- Data, Literal, Prim
    _ -> Nothing

  goBind (NonRec v rhs) = NonRec v <$> go rhs
  goBind (Rec bindings) = Rec <$> goList goBinding bindings

  goBinding (v, rhs) = (,) v <$> go rhs

  goAlternative (pat, alternative) = (,) pat <$> go alternative

  -- Types contain no term variables, so of the tick constructors only
  -- 'Attributes' — which carries a term — needs a traversal.
  goTickInfo (Attributes ty e) = Attributes ty <$> go e
  goTickInfo _ = Nothing

  goList :: (a -> Maybe a) -> [a] -> Maybe [a]
  goList f = goElements
   where
    goElements [] = Nothing
    goElements (x:xs) = case (f x, goElements xs) of
      (Nothing, Nothing) -> Nothing
      (x1, xs1) -> Just (fromMaybe x x1 : fromMaybe xs xs1)
```
And then we can just implement `substGlobalsInTerm` as:
```
substGlobalsInTerm globals = unsafeSubtTm globals mempty
```

martijnbastiaan:
@christiaanb Alright, I was going to drop it because it seems mighty fragile. But if you're happy with it I'll offer it anyway.

I'm going to pull this PR apart into many small PRs to make this reviewable.

martijnbastiaan:
> What's worth trying is that substitution checks for shadowing and falls back to the expensive path when that happens. Depending on how often shadowing actually happens, this might save quite a bit.

I tried this, but the extra bookkeeping cancels out the saved GC.

martijnbastiaan:
I'm closing this PR, I'll open a bunch of smaller ones to make it much easier to review.

PR #3334, opened 2026-07-27
Prevent checkbox comment by Mergify

martijnbastiaan:
See https://docs.mergify.com/changelog/2026-06-19-queue-or-requeue-a-pull-request-with-a-checkbox/

## Still TODO:

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -25,3 +25,8 @@ pull_request_rules:
         merge_conflict_style: diff3
         branches:
           - "1.10"
+
+merge_queue:
+  # Stop Mergify's from commenting "Tick the box to add this pull request to the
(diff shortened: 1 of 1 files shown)


DigitalBrains1 (approved):
It's not just that it adds one comment to a PR... it seems to post the same message again every time after someone posts a new comment or review in the PR. We don't need this message at all, not even once, but the spamming is the really obnoxious part. I can't really imagine that is by design...

DigitalBrains1:
Ah, I think what happens is that it re-posts the message if you hide its earlier message and you post a new comment. So I ended up in a cycle of hide mergify comment - post new review or comment - hide new mergify comment - repeat ad nauseam (which occurred rather quickly)

PR #3335, opened 2026-07-30
Prevent full traverse in `Ord Term` and `Eq Term`

martijnbastiaan:
Bit of context: I was implementing `Hashable Term` and through it found a number of issues with `aeqTerm` and `acmpTerm`, as well as a good opportunity for a performance jump. Most of this PR is therefore spent on setting up a usable test harness. The `Hashable Term` PR will follow soon.

Core algorithm and tests are by and large _artisanal code_, but benchmarks and changes to the test harness are _mostly_  LLM generated.

# Core change
The new method does not need to compute free variables in order to function. Instead, it effectively renames *all* variables based on their position in the tree.

To benchmark this, a downstream design (`wireDemoTest` @ `bittide-hardware`) had its `BindingsMap` dumped and its bindings put in buckets based on the number of `Var` occurences:

| bucket    | size range | bindings | share |
|-----------|------------|----------|-------|
| nullary   | 0          | 4 414    | 35.8% |
| tiny      | 1-10       | 6 374    | 51.7% |
| small     | 11-100     | 1 349    | 10.9% |
| medium    | 101-1000   | 190      | 1.5%  |
| large     | 1001+      | 2        | 0.02% |

On each of these groups, a few tests were run in groups of 30:

  * self: `aeqTerm t t`
  * renamed: `aeqTerm t0 t1` where `t1 = freshenTm t0` (benchmarks *include* `freshenTm` call, so figures look pessimistic for the PR-code)
  * distinct: `aeqTerm t0 t1`
  * dedup sweep: `nubBy aeqTerm`

Timings aren't normalized "per call" / "per binder". Results:

| group               | master       | pr           |
|---------------------|--------------|--------------|
| nullary/self        | 2.221 µs     | **2.272 µs** |
| nullary/renamed     | 2.658 µs     | **2.625 µs** |
| nullary/distinct    | **1.103 µs** | 1.154 µs     |
| nullary/dedup sweep | **9.642 µs** | 10.72 µs     |
| tiny/self           | 288.9 µs     | **11.99 µs** |
| tiny/renamed        | 340.2 µs     | **45.25 µs** |
| tiny/distinct       | 110.7 µs     | **4.089 µs** |
| tiny/dedup sweep    | 196.8 µs     | **19.10 µs** |
| small/self          | 578.4 µs     | **61.98 µs** |
| small/renamed       | 767.7 µs     | **198.3 µs** |
| small/distinct      | 168.6 µs     | **14.84 µs** |
| small/dedup sweep   | 742.3 µs     | **72.52 µs** |
| medium/self         | 7.992 ms     | **764.4 µs** |
| medium/renamed      | 10.03 ms     | **2.405 ms** |
| medium/distinct     | 769.0 µs     | **3.998 µs** |
| medium/dedup sweep  | 2.696 ms     | **24.97 µs** |

"Large" was left out because it only contains two bindings.

As far as end-to-end tests go, this only yields a modest improvement: 4m27 -> 4m19s total, 4m1s -> 3m53s normalization.

## Still TODO:


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -1016,9 +912,34 @@ eqTerm = go
...
+'acmpTermLevels' (and hence 'aeqTerm', 'acmpTerm', @Eq Term@ and @Ord Term@)
+does not compare the result type stored in a 'Case' constructor, and 'eqTerm'
+follows it. This is OK for case expressions with at least one alternative: the
+result type is determined by the alternatives. Every alternative's right-hand
+side has exactly the result type, and (alpha-)equal right-hand sides have
+(alpha-)equal types, so comparing the result type as well would be redundant
+work.
+
+For a case expression with no alternatives the result type is /not/ determined
+by the subterms, so @(case x of {}) :: A@ and @(case x of {}) :: B@ compare
```
christiaanb, line 958: Empty case-expression occurring in Haskell source do not end up in Clash Core, they are filtered out in the GHC Core to Clash Core conversions.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -1035,6 +956,218 @@ instance TypeError (
   ) => Hashable Term where
     hashWithSalt = error "Term.hashWithSalt: unreachable"
 
+{- Note [Numbering binders by level]
```
christiaanb: Can we just say that we convert to a [DeBruijn Index](https://en.wikipedia.org/wiki/De_Bruijn_index), I was confused by the word "level" as there is also such a thing as a "DeBruijn Level" that counts the other way around.

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -1035,6 +956,218 @@ instance TypeError (
...
+  Int ->
+  [Var a] ->
+  [Var a] ->
+  VarEnv Int ->
+  VarEnv Int ->
+  (Int, VarEnv Int, VarEnv Int)
+extendLevels lvl vs1 vs2 envL envR =
+  List.foldl' one (lvl, envL, envR) (List.zipEqual vs1 vs2)
+ where
+  one (!l, eL, eR) (v1, v2) =
```
christiaanb, line 1024: Why only strict on the `l` and not on `eL` and `eR`? Is this to avoid the work of `extendVarEnv` just in case you actually never do a lookup?

clash-lib/src/Clash/Core/Term.hs
```diff
@@ -123,22 +126,30 @@ data TickInfo
...
-      getRank NoDeDup       = 3
-      getRank Attributes {} = 4
+  deriving (Show, Generic, NFData, Binary)
+
+-- | Structural equality on ticks.
+--
+-- 'TickInfo' deliberately has no 'Eq' or 'Ord' instance. Equality on 'Term' and
+-- 'Type' is defined as alpha equivalence. Most callers want structural equality
+-- and callers interested in alpha equivalence can't typically use 'Eq' as they
+-- need to pass their enclosing scope. That leaves only a choice between
```
christiaanb: Why would they need to pass their enclosing scope? None of the alpha-eq/ord functions take such an extra argument.

christiaanb (approved):
Looks good!

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Ah I wasn't aware. This was all developed on intuition 🙈 . Will do!

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: No, just second nature whenever I'm yielding `+1`. It's a tiny tiny operation, memory overhead is (I think) always going to lose against just doing the calculation. I'm not sure about the var envs..

So to be clear, this is just a pattern I've picked up. Do you want me to drop it, or put the bang patterns on `eL` and `eR` too?

clash-lib/src/Clash/Core/Term.hs
martijnbastiaan: My logic was, there are basically two times you want to do an equality test:

1. Nubbing. I.e., deduplicating `Tick`s already at the same level / scope.
2. As part of `aeqTerm`/`acmpTerm`.

For the first you don't want to pay for alpha equality checks. For the second you do need to pass in the environment. I.e., AFAIK in all the cases where you _actually_ want to do alpha equality you'd have to pass it in / use a custom function.

So either you pick `Eq` to be structural equality, which clashes with `Eq Term` or you pick it to be alpha equality, but now you never actually want to use that instance. Given that there is no good option, I'd prefer it to just not have an instance and have consumers pick what they want.

Does that make sense?

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Ah good to know! I'll amend the comment.

clash-lib/src/Clash/Core/Subst.hs
christiaanb: No, it's fine.

clash-lib/src/Clash/Core/Term.hs
christiaanb: Well, ticks don't do any binding, so there's no alpha-equivalence to speak of. I don't see how the new `eqTickInfo` is any different to the old `Eq` instance, given that it has the line:

```haskell
eqTickInfo (Attributes t1 a1) (Attributes t2 a2) = t1 == t2 && a1 == a2
```
and that `a1 == a2` uses `aeqTerm`.

And I really do think we want for `eqTickInfo` on:
```
Attributes t (\x -> x)
```
and
``` 
Attributes t (\y -> y)
```
to return `True`, also for things such as `nub`.

clash-lib/src/Clash/Core/Subst.hs
christiaanb: It’s this line: https://github.com/clash-lang/clash-compiler/blob/df6369060086e126d8a4cc5e6f77622aa190f71c/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs#L434

clash-lib/src/Clash/Core/Term.hs
martijnbastiaan: Yeah, fair!

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -1035,6 +976,268 @@ instance TypeError (
...
 
+{- Note [Numbering binders by De Bruijn index]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Alpha comparison compares two variable occurrences by their De Bruijn index:
+the number of binders crossed on the way out from the occurrence to the binder
+it resolves to.
+
+Rather than renumbering as it descends, the comparison records for each binder
+how many binders enclosed it, its /depth/, and recovers the index at an
+occurrence by subtracting that depth from the depth of the occurrence itself.
```
christiaanb: Ahah! You are actually using De Bruijn levels, my apologies! See https://blueberrywren.dev/blog/debruijn-explanation/

where in the names version we have:
```
Named:   λ f. (λ y f. (f y)) f
```
and then in the De Bruijn levels we get:
```
Levels:  λ (λ λ (2 1)) 0
```

It means we can drop this whole part about "subtracting".

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -1035,6 +976,268 @@ instance TypeError (
...
+  Var a ->
+  Var a ->
+  Ordering
+acmpOccIndices envL envR v1 v2 =
+  case (lookupVarEnv v1 envL, lookupVarEnv v2 envR) of
+    -- Both bound: equal exactly when they have the same index
+    --
+    -- N.B.: Indices are @depth - depthL@ and @depth - depthR@, but a comparison
+    --       on depthR and depthL (flipped!) gives the same result for less work
+    (Just depthL, Just depthR) -> compare depthR depthL
```
christiaanb: Does this flipping really matter? it only matters whether we have a stable order, no?

So whether
```
compare (\x y.x) (\x y.y)
```
returns `GT` or `LT` doesn't really matter, does it?

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Ah, now I'm confused :see_no_evil:. I let Claude do the rename to indices and it snuck in the subtraction. I then read up on it and it seemed fine, but I didn't loop it back again to the reason we were doing the refactor anyway.....

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: It doesn't, yeah. I was trying to prevent confusion by highlighting that you need to flip terms if you rewrite:

```
depth - depthL `compare` depth - depthR
```

to:

```
depthR `compare` depthL
```

but I see that only adds to the confusion. I'll revert back to levels anyway so it should fix itself :-).

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Alright, after reading that blog post: what I originally called levels are actually also "De Bruijn levels", correct?

christiaanb: Yes, correct!

martijnbastiaan: Thanks! Will revert in that case.

PR #3336, opened 2026-07-30
Add `Hashable Term` and `Hashable Type`

martijnbastiaan:
This piggy-backs off the infrastructure made in https://github.com/clash-lang/clash-compiler/pull/3335 and follows the same idea for constructing hashes as the ideas for comparing terms.

This doesn't do anything on its own, but paves the way for using `HashMap`/`HashTable` for quick lookups.

I had originally intended to base this off of [Hashing Module Alpha-Equivalence](https://arxiv.org/abs/2105.02856), but that doesn't make a lot of sense. That paper focuses on hashing in a way that creates stable (position independent) hashes for all subterms. This is super useful for common subexpression elimination, but too computationally complex if you just want a hash for the root of an expression.

## Still TODO:

  - [x] Merge https://github.com/clash-lang/clash-compiler/pull/3335

### DIFF
--- a/clash-lib/src/Clash/Core/Subst.hs
+++ b/clash-lib/src/Clash/Core/Subst.hs
@@ -14,7 +14,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE StandaloneDeriving #-}
-{-# LANGUAGE UndecidableInstances #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
@@ -82,10 +81,9 @@ import qualified Data.List                 as List
 import qualified Data.List.Extra           as List
 import           Data.Ord                  (comparing)
 import           GHC.Stack                 (HasCallStack)
-import           GHC.SrcLoc.Extra          ()
-import           GHC.Types.SrcLoc          (leftmost_smallest)
-import           GHC.TypeLits
-  (TypeError, ErrorMessage (Text, (:<>:)))
+import           GHC.SrcLoc.Extra          () -- Hashable RealSrcSpan
+import           GHC.Types.SrcLoc
+  (SrcSpan (RealSrcSpan, UnhelpfulSpan), leftmost_smallest)
 
 import           Clash.Core.HasFreeVars
 import           Clash.Core.Pretty         (ppr, fromPpr)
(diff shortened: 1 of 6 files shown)


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -1171,3 +1162,171 @@ acmpTermLevels !lvl tmL tmR tyL tyR = go
...
+  Int
+hashOccLevels env salt v =
+  case lookupVarEnv v env of
+    Just lvl -> hashWithSalt salt (0 :: Int, lvl)
+    Nothing -> hashFreeOcc salt v
+
+-- | Hash a variable occurrence that no enclosing binder binds. Such an
+-- occurrence is identified by its 'varKey'.
+hashFreeOcc :: Int -> Var a -> Int
+hashFreeOcc salt v = hashWithSalt salt (1 :: Int, varKey v)
```
christiaanb: What's that `1 :: Int` doing there? Is that in case `varKey v` return zero or something? Worth a comment I think.

christiaanb (approved):
Looks good!

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: This way hashes differentiate between unbound and bound variables when they happen to have the same key/index. It's pretty unlikely of course, but it feels cleaner to do it.

christiaanb: But bound variables are hashed using their debruijn index. So you mean when the debruijn index of a bound variable is the same as the unique of free variable?

My remark stands that his requires a note/comment. Probably also at `hashOccLevels`.

Then finally, don't we want to distinquish between:
1. The deBruijn index of a bound variable
2. The unique of a free variable
3. The unique of a global variable reference

In that case you might also want to have a look at `goVar`.

martijnbastiaan: > But bound variables are hashed using their debruijn index. So you mean when the debruijn index of a bound variable is the same as the unique of free variable?

Yeah.

> My remark stands that his requires a note/comment. Probably also at hashOccLevels.

Will do!

> Then finally, don't we want to distinquish between:

It does, but I think the names are picked poorly. `hashFreeOcc` is also used for global variables ("out of scope" -- but only relative to `env`). That function then uses `varKey`, which does account for globalness.

I'll rework the names.

martijnbastiaan: It's now just:

```haskell
  -- N.B.: Variables are hashed with a "tag" to differentiate between, e.g., a
  --       local free variable with unique "2" and a bound free variable which
  --       happens to be bound at depth 2.
  goVar :: Int -> Id -> Int
  goVar salt i
    -- Global, never bound with respect to 'tmEnv'
    | isGlobalId i = hashWithSalt salt (0 :: Int, varUniq i)
    -- Local, bound variable
    | Just boundDepth <- lookupVarEnv i tmEnv = hashWithSalt salt (1 :: Int, boundDepth)
    -- Free, local variable
    | otherwise = hashWithSalt salt (2 :: Int, varUniq i)
```

which I feel is simpler than scattering it over functions. Same when hashing types.

christiaanb: Ah, I didn't realize that `varKey` was doing more than just getting the unique. That said, I like the new version!

martijnbastiaan: Great, thanks :).

martijnbastiaan:
@mergify rebase

PR #3337, opened 2026-07-30
Compile primitives in a single shared hint session

martijnbastiaan:
Each interpreted blackbox function paid for one or two fresh GHC sessions, back to back. Interpret all of them inside one session, and interpret each distinct function only once.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Driver.hs
+++ b/clash-lib/src/Clash/Driver.hs
@@ -4,6 +4,7 @@
                      2017     , QBayLogic, Google Inc.
                      2020-2024, QBayLogic,
                      2022     , Google Inc.
+                     2026     , Martijn Bastiaan
 
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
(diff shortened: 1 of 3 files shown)


clash-lib/src/Clash/Driver.hs
```diff
@@ -616,7 +726,31 @@ knownTemplateFunctions =
...
+  -> [FilePath]
+  -- ^ Package databases
+  -> FilePath
+  -- ^ The folder in which the GHC bootstrap libraries (base, containers, etc.)
+  -- can be found
+  -> ResolvedPrimMap
+  -- ^ Primitives to compile
+  -> IO CompiledPrimMap
+compilePrimitives idirs pkgDbs topDir primMapR = do
+  let reqs = HashSet.toList $ HashSet.fromList $
```
christiaanb: What's this `HashSet.toList $ HashSet.fromList` business?

christiaanb (approved):
LGTM, just a minor remark/question.

clash-lib/src/Clash/Driver.hs
martijnbastiaan: Ah, I completely missed that. This looks like a very dumb way to deduplicate.. Will fix, thanks.

PR #3338, opened 2026-07-30
Don't call evaluator as often while flattening

martijnbastiaan:
Yields about a 30% performance bump in `bittide-hardware`'s `wireDemoTest`, without changing HDL.

@christiaanb This is something Claude found and I have no idea how to judge whether this is correct or not, except for observing that both the testsuite and bittide survives.. What's your opinion?

## Still TODO:


christiaanb (commented):
From a functionality/completeness point of view, this should be fine, since everything is wrapped in the same `repeatR`. From a performance point of view, perhaps @rowanG077 can chime in, since he touched this code last, introducing the `topdownFixR` rewrite traversal.

rowanG077 (approved):
This looks fine to me. At first glance I wouldn't have expected this would be performance win. But numbers don't lie.

PR #3339, opened 2026-07-30
Dispatch `reduceNonRepPrim` through a `HashMap`

martijnbastiaan:
Yields about a 6% end-to-end speedup in `bittide-hardware`'s `wireDemoTest`.

Very similar to https://github.com/clash-lang/clash-compiler/pull/3331, but "hand" crafted instead of going through scripts.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
+++ b/clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
@@ -41,7 +41,7 @@ import Clash.Core.Term
   ( CoreContext(..), LetBinding, PrimInfo(..), Term(..), TickInfo(..), collectArgs
   , collectArgsTicks, mkApps, mkTicks, mkTmApps)
 import Clash.Core.TyCon (TyCon(..), TyConMap, tyConDataCons)
-import Clash.Core.Type (Type, TypeView(..), mkTyConApp, splitFunForallTy, tyView)
+import Clash.Core.Type (Type, TypeView(..), mkTyConApp, splitFunForallTy, tyView, coreView)
 import Clash.Core.Util (mkVec, shouldSplit, tyNatSize, mkInternalVar)
 import Clash.Core.VarEnv (extendInScopeSet)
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Normalize/Transformations/Reduce.hs
```diff
@@ -156,393 +143,609 @@ reduceConst _ e = return e
...
+    -- primitive, which is expensive to infer, so for a primitive without a
+    -- handler it is only inferred when the primitive's declared type shows a
+    -- @Vec@ result is possible at all. That rules out the vast majority of
+    -- them.
+    if Maybe.isNothing handlerM && not (mayReturnVec tcm (primType p))
+    then return e
+    else do
+      let eTy = inferCoreTypeOf tcm e
+      let resTy = snd (splitFunForallTy eTy)
+      let tv = tyView resTy
```
christiaanb: So in `maybeReturnVec` we over-approximate by allowing applied type families, but then the `tv` we pass to `zeroLenghtVecTerm` still has those un-reduced type family applications. I think we should do `coreView` on `resTy`, see
https://github.com/clash-lang/clash-compiler/blob/df6369060086e126d8a4cc5e6f77622aa190f71c/clash-lib/src/Clash/Core/Type.hs#L228-L236

I realize we didn't do that in the old code. So another option is for `maybeReturnVec` not to over-approximate for type families, and just under-approximate.

martijnbastiaan: > I realize we didn't do that in the old code. So another option is for `maybeReturnVec` not to over-approximate for type families, and just under-approximate.

Nah, that sounds confusing to me. I'll add a `coreView` call and make it a separate commit.

martijnbastiaan: Done.

martijnbastiaan:
@mergify rebase

PR #3340, opened 2026-07-30
Memoize GHC-to-Clash type conversion on the GHC type

martijnbastiaan:
Cuts the loading stage from 14.1s to 6.9s on `bittide-hardware`'s `wireDemoTest`.

It's slightly scrappy in that it ignores coercions, but it's better than getting it wrong I suppose.

## Still TODO:

### DIFF
--- a/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
@@ -35,7 +35,7 @@ module Clash.GHC.GHC2Core
 where
 
 -- External Modules
-import           Control.Lens                ((^.), (%~), (&), (%=), (.~), view, makeLenses)
+import           Control.Lens                ((^.), (%~), (&), (%=), (.~), use, view, makeLenses)
 import           Control.Applicative         ((<|>))
 import           Control.Monad.Extra         (ifM, andM)
 import           Control.Monad.RWS.Strict    (RWS)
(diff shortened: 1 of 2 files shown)


clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -903,11 +969,27 @@ annotateType ty cty = do
 
 -- | Converts GHC Type to a Clash Type. Strips newtypes and signals, with the
 -- exception of newtypes used as annotations (see: SynthesisAttributes).
+--
+-- Conversions are memoized on the GHC type, which also shares the converted
+-- types. The conversion runs 'withoutSrcSpan' so that it depends on the GHC
+-- type alone. Under '_typeSrcSpans' it keeps the source spans and skips the
+-- cache instead.
```
christiaanb: I don't think we do anything with the source spans for type declarations, do we?

I don't see anywhere in `HWType` where we would store this info:
https://github.com/clash-lang/clash-compiler/blob/df6369060086e126d8a4cc5e6f77622aa190f71c/clash-lib/src/Clash/Netlist/Types.hs#L427-L486

i.e. I think we can always cache, regardless of the debug level.

martijnbastiaan: Yeah I didn't see anything either, but I was thinking: _if_ we ever try to use them, it might be discovered way too late that some of the `SrcSpan`s are wrong (due to caching).

Now that I write that out though, I guess that's just an argument for always suppressing them -- debugging mode or not.

martijnbastiaan: Done :)

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -636,6 +688,12 @@ addUsefullR x m =
   then RWS.local (srcSpan .~ x) m
   else m
 
+-- | Convert without an ambient source span: names GHC gives no location of
+-- their own get 'noSrcSpan' instead of the location of the binder under
+-- conversion.
+withoutSrcSpan :: C2C a -> C2C a
+withoutSrcSpan = RWS.local (srcSpan .~ noSrcSpan)
+
```
christiaanb, line 696: Why not just take this out completely?

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
```diff
@@ -903,11 +961,25 @@ annotateType ty cty = do
 
 -- | Converts GHC Type to a Clash Type. Strips newtypes and signals, with the
 -- exception of newtypes used as annotations (see: SynthesisAttributes).
+--
+-- Conversions are memoized on the GHC type, which also shares the converted
+-- types. The conversion runs 'withoutSrcSpan' so that it depends on the GHC
+-- type alone; nothing ever reads back the source spans of names inside a
+-- type, so no information is lost.
```
christiaanb, line 968: Why not take this out completely?

christiaanb (commented):
Why not take all this srcSpan related stuff? Given that it doesn't affect GHC Type to Core Type conversion.

clash-ghc/src-ghc/Clash/GHC/GHC2Core.hs
martijnbastiaan: Because type variables and type constructors (indirectly) carry a `Name` that carries a `SrcSpan`. If you'd take it out, `nameLoc` ends up pointing to the first time you cached it for all. I suppose that's fine for type constructors, but not for type variables.

So IMO, rather than having a sometimes confusing provenance, I'd rather have no provenance at all (and we don't use it at the moment anyway).

christiaanb (approved):

PR #3341, opened 2026-07-30
Add and use `unsafeSubstTm` in `appProp` beta reduction

martijnbastiaan:
`appProp` used `substTm` on already deshadowed terms, leading to unneeded work. This patch introduces `unsafeSubstTm` for exactly those situations.

Also see discussion in #3333.

## Still TODO:

### DIFF
--- a/clash-lib/src/Clash/Core/Subst.hs
+++ b/clash-lib/src/Clash/Core/Subst.hs
@@ -47,6 +47,7 @@ module Clash.Core.Subst
     -- ** Applying substitutions
   , substTm
   , maybeSubstTm
+  , unsafeSubstTm
   , substAlt
   , substId
     -- * Variable renaming
@@ -79,6 +80,7 @@ import           Data.Text.Prettyprint.Doc
 import           Data.Hashable             (Hashable (hashWithSalt))
 import qualified Data.List                 as List
 import qualified Data.List.Extra           as List
+import           Data.Maybe                (fromMaybe)
 import           Data.Ord                  (comparing)
 import           GHC.Stack                 (HasCallStack)
 import           GHC.SrcLoc.Extra          () -- Hashable RealSrcSpan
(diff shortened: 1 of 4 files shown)


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -587,6 +589,68 @@ substTm doc subst = go where
   goTick t@NoDeDup      = t
   goTick (Attributes ty tm) = Attributes (substTy subst ty) (go tm)
 
+-- | Like 'substTm', but doesn't account for shadowing.
```
christiaanb: It's not just about not accounting for shadowing, it's also about not avoiding capture of free variables in the replacement terms; it's that second part that makes it unsafe!

christiaanb (approved):
Looks good, but `unsafeSubstTm` should come with a bigger warning about why it is unsafe in relation to free variable capture in the replacement term!

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Done! Could you check?

clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -595,6 +589,90 @@ substTm doc subst = go where
...
+-- yield @(5, \x -> x)@.
+--
+-- An example of free variable capture: in @\x -> y@ and a substitution
+-- @y |-> f x@, 'unsafeSubstTm' will yield @\x -> f x@, whereas a safe substitution
+-- would yield @\x' -> f x@.
+--
+-- You should therefore only use this function if:
+--
+--   1. No binder in the term shadows a variable in the domain of either
+--      substitution, i.e., every occurrence of a variable in the domain is a
```
christiaanb: Actually, this doesn't matter for `globals`. Global variable references cannot be shadowed by local binders. So rule 1 and 2 only hold for `locals`. It's actually always safe to use `unsafeSubstTm` for globals because:

1. Binders in the "term to substitute in" cannot shadow global variable references.
2. The `globals` should only contain closed expression, i.e. no free variables.

On that last point, that aspect should actually be documented! That the replacement terms in `globals` must be closed.

christiaanb (commented):
The docs still need some minor adjustments.

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: Ah yeah, good point. Will do and then merge.

PR #3343, opened 2026-08-02
Read family instance interfaces for precompiled modules

### LINKED ISSUE #1534: Cannot reduce to an integer: closed type family
```haskell

module Issue where

import Clash.Prelude

type family Width a where
  Width a = BitSize a

topEntity :: BitVector (Width Bool) -> ()
topEntity _ = ()
```

gives:

```
    Clash.Netlist.Util(910): Clash.Core.Util(322): Cannot reduce to an integer:
    Clash.Class.BitPack.BitSize[8214565720323786306]
```

however:

```haskell
*Issue> :kind! (BitVector (Width Bool))
(BitVector (Width Bool)) :: Type
= BitVector 1
```

Apparently this violates one of Clash's invariants. When using:

```haskell
topEntity :: BitVector (Width Bool)
topEntity = 0
```

It gives me a proper (but still wrong?) error message:

```haskell
    Clash.Normalize(223): This bndr has a non-representable return type and can't be normalized:
    Issue.topEntity :: Clash.Sized.Internal.BitVector.BitVector
                         (Issue.Width GHC.Types.Bool)
```

martijnbastiaan:
Fixes #1534

## Still TODO:


clash-ghc/src-ghc/Clash/GHC/LoadModules.hs
```diff
@@ -215,6 +217,26 @@ loadExternalModule hdl modName0 mainIsM = MC.try $ do
   -- 'loadModules' is unaffected by the pruning above.
   return (rootIds, FamInstEnv.emptyFamInstEnv, modName1, loaded, allBinders)
```
christiaanb, line 218: Why not return something else than a `FamInstEnv.emptyFamInstEnv` here instead of the `loadFamilyInstanceModules foundMod` action?

martijnbastiaan: I think this is mostly here to make it have the same type signature as `loadLocalModule`, which makes it a bit easier to call. The latter will still need a call to `tcGetFamInstEnvs` after. So either:

* We do it like it is now (call it in one place)
* We make `loadLocalModule` also call `tcGetFamInstEnvs`

christiaanb (approved):

PR #3345, opened 2026-08-05
pack/unpack no longer instantiate any HDL logic. 

### LINKED ISSUE #3146: `unpack` defaults to last constructor for larger values
`unpack` currently defaults to the last constructor for any value beyond the last value. This can be a bit counterintuitive, and restricts optimizations.

Example:
```hs
ghci> data X = A | B | C deriving (Show,Generic,BitPack)
ghci> unpack 0 :: X
A
ghci> unpack 1 :: X
B
ghci> unpack 2 :: X
C
ghci> unpack 3 :: X
C
```

Stemming from an implementation that looks like:
```hs
\case
  0 -> A
  1 -> B
  _ -> C
```

Proposal: change this to something like:
```hs
\case
  0 -> A
  1 -> B
  2 -> C
  _ -> errorX "Undefined constructor index"
```

rowanG077:
Based of https://github.com/clash-lang/clash-compiler/pull/3177

This PR essentially make `pack` and `unpack` id in HDL. From experiments synthesizer were not able to eliminate the logic for `pack` and `unpack` which is a footgun in scenarios where you really want it to be logic less. This is not a free change since now we need to enforce that user written `BitPack` instances are illegal, unless they match bit packing of the compiler one to one.

TODOs:

- [ ] Make documentation even more explicit that `BitPack` is not to be manually implemented.
- [ ] Investigate if it's possible to make user written instances impossible. Or at the very least very clear that it's highly unsafe to write custom instances.

This fixes https://github.com/clash-lang/clash-compiler/issues/3146

### DIFF
--- a/changelog/2026-07-31T14_00_00+02_00_generic_unpack_undefined
+++ b/changelog/2026-07-31T14_00_00+02_00_generic_unpack_undefined
@@ -1,0 +1,4 @@
+CHANGED: The `Generic` default implementation of `unpack` now returns an
+undefined value for bit patterns that do not encode a value of the target
+type, instead of silently decoding them to one of the constructors. See
+[#3146](https://github.com/clash-lang/clash-compiler/issues/3146).
--- a/clash-prelude/src/Clash/Class/BitPack/Internal.hs
+++ b/clash-prelude/src/Clash/Class/BitPack/Internal.hs
@@ -54,6 +54,7 @@ import Clash.Promoted.Nat             (SNat(..), snatToNum)
 import Clash.Sized.Internal.BitVector
   (pack#, split#, checkUnpackUndef, undefined#, unpack#, unsafeToNatural, isLike#,
    BitVector, Bit, (++#), xToBV)
+import Clash.XException              (errorX)
 
 {- $setup
 >>> :m -Prelude
@@ -142,9 +143,11 @@ class KnownNat (BitSize a) => BitPack a where
 
   -- | Convert a 'BitVector' to an element of type @a@
   --
-  -- When the bit pattern is not the representation of any value of type @a@,
-  -- @unpack@ can return an invalid value, or some value @y@ where @'pack' y@
-  -- is not equal to the input. Use 'maybeUnpack' to detect such bit patterns.
+  -- If the bit pattern is not the representation of any value of type @a@,
+  -- i.e. there is no @x@ such that @bv == 'pack' x@, then @unpack bv@ is
+  -- undefined: an unused constructor encoding returns a fully undefined
+  -- value, while an invalid field encoding returns a value whose affected
+  -- fields are undefined. Use 'maybeUnpack' to detect such bit patterns.
   --
   -- >>> pack (-5 :: Signed 6)
   -- 0b11_1011
@@ -165,7 +168,7 @@ class KnownNat (BitSize a) => BitPack a where
        )
     => BitVector (BitSize a) -> a
   unpack b =
-    to (gUnpack sc 0 bFields)
+    to (snd (gUnpack True sc 0 bFields))
    where
     (checkUnpackUndef unpack . resize -> sc, bFields) = split# b
 
@@ -174,8 +177,8 @@ class KnownNat (BitSize a) => BitPack a where
   --
   -- Not every bit pattern is necessarily the representation of a value of type
   -- @a@. If there is no @x@ such that @bv == 'pack' x@, then @maybeUnpack bv@
-  -- returns @Nothing@. Conversely, 'unpack' @bv@ would return an invalid value,
-  -- or some value @y@ where @'pack' y@ is not equal to @bv@.
+  -- returns @Nothing@. Conversely, 'unpack' @bv@ would return an undefined
+  -- value.
   --
   -- >>> pack (maxBound :: Index 13)
   -- 0b1100
@@ -218,7 +221,9 @@ class KnownNat (BitSize a) => BitPack a where
        , (constrSize + fieldSize) ~ BitSize a
        )
     => BitVector (BitSize a) -> Maybe a
-  maybeUnpack b = to <$> gMaybeUnpack True sc 0 bFields
+  maybeUnpack b =
+    let (valid, x) = gUnpack True sc 0 bFields in
+    if valid then Just (to x) else Nothing
    where
     (checkUnpackUndef unpack . resize -> sc, bFields) = split# b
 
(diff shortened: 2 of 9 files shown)


martijnbastiaan:
>  Investigate if it's possible to make user written instances impossible. Or at the very least very clear that it's highly unsafe to write custom instances.

How about: rename `pack`/`unpack` -> `pack#`/`unpack#`. Add `wasDerived = False`. Add `pack = if wasDerived then clashCompilerError ">:(" else pack#`.

martijnbastiaan:
@mergify rebase

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -230,6 +246,42 @@ packXWith
 packXWith f = xToBV . f
 {-# INLINE packXWith #-}
 
+-- | In Haskell, apply the first argument to the second argument. In HDL,
+-- convert the second argument to the result type directly instead -- a no-op
+-- in hardware. This is sound because a lawful 'BitPack' instance accurately
```
martijnbastiaan, line 251: I don't think this is _quite_ right. VHDL doesn't specify a bit representation for its enums, so synthesizers try to pick "cleverly". Maybe this is a case for dropping rendering enums, but it is something to be aware of. A second case where this might diverge is in context of custom bit representations (I think).

I'm also surprised to find this in `Clash.Class.BitPack.*`. I think the documentation doesn't properly address this.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -596,17 +641,16 @@ class GBitPack f where
     -- ^ Current constructor
     -> BitVector (GFieldSize f)
     -- ^ BitVector containing fields
-    -> Maybe (f a)
-    -- ^ Possibly unpacked result
+    -> (Bool, f a)
+    -- ^ (Bit pattern encodes a value, unpacked result)
```
martijnbastiaan, line 645:
```suggestion
    -- ^ (Unpack successful, unpacked result)
```

I can't really say why, but "bit pattern encodes a value" reads awkward to me.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
```diff
@@ -666,33 +698,32 @@ instance (KnownNat (GFieldSize g), KnownNat (GFieldSize f), GBitPack f, GBitPack
...
   type GConstructorCount (K1 i c)  = 1
 
   gPackFields cc (K1 i) = (cc, pack i)
-  gUnpack _c _cc b      = K1 (unpack b)
-  gMaybeUnpack _checkBounds _c _cc b = K1 <$> maybeUnpack b
+  gUnpack _checkBounds _c _cc b =
+    case maybeUnpack b of
+      Just x -> (True, K1 x)
+      Nothing ->
+        (False, K1 (errorX "unpack: bit pattern does not encode a value of the target type"))
```
martijnbastiaan, line 719:
```suggestion
        (False, K1 (errorX "unpack: failed: {b}"))
```

Maybe this is a little bit too short. In any case, definitely mention the bitvector.

changelog/2026-07-31T14_00_00+02_00_generic_unpack_undefined
```diff
@@ -0,0 +1,4 @@
+CHANGED: The `Generic` default implementation of `unpack` now returns an
+undefined value for bit patterns that do not encode a value of the target
+type, instead of silently decoding them to one of the constructors. See
+[#3146](https://github.com/clash-lang/clash-compiler/issues/3146).
```
martijnbastiaan, line 4: I think this changelog is  superseded by the other one, or am I missing something?

martijnbastiaan (changes requested):
I think your references to "undefined" should be more specific: `XException`. You can link to the modules documentation.

I'm also finding the nomenclature "bit pattern" a bit unusual. Maybe that's just me, but I find "bits" or "bit vector" much more neutral.

changelog/2026-07-31T14_00_00+02_00_generic_unpack_undefined
rowanG077: Yes this is just fallout from before I did not yet fix.

clash-prelude/src/Clash/Class/BitPack/Internal.hs
rowanG077: For enums you are totally correct. 

> I'm also surprised to find this in Clash.Class.BitPack.*. I think the documentation doesn't properly address this.

What do you mean by this? This is an internal function not exposed to users. What should the documentation say?

martijnbastiaan: Oh. I was under the impression that this was a public function. 

The reason the location surprises me is because on the surface it does not seem obvious why this is related to bit packing, given the type signature. Given the implementation (`~TOBV`, `~FROMBV`), it makes more sense.

DigitalBrains1: This function used to be in `Clash.Annotations.BitRepresentation.Deriving` which was the only location it was used. It was documented as
```haskell
-- | In Haskell apply the first argument to the second argument,
--   in HDL just return the second argument.
--
-- This is used in the generated pack/unpack to not do anything in HDL.
```

It used to have a black box just saying `~ARG[1]`, not the new pull-through-BV black box.

One thing that seems odd on first glance, but I suspect there's a good reason, is that the _used arguments_ of the black box is defined as `[0, 1]`, but why is 0 used? https://github.com/clash-lang/clash-compiler/blob/a06d3452bd8317ab3d01b75b13c59090f9d59fc1/clash-lib/src/Clash/Normalize/Transformations/Letrec.hs#L141-L143

Its source is https://github.com/clash-lang/clash-compiler/commit/a098b966d1ae3ef9359bf27b45fa816b836e8d9c , which doesn't explicitly spell out why `f` is needed at all.

But I suppose if you're trying to do constant folding in Clash Core, you do actually need the `f` (and you might end up not using the black box since you just constant-folded it out of the way).

rowanG077: I think a clean design would be: If a type has a `BitPack` instance -> no enum.

It's probably not straightforward to determine that during loading.

martijnbastiaan:
This will break https://github.com/clash-lang/clash-cores/blob/5d084eff6cdc630ecd14be53bdc4fa30a5a4fd95/src/Clash/Cores/LatticeSemi/Ice40/IO.hs#L96-L132. Not saying I'm against it, but just a heads-up.

rowanG077:
@martijnbastiaan Mind if I hijack this PR: https://github.com/clash-lang/clash-cores/pull/66 and use bitpack annotation to properly derive a bitpack? That will solve the `maybeUnpack` case that's currently in master and also this PR.

martijnbastiaan:
I don't even remember making that PR, hijack away, the pilot was long gone.

PR #3346, opened 2026-08-05
Set default timeout for `clash-testsuite` to 5min

martijnbastiaan:
Slowest test on my 6-year-old CPU takes 90 seconds, so a 5 minute timeout seems reasonable.

While hacking on the compiler I've had a few times where tests would get stuck and I would only realize after waaay to long..

## Still TODO:


DigitalBrains1 (approved):
I suggest backporting

Issue #2193, opened 2022-05-08
VHDL's `bool` hides X-propagation


martijnbastiaan:
Consider:

```haskell
bool :: Bool
bool = errorX "help"
```

This produces:

```vhdl
architecture structural of bool is

begin
  result <= true;
end;
```

Even though we don't assign any value to it in the Clash design (well, bottom), the HDL ends up getting a definitive value. As a consequence, it is indistinguishable from `bool = True`.

I therefore believe we should use `std_logic` for `Bool` in HDL. This would give us:

```vhdl
architecture structural of bool is

begin
  result <= '-';
end;
```

which would be much closer to the designer's intent.

christiaanb:
You still get bad `X` propagation regardless of this change. Currently we would translate:
```
if e then p else q
```
to:
```
with e select
  q when False,
  p when others;
```
With the proposed change, we would instead translate it to:
```
with e select
  q when '0'
  p when others;
```
i.e. we would still select `p` when `e` is `X`.

martijnbastiaan:
Yes, that's definitely related. See the other issue I wrote up on X-propagation: https://github.com/clash-lang/clash-compiler/issues/2194.

christiaanb:
But this is inherent to synthesisable VHDL: unless we match on `x` we can't propagate it through a choice construct.

martijnbastiaan:
I'd be in favor of using `X` over `-` for choice constructs. This is a conservative approach, but miles better than what we do now. And it wouldn't involve matching on `X`. Please see the other issue and the paper attached.

Issue #431, opened 2018-11-25
Generate sum types and use user-supplied record field names in VHDL


hesiod:
VHDL supports sum (as enums) and product (record) types, e.g.
```vhdl
type example_enum_type is (value, another_value);
```
and
```vhdl
type example_record is record
   member: std_logic;
   another_member: std_logic;
end record;
```
Most VHDL simulators/waveform viewers display enum values, aiding in debugging. If Clash could map Haskell sum types to enums, the generated VHDL would be more readable and viewing waveforms of Clash-generated VHDL would be more pleasant.
Clash already generates records for product types, however, the individual field names have generic names (`<name>_sel0`, `<name>_sel1` etc). If Clash could use field names of Haskell record types to give VHDL record types more appropriate names, the resulting VHDL would be even more readable.

I'm not quite sure how involved implementing these features would be or whether it is at all possible to realize them. If I understand the VHDL backend correctly, sum types are currently [normalized to bit vectors](https://github.com/clash-lang/clash-compiler/blob/67d36495cc108afad767e92d1ee21e12411cc9e2/clash-lib/src/Clash/Backend/VHDL.hs#L318-L337).
It should be possible to remove this normalization and instead map sum types to VHDL enum types, changing code generation in other place where appropriate, but I don't know enough about the Backend to tell whether that's enough.
Regarding field names, Clash's [`DataCon` type](https://hackage.haskell.org/package/clash-lib-0.99.3/docs/Clash-Core-DataCon.html#t:DataCon) as well as [the `Product` constructor of `HWType`](https://hackage.haskell.org/package/clash-lib-0.99.3/docs/Clash-Netlist-Types.html#t:HWType) does not contain the field names (in GHC, the field labels are [saved in `DataCon`s](https://downloads.haskell.org/~ghc/8.6.2/docs/html/libraries/ghc-8.6.2/DataCon.html#v:dataConFieldLabels)), so right now the backend doesn't know about them. Afterwards, changing the generated field names seems to be quite doable (e.g. [here](https://github.com/clash-lang/clash-compiler/blob/67d36495cc108afad767e92d1ee21e12411cc9e2/clash-lib/src/Clash/Backend/VHDL.hs#L367), [here](https://github.com/clash-lang/clash-compiler/blob/67d36495cc108afad767e92d1ee21e12411cc9e2/clash-lib/src/Clash/Backend/VHDL.hs#L1098-L1099) and in some other places).

martijnbastiaan:
[...]

PR #3355, opened 2026-08-11
Remove enum generation in VHDL backend

rowanG077:
This PR is primarily a proposal, and I’d like to hear what the rest of the Clash developers think.

VHDL enums have a few user-facing problems, most notably X-hiding, see https://github.com/clash-lang/clash-compiler/issues/2193, and unstable bit representations. The latter also means they don’t behave lawfully with Clash’s `BitPack` type class. 

I don’t think there is a compelling reason for us to generate enums. They do make the generated code slightly nicer, and the original feature request was motivated by wanting VHDL simulations to display nicer values (https://github.com/clash-lang/clash-compiler/issues/431). However, I think we’ll have a much better story for this with clash-shockwaves once it supports native simulation.

Given the issues with enums, I propose removing VHDL enum generation.

In a follow up PR I aim to address the x-hiding through comparison expression as comparison or equality expression results in a VHDL boolean which with this PR is still x-hiding. 

### DIFF
--- a/tests/shouldwork/Issues/T431.hs
+++ b/tests/shouldwork/Issues/T431.hs
@@ -30,16 +30,21 @@ assertNotIn needle haystack
                                                     , "\n\nIn:\n\n", haystack ]
   | otherwise = pure ()
 
+-- Enums were requested in #431 but have since been removed again. Guard that
+-- sum types stay rendered as bit vectors.
 mainVHDL :: IO ()
 mainVHDL = do
   [topDir] <- getArgs
   content <- readFile (topDir </> show 'topEntity </> "topEntity.vhdl")
 
-  -- no bit vector literals appear, so no double quotes appear
-  assertNotIn "\"" content
+  -- no enum variants appear in the design
+  assertNotIn "TrafficLight'(Red)" content
+  assertNotIn "TrafficLight'(RedAmber)" content
+  assertNotIn "TrafficLight'(Amber)" content
+  assertNotIn "TrafficLight'(Green)" content
 
-  -- enum variants appear in the design
-  assertIn "TrafficLight'(Red)" content
-  assertIn "TrafficLight'(RedAmber)" content
-  assertIn "TrafficLight'(Amber)" content
-  assertIn "TrafficLight'(Green)" content
+  -- the constructors are encoded as bit vector literals instead
+  assertIn "\"01\" when \"00\"" content
+  assertIn "\"11\" when \"01\"" content
+  assertIn "\"00\" when \"10\"" content
+  assertIn "\"10\" when others" content
--- a/changelog/2026-08-06T10_00_00+02_00_remove_vhdl_enums
+++ b/changelog/2026-08-06T10_00_00+02_00_remove_vhdl_enums
@@ -1,0 +1,10 @@
+CHANGED: Sum types and `Bool` are now rendered as `std_logic_vector` and `std_logic` in VHDL instead of enums.
+
+This includes deprecating the "-fclash-no-render-enums" flag. The flag still is accepted but generates a warning. The flag will be fully removed in a future release.
+
+VHDL Enums have a few problems:
+
+- Unstable bit representations which makes them incompatible with the Clash `BitPack` type class.
+- X-hiding as they cannot take on metavalues.
+
+Partially fixes [#2193](https://github.com/clash-lang/clash-compiler/issues/2193).
--- a/clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
+++ b/clash-ghc/src-ghc/Clash/GHC/ClashFlags.hs
@@ -1,7 +1,7 @@
 {-|
   Copyright   :  (C) 2015-2016, University of Twente,
                      2016-2017, Myrtle Software Ltd,
-                     2021,      QBayLogic B.V.,
+                     2021-2026, QBayLogic B.V.,
                      2022,      Google Inc.,
   License     :  BSD2 (see the file LICENSE)
   Maintainer  :  QBayLogic B.V. <devops@qbaylogic.com>
@@ -83,7 +83,7 @@ flagsClash r = [
   , defFlag "fclash-aggressive-x-optimization-blackboxes" $ NoArg (liftEwM (setAggressiveXOptBB r))
   , defFlag "fclash-inline-workfree-limit"       $ IntSuffix (liftEwM . setInlineWFLimit r)
   , defFlag "fclash-edalize"                     $ NoArg (liftEwM (setEdalize r))
-  , defFlag "fclash-no-render-enums"             $ NoArg (liftEwM (setNoRenderEnums r))
+  , defFlag "fclash-no-render-enums"             $ NoArg (deprecatedNoOp "no-render-enums" "sum types are always rendered as bitvectors")
   , defFlag "fclash-timescale-precision"         $ SepArg (setTimescalePrecision r)
   , defFlag "fclash-ignore-broken-ghcs"          $ NoArg (liftEwM (setIgnoreBrokenGhcs r))
   , defFlag "fclash-no-concurrent-topentity-compilation" $ NoArg (liftEwM (setNoConcurrentTopEntities r))
@@ -105,6 +105,20 @@ deprecated wrong right f a = do
                              ++ right
                              ++ "' instead.")
   liftEwM (f a)
+
+-- | Print deprecation warning for a flag that no longer has any effect
+deprecatedNoOp
+  :: String
+  -- ^ Deprecated flag
+  -> String
+  -- ^ Why the flag is a no-op now
+  -> EwM IO ()
+deprecatedNoOp flag reason =
+  addWarn ("Using '-fclash-" ++ flag
+                             ++ "' is deprecated, it no longer has any effect: "
+                             ++ reason
+                             ++ ". It is accepted for backwards compatibility \
+                                \and will be removed in a future release.")
 
 setInlineLimit :: IORef ClashOpts
                -> Int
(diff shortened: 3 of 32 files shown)


martijnbastiaan:
A couple of things:

* Have you considered disabling the flag by default, instead of completely removing it? It could even be called `-fclash-unsafe-render-enums`, where we detail the downsides (potential introduction of logic). This would still leave the option open for hairy HDL-only debug runs.
* I support the general direction, we shouldn't hide x-propagation as much as possible (though I believe in some cases it is impossible, I forgot the details). Related material we might want to review:
  * [The Dangers of Living with an X](https://support.arm.com/documentation/arp0009/a/)
  * [I'm still in love with my X!](https://dvcon-proceedings.org/document/im-still-in-love-with-my-x-but-do-i-want-my-x-to-be-an-optimist-a-pessimist-or-eliminated/)
* I believe the original motivation for enums in VHDL was just source readability, not necessarily waveform output. Have you considered removing enums in favor of just constants? I.e., instead of generating:

```vhdl
"01" when "00"
```

You'd get something more similar to:


```vhdl
TrafficLight_Types.RED when TrafficLight_Types.Green
```

Edit:

> (though I believe in some cases it is impossible, I forgot the details

I just remembered this is at least partly about things like `rising_edge` and the like.

leonschoorl:
> > (though I believe in some cases it is impossible, I forgot the details
> 
> I just remembered this is at least partly about things like `rising_edge` and the like.

Isn't is the case that anything that uses the BOOLEAN type in VHDL (`rising_edge` being on of them) can't represent X's?

DigitalBrains1:
However, that means that if you wish to eliminate all X hiding from the VHDL you generate, you will need to write alternatives for functions such as `rising_edge`. And the question arises :-) whether your alternative will still synthesise properly in that case.

For `rising_edge` specifically that might not be an issue as possibly all our uses of that function trigger on a clock signal and we don't ever have an X in the clock in Haskell simulation either.

rowanG077:
> Have you considered disabling the flag by default, instead of completely removing it? It could even be called -fclash-unsafe-render-enums, where we detail the downsides (potential introduction of logic). This would still leave the option open for hairy HDL-only debug runs.

The main reason I decided it's not worth it is that I really don't want haskell `Bool` to translate to either VHDL `std_logic` or `boolean` depending on a flag. Because this has far reaching consequences for blackboxes. If we restrict it to user types I'm more on board with it.

> Isn't is the case that anything that uses the BOOLEAN type in VHDL (rising_edge being on of them) can't represent X's?

`rising_edge` should not be a problem. For the conventional operators like equality and ordering I aim to write wrappers that do proper x-propagation in a follow up PR.

DigitalBrains1:
> However, I think we’ll have a much better story for this with clash-shockwaves once it supports native simulation.

Given that we've been generating enums for a while already, wouldn't the more logical moment to remove them then be when clash-shockwaves does in fact support native simulation?

Or are you worried that things have gotten worse with the `BitPack` changes?

If I understand correctly, before #3345, `BitPack` for VHDL enums gave the same results for both Clash simulation and HDL, but it potentially did not correspond to the bit encoding of the VHDL enum itself in hardware.

Also, if an enum was used in a custom `BitRepresentation`, then `BitPack` in HDL would differ between Clash simulation and HDL simulation.

But after #3345, Clash simulation and HDL potentially show different `BitPack`s, and the HDL `BitPack` will correspond to the bit encoding of the VHDL enum itself.

So you've moved the issue; there was an issue before and there will be an issue after, it's just a different one.

DigitalBrains1:
> The main reason I decided it's not worth it is that I really don't want haskell `Bool` to translate to either VHDL `std_logic` or `boolean` depending on a flag.

Have you considered making `-fclash-no-render-enums` the default and potentially renaming the flag `-fclash-unsafe-render-enums`?

[edit]
I merely rephrased Martijns question because I thought Rowan misinterpreted the question. Turns out the confusion was on a different aspect. Rowan interpreted it as _rename the flag and also make it apply to `Bool`_, as this PR aims for parity between `Bool` and user-specified enums, whereas I meant it as _`Bool` keeps being unaffected by the flag as before_.
[/edit]

rowanG077:
> Given that we've been generating enums for a while already, wouldn't the more logical moment to remove them then be when clash-shockwaves does in fact support native simulation?

I think the other problems with enum warrant its removal anyway. The original user that reported that they wanted enum wanted it for simulation clarity. And we have a better story there: clash-shockwaves. I'm not too worried that we leave a gap here in the short term. 

> But after https://github.com/clash-lang/clash-compiler/pull/3345, Clash simulation and HDL potentially show different BitPacks, and the HDL BitPack will correspond to the bit encoding of the VHDL enum itself.
> 
> So you've moved the issue; there was an issue before and there will be an issue after, it's just a different one.

That's correct. But consider that https://github.com/clash-lang/clash-compiler/pull/3345 does not aim to fix either issue. The real fix is to not generate `Enum` which this PR is for. 

If this PR is merged and https://github.com/clash-lang/clash-compiler/pull/3345 is merged the actual underlying problem is fixed.

> Have you considered making -fclash-no-render-enums the default and potentially renaming the flag -fclash-unsafe-render-enums?

Currently Haskell `Bool` always render as VHDL `boolean`. And your question is basically the same as @martijnbastiaan question?

leonschoorl:
I wonder if anyone cares about VHDL booleans on the the (top)entity boundaries.

I could imagine a situation where parts of a design are implemented in clash and parts in VHDL, and if that interface between the two is specified to use booleans, then it would be useful if there was still a way to get clash the generate VHDL code that still produces or consumes VHDL BOOLEAN signals.

But maybe this is just not a issue, and everybody is knowledgeable enough to use std_logic everywhere.
A quick look through at primitives in vivado seems those don't use any booleans, except for some generic parameters/attributes.

tests/shouldwork/Issues/T431.hs
```diff
@@ -30,16 +30,21 @@ assertNotIn needle haystack
...
+  assertNotIn "TrafficLight'(Red)" content
+  assertNotIn "TrafficLight'(RedAmber)" content
+  assertNotIn "TrafficLight'(Amber)" content
+  assertNotIn "TrafficLight'(Green)" content
+
+  -- the constructors are encoded as bit vector literals instead
+  assertIn "\"01\" when \"00\"" content
+  assertIn "\"11\" when \"01\"" content
+  assertIn "\"00\" when \"10\"" content
+  assertIn "\"10\" when others" content
```
martijnbastiaan, line 50: I would strongly prefer it if the case statements do end up mentioning constants -- or have a comment attached to them referring to the original constructor names.

martijnbastiaan (approved):
I do think we have to bite the bullet in order to get us closer to good (though probably never perfect) X propagation, so LGTM.

I'm unsure about the deprecation strategy. I can see that maintaining both paths is pretty much undoable; if anyone comes up with anything better I'm all ears. I've left one comment.

JvWesterveld:
I'm a little bit concerned about the reduced readability of outputted VHDL when not using enums given that this already isn't great. Having comments with the original constructor names added may help when looking at the VHDL file manually, but probably not when working with certain vendor tools, such as an ILA.

Moreover, for handwritten VHDL, I believe I've only ever seen enums used as the type for the state of FSMs (well, likely as element of a record type). I imagine always using `std_(u)logic_vector` instead would curb the synthesizer's ability to pick the encoding that it finds most appropriate, with all that entails, though I haven't tested this yet. On the other hand, having the ability to assign don't care bits to the state might let the synthesizer make minor optimizations at times.

I'm in doubt about whether improved don't-care/x-propagation in output VHDL simulation is worth this trade-off. Out of curiosity, would it be feasible to refrain from using enums only if the type in question has a `BitPack` instance? If so, perhaps that would be the best of both worlds.

> I wonder if anyone cares about VHDL booleans on the the (top)entity boundaries.
>
> I could imagine a situation where parts of a design are implemented in clash and parts in VHDL, and if that interface between the two is specified to use booleans, then it would be useful if there was still a way to get clash the generate VHDL code that still produces or consumes VHDL BOOLEAN signals.

I think it's fine or even desirable to use `std_(u)logic` over boolean at Clash's top entity boundaries. If compiling as VHDL 2008 or higher, which one can optimistically hope all tools would support in the year 2026, `std_(u)logic` can be used as the condition for an if statement, and the [`??`](https://gitlab.com/IEEE-P1076/packages/-/blob/586ebeb9c3fcefd1ac9a07ce749d0e01e678503e/ieee/std_logic_1164-body.vhdl#L974-977) operator can be used to concisely convert to boolean explicitly.

rowanG077:
@JvWesterveld 

In principle I think it's possible to detect whether a type has a `BitPack` instance and make the decision based on that. That was my initial thought as well. Until I started thinking more deeply about x-prop. 

Another drawback is that I think it's quite weird to have such differences in generated code just when a user adds a new instance. `BitPack` is something that I have used to debug things, in that case it's probably undesirable for the code to change so much just because there is an instance. 

>  I imagine always using std_(u)logic_vector instead would curb the synthesizer's ability to pick the encoding that it finds most appropriate, with all that entails, though I haven't tested this yet. 

In my experience synthesizers freely pick the encoding they want anyway, if they can prove they can. In most cases you might be thinking of state machines that use an enum as the state tag. Then synthesizer is now free to choose for example a one-hot encoding. But that optimization is also happening in verilog where an explicit bit encoding must be given.

Granted you probably make the life of synthesizers easier by it being an enum.

tests/shouldwork/Issues/T431.hs
leonschoorl: We could have clash generate some constants in the .._types.vhdl file:
```
  subtype TrafficLight is std_logic_vector(1 downto 0);  -- this line it generates already
  constant TrafficLight_Red : TrafficLight := "00";
  constant TrafficLight_RedAmber : TrafficLight := "01";
  constant TrafficLight_Amber : TrafficLight := "10";
  constant TrafficLight_Green : TrafficLight := "11";
```

And then the generated case statement could like like:
```
  with (ds) select
    result <= TrafficLight_RedAmber when TrafficLight_Red,
              TrafficLight_Green when TrafficLight_RedAmber,
              TrafficLight_Red when TrafficLight_Amber,
              TrafficLight_Amber when others;
```

martijnbastiaan: Yes I'd like that!

PR #3356, opened 2026-08-12
Perform cheap checks first during normalization (+addition performance debug flag)

martijnbastiaan:
Various improvements to quickly bail doing work wherever we can.

## Still TODO:

  - [x] Review @martijnbastiaan 
  - [x] Wait for benchmark results: https://clash-lang.github.io/clash-benchmarks/?machine=oele&branch=clash-lang%2Fclash-compiler%40martijn%2Fperf%2Fmore-small-improvements

### DIFF
--- a/clash-lib/src/Clash/Rewrite/Util.hs
+++ b/clash-lib/src/Clash/Rewrite/Util.hs
@@ -26,7 +26,7 @@ module Clash.Rewrite.Util
 
 import           Control.DeepSeq
 import           Control.Exception           (throw)
-import           Control.Lens ((%=), (+=), (^.))
+import           Control.Lens                ((%=), (+=), (^.), Lens')
 import qualified Control.Lens                as Lens
 import qualified Control.Monad               as Monad
 import           Control.Monad.IO.Class      (liftIO)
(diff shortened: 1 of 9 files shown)


clash-lib/src/Clash/Rewrite/Util.hs
```diff
@@ -381,15 +381,19 @@ inlineBinders
...
-  if inline && elemFreeVars i res then
-    let inScope1 = extendInScopeSet inScope0 i
-        subst = extendIdSubst (mkSubst inScope1) i x
-     in changed (substTm "inlineBinders" subst res)
-  else
-    return expr
+inlineBinders condition (TransformContext inScope0 _) expr@(Let (NonRec i x) res)
+  -- Substitution would be a no-op for a binder that does not occur in the
+  -- body, so the (potentially expensive) property test can be skipped.
+  | not (elemFreeVars i res) = return expr
```
christiaanb: Since we already know `i` is not used in `res`, perhaps we can just do `changed res` instead of `return expr` here. It saves us time as we don't have to look at any of the dead code in future transformation.

Also, there is a `notElemFreeVars`, though I do not think using it will offer much performance benefit.

martijnbastiaan: Ah, good idea! Will apply before merging.

martijnbastiaan: (Provided an approval, of course.)

martijnbastiaan:
@mergify rebase

christiaanb (approved):
Looks good!

PR #3357, opened 2026-08-14
Fix performance regressions on `master`

martijnbastiaan:
Fixes performance regressions introduced by previous PRs. See https://clash-lang.github.io/clash-benchmarks/#machine=volthe&head=22505bfbe2aebb46875f8e86d3b04cbe41eeaed3.

## Still TODO:

  - [x] ~~Write a changelog entry (see changelog/README.md)~~ I don't think this warrants one, though we should probably write one big performance changelog with the accumulated end results.
  - [x] Remove Claudisms / review (@martijnbastiaan)

### DIFF
--- a/clash-lib/src/Clash/Normalize.hs
+++ b/clash-lib/src/Clash/Normalize.hs
@@ -347,6 +347,37 @@ flattenNode b@(CBranch (nm,(Binding _ _ _ _ e _)) us) = do
            then return (Right ((nm,e),us))
            else return (Left b)
 
+{-
+Note [flatten pass structure]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Through experimentation we've learned the following:
+
+1. The evaluator-backed rewrites ('reduceConst', 'reduceNonRepPrim') must not
+   sit inside the top-down propagation bundle. 'topdownFixR' settles that bundle
+   at each node with 'repeatR', so a bundled 'reduceConst' is re-attempted -
+   evaluator call and all - for every 'appProp' or 'caseCon' that fires there.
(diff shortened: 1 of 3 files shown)


rowanG077 (changes requested):
It's non obvious to me why the `flatten` is equivalent as it was before. 

As in we potentially end with a different AST whereas we didn't before? I'm curious if you considered this.

martijnbastiaan:
My understanding is that as long as the same set of transformations run in a `repeatR` structure, in whatever order, it should reach the same fixed point (perhaps modulo uniqs / non-terminating behavior). I inferred this from earlier comments from @christiaanb though, so I might be entirely incorrect.

christiaanb:
[Church-Rosser](https://en.wikipedia.org/wiki/Confluence_(abstract_rewriting)#Church%E2%80%93Rosser_property) is not proven for the transformations. So there can be multiple normal forms. `repeatR`  will ensure you get to one of them (modulo termination triggers), so @rowanG077 is right we might end up with a different AST; that said, it should still be in normal-form meaning we can still translate it to HDL.

martijnbastiaan:
FWIW, on `bittide` 33 out of 51 HDL files are byte identical. For the others they change, but the changes are of the form:

```
t = c$t_case_scrut[7:0];
result = b ? t : X;
```

=>

```
t = c$t_case_scrut[7:0];
result = b ? c$t_case_scrut[7:0] : X;
```

Experimentation shows that this is caused by changing `topdownFixR` to `topdownR`, not by the flattening changes. Just the flattening changes produce byte-identical files across all files.

rowanG077 (approved):

PR #3360, opened 2026-08-18
Correct `RTree` offset calculations in SV backend

### LINKED ISSUE #3359: SystemVerilog backend emits out-of-bounds array index for nested `RTree` selection
The SystemVerilog backend computes the size of an `RTree` half with the base and
exponent flipped:

```haskell
-- clash-lib/src/Clash/Backend/SystemVerilog.hs:1459,1467-1468
lhsSz = (d-1)^(2 :: Int)     -- should be 2^(d-1)
rhsS  = (d-1)^(2 :: Int)     -- should be 2^(d-1)
rhsE  = d^(2 :: Int)-1       -- should be 2^d - 1
```

The VHDL backend has the structurally identical element-index (`Slice`) path and
gets it right:

```haskell
-- clash-lib/src/Clash/Backend/VHDL.hs:2229,2244-2245
z  = 2^(d - 1)
z' = 2^d
```

An `RTree d` has `2^d` elements (`Netlist/Util.hs:707`), so each half has
`2^(d-1)`. Driving the same modifier through all three backends:

| Backend | Emitted | |
|---|---|---|
| SystemVerilog | `tree_sel[0][4:8]` | **out of bounds** on an 8-element array (valid `0..7`) |
| VHDL | `tree_sel(0)(4 to 7)` | correct |
| Verilog | `tree_sel[11:8]` | correct (bit range, not element index) |

Verilog is unaffected because it has no element-index path.

## Fix

```haskell
lhsSz = 2^(d-1)
rhsS  = 2^(d-1)
rhsE  = 2^d - 1
```

martijnbastiaan:
Fixes #3359

## Still TODO:

### DIFF
--- a/changelog/2026-08-18T11_53_09+02_00_fix_3359
+++ b/changelog/2026-08-18T11_53_09+02_00_fix_3359
@@ -1,0 +1,1 @@
+FIXED: The SystemVerilog backend no longer emits out-of-bounds array indices when a nested modifier selects a half of an `RTree`. The size of a tree half was computed as `(d-1)^2` instead of `2^(d-1)`, so for an `RTree 3` the right half was emitted as `[4:8]` on an array whose valid indices are `0..7`. See [#3359](https://github.com/clash-lang/clash-compiler/issues/3359).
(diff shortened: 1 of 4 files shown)


rowanG077 (approved):
Straightforward

PR #3362, opened 2026-08-18
Compare type variables structurally, not just by unique

### LINKED ISSUE #3361: `eqType` doesn't take types into account
`eqType` is supposed to do structural equality, but it fails to do so when it encounters a `Var`, because its `Eq` instance is based on `varKey` which ignores types. This is fine in "alpha blind" contexts, where the caller is supposed to make sure that terms that are being compared are scoped in the same way, but not for situations where it's not. In https://github.com/clash-lang/clash-compiler/pull/3182 this exposed a bug where type caching (https://github.com/clash-lang/clash-compiler/pull/3340) would fail because of it.

https://github.com/clash-lang/clash-compiler/blob/7328d55979ec49f061c7a5b80a9cbc294622fb12/clash-lib/src/Clash/Core/Subst.hs#L926-L930

https://github.com/clash-lang/clash-compiler/blob/7328d55979ec49f061c7a5b80a9cbc294622fb12/clash-lib/src/Clash/Core/Var.hs#L63-L67

This also goes for `GHC2Core`'s `eqType`, where I originally discovered the issue.

martijnbastiaan:
Add `eqName`/`ordName` for structural comparison of names, `eqVar`/`ordVar` for variables, and `ordType` to go with the existing structural `eqType`. Use `eqVar` in `eqType` and `eqTerm`, which had the same hole at occurrences: they already compared `varType` by hand at binders, so `eqVar` replaces those.

Note that `Name`s now also get compared. I'm not 100% sure that this is needed, but better safe than sorry..

Fixes #3361

## Still TODO:

    - Note that while the caching behavior isn't part of any release, the subtle error is still there. So I figured to include a changelog entry still.
  - [x] Wait for https://clash-lang.github.io/clash-benchmarks/#machine=oele&head=149ca37f4a25659fed68570d64508d414927dac5. Hopefully this doesn't undo the performance wins of https://github.com/clash-lang/clash-compiler/pull/3340
    - No performance regression, or very small.

### DIFF
--- a/clash-lib/src/Clash/Core/Subst.hs
+++ b/clash-lib/src/Clash/Core/Subst.hs
@@ -66,6 +66,9 @@ module Clash.Core.Subst
     -- * Structural equivalence
   , eqTerm
   , eqType
+  , ordType
+  , eqVar
+  , ordVar
   )
 where
 
@@ -88,6 +91,7 @@ import           GHC.Types.SrcLoc
   (SrcSpan (RealSrcSpan, UnhelpfulSpan), leftmost_smallest)
 
 import           Clash.Core.HasFreeVars
+import           Clash.Core.Name           (eqName, ordName)
 import           Clash.Core.Pretty         (ppr, fromPpr)
 import           Clash.Core.Term
   (Alt, Bind(..), Pat (..), Term (..), TickInfo (..), PrimInfo(primName))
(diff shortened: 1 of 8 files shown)


clash-lib/src/Clash/Core/Subst.hs
```diff
@@ -996,15 +1036,14 @@ eqTerm = go
...
       List.all2
         (\(b1,r1) (b2,r2) ->
-          b1 == b2 &&
           -- We need to check the types of Rec bindings, because:
           --
           -- letrec (x : Bool) = x in X
           --
           -- is not structurally equivalent to
           --
           -- letrec (x : Int) = x in x
```
christiaanb: Do we still need this comment?

christiaanb (approved):

clash-lib/src/Clash/Core/Subst.hs
martijnbastiaan: No

clemenslosbichler-cloud:
quick q on the test side

changelog points at #3361 , and the actual fix for that is eqVar/eqVarName in GHC2Core's eqType, the one keying _convertedTypeCache. but the new StructuralEquivalence tests all target the Clash.Core.Subst mirrors, so afaics nothing actually exercises the GHC2Core path the changelog names

is that covered in the integration suite somewhere, or is GHC2Core just too painful to unit test directly? might be missing smth

PR #3393, opened 2026-09-01
nix: Update flake

rowanG077:
This nixpkgs updates removes `x86_64-darwin`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -100,11 +100,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1779877693,
-        "narHash": "sha256-NOF9NAREhxr50bbBfVcVOq+ArCMSoe8dP79Pk2uyARk=",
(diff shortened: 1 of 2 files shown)


jaschutte (approved):

DigitalBrains1:
I suspect we need to do the exact same on [`stack-templates/.../simple-nix`](https://github.com/clash-lang/clash-starters/tree/main/simple-nix).

rowanG077:
only once we release 1.10.2

DigitalBrains1:
Ah, you want to support `x86-64_darwin` users to the final drop, the day when they can no longer use `nix flake init` because the world has moved on? :-)

We could add a branch where we keep a copy of the last version that supported `x86_64-darwin`. Would be good to do that _now_ though, as the `clash-starters` repo has _no history_ on the `main` branch. Or we could simply add another template that we don't update.

PR #3398, opened 2026-09-03
Retry `cabal update` on CI

martijnbastiaan:
Both Hackage and GitHub can be unreliable, making `cabal update` fail.

## Still TODO:


DigitalBrains1 (approved):
Good idea!

PR #3408, opened 2026-09-04
Run `deadCode` after flattening loop

### LINKED ISSUE #3407: Dead, non-rep arguments of lifted local functions are not removed
Found and reported by @lmbollen.

Reproducer:

```haskell
{- A dead, non-representable argument of a lifted local function is never
   removed; when the function is inlined during flattening, the argument's
   definition stays behind as a dead let-binding and reaches netlist generation:

     No blackbox found for: GHC.Prim.newMutVar#
-}
module TXXXX where

import Clash.Prelude
import Data.IORef (IORef, newIORef)
import System.IO.Unsafe (unsafePerformIO)

topEntity :: Unsigned 8 -> Unsigned 8 -> Unsigned 8
topEntity k x = step (unsafePerformIO (newIORef 0)) x + step (unsafePerformIO (newIORef 0)) (x + 1)
 where
  -- Needs: (1) a free variable ('k') so it is not the first argument that is
  -- non-representable, (2) two call sites so GHC keeps it as a let-bound
  -- lambda, (3) a body larger than -fclash-inline-function-limit.
  step :: IORef Int -> Unsigned 8 -> Unsigned 8
  step _ a = (a + k) * (a - k) + a * k
```

The original issue showed this behind a `clashSimulation` flag which made it quite confusing, though it boils down to this.

martijnbastiaan:
Fixes #3407

## Still TODO:

### DIFF
--- a/changelog/entries/2026-09-04_T17:48:40_run_dead_code_elimination_after_flattening_loop.md
+++ b/changelog/entries/2026-09-04_T17:48:40_run_dead_code_elimination_after_flattening_loop.md
@@ -1,0 +1,7 @@
+---
+issues: [3407]
+prs: []
+---
+
+# FIXED
+Clash no longer errors when unused arguments of local functions are non-representable in very specific circumstances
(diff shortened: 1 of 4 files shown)


martijnbastiaan:
I talked to @christiaanb in the office and we concluded that this is the right patch.

PR #3409, opened 2026-09-04
Avoid `Integer` in `popCount` and `count{Leading,Trailing}Zeros`

martijnbastiaan:
`popCount`, `countLeadingZeros` and `countTrailingZeros` on `BitVector` (and hence on `Unsigned`, `Signed` and `Index`, which delegate to it) went through `Integer`, which made Clash emit a `Dubious primitive instantiation` warning for `GHC.Num.Integer.integerToInt#` and put an `Integer` in the generated HDL. They now use
`Clash.Sized.Internal.Index.fromEnum#`, which has a black box of its own.

Picked from #3351.

## Still TODO:

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -474,6 +474,8 @@ runClashTest = defaultMain
         , runTest "ReduceOne" def
         , runTest "ExtendingNumZero" def
         , runTest "AppendZero" def
+        , runTest "PopCountNoInteger"
+            def{clashFlags=["-Werror=clash-dubious-primitive"]}
         , runTest "PackGHCNums" def
         , runTest "UnpackGHCNums" def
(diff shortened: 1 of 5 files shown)


rowanG077 (approved):
awesome!

tests/Main.hs
```diff
@@ -474,6 +474,10 @@ runClashTest = defaultMain
         , runTest "ReduceOne" def
         , runTest "ExtendingNumZero" def
         , runTest "AppendZero" def
+        , runTest "SplitZeroWidth"
+            def{clashFlags=["-Werror=clash-unmatchable-constant"]}
```
martijnbastiaan: Woops, messed up a rebase 😳

PR #3410, opened 2026-09-04
Warn about `resetGenN` on every backend

martijnbastiaan:
Its VHDL and Verilog black boxes render nothing outside `translate_off`, exactly like the SystemVerilog one, which was the only one marked non-synthesizable.

Picked from #3351

## Still TODO:


DigitalBrains1 (approved):
"cannot" does not have a space in it, but that's pretty minor. It's just a bit too much nothing after all.

martijnbastiaan:
I'll fix it anyway and check whether there are any others lingering.

martijnbastiaan:
There are others, but they're in code we inherited from GHC..

PR #3411, opened 2026-09-04
Promote "unmatchable constant" warning to proper warning and improve warning traceability

martijnbastiaan:
More or less ad-hoc additions to warnings in order to track down the provenance of warnings seen in (a downstream project) `bittide-hardware`.

Picked from #3351.

## Still TODO:

  - [ ] Think about whether we want more/less (unless a reviewer says OK :-))


rowanG077 (approved):
I'm on board with this. But if I look at the messages this is still pretty esoteric. But it's better if we get bug reports with more info :).

martijnbastiaan:
It very much is. But it also shouldn't happen, they're mostly there to help us..

PR #3419, opened 2026-09-07
qlog entries

marijn-qbaylogic:
This PR converts all existing entries on `master` to the qlog format.

The commits are spread out so git can detect the change as MOVE+MODIFY, rather than DELETE+ADD. This ensures that if the branch is rebased on `master` after a release has deleted an entry, the conflicts result in the new-format file getting deleted. This is not relevant for _this_ PR, but the same method may be used if other old-style entries pop up.

Backporting directly is likely not possible - see #3420.


rowanG077 (approved):
Awesome!

PR #3421, opened 2026-09-07
Add pipeline for rejecting PRs with old changelog entries.

marijn-qbaylogic:
A simple pipeline that errors if files are added to the `changelog/` directory (not the `changelog/entries/` subdirectory).
This file is almost certainly an old-style entry and you should convert it before merging, to avoid making a mess of everything.


DigitalBrains1:
You could add it to [`/.ci/small_tests.sh`](https://github.com/clash-lang/clash-compiler/blob/a00b694530056f2d6ab1bf68f27493a423a28f82/.ci/small-tests.sh) which is invoked [here](https://github.com/clash-lang/clash-compiler/blob/a00b694530056f2d6ab1bf68f27493a423a28f82/.github/workflows/on-push.yml#L89-L90).

DigitalBrains1:
If we add something like

```yaml
env:
    BASE_REF: ${{ github.base_ref }}
```

to the `step` that runs `small_tests.sh`, you could check that you are running CI for a PR, by checking that var is non-empty.

[edit]
Maybe like this?

```diff
$ git diff
diff --git a/.ci/small-tests.sh b/.ci/small-tests.sh
index f551ddd1a..840d6f3b9 100755
--- a/.ci/small-tests.sh
+++ b/.ci/small-tests.sh
@@ -41,3 +41,15 @@ fi
 mv $GIT_ROOT/cabal.project.local $GIT_ROOT/cabal.project.local.disabled || exit $?
 cabal v2-build -j$THREADS --dry-run all > /dev/null || { echo "Maybe state index should be updated?"; exit 1; }
 mv $GIT_ROOT/cabal.project.local.disabled $GIT_ROOT/cabal.project.local || exit $?
+
+# XXX: Remove this test once we're certain there will be no more new PR's with
+# old-style changelog entries
+if [[ -n $BASE_REF ]]; then
+    git fetch --shallow-exclude="$BASE_REF^@" origin HEAD || exit $?
+    # List added files in changelog directory
+    changelog_diff=$(git diff --name-status "$BASE_REF"..HEAD | grep '^[A]\s\+changelog/' | grep -v 'changelog/entries/')
+    if [[ -n $changelog_diff ]]; then
+        echo "::warning title=Old-style changelog entries found::$changelog_diff"
+        exit 1
+    fi
+fi
diff --git a/.github/workflows/on-push.yml b/.github/workflows/on-push.yml
index d2d6bb911..76c0b891f 100644
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -87,6 +87,8 @@ jobs:
           cabal v2-build all -j$(.ci/effective_cpus.sh)
 
       - name: Run small tests
+        env:
+          BASE_REF: ${{ github.base_ref }}
         run: .ci/small-tests.sh
 
       - name: Clash-dev test
```

Hmmm this doesn't work yet. I can't come up with the command to fetch the history from `$BASE_REF` up to `HEAD`. I get a weird error: `fatal: expected 'acknowledgments'`.

But we could just do `git fetch origin "$BASE_REF" HEAD` which will make the whole checkout unshallow just like the current PR does.
[/edit]

martijnbastiaan:
Isn't it much simpler just to check for any illegal changelog files? Don't check if it's added in the PR, just like we don't check whether compilation warnings are added in a certain PR.

DigitalBrains1:
I think it's wasteful to spin up a complete job for this, even if we move it to our own runners.

I also think it's wasteful to have to download the full git history just for this.

I have a few ideas and I'd like to have a go at it tomorrow. I actually think I'm almost there.

.ci/small-tests.sh
```diff
@@ -41,3 +41,11 @@ fi
 mv $GIT_ROOT/cabal.project.local $GIT_ROOT/cabal.project.local.disabled || exit $?
 cabal v2-build -j$THREADS --dry-run all > /dev/null || { echo "Maybe state index should be updated?"; exit 1; }
 mv $GIT_ROOT/cabal.project.local.disabled $GIT_ROOT/cabal.project.local || exit $?
+
+# Test for old changelog entries
+old_entries=$(find changelog/ -maxdepth 1 -type f -name '20*')
+
+if [ -n "$old_entries" ]; then
+    echo "::warning title=Old-style changelog entries found::$old_entries"
```
martijnbastiaan, line 49:
```suggestion
    echo "::error title=Old-style changelog entries found::$old_entries"
```

:-)

martijnbastiaan (approved):
OK! But let's wait for @DigitalBrains1 too.

martijnbastiaan:
@marijn-qbaylogic You didn't wait for @DigitalBrains1 and you did not apply my suggestion.

rowanG077:
@marijn-qbaylogic usually enables auto merge. So I think it's the approval that triggered the merge. 

Edit: Ah no the timeline does not match up.

martijnbastiaan:
We don't require reviews on this repo.

marijn-qbaylogic:
Ah.

Seems I accidentally merged without approval.

Sorry, I thought reviews were required, and either had it on auto merge, merging it when the pipeline succeeded, or clicked merge myself, because I thought the fact I _could_ do that meant it was approved.

DigitalBrains1:
I think this should be backported.

marijn-qbaylogic:
It would be possible to fix the `warning -> error` change, but imho it's not worth the hassle at this point. Can we just backport?

PR #3422, opened 2026-09-07
Fix GHC 9.14 unary-class to include superclass dictionaries

rowanG077:
Reported by @kleinreact as he hit this in the clash-crypto project

## Still TODO:

### DIFF
--- a/tests/shouldwork/Polymorphism/UnaryClass.hs
+++ b/tests/shouldwork/Polymorphism/UnaryClass.hs
@@ -1,0 +1,16 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module UnaryClass where
+
+import Clash.Prelude
+
+class KnownNat n => Alias n
+instance KnownNat n => Alias n
+
+increment :: Alias n => Unsigned n -> Unsigned n
+increment x = x + 1
+{-# OPAQUE increment #-}
+
+topEntity :: Unsigned 8 -> Unsigned 8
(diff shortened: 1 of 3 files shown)


changelog/entries/2026-09-07_T19:57:53_fix_unary_class_superclasses.md
```diff
@@ -0,0 +1,8 @@
+---
+issues: []
+prs: [3422]
+no-links: true
+---
+
+# FIXED
+Fix compilation of classes with a single superclass and no methods on GHC 9.14.
```
martijnbastiaan: We don't have a release with 9.14 in it yet, so I don't think this needs a changelog entry.

tests/shouldwork/Polymorphism/UnaryClass.hs
```diff
@@ -0,0 +1,17 @@
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module UnaryClass where
+
+import Clash.Prelude
+
+-- A superclass dictionary is stored in dataConTheta, not dataConOrigArgTys.
```
martijnbastiaan:
```suggestion
-- A superclass dictionary is excluded from @dataConOrigArgTys@. GHC2Core should
-- use @dataConRepArgTys@.
```

I don't think it makes sense to involve `dataConTheta`, because we don't use that (only indirectly).

martijnbastiaan (approved):
LGTM, two minor comments.

PR #3424, opened 2026-09-08
Add a CI job to check changelog entries in PRs

marijn-qbaylogic:
Added a job that runs on PRs and checks any changelog entries that have been modified.


rowanG077 (approved):
It looks good to me but I was not too involved here so I will defer to @DigitalBrains1

DigitalBrains1 (approved):
The end result is the combined effort of Marijn and me. We both approve of the final version (the PR author can't approve their own PR, so I'm adding this information in this fashion).

DigitalBrains1:
And the `flake.nix` change was provided by Rowan. But I realise this too late now to add him as a co-author, even though that would have been better.

PR #3434, opened 2026-09-14
Allow `ensureSpine` on `Signal`s, add `dontTranslate` to X-testing functions

### LINKED ISSUE #3432: Signal has an `NFDataX` instance, but `ensureSpine` crashes Clash with an indecipherable error
The following makes use of `Signal`s `NFDataX` instance:

```haskell
{-# LANGUAGE NoImplicitPrelude #-}

module Top where

import Clash.Prelude

topEntity :: Signal System ()
topEntity = ensureSpine undefined
```

I would expect that either `ensureSpine` is supported by clash by propagation to the signal's elements, or that I get an error that it is only supported in simulation. Instead, Clash reports:

```
GHC: Setting up GHC took: 0.133s
GHC: Compiling and loading modules took: 0.926s
Clash: Parsing and compiling primitives took 0.661s
GHC+Clash: Loading modules cumulatively took 1.853s
Clash: Compiling Top.topEntity
<no location info>: error:
    Other error:
    Internal error: 'reduceBindersCleanup' encountered a variable reference that was 
    neither in 'doneInl', 'origInl', or in the transformation's in scope set. Unique 
    was: '6989586621679028736'.
```

martijnbastiaan:
Two things:

* `ensureSpine` on a `Signal` no longer crashes Clash with an internal error. When generating HDL, `ensureSpine` on a `Signal` is the identity function. See https://github.com/clash-lang/clash-compiler/issues/3432.
* Functions testing for `XException` (`isX`, `hasX`, `maybeIsX`, and `maybeHasX`) are now marked non-translatable. This makes the Clash compiler throw a proper error, instead of an internal one when it encounters their use.

Fixes #3432

## Still TODO:


tests/shouldwork/Signal/T3432.hs
```diff
@@ -0,0 +1,27 @@
...
+
+-- 'ensureSpine' on a 'Signal' used to expose the signal constructor ':-' to
+-- Clash, which crashed Clash with an internal error. In HDL, 'ensureSpine' on
+-- a 'Signal' should be the identity function.
+--
+-- See https://github.com/clash-lang/clash-compiler/issues/3432
+topEntity ::
+  Signal System (Unsigned 8, Bool) ->
+  Signal System (Unsigned 8, Bool)
+topEntity = ensureSpine
```
rowanG077: Should this not match the repro?

tests/shouldwork/Signal/T3432.hs
```diff
@@ -0,0 +1,27 @@
...
+{-# OPAQUE topEntity #-}
+
+testBench :: Signal System Bool
+testBench = done
```
rowanG077: Does it need a testbench? I think a clash compile is enough.

tests/shouldwork/Signal/T3432.hs
martijnbastiaan: Yeah, fair enough. The completely void type kinda felt iffy as it might (in some future implementation) cause Clash to perhaps clean up early.

tests/shouldwork/Signal/T3432.hs
martijnbastiaan: Fair! I've also restricted it to just run against Verilog.

rowanG077 (approved):

clash-prelude/src/Clash/XException.hs
```diff
@@ -307,6 +308,7 @@ maybeHasX = maybeX hasX
 -- > maybeIsX _|_                 = _|_
 maybeIsX :: a -> Maybe a
 maybeIsX = maybeX isX
+{-# ANN maybeIsX dontTranslate #-}
```
DigitalBrains1, line 311: Doesn't this also need `OPAQUE` to prevent GHC from mangling it in some way that prevents the annotation from springing into action?

martijnbastiaan: Oh flip, yes it does! I'll add a check to CI too when I fix this.

pieter-bos:
> When generating HDL, ensureSpine on a Signal is the identity function.

What do we think about something like this?

``` haskell
{-# LANGUAGE GHC2024 #-}
module SpinedNonEmptyBitRepr where

import Clash.Prelude
import Clash.Annotations.BitRepresentation
import Clash.Annotations.BitRepresentation.Deriving

data PrefixedBox = PrefixedBox (BitVector 2) (BitVector 2)
    deriving (Generic, NFDataX, Show)

{-# ANN module (
    DataReprAnn
        $(liftQ [t|PrefixedBox|])
        8
        [ ConstrRepr 'PrefixedBox 0b1111_0000 0b0101_0000 [0b0000_1100, 0b0000_0011] ]
 ) #-}

$(deriveBitPack [t|PrefixedBox|])

topEntity :: Signal System PrefixedBox
topEntity = ensureSpine undefined
```

For example:

```
clashi> pack <$> topEntity
0b0101_.... 0b0101_.... 0b0101_.... 0b0101_.... 0b0101_.... [...]
```

rowanG077:
@pieter-bos 

I went in here with the understanding that `ensureSpine` IS the identity function in HDL. But that's wrong, even for things that do not have custom bitpacks:

```Haskell
topEntity :: Maybe Bool -> Maybe Bool
topEntity = ensureSpine
```

Clash generates

```verilog
assign result = eta[1:1] ? {1'b1,g1} : {1'b0,1'bx};
```

So I think the PR is wrong. We either should apply `ensureSpine` to signals OR we need to make clear this is a sim only construct and also not generate logic at all. 

Even worse. Wrapping the `Maybe Bool` in a `Signal` makes it actually identity.

martijnbastiaan:
Rowan, could you spell out for me how your example is wrong? 

Pieter, very interesting example! I've always thought about spines as "the things not existing in HDL". So for your product type Clash should never actually generate HDL that discriminates on the constructor bits. Of course, that breaks down at design boundaries.

rowanG077:
```Haskell
topEntity :: Maybe Bool -> Maybe Bool
topEntity = ensureSpine`
```

Produces

```
assign result = eta[1:1] ? {1'b1,g1} : {1'b0,1'bx};
```

Morally I believed that `ensureSpine` should not introduce logic.

DigitalBrains1:
Please note that these two are both definitions of identity functions:

```haskell
f0 :: Maybe Bool -> Maybe Bool
f0 = id

f1 :: Maybe Bool -> Maybe Bool
f1 = \case
  Just b -> Just b
  Nothing -> Nothing
```

The problem is that `(GEnsureSpine a, GEnsureSpine b) => GEnsureSpine (a :+: b)` does what `f1` does, and so introduces logic. But the logic is still the identity function.

I understand that it might generate more hardware, though: I understand that that's an issue. The identity function on `Maybe Bool` is not the identity function on the underlying Verilog type `wire [1:0]`.

[edit]
Ah, discovering after I write that message that Rowan indeed writes _identity function **in HDL**_ is a bit late, isn't it?
[/edit]

DigitalBrains1:
Perhaps we should add a law to `NFDataX` that it must not generate logic (it may fix wires at a fixed value, for the custom bit representation above). And make `GEnsureSpine (a :+: b)` this:

```
instance (GEnsureSpine a, GEnsureSpine b) => GEnsureSpine (a :+: b) where
  gEnsureSpine
    | clashSimulation
    = \case
        (L1 x) -> L1 (gEnsureSpine x)
        (R1 x) -> R1 (gEnsureSpine x)
    | otherwise
    = id
  {-# INLINEABLE gEnsureSpine #-}
```

The other `GEnsureSpine`s already don't generate _logic_ I think (but they'll happily generate the fixed wires of the custom data representation above, since they transform `ensureSpine undefined` into ``ensureSpine undefined `PrefixedBox` ensureSpine undefined``).

[edit]
Aaand Pieters example slightly modified already uncovers a new bug....

<details><summary><code>SpinedNonEmptyBitRepr.hs</code></summary>

```haskell
{-# LANGUAGE GHC2024 #-}
module SpinedNonEmptyBitRepr where

import Clash.Prelude
import Clash.Annotations.BitRepresentation
import Clash.Annotations.BitRepresentation.Deriving

data PrefixedBox = PrefixedBox (BitVector 2)
    deriving (Generic, NFDataX, Show)

{-# ANN module (
    DataReprAnn
        $(liftQ [t|PrefixedBox|])
        6
        [ ConstrRepr 'PrefixedBox 0b1111_00 0b0101_00 [0b0000_11] ]
 ) #-}

$(deriveBitPack [t|PrefixedBox|])

topEntity :: PrefixedBox
topEntity = ensureSpine undefined
```
</details>

works in clashi:
```haskell
>>> pack topEntity
0b01_01..
```

but unfortunately, generating Verilog:
```
Clash: Compiling SpinedNonEmptyBitRepr.topEntity
<no location info>: error:
    Other error:
    Unexpected HWType: BitVector 2
```

Or am I doing something silly? I don't think so.
[/edit]

martijnbastiaan:
> Morally I believed that `ensureSpine` should not introduce logic.

Aha, I think I understand the miscommunication then. I was thinking post-synth.

FWIW, I've pulled:

```verilog
module dut_x (input [1:0] eta, input g1, output [1:0] result);
  assign result = eta[1:1] ? {1'b1,g1} : {1'b0,1'bx};
endmodule
```

through Vivado, Quartus, and Yosys (`synth_xilinx` and `synth_ice40`) and none of them produce LUTs.

PR #3436, opened 2026-09-14
Cache Verilator builds in `clash-testsuite` with `ccache`

martijnbastiaan:
On another PR, I noticed that the (System)Verilog test suites were taking much longer than the VHDL ones. Turns out, Verilator is doing a bunch of repetitive work which can be cached using `ccache`.

For N=1:

Before:

Screenshot 2026-09-14 at 20-59-14 Allow `ensureSpine` on `Signal`s when translating to HDL · clash-lang_clash-compiler@c52683f

After:

Screenshot 2026-09-14 at 20-59-33 Cache Verilator builds in `clash-testsuite` with `ccache` · clash-lang_clash-compiler@798d19e

## Still TODO:


rowanG077 (approved):
Nice!

PR #3438, opened 2026-09-15
Fixed several compliation issues

marijn-qbaylogic:
Fixed several tests


rowanG077:
Please do not use PR to merge things into other development branches like this. We never done it like this and in general cherry picks are sufficient.

PR #3443, opened 2026-09-16
Fix zero-width extending arithmetic

### LINKED ISSUE #3310: Zero width `Index 1` arguments break ExtendingNum functions
```haskell
topEntity :: Index 4 -> Index 1 -> Index 4
topEntity x y = x `add` y
```

translates into 

```verilog
module topEntity
    ( // Inputs
      input wire [1:0] c$arg


      // Outputs
    , output wire [1:0] result
    );


  assign result = c$arg + __VOID__;
endmodule
```

Clash eliminates the zero width `Index 1` argument from the function, but it can eliminate it from the primitive.
The corresponding BitVector/Signed/Unsigned primitives are coded to check the width of of their inputs, but those for Index are not.
And the same for `sub` and `mul`.

We should expand the [ExtendingNumZero test](https://github.com/clash-lang/clash-compiler/blob/174eac30d07fde4e4325030d1b24d85f7089f599/tests/shouldwork/BitVector/ExtendingNumZero.hs) to verify this works all our number types.

I also already saw there is also a problem in VHDL when doing:
 `sub (0 :: Unsigned 0) (n :: Unsigned 16)`

rowanG077:
Fixes https://github.com/clash-lang/clash-compiler/issues/3310

## Still TODO:


clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives.yaml
```diff
@@ -448,8 +448,20 @@
...
+          - resize(unsigned(~ARG[3]), ~SIZE[~TYPO]))
+      ~ELSE
+        ~IF ~SIZE[~TYP[2]] ~THEN
+          std_logic_vector(resize(unsigned(~ARG[2]), ~SIZE[~TYPO]))
+        ~ELSE
+          std_logic_vector(
+            to_unsigned(0, ~SIZE[~TYPO])
+            - resize(unsigned(~ARG[3]), ~SIZE[~TYPO]))
+        ~FI
+      ~FI
```
DigitalBrains1, line 464: I think we optimised it for what the output of the template looks like instead of what the template looks like. There's a lot of whitespace in the result now. My preference would be to keep optimising for the result.

This applies to all the primitives in this PR; I'm merely picking out this one because it shows how we did the whole `~IF` stuff before, as one solid chunk of unreadable mass.

This is what the VHDL looks like with this PR:

```vhdl
  result <= ( Tuple6_sel0_std_logic_vector_0 => std_logic_vector(resize(unsigned(n),17))
            , Tuple6_sel1_std_logic_vector_1 => std_logic_vector(resize(unsigned(n),17))
            , Tuple6_sel2_std_logic_vector_2 =>

              std_logic_vector(resize(unsigned(n), 17))


            , Tuple6_sel3_std_logic_vector_3 =>

              std_logic_vector(
                to_unsigned(0, 17)
                - resize(unsigned(n), 17))


            , Tuple6_sel4_std_logic_vector_4 => (16-1 downto 0 => '0')
            , Tuple6_sel5_std_logic_vector_5 => (16-1 downto 0 => '0') );
```
(this is `topBitVector` from `tests/.../ExtendingNumZero.hs` in this PR)

DigitalBrains1 (commented):
The characters look fine but I would prefer less room between the characters X-D

Also, I found these surprising except for the last one:
```haskell
>>> :k! AResult (BitVector 16) (BitVector 0)
AResult (BitVector 16) (BitVector 0) :: Type
= BitVector 17
>>> :k! AResult (Signed 16) (Signed 0)
AResult (Signed 16) (Signed 0) :: Type
= Signed 17
>>> :k! AResult (Unsigned 16) (Unsigned 0)
AResult (Unsigned 16) (Unsigned 0) :: Type
= Unsigned 17
>>> :k! AResult (Index 65536) (Index 1)
AResult (Index 65536) (Index 1) :: Type
= Index 65536
```

And for the following it could be argued that the result should be 0 bits wide (which it already is for `Index`):
```haskell
>>> :k! MResult (BitVector 16) (BitVector 0)
MResult (BitVector 16) (BitVector 0) :: Type
= BitVector 16
>>> :k! MResult (Signed 16) (Signed 0)
MResult (Signed 16) (Signed 0) :: Type
= Signed 16
>>> :k! MResult (Unsigned 16) (Unsigned 0)
MResult (Unsigned 16) (Unsigned 0) :: Type
= Unsigned 16
>>> :k! MResult (Index 65536) (Index 1)
MResult (Index 65536) (Index 1) :: Type
= Index 1
```

This is only related to the subject of this PR, but it caught my attention while reading the test benches and thinking "wait, why would you even need `resize` in this code?".

clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives.yaml
rowanG077: Yes that's terrible. I will fix that. The template is truly impenetrable but I guess you can format it if you need to touch it.

DigitalBrains1: Yes, that's how I do it as well. It's not a nice dev experience at all. But a lot of our primitives would be better expressed in the other options we have instead of the template language, so it's not obvious that improving the template language would be the best way to spend developer time.

rowanG077:
> The characters look fine but I would prefer less room between the characters X-D
> 
> Also, I found these surprising except for the last one:
> 
> ```haskell
> >>> :k! AResult (BitVector 16) (BitVector 0)
> AResult (BitVector 16) (BitVector 0) :: Type
> = BitVector 17
> >>> :k! AResult (Signed 16) (Signed 0)
> AResult (Signed 16) (Signed 0) :: Type
> = Signed 17
> >>> :k! AResult (Unsigned 16) (Unsigned 0)
> AResult (Unsigned 16) (Unsigned 0) :: Type
> = Unsigned 17
> >>> :k! AResult (Index 65536) (Index 1)
> AResult (Index 65536) (Index 1) :: Type
> = Index 65536
> ```
> 
> And for the following it could be argued that the result should be 0 bits wide (which it already is for `Index`):
> 
> ```haskell
> >>> :k! MResult (BitVector 16) (BitVector 0)
> MResult (BitVector 16) (BitVector 0) :: Type
> = BitVector 16
> >>> :k! MResult (Signed 16) (Signed 0)
> MResult (Signed 16) (Signed 0) :: Type
> = Signed 16
> >>> :k! MResult (Unsigned 16) (Unsigned 0)
> MResult (Unsigned 16) (Unsigned 0) :: Type
> = Unsigned 16
> >>> :k! MResult (Index 65536) (Index 1)
> MResult (Index 65536) (Index 1) :: Type
> = Index 1
> ```
> 
> This is only related to the subject of this PR, but it caught my attention while reading the test benches and thinking "wait, why would you even need `resize` in this code?".

Yes I think that's out of scope for this PR. But my thoughts are:

For `AResult` I think it's reasonable as it also needs to cover subtraction. So `0 :: Signed 0 `sub` (minBound :: Signed 16)` does require 17 bits. For `Unsigned`/`BitVector`, removing the extra bit would change subtraction’s wraparound behavior.

`MResult` could be tighter but I it's a sweeping API change, and perhaps worse. Implementing it as a typeclass will break a lot of `ghc-typelits-*` reasoning.

clash-lib/prims/vhdl/Clash_Sized_Internal_BitVector.primitives.yaml
rowanG077: Fixed now.

DigitalBrains1:
> For `AResult` I think it's reasonable as it also needs to cover subtraction.

Whoops completely forgot about that, good point!

> For `Unsigned`/`BitVector`, removing the extra bit would change subtraction’s wraparound behavior.

I think you could compensate for that, though. The question `AResult` should answer is: how many bits do I need to represent the _result_? Nothing more. It would not dictate a particular implementation of the operation.

> `MResult` could be tighter but I it's a sweeping API change, and perhaps worse. 

I think we should only improve this edge case if it would be worth it compared to the added complexity. At a first glance, it's too much complexity.

DigitalBrains1 (approved):

